PR #185, opened 2010-12-02
"switch" statement

Dsls:
Hello,

While testing Twig, I have added a "switch" statement, that can enable tests such as :
{% switch a %}
{% case 1 %}
Action 1
{% case 2 %}
Action 2
{% default %}
Default display
{% endswitch %}

It is copied from php syntax (except that the "break" is always inserted after each case), and heavily inspired from the "if" statement.

Regards,
Bruno

### DIFF
--- a/lib/Twig/Node/Switch.php
+++ b/lib/Twig/Node/Switch.php
@@ -1,0 +1,64 @@
+<?php
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) 2009 Fabien Potencier
+ * (c) 2009 Armin Ronacher
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+/**
+ * Represents an if node.
+ *
+ * @package    twig
+ * @author     Fabien Potencier <fabien.potencier@symfony-project.com>
+ */
+class Twig_Node_Switch extends Twig_Node
+{
+    public function __construct(Twig_NodeInterface $value, Twig_NodeInterface $cases, Twig_NodeInterface $default = null, $lineno, $tag = null)
+    {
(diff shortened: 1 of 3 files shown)


Levdbas:
I know this goes way back but I would like to ask you to reconsider your statement. I do think a switch statement will lead to cleaner and easier to understand code.

>  In many occasions, you may want to compare the same variable (or expression) with many different values, and execute a different piece of code depending on which value it equals to. This is exactly what the switch statement is for. 

As stated in the PHP website. https://www.php.net/manual/en/control-structures.switch.php

This is something different then the suggested way-to-go, namely `if/elseif ` statements where you can choose which variable you want to evaluate.

warudin:
Agreed with the above. 
Mimicking PHP syntax might not be the main goal for Twig.
But readability is a main reason for using Twig, no?

I think everyone agrees that the switch statement is way more readable than if .. elseif.

```
<?php
// This switch statement:

switch ($i) {
    case 0:
        echo "i equals 0";
        break;
    case 1:
        echo "i equals 1";
        break;
    case 2:
        echo "i equals 2";
        break;
}

// Is equivalent to:

if ($i == 0) {
    echo "i equals 0";
} elseif ($i == 1) {
    echo "i equals 1";
} elseif ($i == 2) {
    echo "i equals 2";
}
?>
```

LiamAshdown:
> That's not something I will add to the core of Twig.
> 
> The goal is not to have the same syntax as PHP, but to provide a simple syntax for web designer. Simplicity also means that we need to provide only one way to do things, and the same can be easily accomplish with if statements (I would even say that ifs statements would be easier and more concise).
> 
> But the good news is that it is really easy to add this support if you want to as demonstrated by your pull request.

This is preposterous - You are making it sound like "switch statements" originated from PHP. Switch statements is one of most commonly used statement tool implemented across a vast amount of programming languages and template libraries.

It's a shame, really.

PR #353, opened 2011-06-05
added {% line \d+ %} directive

arnaud-lb:
This directive allows to change the current line number in the tokenizer.

This allows to generate Twig code from some other Twig or partially-Twig script, and still get relevant line numbers in error messages, when the line numbers of the generated Twig script do not match those of the original script.

For example in the following script:

```
foo
{% line 10 %}
bar
{{ baz }}
```

`"foo"` is on line 1, `"bar"` is considered to be on line 10 and `{{ baz }}` is considered to be on line 11.

This is similar to `# line ...` directives in C.

### DIFF
--- a/lib/Twig/Lexer.php
+++ b/lib/Twig/Lexer.php
@@ -162,6 +162,11 @@ class Twig_Lexer implements Twig_LexerInterface
                     $this->moveCursor($match[0]);
                     $this->lexRawData();
                     $this->state = self::STATE_DATA;
+                // {% line \d+ %}
+                } else if (preg_match('/\s*line\s+(\d+)\s*'.preg_quote($this->options['tag_block'][1], '/').'/As', $this->code, $match, null, $this->cursor)) {
+                    $this->moveCursor($match[0]);
+                    $this->lineno = (int) $match[1];
+                    $this->state = self::STATE_DATA;
                 } else {
                     $this->pushToken(Twig_Token::BLOCK_START_TYPE);
                     $this->state = self::STATE_BLOCK;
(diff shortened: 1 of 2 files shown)


ericmorand:
I know it's 8 years later but isn't it the purpose of source maps to actually do the mapping between a source and a compiled code?

When an error occurs at line 5, then we query the source-map generated by the aforementioned HAML compiler and we get line 3 as a result.

The `line` tag attempts to solve an issue that doesn't exist (and didn't in 2011 either since source maps were already a thing) and confuses the purposes of lexers, parsers and compilers.

stof:
This tag is meant to solve the case where your loader pre-processes the source code (removing a frontmatter for instance), so that the original line might still be reported.
Given that PHP does not have any sourcemap-based tooling, this was an easy way to solve the issue.

ericmorand:
> Given that PHP does not have any sourcemap-based tooling, this was an easy way to solve the issue.

A source-map is just an array of mappings, there is no need for a specialized tool to generate then. It's trivial to generate a source-map:

```php
// Our pre-processor generated the line 10, column 5 from line 6, column 1 of the source file "foo.whatever"

$sourceMap = [
    [
        'source' => 'foo.whatever',
        'originLine' => 6,
        'originalColumn' => 1,
        'generatedLine' => 10,
        'generatedColumn' => 5
    ]
]

return $sourceMap;
```

I can understand that it made things easier because people didn't want to output source-maps from their custom pre-processors, but the excuse `PHP does not have any sourcemap-based tooling` is not valid. What `line` tag looks like to me is a bad compromise for the sake of not asking pre-processor developers to output source-maps.

Anyway, now we end up with that:

```
{% line 5 %}
bar
{% line 1 %}
foo
```

Syntactically valid but doesn't convey any useful information and is conceptually wrong - `foo` is located at line 4, `{% line 1 %}` at line 3, that's the truth.

PR #610, opened 2012-01-24
Make the parser Inception-Proof

Seldaek:
Spent half a day debugging before I realized what happened, but I'll try to keep a long story short: 

When the cache is empty, and the first template containing an assetic `{% javascripts %}` or similar tag is parsed, it will build up the assetic "assets" or recipes cache, this in turn will tokenize and parse all your templates to find assetic tags and cache that information.

At this point the parser is parsing something else in the middle of a parse() call, and since there is a single instance in the environment, it means all the instance vars are messed up and contain incorrect references to the latest TokenStream that was parsed by assetic, etc.

This had two effects on my application, both appearing seemingly randomly because it highly depends on the order of things, the state of your cache and probably other factors:
- The first thing that happened is that a template was compiled using the wrong template filename, which means I had a `__TwigTemplate_abcd` in the file named `dcba.php`, and it would never find the right class.
- The second issue (could not reproduce but I assume it was caused by this as well) is that the parse tree is completely broken and you end up with a parse error because it thinks it's at the end when it's not, or similar problem.

The proposed fix basically pushes/pops all the vars into a stack whenever the parser starts/stops, which worked very effectively here and does not introduce much breakage or complexity.

### DIFF
--- a/lib/Twig/Parser.php
+++ b/lib/Twig/Parser.php
@@ -62,6 +62,7 @@ class Twig_Parser implements Twig_ParserInterface
      */
     public function parse(Twig_TokenStream $stream)
     {
+        // push all variables into the stack to keep the current state of the parser
         $vars = get_object_vars($this);
         unset($vars['stack'], $vars['env']);
         $this->stack[] = $vars;
@@ -109,6 +110,7 @@ class Twig_Parser implements Twig_ParserInterface
 
         $node = $traverser->traverse($node);
 
+        // restore previous stack so previous parse() call can resume working
         foreach (array_pop($this->stack) as $key => $val) {
             $this->$key = $val;
         }


stripyshirtguy:
We're still experiencing the 

`Unexpected token "text" of value "" ("end of statement block" expected)`

issue, reported in https://github.com/symfony/assetic-bundle/issues/29 but references this PR.

We're using Twig v1.22.2 and Assetic v1.1.3. I've confirmed this PR is merged in to the tag we're using and actually checked the code and I can see it in there. It happens in dev, doesn't seem to happen in production (although we warm the cache and errors/debugging are off/suppressed so maybe that helps).

We only get the error on first load of a modified template (or after clearing the Symfony file cache). After that it's perfectly fine. Narrowed it down to a line `{% javascripts '@version_javascript_includes' %}{% endjavascripts %}` where `@version_javascript_includes` is a configured list of input files as per Assetic usage. After refreshing and the error disappearing, the javascripts all load as they should.

Any ideas what could be causing this?

stof:
This may be related to the fact that this PR forgot to reset the state in case of an exception. It only resets it on successful parsing

stripyshirtguy:
Thanks @stof 

Ah, so there might be an underlying error/exception that is not being bubbling up? How would I be able to see that, even if just temporarily patching twig/assetic?

EmmanuelVella:
Hi, just wanted to inform you that I have this issue too (with twig 1.24.0 and assetic 1.3.1).

razorness:
Same issue:

* twig/twig - v1.24.1
* symfony/assetic-bundle - v2.8.0

stof:
@razorness please open an issue instead of commenting on a merged PR (merged PRs are not bug reports, and do not appear in the list of open issues). And it would be great to provide a reproducing case (this can be done as a small project based on the symfony standard edition with the minimal amount of code needed to trigger it, in case it only happens with AsseticBundle).

PR #780, opened 2012-07-19
Inline optimization

mcuelenaere:
This optimization executes filters or functions at build time when possible.

Please verify the second commit whether these are allowed to be inlined. I've left out the obvious ones (date, random) but I'm not sure all of these can be inlined.

### DIFF
--- a/lib/Twig/Extension/Core.php
+++ b/lib/Twig/Extension/Core.php
@@ -149,9 +149,9 @@ class Twig_Extension_Core extends Twig_Extension
             'merge'   => new Twig_Filter_Function('twig_array_merge', array('is_consistent' => true)),
 
             // string/array filters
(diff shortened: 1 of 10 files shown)


acasademont:
Hi all, I reworked this patch but the gains are negligible, inlining constants doesn't make much sense. Right now what the patch does is tranforming `"foo"|upper` directly to "FOO". The thing is that almost nobody does this, you would already write "FOO" directly in your template.

That being said, it would definitely provide a nice gain if we could inline translations and routing. Inlining translations and routing  would require to have different cached templates for each supported language but that's known at compile time. It would be a nice addition to Twig 2, hope to find some time to work on a POC.

fabpot:
Closing this old PR.

PR #934, opened 2012-12-13
[RFC] Support foo.some_thing === foo.someThing

simensen:
I want to be able to support all underscores in my Twig templates. Most of my data in one of my projects comes from YAML files that are usually formatted `like_this` so it is a little jarring when in my template I need to mix with `someMethod` on an object:

    paginated_items.theFirstItem.some_description

This PR allows for mapping `the_first_item` to `theFirstItem`.

    paginated_items.the_first_item.some_description

I started a simple implementation of this. Happy to change it if there is a better way to go about it. I understand speed is a factor here so ideas on how I can minimize impact would be helpful.

The testing is pretty complicated. I did the bes tI could to try and figure out how to test this. Suggestions on how I can do that better would be helpful as well.

### DIFF
--- a/lib/Twig/Template.php
+++ b/lib/Twig/Template.php
@@ -405,12 +405,20 @@ abstract class Twig_Template implements Twig_TemplateInterface
         }
 
         $lcItem = strtolower($item);
+        $deunderscoreItem = str_replace('_', '', $lcItem);
+        $checkDecamelItem = $lcItem !== $deunderscoreItem;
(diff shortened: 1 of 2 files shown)


RyanThompson:
Looks like the PR about there - would LOVE to see this merged in. @vicb You are right PSR is not mandatory but standardized conventions make stuff like this a lot easier and reliable. We can assume the conventions we want to cover instead of a whole bag of them by adhering to standard rules ^_^

fabpot:
Closing in favor of #1299

Issue #935, opened 2012-12-13
xor operator


shieldo:
Is there a reason there is no xor logical operator in Twig? I'd submit a PR but I'd like to know first whether the feeling is that this would seem like a valuable addition (for the relatively rare cases it applies to) or something that could be considered a step too far for a templating language. There's already bitwise xor, which suggests the former.

PR #972, opened 2013-01-22
Added support for xor operator

kpobococ:
This should close issue #935


denisvmedia:
Will this be ever merged?

fabpot:
Still not convinced that it should be in core, so closing.

Issue #1003, opened 2013-02-20
Consider introducing the inline tag to remove line breaks


marcospassos:
The `{ % spaceless %}` tag is so useful, but sometimes we need to strip all line breaks. So, my suggestion is introducing a new `{% inline %}` tag that does this job.

Use cases:

``` html
<div id="product-description">
{# Strips <br/> tags #}
{% inline 'html' %}
{{ product.description>
{% endinline %}
</div>
```

``` html
{% inline %}
<table
  for="for" 
  javascript="javascript" 
  frameworks="framework" 
  like="like"
  angular="angular"
  knockoutjs="knockoutjs"
  etc="etc"
>
  <thead>
...
</table>
{% endinline %}
```

``` html
{# cache.js #}
{% inline %}
cache['template/home.html'] = "
{% autoescape 'js' %}
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01//EN"
   "http://www.w3.org/TR/html4/strict.dtd">
<HTML>
   <HEAD>
      <TITLE>My first HTML document</TITLE>
   </HEAD>
   <BODY>
      <P>Hello world!
   </BODY>
</HTML>
{% endautoescape %}
";
{% endinline %}
```

PR #1018, opened 2013-02-27
Added "normalize_whitespace" filter

Koc:
Fixed tickets: #1003

Use case:

``` jinja
{# <title> generation #}
{% block title %}
    {%- filter normalize_whitespace %}
        {% if foo %}
            {{ foo.title }}
            {% if bar %}
                / {{ bar.title }}
            {% endif %}
            /
        {% endif %}

        {{ baz }} / {{ parent() }}
    {% endfilter -%}
{% endblock %}
```

TODO:
- [x] add documentation if this PR can be accepted
- [ ] consider adding "spaceless" filter (IMHO actually it is filter, not tag)

### DIFF
--- a/doc/filters/normalize_whitespace.rst
+++ b/doc/filters/normalize_whitespace.rst
@@ -1,0 +1,27 @@
+``normalize_whitespace``
+========
+
+.. versionadded:: 1.25.1
+    The normalize_whitespace filter was added in Twig 1.25.1.
+
+The ``normalize_whitespace`` filter replaces duplicated spaces and/or linebreaks with single space. It also remove whitespace from the beginning
+and end of a string:
+
+.. code-block:: jinja
+
+    {# <title> generation #}
+    {% block title %}
+        {%- filter normalize_whitespace %}
+            {% if foo %}
+                {{ foo.title }}
+                {% if bar %}
+                    / {{ bar.title }}
+                {% endif %}
+                /
+            {% endif %}
+
+            {{ baz }} / {{ parent() }}
+        {% endfilter -%}
+    {% endblock %}
+
+    {# outputs something like 'foo.title / bar.title / baz / parent' #}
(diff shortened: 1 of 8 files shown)


lib/Twig/Extension/Core.php
```diff
@@ -575,6 +576,11 @@ function twig_jsonencode_filter($value, $options = 0)
     }
 }
 
+function twig_normalize_whitespace_filter($value)
+{
+    return preg_replace('/([\s]+)/', ' ', $value);
```
stof: `/\s+/` is enough

doc/filters/normalize_whitespace.rst
```diff
@@ -0,0 +1,27 @@
+``normalize_whitespace``
+========
+
+.. versionadded:: 1.12.3
+    The normalize_whitespace filter was added in Twig 1.12.3.
```
stof: this should be changed

lib/Twig/Extension/Core.php
```diff
@@ -146,6 +146,7 @@ public function getFilters()
             new Twig_SimpleFilter('striptags', 'strip_tags'),
             new Twig_SimpleFilter('trim', 'trim'),
             new Twig_SimpleFilter('nl2br', 'nl2br', array('pre_escape' => 'html', 'is_safe' => array('html'))),
+            new Twig_SimpleFilter('normalize_whitespace', 'twig_normalize_whitespace_filter', array('is_safe' => array('html'))),
```
stof: this is safe for HTML only if the input is safe. Otherwise, it is a hole in the escaping, as the filter is not escaping anything.

So the right option is not `is_safe` but [`preserves_safety`](https://github.com/fabpot/Twig/blob/master/lib/Twig/SimpleFilter.php#L34)

petemolinero:
I'd love to see this added as well.

stof:
the doc for the new ``spaceless`` filter is missing

Koc:
@stof yes, I know. Tests are also missing. Are you not against of adding it? Should I deprecate spaceless tag?

stof:
Well, it could indeed be deprecated in favor of the (not widely known) filter tag (e.g. ``{% filter spaceless %}``. but we should get @fabpot's mind on that.

fabpot:
The `spaceless` tag has recently been deprecated and replaced with a `spaceless` filter, so that part of the PR is done now. Regarding the new filter, I'm not sure it's something we want in core.

Koc:
@fabpot can it be part of text extension https://twig-extensions.readthedocs.io/en/latest/text.html ?

Issue #719, opened 2012-05-08
Add a "with" tag


fabpot:
see http://jinja.pocoo.org/docs/templates/#with-statement

and http://groups.google.com/group/twig-users/browse_thread/thread/140ec863930f4fe1

PR #1054, opened 2013-04-12
Add {% with %} tag

xphere:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | no |
| Deprecations? | no |
| Tests pass? | yes |
| Fixed tickets | #719 |
| License | MIT |
| Doc PR | - |

Jinja flavoured `{% with %}` tag implementation.
PR related to issue https://github.com/fabpot/Twig/issues/719

### DIFF
--- a/doc/tags/index.rst
+++ b/doc/tags/index.rst
@@ -22,3 +22,4 @@ Tags
     spaceless
     use
     verbatim
+    with
--- a/doc/tags/with.rst
+++ b/doc/tags/with.rst
@@ -1,0 +1,26 @@
+``with``
+========
+
+Enforces a nested scope where you can safely set variables without bloating the main context and avoiding potential name collisions.
+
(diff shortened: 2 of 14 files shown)


stof:
@hason the first syntax is totally useless though

fabpot:
The first syntax is useful as it marks the beginning of a new scope and variables defined inside won't survive the end of the with tag:

```twig
{% with %}
  {% set foo = 'bar' %}
{% endwith %}
{# foo is not defined here #}
```

stof:
hmm indeed. I missed that. And 👍 for that. I have use cases for that (I have a few dummy blocks in my work projects defined only to isolate some scope in the base layout).

fabpot:
Closing in favor of #2246, where we only support the same logic as the `include` tag as proposed by @hason

Issue #861, opened 2012-10-15
Consider auto-importing macros in the same file


schmittjoh:
I've always found it a bit weird that you cannot directly call macros defined in the same file.

Having to "import" something that is in the technically in the same file feels a bit weird (especially as you import on a per-file basis).

Can't we simply auto-import them? Obviously, if you define them in the same file, you probably want to use them there.

chetzof:
Did the changes in #1215 mature enough? :)

fabpot:
Implemented in #3019.

Issue #867, opened 2012-10-21
Calling a macro from within a macro in the same file fails


djc:
It looks like this doesn't work. I guess it might be related to #861.

lib.html:

```
{% import "lib.html" as lib %}

{% macro inner() %}
    FOO
{% endmacro %}

{% macro main() %}
    {{ lib.inner() }}
{% endmacro %}
```

base.html:

```
{% import "lib.html" as lib %}
{{ lib.main() }}
```

Setup:

```
<?php

require_once('/usr/share/php/Twig/Autoloader.php');
Twig_Autoloader::register();
$loader = new Twig_Loader_Filesystem('test');
$twig = new Twig_Environment($loader, array());
$tpl = $twig->loadTemplate("base.html");

header('Content-Type: text/html; charset=utf-8');
print($tpl->render(array()));

?>
```

PR #1085, opened 2013-05-11
Added support for directly call macros defined in the same template

hason:
Fixes: #861, #867
Depends on #933


matthew-dean:
So calling macros from the same file never got merged in?

SpacePossum:
>So calling macros from the same file never got merged in?

It was but never released as stable before it was reverted, more background on this can be found here https://github.com/twigphp/Twig/issues/861

PR #1197, opened 2013-09-22
Treat object as string when used as array offset

ob-ivan:
Currently if you call `{{ attribute(object, item) }}` where `object` argument is a true array (not an object acting as array) and `item` is an object, an error will be raised: "array_key_exists(): The first argument should be either a string or an integer".

In contrast, if both `object` and `item` are objects, then `item` gets converted to string and everything goes smoothly.

This PR eliminates this incosistency. Now objects that are string-convertible will be converted to strings when passed as `item` argument.


Tobion:
:-1: for the reasons also stated by stof

PR #1299, opened 2013-12-19
Added support for foo.some_thing === foo.someThing

hason:
Alternative implementation of #934. Works well for methods, properties and array calls of objects with minimal performance impact. Results of twig benchmark ([fabpot/twig-perf](https://github.com/fabpot/twig-perf)):

```
                               |       v1.12.3 |       v1.16.3 | origin/master | origin/methods | 
------------------------------------------------------------------------------------------------
empty                          |     13.5  0.0 |      8.8  0.0 |     14.4  0.0 |     20.5  0.0 | 
empty/B                        |      7.9  0.0 |     14.1  0.0 |     14.6  0.0 |     18.9  0.0 | 
simple_attribute               |     88.2  0.9 |     99.8  0.9 |     99.2  0.8 |     99.3  0.9 | 
simple_array_access            |    180.2  1.3 |    132.8  0.9 |    120.6  0.9 |    127.1  1.0 | 
simple_method_access           |     94.4  2.4 |     92.9  2.7 |    101.7  2.7 |    119.7  2.9 | 
simple_attribute_big_context/B |     90.5  0.9 |    101.3  0.9 |    100.5  1.0 |    116.1  1.3 | 
simple_variable                |    710.0  0.7 |    698.6  0.7 |    711.9  0.8 |    680.2  0.7 | 
simple_variable_big_context/B  |    702.2  1.3 |    668.3  1.3 |    795.9  1.2 |    674.3  1.2 | 
simple_foreach                 |     10.9  0.0 |     21.0  0.0 |     15.2  0.0 |     10.7  0.0 | 
simple_foreach/B               |     10.1  5.7 |     16.3  7.7 |      9.9  5.0 |     14.5  4.7 | 
empty_extends                  |      9.5  0.0 |     17.4  0.0 |     17.6  0.0 |     13.9  0.0 | 
empty_extends/B                |     16.1  0.0 |     17.5  0.0 |     10.2  0.0 |     22.6  0.0 | 
empty_include                  |     10.3  0.0 |     10.0  0.0 |     12.1  0.0 |      9.5  0.0 | 
empty_include/B                |     14.8  0.0 |     14.5  0.0 |     15.4  0.0 |     14.4  0.0 | 
standard                       |     22.1  0.0 |     12.6  0.0 |     12.9  0.0 |     14.0  0.0 | 
escaping                       |    101.8  1.2 |    110.9  1.2 |    107.5  0.8 |     99.2  1.0 |
```

### DIFF
--- a/lib/Twig/Template.php
+++ b/lib/Twig/Template.php
@@ -504,7 +504,7 @@ abstract class Twig_Template implements Twig_TemplateInterface
         }
 
         $cache = array_combine($methods, $methods);
-        $keys = array_merge(preg_replace('/^(?:get|is)(.++)$/i', '\\1', $methods), $methods);
+        $keys = array_merge(preg_replace('/^(?:get|is)_?(.++)$/i', '\\1', $methods), $methods);
         $keys = array_merge(preg_replace('/((?<=[a-z]|\d)[A-Z]|(?<!^)[A-Z](?=[a-z]))/', '_\\1', $keys), $keys);
 
         return array('methods' => $cache + array_change_key_case(array_combine($keys, array_merge($methods, $methods, $methods, $methods))));
--- a/lib/Twig/Template.php
+++ b/lib/Twig/Template.php
@@ -504,8 +504,9 @@ abstract class Twig_Template implements Twig_TemplateInterface
         }
 
         $cache = array_combine($methods, $methods);
-        $keys = preg_replace('/^(?:get|is)(.++)$/i', '\\1', $methods);
+        $keys = array_merge(preg_replace('/^(?:get|is)(.++)$/i', '\\1', $methods), $methods);
+        $keys = array_merge(preg_replace('/((?<=[a-z]|\d)[A-Z]|(?<!^)[A-Z](?=[a-z]))/', '_\\1', $keys), $keys);
 
-        return array('methods' => $cache + array_change_key_case($cache + array_combine($keys, $methods)));
+        return array('methods' => $cache + array_change_key_case(array_combine($keys, array_merge($methods, $methods, $methods, $methods))));
     }
 }
--- a/lib/Twig/Template.php
+++ b/lib/Twig/Template.php
@@ -361,13 +361,32 @@ abstract class Twig_Template implements Twig_TemplateInterface
      */
     protected function getAttribute($object, $item, array $arguments = array(), $type = Twig_Template::ANY_CALL, $isDefinedTest = false, $ignoreStrictCheck = false)
     {
+        if (is_object($object)) {
+            $class = get_class($object);
+            if (!isset(self::$cache[$class])) {
+                self::$cache[$class] = $this->getCacheForClass($class);
+            }
+        }
+
         // array
         if (Twig_Template::METHOD_CALL !== $type) {
             $arrayItem = is_bool($item) || is_float($item) ? (int) $item : $item;
+            $hasArrayItem = false;
 
-            if ((is_array($object) && array_key_exists($arrayItem, $object))
-                || ($object instanceof ArrayAccess && isset($object[$arrayItem]))
-            ) {
+            if (is_array($object) && array_key_exists($arrayItem, $object)) {
+                $hasArrayItem = true;
+            } elseif ($object instanceof ArrayAccess) {
+                if (isset($object[$arrayItem])) {
+                    $hasArrayItem = true;
+                } elseif (isset(self::$cache[$class]['properties'][$arrayItem])
+                    && isset($object[self::$cache[$class]['properties'][$arrayItem]])
+                ) {
+                    $arrayItem = self::$cache[$class]['properties'][$arrayItem];
+                    $hasArrayItem = true;
+                }
+            }
+
+            if ($hasArrayItem) {
                 if ($isDefinedTest) {
                     return true;
                 }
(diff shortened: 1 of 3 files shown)


lib/Twig/Template.php
```diff
@@ -348,13 +348,31 @@ protected function displayWithErrorHandling(array $context, array $blocks = arra
...
+                    $hasArrayItem = true;
+                } elseif (isset(self::$cache[$class]['properties'][$arrayItem])
+                    && isset($object->{self::$cache[$class]['properties'][$arrayItem]})
+                ) {
+                    $arrayItem = self::$cache[$class]['properties'][$arrayItem];
+                    $hasArrayItem = true;
+                }
+            }
+
+            if (isset($hasArrayItem)) {
```
stof: instead of a `isset` on a boolean variable, I would rather set it to false initially

lib/Twig/Template.php
```diff
@@ -348,13 +348,31 @@ protected function displayWithErrorHandling(array $context, array $blocks = arra
...
-            if ((is_array($object) && array_key_exists($arrayItem, $object))
-                || ($object instanceof ArrayAccess && isset($object[$arrayItem]))
-            ) {
+            if (is_array($object) && array_key_exists($arrayItem, $object)) {
+                $hasArrayItem = true;
+            } elseif ($object instanceof ArrayAccess) {
+                if (isset($object[$arrayItem])) {
+                    $hasArrayItem = true;
+                } elseif (isset(self::$cache[$class]['properties'][$arrayItem])
+                    && isset($object->{self::$cache[$class]['properties'][$arrayItem]})
```
stof: This looks weird to me; It is not testing the array item here

lib/Twig/Template.php
```diff
@@ -395,36 +413,35 @@ protected function getAttribute($object, $item, array $arguments = array(), $typ
...
         if (Twig_Template::METHOD_CALL !== $type) {
             if (isset($object->$item) || array_key_exists((string) $item, $object)) {
+                $property = $item;
+            } elseif (isset(self::$cache[$class]['properties'][$item])
+                && isset($object->{self::$cache[$class]['properties'][$item]})
+            ) {
+                $property = self::$cache[$class]['properties'][$item];
+            }
+
+            if (isset($property)) {
```
stof: I would also avoid `isset` here. Relying on the fact that the variable was not set at the beginning of the block is rather fragile in case of changes in the previous code

lib/Twig/Template.php
```diff
@@ -469,6 +486,45 @@ protected function getAttribute($object, $item, array $arguments = array(), $typ
...
+            $keys = array_merge(preg_replace('/^(?:get|is)_?(.++)$/i', '\\1', $methods), $methods);
+            $keys = array_merge(preg_replace('/((?<=[a-z]|\d)[A-Z]|(?<!^)[A-Z](?=[a-z]))/', '_\\1', $keys), $keys);
+            $cache['methods'] += array_change_key_case(array_combine($keys, array_merge($methods, $methods, $methods, $methods)));
+        }
+
+        $properties = array_keys(get_class_vars($class));
+        $implements = class_implements($class, false);
+        if (isset($implements['ArrayAccess']) || (isset($cache['methods']['__isset']) && isset($cache['methods']['__get']))) {
+            $reflection = new ReflectionClass($class);
+            foreach ($reflection->getProperties(ReflectionProperty::IS_PROTECTED) as $property) {
```
stof, line 540: why protected and not private if your goal is to try to guess what will be allowed for `__get` and array access ?

And this seems weird to be, because the implementation of ArrayAccess or magic method can do things in a totally different way than accessing the protected properties (sotring the data in a single private property containing an array for instance, as done in Pimple)

lib/Twig/Template.php
hason: Fixed, good catch!

lib/Twig/Template.php
hason, line 540: I want to support very usual use case:

``` php
/**
 * @MappedSuperclass
 */
abstract class BaseEntity {
    /** @Id @Column(type = "integer") @GeneratedValue */
    protected $id;

    public function __get($property) {
        return $this->$property;
    }
    ....
}

/**
 * @Entity
 */
class Post extends BaseEntity {
    /** @Column(length = 80) */
    protected $name;

    /** @Column(type = "datetime") */
    protected $createdAt;

    /** @ManyToMany(targetEntity = "User", fetch = "EAGER") */
    protected $authors;

    ...
}
```

hason:
@fabpot ping

This PR does not hurt performance https://github.com/twigphp/Twig/pull/934#issuecomment-13505353 and works with properties and arrays https://github.com/twigphp/Twig/pull/934#issuecomment-13523606

composer.json
```diff
@@ -27,7 +27,8 @@
         "forum": "https://groups.google.com/forum/#!forum/twig-users"
     },
     "require": {
-        "php": ">=5.2.4"
+        "php": ">=5.2.4",
+        "ext-pcre": "*"
```
fabpot: I would remove this one.

fabpot:
@tucksaun @jpauli Can you have a look at the C part of this PR?

lib/Twig/Template.php
```diff
@@ -361,13 +361,32 @@ protected function displayWithErrorHandling(array $context, array $blocks = arra
      */
     protected function getAttribute($object, $item, array $arguments = array(), $type = Twig_Template::ANY_CALL, $isDefinedTest = false, $ignoreStrictCheck = false)
     {
+        if (is_object($object)) {
+            $class = get_class($object);
+            if (!isset(self::$cache[$class])) {
+                self::$cache[$class] = $this->getCacheForClass($class);
```
fabpot: for the non-debug mode, what about keeping this cache on disk? Would that makes things faster? It looks like there is a lot going on in the `getCacheForClass()` method and I cannot believe that it does not have any impacts on real-world templates.

hason: Support for the cache may be added in another PR.

lsmith77:
can we give this PR some life again?

elnur:
😞

fabpot:

fabpot:

fabpot:
This one never moved forward as we did not get feedback on the C implementation. I propose to rebase it on 2.x where we don't have the C extension.

hason:
@fabpot The C extension in 1.x is no longer used. Should I update this PR on 1.x or 2.x?

xabbuh:
@hason The C extension is gone.

PR #1309, opened 2014-01-02
Fixed fatal error for unknown macro

### LINKED ISSUE #1052: Acessing globally imported macro in another macro fails with PHP fatal error
**Twig 1.12.3 (2013-04-08)**

Assume we have this template:

``` twig
{% import "macros/clipart.twig" as clm %}

{% macro show(something) %}
    {{ unknown_variable.moody }}
    {{ clm.moody }}
{% endmacro %}
```

The docs say, that macros have no access to global scope. That's fine. But they still track imported macroses!

Line `{{ unknown_variable.moody }}` runs without errors producing an empty strings (with strict variables disabled)

The other line `{{ clm.moody }}` fails with fatal error **Call to a member function getmoody() on a non-object**

As far as I could dig into the sources, I found that in _ExperssionParser.php:368_ uses `$this->parser->getImportedSymbol` to look for `clm` variable, that uses `Parser::importedSymbols` which is global. But `clm` isn't included in the `$context` of the generated template method. And `strict_variables` don't help here.

And in generated template method we have `echo $_clm_->moody();` (and `$_clm_` is `null` since it doesn't exists in `$context`) with no sanity checks.

Solutions:
- make imported modules visible within macros context (against the docs)
- don't recognize them as modules within macros (according to the docs)

hason:
Fixes #1052

### DIFF
--- a/lib/Twig/Node/Expression/MacroCall.php
+++ b/lib/Twig/Node/Expression/MacroCall.php
@@ -1,0 +1,36 @@
+<?php
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) 2012 Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+/**
+ * Represents a macro call node.
+ *
+ * @author Martin Hasoň <martin.hason@gmail.com>
+ */
+class Twig_Node_Expression_MacroCall extends Twig_Node_Expression
+{
+    public function __construct(Twig_Node_Expression $template, $name, Twig_Node_Expression_Array $arguments, $lineno)
+    {
+        parent::__construct(array('template' => $template, 'arguments' => $arguments), array('name' => $name), $lineno);
+    }
+
+    public function compile(Twig_Compiler $compiler)
+    {
+        $compiler
+            ->raw('$this->callMacro(')
+            ->subcompile($this->getNode('template'))
+            ->raw(', ')
+            ->repr($this->getAttribute('name'))
+            ->raw(', ')
+            ->subcompile($this->getNode('arguments'))
+            ->raw(')')
+        ;
+    }
+}
--- a/lib/Twig/Template.php
+++ b/lib/Twig/Template.php
@@ -24,6 +24,7 @@ abstract class Twig_Template implements Twig_TemplateInterface
     protected $env;
     protected $blocks;
     protected $traits;
+    protected $macros;
 
     /**
      * Constructor.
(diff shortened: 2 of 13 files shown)


lib/Twig/Template.php
```diff
@@ -476,6 +478,33 @@ protected function getAttribute($object, $item, array $arguments = array(), $typ
...
+     * @throws Twig_Error_Runtime if the macro is not defined
+     */
+    final protected function callMacro(Twig_Template $template, $macro, array $arguments)
+    {
+        if (!isset($template->macros[$macro]['reflection'])) {
+            if (!isset($template->macros[$macro])) {
+                throw new Twig_Error_Runtime(sprintf('Macro "%s" is not defined in the template "%s".', $macro, $template->getTemplateName()));
+            }
+
+            $template->macros[$macro]['reflection'] = new ReflectionMethod($template, $template->macros[$macro]['method']);
```
sstok: Reflection is always slower, so I think we should benchmark this.
A macro definition does not change after its compiled so keeping this optimized is a better option.

stof: I think using the followign code would be better:

``` php
call_user_func_array(array($template, $template->macros[$macro]['method']), $arguments)
```

hason: Benchmark http://3v4l.org/W5EEu Reflection is used for the possible support of named arguments.

stof: your benchmark is not relevant. The expensive part of using reflection is the instantiation, not the usage. And PHP caches the data when asking for reflection for the same object again, meaning it does the expensive work only once.
In a real usage of Twig, you will probably get reflection for different macros

hason: @stof Benchmark for 7 macros http://3v4l.org/j5F2j

lib/Twig/Node/Expression/MacroCall.php
```diff
@@ -0,0 +1,36 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+/**
+ * Represents a macro call node.
+ *
+ * @author Martin Hasoň <martin.hason@gmail.com>
+ */
+class Twig_Node_Expression_MacroCall extends Twig_Node_Expression
```
stof: MethodCall is already used only for macros currently. no need for a need node

fabpot:

fabpot:

fabpot:
Closing as it is now fixed with the recent changes done in Twig 2.x.

Issue #1159, opened 2013-08-06
Add support for unconsumed arguments (varargs and kwargs) in macros


hason:
http://jinja.pocoo.org/docs/templates/#macros

Better suggestion inspired by python:

```jinja
{% macro tag(name, **attributes) %}
<{{ name }}
{% for attribute, value in attributtes %}
  {{ attribute }}="{{ value|e('html_attr') }}"
{% endfor %}
>
{% endmacro %}

{{ tag('a', href="http://example.org", title="Example") }}
```

PR #1317, opened 2014-01-16
Added variadic filters, tests, and functions

hason:
Similar to #1159

```php
$env->addFunction(new Twig_SimpleFunction('foo', function ($a1, array $args = array()) {
}, array('is_variadic' => true));
```
```jinja
{{ foo(1, 2, a="a", b="b") }}
{# foo(1, array(0 => 2, "a" => "a", "b" => "b")); #}
```

### DIFF
--- a/doc/advanced.rst
+++ b/doc/advanced.rst
@@ -224,6 +224,23 @@ through your filter::
 
     $filter = new Twig_SimpleFilter('somefilter', 'somefilter', array('pre_escape' => 'html', 'is_safe' => array('html')));
 
+Variadic Filters
+~~~~~~~~~~~~~~~~
+
+.. versionadded:: 1.19
+    Support for variadic filters was added in Twig 1.19.
+
+If you want to pass a variable number of positional or named arguments to the filter,
+set the ``is_variadic`` option to ``true``; Twig will pass the array of arbitrary arguments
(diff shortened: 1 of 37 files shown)


fabpot:
I would add only one argument as there is no difference between maps/sequences/hashes in PHP. Also, do we really need to add a configuration setting for that?

hason:
@fabpot PR is updated

doc/advanced.rst
```diff
@@ -224,6 +224,21 @@ through your filter::
...
+
+If you want to pass a variable number of positional or named arguments to the filter,
+set the ``is_variadic`` option to ``true``; Twig will pass the array of arbitrary arguments
+as the last argument to the filter call that is defined as an array with an empty default value::
+
+    $filter = new Twig_SimpleFilter('thumbnail', function ($file, array $options = array()) {
+        ...
+    }, array('is_variadic' => true));
+
+
```
fabpot: I suppose that this should also be documented for Functions and Tests.

You should also probably mention that the downside is that we cannot check anymore if the named arguments passed to the filter are valid or not as if they are not valid, they will end up in the option array.

hason:
@fabpot The documentation is enhanced. Should I open a new PR for 1.x branch?

hason:
Closing in favor of #1699

PR #1333, opened 2014-02-07
Added disambiguation on using raw in expressions

OwlyCode:
I recently got in trouble using the `raw` filter in a ternary, leading to a value being escaped and me not expecting it. The code was : 

```
{{ foo|striptags|length > 250 ? foo|striptags|slice(0, 250) ~ '...' : foo|raw }}
```

When foo's length was under 250 characters, the result of this expression was escaped. Here is the updated documentation to explain why.


fabpot:
Will finally be fixed by #2934

Issue #1415, opened 2014-05-29
Separate cache class...yes, again!


andrewtch:
As for me, I see no need for memcache / redis / whatsoever cache in real life, yet I've stumbled upon a real-life situation where it is necessary:
- twig is used for a complex (8-10 levels of inheritance) system of landings
- landings are deployed automatically on read-only file system (no file cache)
- it is not allowed to have any directories writable by web server (since landings include partner scripts, this is a security measure)
- 'warming up' cache breaks continious deployment as it is now
- parsing templates takes considerable amount of cpu time

I can PR in a following way:
- add Twig_Cache_Interface
- add Twig_Cache_Filesystem mimicing current logic
-  alter Twig_Environment -> setCache call to automatically instantiate Twig_Cache_Filesystem($cache) if $cache is not instanceof Twig_Cache_Interface

that's all. Simply one cache, minor changes in loadTemplate, no cache - dependent implementations (We leave Twig_Cache_Whatsoever to user having such issues as this code should not be in a core cos mostly opcode cachers perform faster).

tl;dr: I want to refactor twig cache to support custom cache drivers, but not because of some heretic reasons like "memcache is faster than ZendOpcode", but for those users who have read-only filesystem but still want to save time on template parsing (like me).

PR can be done in some hours )

PR #1421, opened 2014-06-10
Implemented filesystem cache, one and only (see #1415)

andrewtch:


guilhermeaiolfi:
So, what's the status on this?

andrewtch:
I'm still wating for @fabpot to answer.

F21:
It this still being targeted for 2.0?

andrewtch:
I can try to update it if anyone is still interested.

F21:
@andrewtch Would love to see it happen. I am planning to build a twig rendering microservice that will render templates uploaded by users. The service will be deployed to a docker container and will be immutable. It makes sense to be able to place the template cache somewhere else so that we don't have to warm the cache during every deploy.

stof:
@F21 Put the cache in a folder which is not cleared between deployments (but take care to clear it when you deploy a Twig update as the cached templates are not meant to be compatible with any Twig version but only with the version generating them, so some Twig updates may not allow reusing the same cached template)

fabpot:
closing in favor of #1812 which is built on top of this PR.

PR #1433, opened 2014-07-05
Added support for custom variables in block function

### LINKED ISSUE #1302: Pass variables to block function
Please add ability to pass variables to `block` function. 

For example this block could be rewritten

```
{% block integer_widget %}
{% spaceless %}
    {% set type = type|default('number') %}
    {{ block('form_widget_simple') }}
{% endspaceless %}
{% endblock integer_widget %}
```

to

```
{% block integer_widget %}
{% spaceless %}
    {{ block('form_widget_simple', {type: type|default('number')}) }}
{% endspaceless %}
{% endblock integer_widget %}
```

and this:

```
{% block tree_node %}
<ul>
    {# we need to save original root variable #}
    {% set oldRoot = root %}
    {% for node in root.children  %}
        {{ node.title }}
        {% set root = node %}
        {{ block('tree_node') }}
    {% endfor %}
    <!-- {{ oldRoot.children|length }} -->
</ul>
{% endblock %}
```

to

```
{% block tree_node %}
<ul>
    {% for node in root.children  %}
        {{ node.title }}
        {{ block('tree_node', {root: node}) }}
    {% endfor %}
    <!-- {{ root.children|length }} -->
</ul>
{% endblock %}
```

hason:
If you must use this workaround:

``` jinja
{# blocks.twig #}
{% block foo %}
{{ var }}
{% endblock %}

{# page.twig #}
{% use "blocks.twig" %}
{% set oldvar = var %}
{% set var = "newvalue" %}
{{ block("foo") }}
{% set var = oldvar %}
```

After this PR everything will be easier:

``` jinja
{# page.twig #}
{% use "blocks.twig" %}
{{ block("foo", {"var": "newvalue"}) }}
```

Fixes #1302 

TODO:
- [x] Add same functionality for `parent()`
- [x] Documentation

### DIFF
--- a/doc/functions/block.rst
+++ b/doc/functions/block.rst
@@ -1,6 +1,9 @@
 ``block``
 =========
 
+.. versionadded:: 1.28
+    Ability to pass variables to block function was added in Twig 1.28.
+
 When a template uses inheritance and if you want to print a block multiple
 times, use the ``block`` function:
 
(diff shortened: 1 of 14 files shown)


gregquat:
Any news on this?

HeahDude:
For what it worth, I've made a global simple twig function to autoload macros from anywhere (views, tags, blocks) at compile time, you can find my public gist [here](https://gist.github.com/HeahDude/f9aeb128fcac309dddc8) any comments or suggestions are welcome.

mpdude:
Also would like to see this or hear what the recommended superior alternative is.

Additionally, `use` suffers from the same problem.

qkdreyer:
+1

hason:
@fabpot What do you think? So far nine votes :+1:.

caponica:
@hason - I'm not sure @fabpot will change his mind, but I would really like to know the "correct" way to do this!

javiereguiluz:
@hason if this feature is accepted, the `{% with %}` tag seems the most natural solution to me. Moreover, the original Jinja templates support the [with statement](http://jinja.pocoo.org/docs/dev/templates/#with-statement) and there is even a pending PR to add them to Twig: #1054.

fabpot:
Closing this PR as I prefer to support the `with` tag instead.

PR #1435, opened 2014-07-07
Improved named arguments

hason:


hason:
@fabpot ping

fabpot:
The second commit is fine by me and I can cherry-pick it, but I still don't understand the utility of the first one.

lib/Twig/Node/Expression/Call.php
```diff
@@ -139,34 +142,49 @@ protected function getArguments($callable, $arguments)
...
+        $optionalArguments = array();
         $pos = 0;
         foreach ($definition as $param) {
             $names[] = $name = $this->normalizeName($param->name);
 
             if (array_key_exists($name, $parameters)) {
                 if (array_key_exists($pos, $parameters)) {
-                    throw new Twig_Error_Syntax(sprintf('Argument "%s" is defined twice for %s "%s".', $name, $this->getAttribute('type'), $this->getAttribute('name')));
+                    throw new Twig_Error_Syntax(sprintf('Argument "%s" is defined twice for %s "%s".', $name, $callType, $callName));
+                } elseif (!empty($missingArguments)) {
```
stof: just ``if`` here. No need for ``elseif``

hason:
The first commit solves an example from documentation http://twig.sensiolabs.org/doc/advanced.html#filters

We register custom filter as a simple PHP function
```php
$filter = new Twig_SimpleFilter('compare', 'substr_compare');
```
A simple example works
```twig
{{ "foobar"|compare("bar", 3) }} {# output is 0 #}
```

If I set a second optional argument (note: in PHP documentation is a mistake ``case_sensitivity`` vs ``case_insensitivity``)

```twig
{{ "foobar"|compare("bar", 3, case_sensitivity=false) }}
```

Twig throws an exception with wrong message:

```
Twig_Error_Syntax: Unknown argument "case_sensitivity" for filter "compare(main_str, str, offset, length)".
```

This PR suggests a message:

```
Twig_Error_Syntax: Argument "case_sensitivity" could not be assigned for filter "compare(main_str, str, offset, length, case_sensitivity)" because it is mapped to internal PHP function which cannot determine default value for optional argument "length".
```

fabpot:
Thank you @hason.

PR #1436, opened 2014-07-07
Added a decorator class for Twig_Node_Module

### LINKED ISSUE #1402: There's no easy way to decorate Twig_Node_Module
To show the issue, I've create [a branch with a failing test](https://github.com/rybakit/Twig/tree/module_node), which fails because `Twig_Node_SandboxedModule` ignores custom module instance. It would be nice to have a way to decorate a module node instead of replacing it to fix that.

hason:
Fixes #1402

### DIFF
--- a/lib/Twig/Node/ModuleDecorator.php
+++ b/lib/Twig/Node/ModuleDecorator.php
@@ -1,0 +1,174 @@
+<?php
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) 2009-2014 Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+/**
+ * The decorator class for module node.
+ *
+ * @author Martin Hasoň <martin.hason@gmail.com>
+ */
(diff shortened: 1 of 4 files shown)


stof:
@fabpot ping

lib/Twig/Node/ModuleDecorator.php
```diff
@@ -0,0 +1,174 @@
...
+     * @param array  $arguments The parameters to be passed to the method
+     *
+     * @return mixed
+     *
+     * @throws BadMethodCallException
+     */
+    public function __call($method, array $args)
+    {
+        if (!method_exists($this->module, $method) && !method_exists($this->module, '__call')) {
+            throw new \BadMethodCallException(sprintf('Method "%s::%s()" does not exist.', get_class($this->module), $method));
```
fabpot: You need to remove the `\` to keep BC with PHP 5.2.

fabpot:
Even if the idea is good, I don't like the implementation as it exposes the internals of the Module class. I will see if I have a better idea but what I've done in the Profiler (adding empty nodes you can override) seems a better approach to me.

fabpot:
Closing in favor of #1603

PR #1440, opened 2014-07-10
[2.0] Added support for floating-point numbers for "%" and "divisible by" operators

### LINKED ISSUE #1353: % operator returns the integer result
Apologies if this has already been mentioned - I wasn't sure exactly what to search for.

Currently, `{{ 1.5 % 1 }}` will display 0 instead of 0.5 because it is a straight proxy for PHP's `%` operator instead of `fmod`.

This might be expected behaviour for a PHP developer but I personally believe that the majority of template designers would expect the latter.

This issue also affects `divisibleby`, more noticeably, with `{{ 1.5 is divisibleby(1) ? 'true' : 'false' }}` displaying true instead of the expected false.

It also makes things act strangely when the operands exceed the maximum integer size, though I'd imagine that's pretty rare.

hason:
Fixes #1353


doc/tests/divisibleby.rst
```diff
@@ -5,6 +5,9 @@
     The ``divisible by`` test was added in Twig 1.14.2 as an alias for
     ``divisibleby``.
 
+.. versionadded:: 1.16.1
+    Support for floating-point numbers was added in Twig 1.16.1.
```
stof: As it is a new feature, not a bugfix, it would be 1.17

fabpot: More than that, it's a BC break, so it won't be merged before 2.0.

fabpot:

fabpot:

fabpot:
Let's take a decision here. Implement the change as a BC break in 3.0 or close?

connorhu:
Can we plan this BC somewhere at the future? This bug is still exists in Twig 3.0.3.

PR #1472, opened 2014-08-04
Introduced methods in Module and Import nodes to easy extend loading of templates

hason:
In inherited class I want to note that a node requires loading the concrete template. Similar to `Twig_Node_Embed` and `Twig_Node_Include`.

### DIFF
--- a/lib/Twig/Node/Module.php
+++ b/lib/Twig/Node/Module.php
@@ -113,15 +113,7 @@ class Twig_Node_Module extends Twig_Node
         if ($parent instanceof Twig_Node_Expression_Constant) {
             $compiler->subcompile($parent);
         } else {
-            $compiler
-                ->raw("\$this->loadTemplate(")
-                ->subcompile($parent)
-                ->raw(', ')
-                ->repr($compiler->getFilename())
-                ->raw(', ')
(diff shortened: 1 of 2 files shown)


fabpot:
You can already log all loaded templates by looking for `Twig_Node_Module` nodes (like done by the profiler: https://github.com/twigphp/Twig/pull/1597/files#diff-f0f1b1e852bd3d37706cad34af876e88R37)

lib/Twig/Node/Module.php
```diff
@@ -391,33 +375,21 @@ protected function compileDebugInfo(Twig_Compiler $compiler)
...
-                ->write(sprintf("%s = \$this->loadTemplate(", $var))
-                ->subcompile($node)
-                ->raw(', ')
-                ->repr($compiler->getFilename())
-                ->raw(', ')
-                ->repr($node->getLine())
-                ->raw(");\n")
-            ;
-        } else {
-            $compiler
```
fabpot: Looks like this code is lost.

fabpot:
The more I work on Twig, the more I want to restrict the extension point. So, it looks like this PR goes in the opposite direction. I don't want to give too much extension points here as the Module class is/should be an implementation detail on how we currently compile the code as a class.

PR #1522, opened 2014-10-06
minor fix in the Core.php Extension

NAYZO:
A small fix in the "twig_random" function of the "Core" Extension.
We should use "!==" instead of "!=" since it's better and faster in this case.


fabpot:
Closing as there is nothing wrong with the current way.

Issue #1322, opened 2014-01-24
parsing error with negative numbers as default parameters


gromgull:
As far as I can see, there is no way to use a negative number as a default parameter value: 

{% macro image(entry, transform="", tag=-1) %}
{% macro image(entry, transform="", tag = -1) %}

both give me an Unexpected token "operator" of value "-" error

fabpot:
For what it's worth, this has been fixed.

PR #1535, opened 2014-10-11
fixed negative number lexing

### LINKED ISSUE #1368: {{ -3 }} vs {{ (-3) }}
Hello, 
I don't know if it is expected to be like this, but why is 3 interpreted as an expression in this case {{ -3 }} ?  different from {{ (-3) }}

fabpot:
Fixes #1368 and #1322.

I don't see it breaking BC as relying on the current way seems weird to me.


fabpot:
Fixed in #4543

PR #1559, opened 2014-10-25
Added template information on debug mode

desyncr:
Searched on issues and PRs but could not find something like this so I gave it a try. I don't have a deep knowledge of twig code base so I don't know is this approach is the best (or correct) but I think the idea is useful.

On debug mode (with `template_info` set to `true`) all included templates are wrapped by the path information for the template.
This is useful for easily moving in a code base which is not very well know to the developer (ie, a legacy code base or a new developer joining a project).

Example output:

    
    <div class="footer container">
    ...
    </div>

    

The default format is (for html):

    %3$s%2$s%3$s%3$s

Given to sprintf with the following arguments:

    $templatePath, $contents, PHP_EOL

Being `$templatePath` the template path (not resolved to file system, but in the format ie, `AcmeDemoBundle:Header:section/template.html.twig`).

`$contents` is the contents itself of the template already handled by doDisplay (ie, rendered).

`PHP_EOL` is only for formatting.

The format is determined by the option 'autoescape' passed in `$options` to `Twig_Environment`; and can be either a string or a callback.
Available format (by default) are 'html', 'js' and 'css'; adding, removing and changing formats is possible through the `Twig_Extension_Debug` extension.

The configuration argument `debug` and `template_info` must be both in true to enable this functionality.

Edit: Update description

### DIFF
--- a/lib/Twig/Environment.php
+++ b/lib/Twig/Environment.php
@@ -53,6 +53,8 @@ class Twig_Environment
      *  * debug: When set to true, it automatically set "auto_reload" to true as
      *           well (default to false).
      *
+     *  * template_info: Whether to enable template information on debugging (default to false).
+     *
      *  * charset: The charset used by the templates (default to UTF-8).
      *
      *  * base_template_class: The base template class to use for generated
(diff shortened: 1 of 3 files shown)


stof:
even relying on the current autoescape setting will break things: there is many place in HTML where you cannot put an HTML comment for instance.

Issue #1560, opened 2014-11-05
Blocks and macros in templates containing embeds


sebhoerl:
Consider the following setup:

    {# layout.html #}
    <h1>Website</h1>
    {% block content '' %}

    {# panel.html #}
    <div class="panel">{% block content '' %}</div>

    {# index.html #}
    {% extends 'layout.html' %}

    {% block task %}
        t.name
    {% endblock %}

    {% block content %}
        {% embed 'panel.html' %}
            {% block content %}
                {% for t in tasks %} {{ block('task') }} {% endfor %}
            {% endblock %}
        {% endembed %}
    {% endblock %}

Usually I didn't use the `embed` block and used this structure to separate the loop from a markup-heavy rendering of the tasks. Intuitively one could think that the `{{ block('task') }}` would be replaced with the task block defined in `index.html`, but actually, as it is implemented right now, Twig will only search for the `task` block within the `embed` area, so nothing happens. This makes sense, because otherwise there could be a lot of confusion about which blocks to use (e.g. for `content` here). I tried modifying the behavior with several approaches, but always ran into some inconsistencies. However, I think that it could be beneficial to have a strategy for such a problem? 

An alternative approach would be to use a macro, but it yields the same problem. Though in this case one can import it via `{% from 'index.hml' import task %}`.

I don't know if anybody has observed this behavior already, but after a quick search I couldn't find anything. So I hope I could point out something and maybe it might be an idea for a future improvement.

PR #1561, opened 2014-11-05
Added use statement for embed tag

sebhoerl:
My proposal for #1560.

It makes it possible to use the `print` block form the main template within the `embed` area in the following example:

    {# layout.html #}
    <h1>Website</h1>
    {% block content '' %}

    {# panel.html #}
    <div class="panel">{% block content '' %}</div>

    {# index.html #}
    {% extends 'layout.html.twig' %}

    {% block print %}
        {{ x }}
    {% endblock %}

    {% block content %}
        {% embed 'panel.html.twig' use print %}
            {% block content %}
                {% for x in ['a', 'b', 'c'] %} {{ block('print') }} {% endfor %}
            {% endblock %}
        {% endembed %}
    {% endblock %}

What has been added is the `use` keyword for the `embed` tag. Maybe the implementation is not optimal, since I just skimmed over the source to see how stuff is usually done, but I can improve it if you're interested in including this.

### DIFF
--- a/lib/Twig/Node/Module.php
+++ b/lib/Twig/Node/Module.php
@@ -38,10 +38,12 @@ class Twig_Node_Module extends Twig_Node
         if ($this->hasAttribute('embedding_module')) {
             $embeddingModule = $this->getAttribute('embedding_module');
             $useBlocks = $this->getAttribute('embed_use_blocks');
-            
+
             $foundBlocks = array();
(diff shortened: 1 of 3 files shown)


fabpot:
I'm not a big fan of this addition. It makes boundaries less clear and adds some complexity to a already quite complex topic. So, I'm closing this as a won't fix.

PR #1567, opened 2014-11-13
Let's talk about Twig 2.0

fabpot:
No big changes, just the removal of deprecated features and things already documented in the docs.

see #1525 for some discussions.


acasademont:
I would like Twig 2.0 to focus a bit more on performance in production mode. Some ideas that come to my mind:

- Inlining things that provide real performance gains, see #780 
- Make less "echo" calls. Right now we are at least using one echo call per line. This is useful for debugging purposes as it is the only way to match the PHP compiled template line to the Twig line if an error should occur. But in production we normally don't need such things and we could aim for having just the bare minimum number of echo calls (or no echo calls at all and use close/open php tags). Perhaps this should be an optional compilation mode.

fabpot:
@acasademont You should probably open separate issues for what you want to be done in Twig 2.0 as this PR is closed now.

About making less `echo` calls, it won't change anything... except if you can prove me wrong. But I tested various combination a long time ago, and I was not able to see any differences.

jrobeson:
@acasademont : certainly performance improvements are welcome, but they'd need to be proven and not just theoretical.

acasademont:
of course of course, hence my comments on the patch in #780 that I reworked just to find out that it's pretty useless in the current form ;)

vkbansal:
@fabpot Is moving to namespaces finalized for 2.0? I'd really like to contribute in that area

stof:
@vkbansal moving to namespaces would break BC entirely with strictly no benefit (the codebase is perfectly usable today). This would mean that Twig 2.x would be a nightmare for the community by splitting the ecosystem. See #1644 trying to reduce the impact the 2.x migration instead, to make it easier.

Issue #1423, opened 2014-06-21
Whitespace control - trim spaces, but not newlines?


ioleo:
Example code:

``` html+django
  <pre>
       {{- "My lovely poem:" }}

       {{- "I love twig, oh my, let's make it better, don't ask why!" }}
      {# ^ note this minus sign #}
  </pre>
```

The output html would render as:

`My lovely poem:I love twig, oh my, let's make it better, don't ask why!`

Why? Becouse it not only removes all the spaces, but also newline characters.

Is there a way to "remove all spaces/tabs" until a twig block/character OR newline?

The desired output is:

```
My lovely poem:

I love twig, oh my, let's make it better, don't ask why!
```

I imagine it could be achieved by introduceing another whitespace control sign (that could replace the minus marked in comment) which would work similar to the `-`, but it would stop at newline character.

Or is it already possible somehow? If so - how?

etki:
Having same issues rendering some Yaml templates (foreseeing questions - creating it as internal structure and then serializing it would be even worse)

brunoais:
@fabpot May I try picking this one up?

petski:
For smarty, I've been using https://gist.github.com/rodneyrehm/3013807 for this behavior. The feature in the gist gives even more whitespace control and could be an inspiration for the feature proposed here.

If @ioleo could dust off #1569, that would be great. If not, I'd be happy to invest some time in a PR.

ioleo:
@petski I have since moved to Scala/Java world and no longer have any active PHP projects. It is highly unlikely I will find time.

fabpot:
fixed by #2925

PR #1569, opened 2014-11-18
[feature] Line whitespace control

ioleo:
This PR introduces line whitespace control character `~` which can be used the same way standard whitespace control character `-` can be, for example:
- `{{~ "Trim to my left, but only until first line break" }}`
- `{%~ for word in ["Trim to my left, but only until first line break"] %}`

I'm not sure if trimming on the right side is possible. `{% for word in ["Trim to my right, but only until first line break"] ~%}` does not stop at line break - it behaves the same way `-` would.. it seems that the closeing tag is handled by diffrent part of code, but I could not find where.

Fixes issue #1423

### DIFF
--- a/lib/Twig/Lexer.php
+++ b/lib/Twig/Lexer.php
@@ -301,9 +301,19 @@ class Twig_Lexer implements Twig_LexerInterface
         $text = substr($this->code, $this->cursor, $match[0][1] - $this->cursor);
         $this->moveCursor($text.$match[0][0]);
 
-        // if (false !== strpos($match[1][0], $this->options['whitespace_all_trim'])) {
-        //     $text = rtrim($text);
-        // }
+        if (false !== strpos($match[1][0], $this->options['whitespace_all_trim'])) {
+            $text = rtrim($text);
+        }
+
+        if (false !== strpos($match[1][0], $this->options['whitespace_line_trim'])) {
+            if ($pos = strrpos($text, "\n")) {
+                // rtrim only until last line break
+                $text = substr($text, 0, $pos+1).rtrim(substr($text, $pos+1));
+            } else {
+                // rtrim all
+                $text = rtrim($text);
+            }
+        }
 
         $this->pushToken(Twig_Token::TEXT_TYPE, $text);
     }
--- a/lib/Twig/Lexer.php
+++ b/lib/Twig/Lexer.php
@@ -334,11 +334,9 @@ class Twig_Lexer implements Twig_LexerInterface
             $this->pushToken(Twig_Token::INTERPOLATION_START_TYPE);
             $this->moveCursor($match[0]);
             $this->pushState(self::STATE_INTERPOLATION);
-
         } elseif (preg_match(self::REGEX_DQ_STRING_PART, $this->code, $match, null, $this->cursor) && strlen($match[0]) > 0) {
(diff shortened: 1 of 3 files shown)


lib/Twig/Lexer.php
```diff
@@ -51,22 +51,23 @@ public function __construct(Twig_Environment $env, array $options = array())
...
         $this->options = array_merge(array(
-            'tag_comment'     => array('{#', '#}'),
-            'tag_block'       => array('{%', '%}'),
-            'tag_variable'    => array('{{', '}}'),
-            'whitespace_trim' => '-',
-            'interpolation'   => array('#{', '}'),
+            'tag_comment'           => array('{#', '#}'),
+            'tag_block'             => array('{%', '%}'),
+            'tag_variable'          => array('{{', '}}'),
+            'whitespace_all_trim'   => '-',
```
stof: renaming the key is a BC break for people instantiating the lexer with custom options

loostro:
@stof reverted old key name to remove possible bc break

stof:
There is a big part missing: tests

loostro:
The `whitespace_control` is also not tested, at least couldn't find it in [LexerTest.php](https://github.com/twigphp/Twig/blob/master/test/Twig/Tests/LexerTest.php). I'm not sure how to test this.

stof:
This is covered by integration tests in https://github.com/twigphp/Twig/blob/master/test/Twig/Tests/Fixtures/tags/trim_block.test

loostro:
@stof Thanks for pointing me in the right direction :) Tests are ready.

There is only one case that I found weird:

```html+django
Trim line on output tag end:
{% for i in range(1, 9) %}
  {{~ i }}  
{% endfor %}
```

> Note the 2 spaces after `{{~ i }}  `

This results in adding a new line.

```
Trim line on output tag end:
1  
2  
3  
4  
(etc ...)
```

While the same example without trailing spaces:
```html+django
Trim line on output tag end:
{% for i in range(1, 9) %}
  {{~ i }}
{% endfor %}
```

> Note no spaces after `{{~ i }}`

Results in:

```
Trim line on output tag end:
1234(etc ...)
```

But this is not related to `~` tag (and this PR).

pitpit:
+1

brunoais:
Could it be that when it searches for the "}}" it also consumes the following character?

lib/Twig/Lexer.php
```diff
@@ -298,6 +305,16 @@ protected function lexRawData($tag)
             $text = rtrim($text);
         }
 
+        if (false !== strpos($match[1][0], $this->options['whitespace_line_trim'])) {
+            if ($pos = strrpos($text, "\n")) {
```
brunoais: But:
If strrpos() returns 0, it executes rtrim() instead of doing nothing.

cvuorinen:
This feature could greatly improve the readability of templates when using Twig to generate Markdown. Currently can't indent properly and have to start every loop, if, else etc. from the beginning of the line without indentation because whitespace matters in markdown whereas in html it doesn't.

stof:
> But this is not related to ~ tag (and this PR).

this is related to the fact that a Twig markup at the end of a line will consume the newline char too

loostro:
@stof I will not have time to update this branch, so unless someone takes over, you can close this PR.

fabpot:

fabpot:
fixed by #2925

PR #1583, opened 2014-12-23
The results of is_dir() are cached

thofman:
If a production server uses statcache, the results of the first and second call of is_dir() is always the same. 
So before calling is_dir() for the second time, use clearstatcache() to clear the results of the first call of is_dir().

### DIFF
--- a/lib/Twig/Environment.php
+++ b/lib/Twig/Environment.php
@@ -1232,8 +1232,11 @@ class Twig_Environment
     {
         $dir = dirname($file);
         if (!is_dir($dir)) {
-            if (false === @mkdir($dir, 0777, true) && !is_dir($dir)) {
-                throw new RuntimeException(sprintf("Unable to create the cache directory (%s).", $dir));
+            if (false === @mkdir($dir, 0777, true)) {
+                clearstatcache(false, $dir);
+                if (!is_dir($dir)) {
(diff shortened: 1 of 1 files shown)


Tobion:
@THofman based on the docs this cannot because the stat cache is only active for existing files.

> You should also note that PHP doesn't cache information about non-existent files. So, if you call file_exists() on a file that doesn't exist, it will return FALSE until you create the file. If you create the file, it will return TRUE even if you then delete the file. However unlink() clears the cache automatically.

http://php.net/manual/en/function.clearstatcache.php

Also the arugments you pass to clearstatcache(false, $filename) make no sense according the documentation and are not effective.

> filename
Clear the realpath and the stat cache for a specific filename only; only used if clear_realpath_cache is TRUE.

Also doctrine cachee does't use this approach either: https://github.com/doctrine/cache/blob/master/lib/Doctrine/Common/Cache/FileCache.php#L185 And nobody reported problems with this as far as I can see.

Can you provide a failing test case? Otherwise I think this must be reverted.

THofman:
@Tobion Sorry for the late response. We (a colleague and I) thought we fixed it, but apparently we didn't read the docs as well as you did :)
This error does not occur frequently, so we couldn't provide you with a failing test case.

My colleague thinks it might have something to do with MooseFS on our production server.

PR #1591, opened 2015-01-13
added an escaping strategy based on the template filename extension

fabpot:
This PR introduces a new autoescaping strategy based on the template filename extension. This strategy has been extracted from Symfony, where we are using it for quite a long time now.

### DIFF
--- a/lib/Twig/FileExtensionEscapingStrategy.php
+++ b/lib/Twig/FileExtensionEscapingStrategy.php
@@ -1,0 +1,49 @@
+<?php
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) 2015 Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+/**
+ * Default autoescaping strategy based on file names.
+ *
+ * This strategy sets the HTML as the default autoescaping strategy,
+ * but changes it based on the filename.
+ *
+ * Note that there is no runtime performance impact as the
+ * default autoescaping strategy is set at compilation time.
+ *
+ * @author Fabien Potencier <fabien@symfony.com>
+ */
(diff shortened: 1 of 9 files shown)


stof:
I see an issue with your logic: it will also match in folder names if they have dots in it

fabpot:
@stof the new version should work fine (I've added some more tests).

stof:
:+1:

lib/Twig/FileExtensionEscapingStrategy.php
```diff
@@ -0,0 +1,49 @@
...
+    /**
+     * Guesses the best autoescaping strategy based on the file name.
+     *
+     * @param string $filename The template file name
+     *
+     * @return string The escaping strategy name to use
+     */
+    public static function guess($filename)
+    {
+        if (!preg_match('{\.(js|css|txt)(?:\.[^/\\\\]+)?$}', $filename, $match)) {
```
Tobion: the beginning delimiter and ending must be the same, so `{}` does not work

Tobion: how can the tests pass?

stof: ``{}`` is supported by PHP as delimiter (special case).

Tobion: At least https://www.functions-online.com/preg_match.html complained. So maybe not in all versions.

stof: use a non-buggy tool. It looks like they are trying to validate the input on their own before the code (the output said ``'an error occured before execution'``), and their logic does not seem to match the PHP code.
These delimiters are supported on all versions since PHP 4.3: http://3v4l.org/9jPPv

Note that you can also use several other brace pairs as delimiters: ``()``, ``[]`` and ``<>``

PR #1597, opened 2015-01-16
added a profiler

fabpot:
This profiler allows to better understand what's going on when rendering templates. It should be used when debug is true only as it adds some overhead.

Here is a sample output in the text format:

```
main 567.04ms/100%
└ index 463.28ms/82%
  └ base 463.24ms/82%
    └ base::block(header)
    └ index::block(content) 306.29ms/54%
    │ └ included 50.18ms/9%
    │ └ included 45.42ms/8%
    │ └ included 35.42ms/6%
    │ └ base::block(content) 104.97ms/19%
    │ │ └ included 34.90ms/6%
    │ └ included 35.84ms/6%
    └ base::block(footer) 156.84ms/28%
      └ included 38.96ms/7%
      └ base::macro(foo)
```

You can also dump a profile in a Blackfire compatible format to [visualise the data as a graph](http://goo.gl/RwNGFH).

A profile is also a great way to be able to list all templates/blocks/macros that were used when rendering a template.

A PR on Symfony (symfony/symfony#13428) integrates this profiler in the Symfony web profiler to replace the current TimedTwigEngine and to add a new Twig panel.

symfony-web-profiler

### DIFF
--- a/lib/Twig/Extension/Profiler.php
+++ b/lib/Twig/Extension/Profiler.php
@@ -1,0 +1,52 @@
+<?php
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) 2015 Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+class Twig_Extension_Profiler extends Twig_Extension
+{
+    private $actives;
+
+    public function __construct(Twig_Profiler_Profile $profile)
+    {
+        $this->actives = array($profile);
+    }
+
+    public function enter(Twig_Profiler_Profile $profile)
+    {
+        $this->actives[0]->addProfile($profile);
+        array_unshift($this->actives, $profile);
+    }
+
+    public function leave(Twig_Profiler_Profile $profile)
+    {
+        $profile->leave();
+        array_shift($this->actives);
+
+        if (1 === count($this->actives)) {
+            $this->actives[0]->leave();
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getNodeVisitors()
+    {
+        return array(new Twig_Profiler_NodeVisitor_Profiler($this->getName()));
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getName()
+    {
+        return 'profiler';
+    }
+}
--- a/lib/Twig/Profiler/Dumper/Blackfire.php
+++ b/lib/Twig/Profiler/Dumper/Blackfire.php
@@ -1,0 +1,68 @@
+<?php
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) 2015 Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+/**
+ * @author Fabien Potencier <fabien@symfony.com>
+ */
+class Twig_Profiler_Dumper_Blackfire
+{
+    public function dump(Twig_Profiler_Profile $profile)
+    {
+        $data = array();
+        $this->dumpProfile('main()', $profile, $data);
+        $this->dumpChildren('main()', $profile, $data);
+
+        $start = microtime(true);
+        $str = <<<EOF
+file-format: BlackfireProbe
+cost-dimensions: wt mu pmu
+request-start: {$start}
+
+
+EOF;
+
+        foreach ($data as $name => $values) {
+            $str .= "{$name}//{$values['ct']} {$values['wt']} {$values['mu']} {$values['pmu']}\n";
+        }
+
+        return $str;
+    }
+
+    private function dumpChildren($parent, Twig_Profiler_Profile $profile, &$data)
+    {
+        foreach ($profile as $p) {
+            if ($p->isTemplate()) {
+                $name = $p->getTemplate();
+            } else {
+                $name = sprintf('%s::%s(%s)', $p->getTemplate(), $p->getType(), $p->getName());
+            }
+            $this->dumpProfile(sprintf('%s==>%s', $parent, $name), $p, $data);
+            $this->dumpChildren($name, $p, $data);
+        }
+    }
+
+    private function dumpProfile($edge, Twig_Profiler_Profile $profile, &$data)
+    {
+        if (isset($data[$edge])) {
+            $data[$edge]['ct'] += 1;
+            $data[$edge]['wt'] += floor($profile->getDuration() * 1000000);
+            $data[$edge]['mu'] += $profile->getMemoryUsage();
+            $data[$edge]['pmu'] += $profile->getPeakMemoryUsage();
+        } else {
+            $data[$edge] = array(
+                'ct' => 1,
+                'wt' => floor($profile->getDuration() * 1000000),
+                'mu' => $profile->getMemoryUsage(),
+                'pmu' => $profile->getPeakMemoryUsage(),
+            );
+        }
+    }
+}
--- a/lib/Twig/Profiler/Dumper/Text.php
+++ b/lib/Twig/Profiler/Dumper/Text.php
@@ -1,0 +1,68 @@
+<?php
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) 2015 Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+/**
+ * @author Fabien Potencier <fabien@symfony.com>
+ */
+class Twig_Profiler_Dumper_Text
+{
+    private $root;
+
+    public function dump(Twig_Profiler_Profile $profile)
+    {
+        return $this->dumpProfile($profile);
+    }
+
+    protected function formatTemplate(Twig_Profiler_Profile $profile, $prefix)
+    {
+        return sprintf('%s└ %s', $prefix, $profile->getTemplate());
+    }
+
+    protected function formatNonTemplate(Twig_Profiler_Profile $profile, $prefix)
+    {
+        return sprintf('%s└ %s::%s(%s)', $prefix, $profile->getTemplate(), $profile->getType(), $profile->getName());
+    }
+
+    protected function formatTime(Twig_Profiler_Profile $profile, $percent)
+    {
+        return sprintf('%.2fms/%.0f%%', $profile->getDuration() * 1000, $percent);
+    }
+
+    private function dumpProfile(Twig_Profiler_Profile $profile, $prefix = '', $sibling = false)
+    {
+        if ($profile->isRoot()) {
+            $this->root = $profile->getDuration();
+            $start = $profile->getName();
+        } else {
+            if ($profile->isTemplate()) {
+                $start = $this->formatTemplate($profile, $prefix);
+            } else {
+                $start = $this->formatNonTemplate($profile, $prefix);
+            }
+            $prefix .= $sibling ? '│ ' : '  ';
+        }
+
+        $percent = $this->root ? $profile->getDuration() / $this->root * 100 : 0;
+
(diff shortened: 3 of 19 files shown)


ninsuo:
Cool stuff.

May it possible to integrate path of cached compiled files?

Compiled templates are very useful to understand twig's internals and/or debug custom tags.

stof:
@fabpot why does fabbot builds a combined status again here ? I thought this was removed weeks ago when Github rebuilt their UI

fabpot:
@stof not sure why, I've removed it again (or I missed this one back then).

lib/Twig/Extension/Profiler.php
```diff
@@ -0,0 +1,52 @@
...
+    public function enter(Twig_Profiler_Profile $profile)
+    {
+        $this->actives[0]->addProfile($profile);
+        array_unshift($this->actives, $profile);
+    }
+
+    public function leave(Twig_Profiler_Profile $profile)
+    {
+        $profile->leave();
+        array_shift($this->actives);
```
stof: shouldn't you check that the profile received as argument is the right one before shifting it ?

lib/Twig/Profiler/DumperInterface.php
```diff
@@ -0,0 +1,18 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+/**
+ * @author Fabien Potencier <fabien@symfony.com>
+ */
+interface Twig_Profiler_DumperInterface
+{
+    public function dump(Twig_Profiler_Profile $profile);
```
stof: you should document the return type of the method (if the return type cannot be part of the interface, there is no point providing an interface)

lib/Twig/Profiler/Node/EnterProfile.php
```diff
@@ -0,0 +1,40 @@
...
+        parent::__construct(array(), array('extension_name' => $extensionName, 'name' => $name, 'type' => $type));
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function compile(Twig_Compiler $compiler)
+    {
+        $compiler
+            ->write("\$__profiler__ = \$this->env->getExtension(")
```
stof: shouldn't you ask the compiler to give you a variable name instead, to avoid conflicts ?

lib/Twig/Profiler/Profile.php
```diff
@@ -0,0 +1,124 @@
...
+    {
+        return isset($this->ends['wt']) && isset($this->starts['wt']) ? $this->ends['wt'] - $this->starts['wt'] : 0;
+    }
+
+    public function getMemoryUsage()
+    {
+        return isset($this->ends['mu']) && isset($this->starts['mu']) ? $this->ends['mu'] - $this->starts['mu'] : 0;
+    }
+
+    public function getPeakMemoryUsage()
```
stof: you should add the phpdoc on methods

stof:
Tests for the HTML dumper are missing

lib/Twig/Profiler/Dumper/Blackfire.php
```diff
@@ -0,0 +1,70 @@
...
+        }
+
+        return $str;
+    }
+
+    private function dumpChildren($parent, Twig_Profiler_Profile $profile, &$data)
+    {
+        foreach ($profile as $p) {
+            if ($p->isTemplate()) {
+                $name = sprintf('%s', $p->getTemplate());
```
stof: why using a sprintf here ? There is only the template name

lib/Twig/Profiler/Dumper/Blackfire.php
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * @author Fabien Potencier <fabien@symfony.com>
+ */
+class Twig_Profiler_Dumper_Blackfire implements Twig_Profiler_DumperInterface
+{
+    public function dump(Twig_Profiler_Profile $profile)
+    {
+        $data = array();
+        $this->dumpChildren('main()', $profile, $data);
+        $this->dumpProfile('main()', $profile, $data);
```
stof: I see an inconsistency here: the root profile is dumped after its children, while other profiles are dumped before their children. What is the reason ?

lib/Twig/Profiler/Dumper/Blackfire.php
```diff
@@ -0,0 +1,70 @@
...
+                $name = sprintf('%s', $p->getTemplate());
+            } else {
+                $name = sprintf('%s::%s(%s)', $p->getTemplate(), $p->getType(), $p->getName());
+            }
+            $this->dumpProfile(sprintf('%s==>%s', $parent, $name), $p, $data);
+            $this->dumpChildren($name, $p, $data);
+        }
+    }
+
+    private function dumpProfile($edge, Twig_Profiler_Profile $profile, &$data)
```
stof: ``$data`` should be typehinted as array

lib/Twig/Profiler/Dumper/Blackfire.php
```diff
@@ -0,0 +1,70 @@
...
+EOF;
+
+        foreach ($data as $name => $values) {
+            $str .= "{$name}//{$values['ct']} {$values['wt']} {$values['mu']} {$values['pmu']}\n";
+        }
+
+        return $str;
+    }
+
+    private function dumpChildren($parent, Twig_Profiler_Profile $profile, &$data)
```
stof: ``$data`` should be typehinted as array

lib/Twig/Profiler/Profile.php
```diff
@@ -0,0 +1,124 @@
...
+ * (c) 2015 Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+/**
+ * @author Fabien Potencier <fabien@symfony.com>
+ */
+class Twig_Profiler_Profile implements \IteratorAggregate
```
stof: Symfony is serializing this object to store it in the Symfony profiler. So I suggest implementing Serializable here, to control the way it gets serialized (and ensuring BC on the serialization format to allow deserializing older profiles with new versions of the code). The rule in Symfony is that only Serializable classes are guaranteed to be serializable in a BC way (the only way to ensure it for non Serializable classes would be to freeze the class, forbidding to change any private property, which is a no-go)

test/Twig/Tests/Profiler/Dumper/AbstractTest.php
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * (c) Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+abstract class Twig_Tests_Profiler_Dumper_AbstractTest extends PHPUnit_Framework_TestCase
+{
+    abstract public function testDump();
```
stof: I suggest removing this abstract method. It is not really needed given that it is not used in the abstract class

lib/Twig/Profiler/NodeVisitor/Profiler.php
```diff
@@ -0,0 +1,64 @@
...
+    public function enterNode(Twig_NodeInterface $node, Twig_Environment $env)
+    {
+        return $node;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function leaveNode(Twig_NodeInterface $node, Twig_Environment $env)
+    {
```
stof: should we disable this node visitor automatically when the Twig_Environment is not in debug mode, to avoid mistakes adding the overhead in prod ? The Debug extension is already changing its behavior based on the debug mode

lib/Twig/Profiler/Dumper/Blackfire.php
```diff
@@ -0,0 +1,70 @@
...
+            }
+            $this->dumpProfile(sprintf('%s==>%s', $parent, $name), $p, $data);
+            $this->dumpChildren($name, $p, $data);
+        }
+    }
+
+    private function dumpProfile($edge, Twig_Profiler_Profile $profile, &$data)
+    {
+        if (isset($data[$edge])) {
+            $data[$edge] = array(
```
nicolas-grekas: several `+=` whould be faster than reallocating an array here:
`$data[$edge]['ct'] += 1;` etc.

stof:
The HTML dumper relies on the newlines and the indentation of the HTML to render things properly. This means that it should wrap the output in a ``<pre>`` tag itself (your Symfony integration looks good because you wrap it on Symfony side, but it should not be necessary to leave it to the user of the dump)

lib/Twig/Extension/Profiler.php
fabpot: Given that I do the instrumentation automatically, I'm not sure in which case it could become wrong.

lib/Twig/Profiler/Dumper/Blackfire.php
fabpot: It was needed for Blackfire at some point, but that's not the case anymore, changed.

lib/Twig/Profiler/Profile.php
```diff
@@ -0,0 +1,134 @@
...
+ * (c) 2015 Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+/**
+ * @author Fabien Potencier <fabien@symfony.com>
+ */
+class Twig_Profiler_Profile implements \IteratorAggregate, \Serializable
```
stof: this breaks the PHP 5.2 compat (see the Travis failures). You should remove the ``\``

lib/Twig/Profiler/Profile.php
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    public function getIterator()
+    {
+        return new ArrayIterator($this->profiles);
+    }
+
+    public function serialize()
+    {
+        return array($this->name, $this->type, $this->starts, $this->ends, $this->profiles);
```
stof: this must return a string

lib/Twig/Profiler/Dumper/Text.php
```diff
@@ -0,0 +1,68 @@
...
+
+        if ($profile->getDuration() * 1000 < 1) {
+            $str = $start."\n";
+        } else {
+            $str = sprintf("%s %s\n", $start, $this->formatTime($profile, $percent));
+        }
+
+        $nCount = count($profile->getProfiles());
+        foreach ($profile as $i => $p) {
+            $str .= $this->dumpProfile($p, $prefix, $i + 1 !== $nCount, $root);
```
stof: ``$root`` is used only to forward it to the recursive call each time, so it is actually unused and could be removed

stof:
The Profile serialization should be tested, ensuring we don't loose data when serializing and unserializing again (it would also ensure that the serialization works, which I doubt as the implementation does not return a string as expected in the php documentation)

lib/Twig/Profiler/NodeVisitor/Profiler.php
fabpot: The dump() function is a bit different: when a user forgot one in a template, it can be dangerous. Here, I don't see why we should forbid this extension to be used in non-debug mode.

stof: fine with me

lib/Twig/Profiler/Dumper/Blackfire.php
fabpot: I don't see the need for that and this is a private function anyway.

stof: what is the reason to typehint ``$profile`` on the private function then ? IMO we should be consistent: either we typehint all arguments in private methods (assuming they *can* be typehinted of course) or none

fabpot: Because I find it stupid in general to type-hint with array as it forbids using a Traversable later on.

PR #1598, opened 2015-01-16
FIX #1596 FileSystem::findTemplate now applies realpath on the absolute ...

### LINKED ISSUE #1596: Twig_Loader_Filesystem::findTemplate returns absolute path but not realpath
Hi,

I'm using twig in a symfony 2.6 full stack context. I have noticed that the cache classes generated by the twig bundle were not really used by twig and new classes were created. I've found the problem and it appears to be in Twig.

Long story short : I want to avoid having file generated when a user hits my page

Long story long : when symfony warms up the cache, it's made using this kind of names: "```WebProfilerBundle:Collector:memory.html.twig```". The resolution to the filename is made by the ```FileSystemLoader``` (from ```TwigBundle```) and it's ok.
When a user hits my page, the template is loaded with "```@WebProfiler/Collector/memory.html.twig```". The resolution to the filename is the made by ```Twig_Loader_Filesystem```. The problem is that in a symfony context, the base path is "app" and to go in src, there is an extra "../". So the returned full path, even if it's the same file resolved, is not the same. And this path is used to create the twig class name (work of ```Twig_Loader_Filesystem::getCacheKey```). So the cache is generated with the warmup and again when I've a hit.

Actually the solution seems very simple, applying ```realpath``` on the generated value does the trick.

Thanks for your feedback,

Xavier.

xavierleune:
...path before returning it.

This is a purposal to fix the issue #1596.
Please make any feedback needed.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,6 @@
 * 1.17.1 (2015-XX-XX)
 
- * n/a
+ * added a 'realpath' strategy when returning the template filename, if possible.
 
 * 1.17.0 (2015-01-14)
 
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,7 @@
 * 1.17.1 (2015-XX-XX)
 
- * added a 'realpath' strategy when returning the template filename, if possible.
+ * The template filename have realpath applied before returning it. In this way we have the same path between the
+   resolution of Twig and Twing bundle cache warmup, allowing to have one cache key.
 
 * 1.17.0 (2015-01-14)
 
--- a/lib/Twig/Loader/Filesystem.php
+++ b/lib/Twig/Loader/Filesystem.php
@@ -183,8 +183,11 @@ class Twig_Loader_Filesystem implements Twig_LoaderInterface, Twig_ExistsLoaderI
         }
 
         foreach ($this->paths[$namespace] as $path) {
+            if (false !== $realpath = realpath($path.'/'.$shortname)) {
+                return $this->cache[$name] = $realpath;
+            }
             if (is_file($path.'/'.$shortname)) {
-                return $this->cache[$name] = realpath($path.'/'.$shortname);
+                return $this->cache[$name] = $path.'/'.$shortname;
             }
         }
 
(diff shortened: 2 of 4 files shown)


ninsuo:
You should notice the note in [php documentation](http://php.net/manual/en/function.realpath.php):

> Note: the running script must have executable permissions on all directories in the hierarchy, otherwise realpath() will return FALSE.

I personnaly enter this case as my applications are only granted to access their home.

fabpot:
To avoid the `false` issue mentioned by @ninsuo, we can do something like this:

```php
if (false !== $realpath = realpath($path.'/'.$shortname)) {
    return $this->cache[$name] = $realpath;
}
```

xavierleune:
@ninsuo that's a very good point but to be complete, if you can list files in a directory, you have the exec right. Actually without this right you can't make file_get_contents on a file.
To be sure just try this:
```mkdir testrx && chmod a-x testrx/ && sudo touch testrx/truc.txt```
and then in a php -a try to do this:
```var_dump(file_get_contents('testrx/truc.txt'));```

And you'll get the following error : ```Warning: file_get_contents(testrx/truc.txt): failed to open stream: Permission denied in php shell code on line 1```

So in fact I think the purposal by @fabpot is not needed. But i Can throw a Twig_Error_Loader Exception if realpath is false.
What do you think about that ?

**Comment by fabpot** [comment]
@xavierleune I prefer to be on the safe side and use what I proposed.

**Comment by ninsuo** [comment]
I agree that on a classic filesystem php works perfectly. But it really depends on the filesystem you're using. `realpath` behaves very randomly if you're under nfs, or glusterfs.

**Comment by stof** [comment]
another case where realpath returns false: inside phar archives. Making it impossible to use Twig when building your project as a phar is a bad idea (and it would even be a BC break). So the proper handling of ``false`` return values is a must-have

**Comment by xavierleune** [comment]
Thanks for all your feedbacks, the last version now uses realpath and then the standard way on error. To be complete about readpath, it seems that is_file suffers the same limitation : http://php.net/manual/fr/function.is-file.php#107403

I don't know why travis is still working on this PR (this build seems locked out : https://travis-ci.org/twigphp/Twig/builds/47250271) but tests are running fine on my environment.

**Comment by fabpot** [comment]
Can you add an entry in the CHANGELOG?

**Comment by xavierleune** [comment]
Is this entry ok for you ?

File: CHANGELOG [hunk 1]
```diff
@@ -1,6 +1,6 @@
 * 1.17.1 (2015-XX-XX)
 
- * n/a
+ * added a 'realpath' strategy when returning the template filename, if possible.
```
fabpot: I would explain which bug has been fix instead.

xavierleune:
Please let me know if the last changelog entry is not ok for you.

fabpot:
Thank you @xavierleune.

PR #1601, opened 2015-01-20
reduced the number of times error information has to be guessed

fabpot:


lib/Twig/Template.php
```diff
@@ -240,6 +240,28 @@ public function getBlockNames()
         return array_keys($this->blocks);
     }
 
+    protected function loadTemplate($template, $templateName = null, $line = null, $index = null)
+    {
+        try {
+            if (is_array($template)) {
+                return $this->env->resolveTemplate($template);
+            } elseif ($template instanceof Twig_Template) {
```
stof: ``if`` is enough as the previous ``if`` returns

PR #1602, opened 2015-01-20
added Twig_Environment::createTemplate()

### LINKED ISSUE #1506: Render Strings
In addition to rendering strings from an variable within a template as added within #874 
I see the use case of rendering a string e.g. in the Servicelayer to e.g. Prepopulate Entites
or whatever.

This can currently be achieved with several ways which all have some drawbacks:

- Adding the String Loader
````
    acme.demo.twig.stringloader:
        class: Twig_Loader_String
        tags:
            - { name: twig.loader }
````

This has the drawback that all invalid templates will be rendered as strings. If you have an exception in the template or the template does not exist you will just see the rendered template name. 

- Adding a StringTwigService

````
services:
    acme.twig_string_loader:
        class:        "Twig_Loader_String"
    acme.twig_string:
        class:        "%twig.class%"
        arguments:    [@acme.twig_string_loader, %twig.options% ]
````

With this approach you have a separate string loader for the above mentioned purpose, however you do not have access to all your extensions. However for my use cases I usually go with Option (2).

I was just considering if a renderString() method in the Twig Core might make sense. When I was investigation those issues I saw a lot of comments how to render a string.

### PR DESCRIPTION
**Pull request by fabpot**
fixes #1506

### DIFF
--- a/lib/Twig/Environment.php
+++ b/lib/Twig/Environment.php
@@ -347,6 +347,41 @@ class Twig_Environment
     }
 
     /**
+     * Creates a template from source.
+     *
+     * This method should not be used as a generic way to load templates.
+     *
+     * @param string $name  The template name
+     * @param int    $index The index if it is an embedded template
+     *
+     * @return Twig_Template A template instance representing the given template name
+     *
+     * @throws Twig_Error_Loader When the template cannot be found
+     * @throws Twig_Error_Syntax When an error occurred during compilation
+     */
+    public function createTemplate($template)
+    {
+        $name = sprintf('__string_template__%s', hash('sha256', uniqid(mt_rand(), true), false));
+
+        $loader = new Twig_Loader_Chain(array(
+            new Twig_Loader_Array(array($name => $template)),
+            $current = $this->getLoader(),
+        ));
+
+        $this->setLoader($loader);
+        try {
+            $template = $this->loadTemplate($name);
+        } catch (Exception $e) {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by stloyd** [comment]
:+1:

File: lib/Twig/Environment.php [hunk 1]
```diff
@@ -347,6 +347,39 @@ public function loadTemplate($name, $index = null)
     }
 
     /**
+     * Creates a template from source.
+     *
+     * @param string  $name  The template name
+     * @param integer $index The index if it is an embedded template
```
stof: fabbot is complaining here

lib/Twig/Environment.php
```diff
@@ -347,6 +347,39 @@ public function loadTemplate($name, $index = null)
...
+     *
+     * @return Twig_TemplateInterface A template instance representing the given template name
```
stof: I would say ``@return Twig_Template`` given that the interface is deprecated

lib/Twig/Extension/StringLoader.php
```diff
@@ -43,22 +43,5 @@ public function getName()
...
-        $template = $env->loadTemplate($name);
-    } catch (Exception $e) {
-        $env->setLoader($current);
-
-        throw $e;
-    }
-    $env->setLoader($current);
-
-    return $template;
+    return $env->createTemplate($template);
```
stof: you could even save a method call in the compiled template by calling the env method directly instead of using a wrapper function

PR #1603, opened 2015-01-20
removed the special Sandbox Module class

### LINKED ISSUE #1402: There's no easy way to decorate Twig_Node_Module

fabpot:
This PR removes the need for the special Sandbox Module class. This approach can be generalized by adding more "empty" nodes in the Module class to allow third-party extension to easily add code to the generated class. But I think that hacking the module class is always never a good idea and should be avoided (so this "feature" should not even be documented :)).

This is an alternative for #1436, fixes #1402.


stof:
:+1: Using the ``display_enter`` extension point is much better (and allows to have many nodes using it assuming they take care of preserving the previous one too, as done here and in the profiler)

PR #1605, opened 2015-01-21
Small cleanup to use already declared variable

hason:


stof:
:+1:

fabpot:
Good catch, thanks @hason.

PR #1609, opened 2015-01-27
get PHP version info once on construction

SpacePossum:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes

Hi,
I was profiling a project and noticed that `phpversion` was called 5000 times by Twig.
This patch reduces the amount of calls without breaking anything (I hope ;)  )

twig_phpversion


SpacePossum:
(I noticed the file permission change, I'l revert that later ;) )

lib/Twig/NodeVisitor/Optimizer.php
```diff
@@ -45,6 +46,7 @@ public function __construct($optimizers = -1)
         }
 
         $this->optimizers = $optimizers;
+        $this->isPHP540RCorHigher = version_compare(phpversion(), '5.4.0RC1', '>=');
```
fabpot: Can you use `PHP_VERSION_ID >= 50400` instead? and perhaps rename the variable to just `$isPHP54`.

SpacePossum: I've made the change, hope it is better.
Since PHP_VERSION_ID is available since 5.2.7 and Twig requires >=5.2.4 we cannot rely on it to be defined.

stof: I suggest bumping the requirement to 5.2.7 instead, to be able to use the constant all the time (which will also avoid using a property in the class)

SpacePossum: Thank you for the feedback. I've made the adjustments. Let me know what you think :)

stof:
Can you also updated other places using ``phpversion()`` in Twig ? Using the constant is indeed much faster (and can allow OPCache to optimize the whole condition when enabling this optimization)

lib/Twig/NodeVisitor/Optimizer.php
```diff
@@ -56,7 +56,7 @@ public function enterNode(Twig_NodeInterface $node, Twig_Environment $env)
             $this->enterOptimizeFor($node, $env);
         }
 
-        if (!version_compare(phpversion(), '5.4.0RC1', '>=') && self::OPTIMIZE_VAR_ACCESS === (self::OPTIMIZE_VAR_ACCESS & $this->optimizers) && !$env->isStrictVariables() && !$env->hasExtension('sandbox')) {
+        if (!PHP_VERSION_ID >= 50400 && self::OPTIMIZE_VAR_ACCESS === (self::OPTIMIZE_VAR_ACCESS & $this->optimizers) && !$env->isStrictVariables() && !$env->hasExtension('sandbox')) {
```
stof: use ``if (PHP_VERSION_ID < 50400 && ....``

SpacePossum:
(don't worry about the commit count; I'll squash those when done)

@stof What would be the best way to apply this fix on checks against a RC version, for example:
https://github.com/twigphp/Twig/blob/master/lib/Twig/Node/Expression/Name.php#L49

stof:
@SpacePossum nobody is running beta versions of PHP 5.4 anymore (at least no sane people)

lib/Twig/Error.php
```diff
@@ -188,10 +188,10 @@ protected function guessTemplateInfo()
         $template = null;
         $templateClass = null;
 
-        if (version_compare(phpversion(), '5.3.6', '>=')) {
-            $backtrace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS | DEBUG_BACKTRACE_PROVIDE_OBJECT);
-        } else {
+        if (PHP_VERSION_ID < 50306) {
             $backtrace = debug_backtrace();
+        } else {
+        	$backtrace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS | DEBUG_BACKTRACE_PROVIDE_OBJECT);
```
stof: Please don't change the order of the code to avoid unnecessary diffs (and don't use tabs for the indentation)

test/Twig/Tests/Node/Expression/FunctionTest.php
```diff
@@ -78,7 +78,7 @@ protected function createFunction($name, array $arguments = array())
 
     protected function getEnvironment()
     {
-        if (version_compare(phpversion(), '5.3.0', '>=')) {
+    	if (PHP_VERSION_ID >= 50300) {
```
stof: the indentation change is wrong

test/Twig/Tests/Node/Expression/TestTest.php
```diff
@@ -47,7 +47,7 @@ protected function createTest($node, $name, array $arguments = array())
 
     protected function getEnvironment()
     {
-        if (version_compare(phpversion(), '5.3.0', '>=')) {
+    	if (PHP_VERSION_ID >= 50300) {
```
stof: the indentation change is wrong here too

SpacePossum:
@stof good point. So for example:
```php
 if (version_compare(phpversion(), '5.4.0RC1', '>=')) {
```

Should become:
```php
 if (PHP_VERSION_ID >= 5.4.1')) {
```

?

stof:
@SpacePossum no. 5.4.0RC1 is before 5.4.0

fabpot:
Thank you @SpacePossum.

SpacePossum:
:+1:  thanks @fabpot

PR #1616, opened 2015-01-29
fixed named args for static methods

kriswallsmith:
This fixes an obscure reflection error that gets thrown if a functioned backed by a static method is called with named arguments.

### DIFF
--- a/lib/Twig/Node/Expression/Call.php
+++ b/lib/Twig/Node/Expression/Call.php
@@ -120,6 +120,9 @@ abstract class Twig_Node_Expression_Call extends Twig_Node_Expression
         } elseif (is_object($callable) && !$callable instanceof Closure) {
             $r = new ReflectionObject($callable);
             $r = $r->getMethod('__invoke');
+        } elseif (is_string($callable) && false !== strpos($callable, '::')) {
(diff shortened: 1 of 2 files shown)


fabpot:
Thank you @kriswallsmith.

fabpot:
@kriswallsmith Is it related to a break on Assetic with the latest version of Twig?

kriswallsmith:
No, something I discovered independently.

lib/Twig/Node/Expression/Call.php
```diff
@@ -120,6 +120,9 @@ protected function getArguments($callable, $arguments)
         } elseif (is_object($callable) && !$callable instanceof Closure) {
             $r = new ReflectionObject($callable);
             $r = $r->getMethod('__invoke');
+        } elseif (is_string($callable) && false !== strpos($callable, '::')) {
+            $parts = explode('::', $callable, 2);
+            $r = new ReflectionMethod($parts[0], $parts[1]);
```
rybakit: These two lines could be reduced to:
```php
$r = new ReflectionMethod($callable);
```

fabpot: done, thanks.

PR #1617, opened 2015-01-29
Fix time sensitive profiler test (ready)

SpacePossum:
The profiler tests fail some times because of the performance of the machine these are run on.
These changes should provide stability.

### DIFF
--- a/test/Twig/Tests/Profiler/Dumper/AbstractTest.php
+++ b/test/Twig/Tests/Profiler/Dumper/AbstractTest.php
@@ -32,13 +32,13 @@ abstract class Twig_Tests_Profiler_Dumper_AbstractTest extends PHPUnit_Framework
 
         $embedded = clone $embedded;
         $index->addProfile($embedded);
-        $a = range(1, 1000);
+        usleep(500);
         $embedded->leave();
-        $profile->leave();
 
-        usleep(5000);
+        usleep(4500);
         $index->leave();
 
+        $profile->leave();
         return $profile;
     }
 }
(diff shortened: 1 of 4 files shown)


SpacePossum:
Looking at the travis results this still seems unstable..

test/Twig/Tests/Profiler/Dumper/AbstractTest.php
```diff
@@ -32,13 +32,14 @@ protected function getProfile()
 
         $embedded = clone $embedded;
         $index->addProfile($embedded);
-        $a = range(1, 1000);
```
SpacePossum: This is effectively a nonop and will be removed by the compiler removing its use (delay)

test/Twig/Tests/Profiler/Dumper/AbstractTest.php
```diff
@@ -32,13 +32,14 @@ protected function getProfile()
...
+        usleep(500);
         $embedded->leave();
-        $profile->leave();
```
SpacePossum: Im not sure why to leave the `profile` before leaving the `index`, from a DX point of view I would say this a state exception, but maybe I'm not to good in reading this :)

test/Twig/Tests/Profiler/ProfileTest.php
```diff
@@ -69,9 +69,10 @@ public function testGetAddProfile()
     public function testGetDuration()
     {
         $profile = new Twig_Profiler_Profile();
+        usleep(1);
```
SpacePossum: Without this delay and running the test multiple times after each other on fast setup (and so with full op cache) this would fall over

test/Twig/Tests/Profiler/ProfileTest.php
```diff
@@ -69,9 +69,10 @@ public function testGetAddProfile()
...
         $profile->leave();
 
-        $this->assertTrue($profile->getDuration() > 0);
+        $this->assertTrue($profile->getDuration() > 0, sprintf('Expected duration > 0, got: %f', $profile->getDuration()));
```
SpacePossum: just some more information (DX), I can take it out again if wanted

fabpot:
Thank you @SpacePossum.

PR #1618, opened 2015-01-29
replace multibyte characters support, add documention and tests

SpacePossum:
This adds multibyte support to the replace-filter when used with two strings as arguments.
It adds tests for both UTF-8 characters and non-UTF-8 and documents the "hidden" featured.

### DIFF
--- a/doc/filters/replace.rst
+++ b/doc/filters/replace.rst
@@ -11,6 +11,14 @@ The ``replace`` filter formats a given string by replacing the placeholders
     {# outputs I like foo and bar
        if the foo parameter equals to the foo string. #}
 
+Single characters can be replaced by another when passing two strings as translation map
+
+.. code-block:: jinja
+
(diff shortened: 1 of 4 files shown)


stof:
I think the syntax ``|replace("zb", "ta")`` is totally impossible to understand for someone writing templates anyway. So I'm not sure it makes sense to add multibyte support for it. I would rather document only the array syntax.

Tobion:
I agree with @stof :-1:

SpacePossum:
The syntax might be a bit hard, but so are other parts of Twig like macros.
The feature is already in and supported for a long time, so why not document it?

Imho one of the great things about Twig is not having to worry about string operations on multibyte strings (that much). So making this feature work with multibyte chars would help.

SpacePossum:
I think rebasing would be useless giving the changes this will make it. What might be useful would to be stop supporting this hidden feature on 2.0. Thanks for the reviews and insights :)

PR #1624, opened 2015-02-04
Organize and show value types

kevinsimper:
I was a little confused about how `auto_reload` and what kind of types you had to provide. Hope this helps to clarify that for others.

Should it also say something about how `cache: true` and `auto_reload: true` works together, as that is a question i have myself :)

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -72,29 +72,43 @@ options as the constructor second argument::
 
 The following options are available:
 
-* ``debug``: When set to ``true``, the generated templates have a
+* ``debug`` *boolean*
+
+  When set to ``true``, the generated templates have a
   ``__toString()`` method that you can use to display the generated nodes
   (default to ``false``).
 
-* ``charset``: The charset used by the templates (default to ``utf-8``).
+* ``charset`` *string (default to ``utf-8``)*
 
-* ``base_template_class``: The base template class to use for generated
-  templates (default to ``Twig_Template``).
+  The charset used by the templates.
 
-* ``cache``: An absolute path where to store the compiled templates, or
+* ``base_template_class`` *(default to ``Twig_Template``)*
+
+  The base template class to use for generated
+  templates.
+
+* ``cache`` *boolean*
+
+  An absolute path where to store the compiled templates, or
   ``false`` to disable caching (which is the default).
 
-* ``auto_reload``: When developing with Twig, it's useful to recompile the
+* ``auto_reload`` *boolean*
+
+  When developing with Twig, it's useful to recompile the
   template whenever the source code changes. If you don't provide a value for
   the ``auto_reload`` option, it will be determined automatically based on the
   ``debug`` value.
 
-* ``strict_variables``: If set to ``false``, Twig will silently ignore invalid
+* ``strict_variables`` *boolean*
+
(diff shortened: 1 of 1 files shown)


doc/api.rst
```diff
@@ -72,29 +72,43 @@ options as the constructor second argument::
...
+  When set to ``true``, the generated templates have a
   ``__toString()`` method that you can use to display the generated nodes
   (default to ``false``).
 
-* ``charset``: The charset used by the templates (default to ``utf-8``).
+* ``charset`` *string (default to ``utf-8``)*
+
+  The charset used by the templates.
+
+* ``base_template_class`` *(default to ``Twig_Template``)*
```
sstok: maybe add a type here to?

doc/api.rst
```diff
@@ -72,29 +72,43 @@ options as the constructor second argument::
...
+
+  The base template class to use for generated
+  templates.
 
-* ``base_template_class``: The base template class to use for generated
-  templates (default to ``Twig_Template``).
+* ``cache`` *boolean*
```
sstok: this can a boolean and string.

doc/api.rst
```diff
@@ -72,29 +72,43 @@ options as the constructor second argument::
...
-* ``strict_variables``: If set to ``false``, Twig will silently ignore invalid
+* ``strict_variables`` *boolean*
+
+  If set to ``false``, Twig will silently ignore invalid
   variables (variables and or attributes/methods that do not exist) and
   replace them with a ``null`` value. When set to ``true``, Twig throws an
   exception instead (default to ``false``).
 
-* ``autoescape``: If set to ``true``, HTML auto-escaping will be enabled by
+* ``autoescape`` *boolean*
```
stof: this is not a boolean all the time

kevinsimper: Do you know what the values can be?

doc/api.rst
kevinsimper: What is the string values?

stof: read the following sentence. It explains it. It is the path to the cache folder

doc/api.rst
stof: one of the supported escaping strategies (``true`` is an alias for ``html`` for BC reasons)

doc/api.rst
kevinsimper: So it should say `boolean/string`?

doc/api.rst
SpacePossum: Maybe state that is the charset of the `output`  as well?
(as this is something I know others find confusing (ie. think it is charset of the data you use in Twig))

SpacePossum:
nice PR :+1: , may I ask why you target master here? (1.x might benefit from this as well)

kevinsimper:
Don't know if 1.x was there in februar? :)

doc/api.rst
stof: ``string|false`` would be the actual one

doc/api.rst
stof: btw, the following paragraphs are listing them

fabpot:
Thank you @kevinsimper.

fabpot:
I've fixed the remaining issues in d0b7130

PR #1626, opened 2015-02-07
Twig dump is now using Symfony VarDumper Component

Whyounes:


ninsuo:
Twig still uses `var_dump`:

https://github.com/twigphp/Twig/blob/master/lib/Twig/Extension/Debug.php#L46

And as all components should be usable standalone, I don't think this will change.

Whyounes:
But i think it should be using it by now. We require it as a `--dev` and the `twig_var_dump ` function can use it!

stof:
The symfony twig-bridge already provides a drop-in replacement for the Debug extension of twig using the VarDumper component.

Issue #1625, opened 2015-02-05
The title filter on date style phrases shows incorrectly


viion:
If you have a number in-front of a letter, the first letter of the word is capitalized regardless of numbers prior to it. For example

```
{% set var = '3rd of the 7th winning' %}
{{ var|title }}
```
Output: **3Rd Of The 7Th Winning**
Expected: **3rd Of The 7th Winning**

Fiddle: http://twigfiddle.com/0s11rr

PHP's ``` ucwords() ``` does it as expected, apologies if I'm using the wrong filter.

ninsuo:
I :+1:  this

From the source code, it comes from the `mb_convert_case` function.

```php
    /**
     * Returns a titlecased string.
     *
     * @param Twig_Environment $env    A Twig_Environment instance
     * @param string           $string A string
     *
     * @return string The titlecased string
     */
    function twig_title_string_filter(Twig_Environment $env, $string)
    {
        if (null !== ($charset = $env->getCharset())) {
            return mb_convert_case($string, MB_CASE_TITLE, $charset);
        }

        return ucwords(strtolower($string));
    }
```

If you test both functions separately, we get:

```php
$str = "3rd of the 7th winning";

// 3Rd Of The 7Th Winning
echo mb_convert_case($str, MB_CASE_TITLE, "UTF-8");

// // 3rd Of The 7th Winning
echo ucwords($str); 
```

The bug has already been reported 6 years ago: https://bugs.php.net/bug.php?id=46607
So using a workaround should be better.

fabpot:
Closing this issue as the related PR was closed.

PR #1628, opened 2015-02-08
Fix title filter, do not uppercase characters after digits and do not upper 'S

SpacePossum:
For:
https://github.com/twigphp/Twig/issues/1625

### DIFF
--- a/lib/Twig/Extension/Core.php
+++ b/lib/Twig/Extension/Core.php
@@ -1277,7 +1277,15 @@ if (function_exists('mb_get_info')) {
     function twig_title_string_filter(Twig_Environment $env, $string)
     {
         if (null !== ($charset = $env->getCharset())) {
-            return mb_convert_case($string, MB_CASE_TITLE, $charset);
+            $title = mb_convert_case($string, MB_CASE_TITLE, $charset);
+            $title = preg_replace('#\'S#u', '\'s', $title);
+
+            return preg_replace_callback(
+                            '#\b\d+([^\d\s]+)\b#u',
+                            function ($matches) {
+                                return mb_strtolower($matches[0], 'UTF-8');
+                            },
+                            $title);
         }
 
         return ucwords(strtolower($string));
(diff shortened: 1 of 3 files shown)


lib/Twig/Extension/Core.php
```diff
@@ -1277,7 +1277,15 @@ function twig_lower_filter(Twig_Environment $env, $string)
     function twig_title_string_filter(Twig_Environment $env, $string)
     {
         if (null !== ($charset = $env->getCharset())) {
-            return mb_convert_case($string, MB_CASE_TITLE, $charset);
+            $title = mb_convert_case($string, MB_CASE_TITLE, $charset);
+            $title = preg_replace('#\'S#u', '\'s', $title);
+
+            return preg_replace_callback(
```
sstok: Unfortunately Twig still supports PHP 5.2 so we can't use closures here.

lib/Twig/Extension/Core.php
```diff
@@ -1277,7 +1277,15 @@ function twig_lower_filter(Twig_Environment $env, $string)
...
+                            '#\b\d+([^\d\s]+)\b#u',
+                            function ($matches) {
```
sstok: Unfortunately Twig still supports PHP 5.2 so we can't use closures here.

SpacePossum: I should've know that since I even sent a PR to push to 5.2.7 :/
I'll see if can refactor this into something useful that works for 5.2 as well. Let me know if you've an idea about how to solve this ticket, maybe a whole different approach would be better.

fabpot: I'm not too happy to fix a PHP bug, especially if it means adding too much code in Twig itself. It would be much better to fix PHP instead.

SpacePossum: I agree :)
However as was pointed out in https://github.com/twigphp/Twig/issues/1625 this bug is open for 6 years... Maybe someone else can get the developers @ php to pick this one up?

PR #1629, opened 2015-02-18
Clean output buffers when an exception occured

hvt:
When I was testing rendering of a specific block (`renderBlock()`) of a faulty Twig template (non-existent variable) in PHPUnit, it stated that particular test was risky.

After some debugging, I found out that when an exception occurs in the `displayBlock()` method, the output buffer that was opened in `renderBlock()` would still be open. PHPUnit complains about that by saying the test was risky.

I fixed this by closing all open output buffers when an exception occurs. The fix is the same as the one that is already in place in `render()`. Next, I also applied it to `renderParentBlock()`, that has the same behaviour as `renderBlock()`.

### DIFF
--- a/lib/Twig/Template.php
+++ b/lib/Twig/Template.php
@@ -176,8 +176,17 @@ abstract class Twig_Template implements Twig_TemplateInterface
      */
     public function renderParentBlock($name, array $context, array $blocks = array())
     {
+        $level = ob_get_level();
         ob_start();
-        $this->displayParentBlock($name, $context, $blocks);
+        try {
+            $this->displayParentBlock($name, $context, $blocks);
+        } catch (Exception $e) {
+            while (ob_get_level() > $level) {
+                ob_end_clean();
+            }
+
+            throw $e;
+        }
 
         return ob_get_clean();
     }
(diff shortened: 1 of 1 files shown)


hvt:
Did I forget anything to get this included ;]?

stof:
renderBlock is an internal method. This is why it does not include the overhead of catching all exceptions to clear the output buffer. It is not meant to be called directly outside a render call

hvt:
(Shouldn't these methods be `protected` then?)

I think there is quite a valid use-case of using the `renderBlock` methods. For example, I have an `email.twig` template, that contains a `subject`, `plain`, and `html` block and I'd like to render these blocks separately.

stof:
@hvt no, they cannot be protected, because they are called from other Twig classes too (but from inside a template rendering).

Regarding the email use case, I also have implemented such rendering (adding the buffer handling outside the ``renderBlock`` code). And I'm not sure it is the right approach anymore, because I also need different escaping strategies for the different parts (the subject and the plaintext body should not be escaped as HTML, as they are not HTML). My preference now would be to have separate templates for the plaintext and the html versions (``email.txt.twig`` and ``email.html.twig`` for instance).

hvt:
@stof Okay, clear on the `protected` part.

I've overcome that by explicitly starting each `subject` and `plain` block with an `{% autoescape false %}` tag.

What's exactly against merging this? I.e. all tests succeed.

stof:
@hvt try/catch blocks you added are likely to add overhead. It would need a profiling to see the exact impact though.

anyway, I cannot merge this. I'm not a collaborator on the repo.

fabpot:
I'm closing this for the exact reasons explained by @stof: those methods are internal ones, and must not be used by end users.

PR #1637, opened 2015-02-26
Updates slice so that it still accepts a SimpleXMLElement

bob-p:
Since #1503 you cannot pass a SimpleXMLElement to slice as you get:

   Catchable fatal error: Argument 1 passed to LimitIterator::__construct() must implement interface Iterator, instance of SimpleXMLElement given

As SimpleXMLElement implements Traversable but not Iterator

This checks that $item is an instance of Iterator before passing to LimitIterator


fabpot:
Thank you @bob-p.

PR #1638, opened 2015-02-27
added PHP 7 nightly to the test matrix and HHVM nightly

fabpot:


stof:
you should exclude the TWIG_EXT builds for HHVM-nighlty and PHP-nightly (as done for HHVM currently)

PR #1641, opened 2015-03-02
deprecated the String loader

fabpot:
Even if the String loader should never be used, people keep trying to use it. It's too confusing to keep and we don't really need it in tests. So, I propose to deprecate it in 1.x and remove it in 2.x

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
 * 1.18.1 (2015-XX-XX)
 
+ * deprecated Twig_Loader_String
  * fixed the slice filter when used with a SimpleXMLElement object
  * fixed filesystem loader when trying to load non-files (like directories)
 
--- a/doc/deprecated.rst
+++ b/doc/deprecated.rst
@@ -101,6 +101,12 @@ Interfaces
   those constants Twig_Template::ANY_CALL, Twig_Template::ARRAY_CALL,
   Twig_Template::METHOD_CALL)
 
+Loaders
+-------
+
+* As of Twig 1.x, ``Twig_Loader_String`` is deprecated and will be removed in
+  2.0.
+
 Globals
 -------
 
--- a/lib/Twig/Loader/String.php
+++ b/lib/Twig/Loader/String.php
@@ -19,6 +19,8 @@
  * source code of the template). If you don't want to see your cache grows out of
  * control, you need to take care of clearing the old cache file by yourself.
  *
+ * @deprecated since 1.18.1 (to be removed in 2.0)
+ *
  * @author Fabien Potencier <fabien@symfony.com>
  */
 class Twig_Loader_String implements Twig_LoaderInterface, Twig_ExistsLoaderInterface
--- a/test/Twig/Tests/EnvironmentTest.php
+++ b/test/Twig/Tests/EnvironmentTest.php
@@ -46,7 +46,7 @@ class Twig_Tests_EnvironmentTest extends PHPUnit_Framework_TestCase
     public function testGlobals()
     {
         // globals can be added after calling getGlobals
-        $twig = new Twig_Environment(new Twig_Loader_String());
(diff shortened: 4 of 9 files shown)


ninsuo:
Hi, can you elaborate the reason why it is deprecated? I did not noticed this anywhere.

fabpot:
I want to deprecate it is too easy to make mistakes when using it. Basically, it does not do what people expect it to do and as such (and because there are better ways), I prefer to remove it. I think I receive at least an email a month about this topic ;)

stof:
:+1: 

Btw, it would be great to fix this intermittent test failure with the profiler dumper test

stof:
@ninsuo you should use ``$env->createTemplate($string)`` to do what people generally think that the Twig_Loader_String will do. The Twig_Loader_String cannot work properly by design (the first version of the ``template_from_string`` function was using it, and it has been changed because it was broken)

ninsuo:
@stof I do not understand, is the following code potentially invalid?

```php
$env = new \Twig_Environment(new \Twig_Loader_String());
echo $env->render('Hello, {{ name }}', array('name' => 'Bob')); // Hello, Bob
```

I understand the bad practice, but what it the technical problem here?

stof:
@ninsuo your example is probably the only code snippet where the usage of string loader will work. But as soon as you make it more complex, it breaks:

- if your code tries to include or extend another template (or any other feature involving other templates, like ``{% embed %}``, ``{% use %}`` or importing macros for instance), using the string loader will break everything
- if you enable the twig cache, using the string loader will not play with it fine

And doing the same in a working way is not that difficult:

```php
$env = new \Twig_Environment(new \Twig_Loader_Array());
$template = $env->createTemplate('Hello, {{ name }}');
echo $env->render(template, array('name' => 'Bob')); // Hello, Bob
```

ninsuo:
Got it, thanks (y)

dawehner:
Drupal basically needs the dynamicness of \Twig_Loader_String for some special usecases,
but yeah I understand why it should be removed. Its a bad idea to start with

ninsuo:
You can do the same as Twig_Loader_String with Twig_Loader_Array, using @stof's syntax above. This is really as flexible as Twig_Loader_String... The quick fix you provided to Drupal will not take any mentioned advantages.

dawehner:
@ninsuo Sure, we could provide workarounds but its tricky, given that we use twig via a generic render system and have the environment stored in the container

WouterJ:
@stof your code does not work ("Catchable Fatal Error: Object of class __TwigTemplate_dd0d0e2c37a2d0f8b67dcdcd30ecb79029586ee23cd1c76f343d6878b2fdbb35 could not be converted to string"). Transforming this into:

```php
$twig = new \Twig_Environment(...);
$template = $twig->createTemplate('Hello {{ name }}!');
echo $template->render(['name' => 'Bob']);
```

.. suffers from the same problem as Twig_Loader_String, as it will create a cache entry each time you call it. Is there really no way to implement this in a nicer way? (e.g. creating a cache name in `createTemplate` based on the contents of the template, so the same contents results in the same cache name?)

fabpot:
@WouterJ If you need a (hopefully) solid implementation, have a look at `Twig_Environment::createTemplate()`: https://github.com/twigphp/Twig/blob/1.x/lib/Twig/Environment.php#L408-L428

WouterJ:
Thanks @fabpot. I discovered that I mixed up the location of the caching (`Env::loadTemplate()` vs in a loader). Things indeed work as expected using `createTemplate.`

fabpot:
I've added a note in b5c7f5c about `createTemplate()`.

WouterJ:
:+1:

fabpot:
I've just pushed a PR that adds a recipe about this topic as this seems to be a question that people ask all the time: twigphp/Twig#1826

PR #1651, opened 2015-03-11
Twig: adding twig extension debug from config file

Herod:
based on this question
http://stackoverflow.com/questions/18362495/twig-adding-twig-extension-debug-from-config-file

### DIFF
--- a/doc/functions/dump.rst
+++ b/doc/functions/dump.rst
@@ -24,6 +24,14 @@ introspecting its variables:
     Even when enabled, the ``dump`` function won't display anything if the
     ``debug`` option on the environment is not enabled (to avoid leaking debug
     information on a production server).
+    
+    Or defined as following in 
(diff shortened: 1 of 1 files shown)


doc/functions/dump.rst
```diff
@@ -24,6 +24,14 @@ introspecting its variables:
...
     ``debug`` option on the environment is not enabled (to avoid leaking debug
     information on a production server).
+    
+    Or defined as following in 
+    # app/config/config.yml
+    services:
+        custom.twig.extension.debug:
+            class: Twig_Extension_Debug
+            tags:
+                 - { name: 'twig.extension' }
```
stof: You are documenting Symfony here, which has nothing to do in the Twig documentation

stof: and btw, TwigBundle already registers it automatically for you

PR #1655, opened 2015-03-16
[2.0] unclosed "verbatim" block should be litteraly written

### LINKED ISSUE #1654: Lexer -> Tag leftover bug
Hi,

On this line there is a reference to `$tag` in the build of the exception message:
https://github.com/twigphp/Twig/blob/master/lib/Twig/Lexer.php#L291
However the `$tag` variable is no longer passed on.
I don't know how to fix it, but I thought to let you you know (on `master` only btw).

ninsuo:
Fixes #1654
cc @SpacePossum


SpacePossum:
:+1:

fabpot:
Thank you @ninsuo.

PR #1659, opened 2015-03-19
Added support for unconsumed macro arguments

### LINKED ISSUE #1159: Add support for unconsumed arguments (varargs and kwargs) in macros

hason:
Fixes #1159 and supplements #1699

### DIFF
--- a/lib/Twig/Node/Macro.php
+++ b/lib/Twig/Node/Macro.php
@@ -16,8 +16,19 @@
  */
 class Twig_Node_Macro extends Twig_Node
 {
+    const VARARGS_NAME = 'varargs';
+
     public function __construct($name, Twig_NodeInterface $body, Twig_NodeInterface $arguments, $lineno, $tag = null)
     {
+        foreach ($arguments as $argumentName => $argument) {
+            if (self::VARARGS_NAME === $argumentName) {
+                throw new Twig_Error_Syntax(sprintf(
+                    'The argument "%s" in macro "%s" cannot be defined because the variable "%s" is reserved for arbitrary arguments',
+                    self::VARARGS_NAME, $name, self::VARARGS_NAME
+                ), $argument->getLine());
+            }
+        }
+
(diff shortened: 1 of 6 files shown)


lib/Twig/Node/Macro.php
```diff
@@ -16,8 +16,19 @@
  */
 class Twig_Node_Macro extends Twig_Node
 {
+    const VARARGS_NAME = 'varargs';
+
     public function __construct($name, Twig_NodeInterface $body, Twig_NodeInterface $arguments, $lineno, $tag = null)
     {
+        foreach ($arguments as $argumentName => $argument) {
+            if ($argumentName === self::VARARGS_NAME) {
```
sstok: Should be a Yoda expression `self::VARARGS_NAME === $argumentName`.

lib/Twig/Node/Macro.php
```diff
@@ -52,30 +63,33 @@ public function compile(Twig_Compiler $compiler)
...
+        $compiler
+            ->write("\$context = \$this->env->mergeGlobals(array(\n")
+            ->indent()
+        ;
 
+        foreach ($this->getNode('arguments') as $name => $default) {
             $compiler
-                ->outdent()
-                ->write("));\n\n")
+                ->write('')
```
sstok: Is there a reason for using an empty string instead of using addIndentation() ?

lib/Twig/Node/Macro.php
hason: Fixed

stof:
for PHP 5.6+, this could be optimized by using a variadic function when compiling the template.

hason:
@stof Good point. Done.

fabpot:
Thank you @hason.

PR #1661, opened 2015-03-25
Fix leaks and tidy api

jpauli:


stof:
I suggest doing this in the 1.x branch as it is still maintained

fabpot:
@stof: it is indeed for the 1.x branch. The 2.x C ext will see many more major changes.

stof:
then the target of the PR is wrong (which also means it does not run against all the PHP versions supported by 1.x)

fabpot:
@jpauli Can you close this PR and resubmit it on the 1.x branch?

stof:
@fabpot should 1.x be the default branch on github for now ?

fabpot:
@stof That's already the case

PR #1664, opened 2015-03-29
Use mocks for profiles. Fixes #1660

### LINKED ISSUE #1660: Volatile test
The profile dumper tests seem to be volatile: https://travis-ci.org/twigphp/Twig/jobs/55327057

matthieuauger:
| Q             | A
| ------------- | ---
| Bug fix?      | yes #1660 
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

This PR aims to change the getProfile() method to returns mocks.
This fixes #1660

BTW Is there a specific reason why there are no mocks for the tests in the project ?

### DIFF
--- a/test/Twig/Tests/Profiler/Dumper/AbstractTest.php
+++ b/test/Twig/Tests/Profiler/Dumper/AbstractTest.php
@@ -13,32 +13,88 @@ abstract class Twig_Tests_Profiler_Dumper_AbstractTest extends PHPUnit_Framework
 {
     protected function getProfile()
     {
-        $profile = new Twig_Profiler_Profile();
-        $index = new Twig_Profiler_Profile('index.twig', Twig_Profiler_Profile::TEMPLATE);
-        $profile->addProfile($index);
-        $body = new Twig_Profiler_Profile('embedded.twig', Twig_Profiler_Profile::BLOCK, 'body');
-        $body->leave();
-        $index->addProfile($body);
-        $embedded = new Twig_Profiler_Profile('embedded.twig', Twig_Profiler_Profile::TEMPLATE);
-        $included = new Twig_Profiler_Profile('included.twig', Twig_Profiler_Profile::TEMPLATE);
-        $embedded->addProfile($included);
-        $index->addProfile($embedded);
-        $included->leave();
-        $embedded->leave();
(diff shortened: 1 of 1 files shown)


test/Twig/Tests/Profiler/Dumper/AbstractTest.php
```diff
@@ -13,32 +13,88 @@
...
-        usleep(500);
-        $embedded->leave();
-
-        usleep(4500);
-        $index->leave();
-
-        $profile->leave();
+        $profile = $this->getMockBuilder('Twig_Profiler_Profile')->disableOriginalConstructor()->getMock();
+
+        $profile->method('isRoot')->willReturn(true);
```
sstok: It seems you are using Prophecy or a newer version of PHPUnit.
Twig still supports PHP 5.2 which means you can't use the last PHPUnit version.
https://travis-ci.org/twigphp/Twig/jobs/56346922

stof: this should be ``->will($this->returnValue(true))``

matthieuauger:
Thanks @stof and @sstok 

Seems that you can't chain method() without expects() on old phpunit versions, the tests are now green. 

Are you ok with this change ? I can use mocks in other tests if needed. We have a gain of rapidity, memory, the tests are less dependant of the implementation of their dependencies, and we can make better assumptions (here we could change the assertRegex by a simple assertEquals since we control the values).

test/Twig/Tests/Profiler/Dumper/AbstractTest.php
```diff
@@ -13,32 +13,88 @@
...
+    {
+        return $this->generateProfile('foo', 0.0001, false, 'macro', 'index.twig', $subProfiles);
+    }
+
+    /**
+     * @param $name
+     * @param $duration
+     * @param $isTemplate
+     * @param $type
+     * @param $templateName
```
stof: your phpdoc is invalid here. types are missing

matthieuauger: Done. Thanks

fabpot:
Thank you @matthieuauger.

PR #1666, opened 2015-03-31
add ignore_missing agrument to source function

vetruvet:
This basically works the same way as `ignore_missing` on the `include` function/tag.

### DIFF
--- a/lib/Twig/Extension/Core.php
+++ b/lib/Twig/Extension/Core.php
@@ -1387,13 +1387,20 @@ function twig_include(Twig_Environment $env, $context, $template, $variables = a
 /**
  * Returns a template content without rendering it.
  *
- * @param string $name The template name
+ * @param string $name           The template name
+ * @param bool   $ignore_missing Whether to ignore missing templates or not
  *
  * @return string The template source
  */
-function twig_source(Twig_Environment $env, $name)
+function twig_source(Twig_Environment $env, $name, $ignoreMissing = false)
 {
-    return $env->getLoader()->getSource($name);
+    try {
+        return $env->getLoader()->getSource($name);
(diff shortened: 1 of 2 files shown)


sstok:
:+1:

lib/Twig/Extension/Core.php
```diff
@@ -1387,13 +1387,20 @@ function twig_include(Twig_Environment $env, $context, $template, $variables = a
...
-function twig_source(Twig_Environment $env, $name)
+function twig_source(Twig_Environment $env, $name, $ignoreMissing = false)
 {
-    return $env->getLoader()->getSource($name);
+    try {
+        return $env->getLoader()->getSource($name);
+    } catch (Twig_Error_Loader $e) {
+        if (!$ignoreMissing) {
+            throw $e;
+        }
```
stof: you should return an empty string when ignoring missing templates

vetruvet: I agree, but I was following convention. In the `twig_include` function (https://github.com/twigphp/Twig/blob/1.x/lib/Twig/Extension/Core.php#L1409), nothing is returned when ignoring missing templates.

lib/Twig/Extension/Core.php
```diff
@@ -1387,13 +1387,20 @@ function twig_include(Twig_Environment $env, $context, $template, $variables = a
 /**
  * Returns a template content without rendering it.
  *
- * @param string $name The template name
+ * @param string $name           The template name
+ * @param bool   $ignore_missing Whether to ignore missing templates or not
```
fabpot: should be `$ignoreMissing`

fabpot:
Thank you @vetruvet.

PR #1667, opened 2015-04-01
[1.x] Updated exception messages for null vars

phansys:
Updated exception messages for null vars in ```Twig_Template::getAttribute()```.

Example:
**Before:**
```
Impossible to invoke a method ("myMethod") on a NULL variable ("") in My:template.html.twig at line 1
```

**After:**
```
Impossible to invoke a method ("myMethod") on a null variable in My:template.html.twig at line 1
```

### DIFF
--- a/lib/Twig/Template.php
+++ b/lib/Twig/Template.php
@@ -419,7 +419,13 @@ abstract class Twig_Template implements Twig_TemplateInterface
                         $message = sprintf('Key "%s" for array with keys "%s" does not exist', $arrayItem, implode(', ', array_keys($object)));
                     }
                 } elseif (Twig_Template::ARRAY_CALL === $type) {
-                    $message = sprintf('Impossible to access a key ("%s") on a %s variable ("%s")', $item, gettype($object), $object);
+                    if (null === $object) {
+                        $message = sprintf('Impossible to access a key ("%s") on a null variable', $item);
+                    } else {
+                        $message = sprintf('Impossible to access a key ("%s") on a %s variable ("%s")', $item, gettype($object), $object);
+                    }
+                } elseif (null === $object) {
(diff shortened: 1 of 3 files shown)


phansys:
Tests are failing because inconsistencies between C ext dump output for ```null``` (PHP code returns ```NULL``` while C ext returns  ```null```). How I should treat this issue? IMO, we could convert output from PHP to lower or C to upper.

phansys:
Some refs:
http://php.net/manual/en/function.gettype.php#refsect1-function.gettype-returnvalues
https://github.com/php/php-src/blob/master/Zend/zend_API.c#L180

lib/Twig/Template.php
```diff
@@ -419,7 +419,13 @@ protected function getAttribute($object, $item, array $arguments = array(), $typ
                         $message = sprintf('Key "%s" for array with keys "%s" does not exist', $arrayItem, implode(', ', array_keys($object)));
                     }
                 } elseif (Twig_Template::ARRAY_CALL === $type) {
-                    $message = sprintf('Impossible to access a key ("%s") on a %s variable ("%s")', $item, gettype($object), $object);
+                    if (null === $object) {
+                        $message = sprintf('Impossible to access a key ("%s") on a %s variable', $item, gettype($object));
```
GromNaN: Here you know that `$object = null`. Then you can write it directly instead of relying on `gettype`.

```php
$message = sprintf('Impossible to access a key ("%s") on a null variable', $item);
```

phansys: You're right @GromNaN. I will update this messages based on the definition between ```NULL``` and ```null```.

phansys:
I just updated the PR with @GromNaN's suggestion. Failed build is not related to this change (see https://travis-ci.org/phansys/Twig/builds/56965336).

fabpot:
Thank you @phansys.

Issue #1644, opened 2015-03-06
Compatibility between Twig 1.last and 2.0 for packages extending Twig


stof:
Currently, Twig 2.0 introduces a few hard BC breaks in classes where the typehints are changed because of the removal of a bunch of interfaces:

-  Twig_NodeVisitorInterface changed signature
- removal of ``Twig_ExistsLoaderInterface`` (making it impossible to write a loader compatible with Twig 2.x+ while being efficient on 1.x as it cannot implement the removed interface even if it implements the method)

These 2 changes mean that it is impossible to write a package depending on these features and supporting both Twig 1.x and 2.x, thus forcing to split the community: once a package migrate to Twig 2.x, you can use it in your project only if *all* other packages you are using are migrated too.
This would be an issue for Symfony for instance, which would have to break the Twig 1.x compatibility at some point when switching to 2.x (forcing the whole Symfony community to switch at the same point and to maintain older versions of their lib for Twig 1.x because of the Symfony LTS).

It would be much better to provide BC/FC layers for these:

- keeping the ``Twig_ExistsLoaderInterface`` for now (as an empty interface extending ``Twig_LoaderInterface`` so that loaders can be compatible with Twig 1.x in an efficient way (the ``exists`` method has been added to improve performance), and removing it in 3.0 only (as done for ``Twig_Simple(Filter|Test|Function)``)
- for the node visitor typehint change, a way to handle it would be to provide a base node visitor class in Twig 1.19 which would implement the methods and delegate the logic to protected methods using the Twig_Node typehint, and updating the public methods in 2.x to change the typehint. This way, a library could be compatible with both 1.19+ and 2.x by implementing the protected methods instead of the public ones.

this would make the adoption of Twig 2.x much easier by allowing extension packages to support both versions without branching different versions of themselves (which would either require abandonning the 1.x support or abandonning semver to add features in 2 versions in parallel based on Twig versionning)

Green-Cat:
:+1:

joelpittet:
👍

fabpot:
Anyone willing to implement those changes?

joelpittet:
I'm not super strong on the internals, but I'll try to send some pull requests to get there.

joelpittet:
[...]

PR #1673, opened 2015-04-18
Add Twig_ExistsLoaderInterface back for 1.x BC layer #1644

joelpittet:
This is the 2.x change for 1.x BC layer of Twig_ExistsLoaderInterface for #1644

### DIFF
--- a/lib/Twig/ExistsLoaderInterface.php
+++ b/lib/Twig/ExistsLoaderInterface.php
@@ -1,0 +1,24 @@
+<?php
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) 2009 Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+/**
+ * Adds an exists() method for loaders. Extends Twig_LoaderInterface for loaders
+ * can be compatible with Twig 1.x and 2.x.
+ *
+ * 1.x BC layer.
+ *
+ * @author Florin Patan <florinpatan@gmail.com>
+ *
+ * @deprecated since 1.12 (to be removed in 3.0)
+ */
+interface Twig_ExistsLoaderInterface extends Twig_LoaderInterface
+{
+}


lib/Twig/ExistsLoaderInterface.php
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * Adds an exists() method for loaders. Extends Twig_LoaderInterface for loaders
+ * can be compatible with Twig 1.x and 2.x.
+ *
+ * 1.x BC layer.
+ *
+ * @author Florin Patan <florinpatan@gmail.com>
+ *
+ * @deprecated since 1.12 (to be removed in 3.0)
```
stof: it is not deprecated in 1.12 but in 2.0 (in 1.12, it is the only way to get this feature)

joelpittet: Thanks, I thought so but I copied that from what is in 1.x, likely need to change in both branches?

stof:
:+1:

lib/Twig/ExistsLoaderInterface.php
stof: yeah, the deprecation message in 1.x looks wrong too then.

lib/Twig/ExistsLoaderInterface.php
```diff
@@ -0,0 +1,24 @@
...
+ * This file is part of Twig.
+ *
+ * (c) 2009 Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+/**
+ * Adds an exists() method for loaders. Extends Twig_LoaderInterface for loaders
```
Tobion: language problem: "for loaders can be"

stof: this does not add this method anymore :)

fabpot:
Thank you @joelpittet.

PR #1677, opened 2015-04-24
Remove docblock about an argument that does not exists.

yvoyer:
The argument do not exists anymore, no point in keeping it.


stof:
#1678 includes this change, alongside lots of other phpdoc fixes.

PR #1679, opened 2015-04-25
[feature] Line whitespace control

brunoais:
This PR introduces line whitespace control character `~` which can be used the same way standard whitespace control character `~` can be, for example:

```
{{~ "Trim to my left, but only until first line break" }}
{%~ for word in ["Trim to my left, but only until first line break"] %}
```

Based on PR by loostro #1569
Except this one properly handles the use on the left and the right side of the tags.


brunoais:
If anyone can give a close look to the tests, it would be great. It's the first time I write texts in this syntax so I'm clueless if I'm doing it right. I only know it passes.

brunoais:
I think I got to understand how tests work. Please do some final checks.

As a final note, there's some availability for extra functionality for this operation as follows:

- If a tag is at the leftmost of the line (the character to the left of a tag, e.g. `{%~` or `{{~`, is a newline (`\n`) character, then remove that newline character and the rest of the previous line from right to left until a  non-newline whitespace character is found (I.e.

	abc  
	
	{%\~...}

Becomes:

	abc  
	

Instead of:

	abc  
	
	

As it is now.
)
Edit: (Sheesh, Markdown doesn't help with whitespace... I can't show you the example).

-  If a tag is at the rightmost most of the line (the character to the right of a tag, e.g. `~%}` or `~}}`, is a newline (`\n`) character, then remove that newline character and the rest of the previous line from right to left until a non-newline whitespace character is found.

In practice, what these means that the match made for them is (simplified form):
`\s[ \t\r]*`, instead of `[ \t\r]*`
and
`[ \t\r]*\s\z` instead of `[ \t\r]*\z`
, respectively.

brunoais:
Any updates?

brunoais:
@fabpot  Any updates on the review?

brunoais:
@fabpot If I adapt this to 2.x, will it be included in Twig?

brunoais:
@fabpot Still waiting for an answer.

> If I adapt this to 2.x, will it be included in Twig?

PR #1686, opened 2015-05-07
Fix two failed unit tests on Windows:

johnnytemp:
1. fopen() cannot handle directory, replaced by opendir() for getting a testing resource. (also double check a resource is opened)
2. comparing paths by assertEquals() which has mixed '\' and '/' on Windows.


fabpot:
Thank you @johnnytemp.

SpacePossum:
:+1:  thanks @johnnytemp

PR #1690, opened 2015-05-13
Add an anchor to a link

Zzortell:


stof:
using a Sphinx ref to point to the right place would be cleaner IMO.

Zzortell:
I'm not sure of the syntax, I don't know rst nor Sphinx, so please check my commit.

fabpot:
fixed in df9da8e8

PR #1693, opened 2015-05-18
Method "hasLoader()" added to a Twig_Environment

TheCelavi:


TheCelavi:
This method should improve convenience of setting/adding/modifying loader later on during the execution depending of the fact that Environment has or has not Loader defined.

Now, this is possible only by doing this:

    try {
        $hasLoader = (bool) $env->getLoader();
    } catch(LogicException $e) {
        $hasLoader = false;
    }

PR does not cause any BC breaks on the table.

Tobion:
:-1: it doesn't add anything that is not already possible. And the idea is that the evironment always has a loader to be usable. So exposing a hasLoader method would imply wrong semantics (namely that the loader is optional, which it isn't).

TheCelavi:
@Tobion The very constructor of Environment allows for loader to be omitted, lib/Twig/Environment.php, line no. 85:

    public function __construct(Twig_LoaderInterface $loader = null, $options = array()) {
        ....
    }

which means that loader can be added after construction of the Environment -> ergo, something can happen in between. 

The purpose of 'hasLoader' is to enable introspection of this occurrence when loader is not provided in construction process. I do not have any other way to check that except with try/catch block.

Or, if that semantic 'loader is not optional' is so important - why constructor allows for loader to be optional?

stof:
IMO, Twig 2.0 should make the loader mandatory.

TheCelavi:
It is not mandatory now. So 'getLoader' can/will throw exception if loader is not provided, and we can determine if loader is provided without try/catch block.

fabpot:
Closing for the various reasons explained by other commenters.

TheCelavi:
Reason given here is as follows: "exposing a hasLoader method would imply wrong semantics (namely that the loader is optional, which it isn't)".

However, constructor allows loader to be null:

    public function __construct(Twig_LoaderInterface $loader = null, $options = array()) {
        ....
    }


Given reason contradicts with current constructor implementation. I have checked: https://github.com/twigphp/Twig/blob/1.x/lib/Twig/Environment.php, constructor signature is not changed.

Does that means that constructor will be changed, should someone create PR with new constructor signature (or I can do it as well)?

Or - things will be left as they are and wether loader is provided or not can be examined only in try/catch block, e.g:

    try {
        $hasLoader = (bool) $env->getLoader();
    } catch(LogicException $e) {
        $hasLoader = false;
    }

PR #1699, opened 2015-06-02
Added variadic filters, tests, and functions

hason:
Replaces #1317

```php
$env->addFunction(new Twig_SimpleFunction('foo', function ($a1, array $args = array()) {
}, array('is_variadic' => true));
```
```jinja
{{ foo(1, 2, a="a", b="b") }}
{# foo(1, array(0 => 2, "a" => "a", "b" => "b")); #}
```


fabpot:
Thank you @hason.

PR #1702, opened 2015-06-02
[Docs] [Tags] [Spaceless] - Fixed link to whitespace control section

ureimers:
| Q             | A
| ------------- | ---
| Doc fix?      | yes
| New docs?     | no
| Fixed tickets | none

The link to the whitespace control section was just pointing to the whole templates documentation instead of deep linking to the whitespace section. I added a label for the whitespace control section and referenced it from the spaceless documentation.
I also changed the linked text from "dedicated" to "dedicated section" to make the link an actual word.

I hope the syntax is correct. I double checked with Sphinx' online documentation and the existing references in the Symfony documentation as I don't have Sphinx installed on my system.


ureimers:
Hmm... don't know why my old (and already merged) commits are in this PR again... Hope that doesn't break anything.

stof:
This is because your master branch was not matching the twigphp one. You should not create PRs from the twigphp repo to your fork to update the master branch. and btw, you should generally avoid using the master branch to send PRs. It makes things a pain when a PR got rejected, when a PR gets squashed before merging (because your master branch diverges from the upstream one in these cases) or when you want to open multiple PRs in parallel.

ureimers:
Thanks Christophe. I fear I might have to read up big time on how to use GIT correctly. I've been using SVN since I don't know when and GIT(HUB) is absolutely not my turf.

So I'll use the 1.x branch then. I tried to create one PR for the two files I changed, thinking that only the files, that actually differ from the base get commited. But that seems to base on a misunderstanding on how GIT works.

The thing is: I'd really like to help improve Symfony and its components but all this GIT mumbo jumbo is a real big hurdle. There are many things to do wrong and I admit that it's a bit frightening. http://symfony.com/doc/current/contributing/code/patches.html is a nice start but for small fixes like this one or typos I think it would really help to create a small tutorial for the GitHub web frontend (which is what I've been using, with more or less success...).

Well, as I said, I'll try to read up on how to work with GIT correctly.

Is there anything I can do right now, to clean things up?

fabpot:
@ureimers No worries, I will clean things up for you.

fabpot:
merged, thanks

stof:
@ureimers would need to clean the master branch of his fork though, to avoid similar issue in the future

ureimers:
@stof: Wouldn't it be the easiest way if if simply delete my fork for now? All of my changes have been merged and I don't need any history for them.

stof:
if everything is merged, yeah, it could be the easiest way

ureimers:
Done. Thanks for your help.

PR #1703, opened 2015-06-02
Added note about case sensitivity of logic operators

biozshock:


fabpot:
Thank you @biozshock.

PR #1704, opened 2015-06-02
Added twig keyword to composer.json

hason:
See https://packagist.org/search/?tags=twig


Tobion:
The library is called twig. No point in adding the name as keyword :-1:

PR #1706, opened 2015-06-06
[WIP] Added renderString method 

peterrehm:
As the String loader has been deprecated I think it makes sense to offer out of the box a way to render strings with twig as @stof suggested here https://github.com/twigphp/Twig/pull/1641#issuecomment-76732213.

This is a wip to find out if this is generally wanted and if so I would have a closer look at it and would provide also the proper documentation. I think it would be great to have a string renderer by default.

The symfony2 bridge would need to be updated accordingly.

### DIFF
--- a/lib/Twig/Environment.php
+++ b/lib/Twig/Environment.php
@@ -294,6 +294,24 @@ class Twig_Environment
     }
 
     /**
+     * Renders a string.
+     *
+     * @param string $string  The template
+     * @param array  $context An array of parameters to pass to the template
+     *
+     * @return string The rendered template
+     *
+     * @throws Twig_Error_Syntax  When an error occurred during compilation
+     * @throws Twig_Error_Runtime When an error occurred during rendering
+     */
+    public function renderString($string, array $context = array())
+    {
+        $template = $this->createTemplate($string);
+        
+        return $template->render($context);
+    }
+
+    /**
      * Displays a template.
      *
      * @param string $name    The template name


stof:
I don't think this method is really needed in the Twig_Environment itself.

peterrehm:
@stof Don't you think this makes this easier accessible? There are easy workarounds around that but looking at other implementations this might guide the people into the right direction for such a need. 

Or at least we should document the proper way you suggested.

weaverryan:
So is the way @stof described at https://github.com/twigphp/Twig/pull/1641#issuecomment-76732213 the new proper way of rendering a string? Then having a small section on http://twig.sensiolabs.org/doc/recipes.html might do it.

peterrehm:
@weaverryan And in addition to that a brief section in the symfony docs of how to approach this (wrapped into another service) if that is what we want to show. This way would just be more convenient.

I am actually not sure if the snippet from @stof works, I assume it should be as follows:

````php
$env = new \Twig_Environment(new \Twig_Loader_Array());
$template = $env->createTemplate('Hello, {{ name }}');
echo $template->render(array('name' => 'Bob')); // Hello, Bob
````

Calling the render method on the template rather than going through loadTemplate() again. This caused some issues when I tried around that.

fabpot:
Closing this one as the `createTemplate()` method is enough.

fabpot:
see 4e58c48

PR #1708, opened 2015-06-07
Bugfix/constructor should not have optional loader

TheCelavi:
Following #1693 and the fact that "hasLoader" contains semantical error that loader is optional, this PR removes optionality of loader from Environment.

### DIFF
--- a/test/Twig/Tests/EnvironmentTest.php
+++ b/test/Twig/Tests/EnvironmentTest.php
@@ -11,16 +11,6 @@
 
 class Twig_Tests_EnvironmentTest extends PHPUnit_Framework_TestCase
 {
-    /**
-     * @expectedException        LogicException
-     * @expectedExceptionMessage You must set a loader first.
-     */
-    public function testRenderNoLoader()
-    {
-        $env = new Twig_Environment();
-        $env->render('test');
-    }
-
     public function testAutoescapeOption()
     {
         $loader = new Twig_Loader_Array(array(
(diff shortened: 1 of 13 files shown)


TheCelavi:
I just realised that Tests are written as if loader is optional (even though it is claimed that it isn't - more over, there is a test that actually tests if loader is provided) so I will update this PR with fixed tests, sorry...

test/Twig/Tests/EnvironmentTest.php
```diff
@@ -11,16 +11,6 @@
 
 class Twig_Tests_EnvironmentTest extends PHPUnit_Framework_TestCase
 {
-    /**
```
TheCelavi: Test does not makes any sense anymore.

TheCelavi:
Sorry about number of commits -> I was to lazy to setup testing env :)

stof:
not an issue. @fabpot has a tool which can squash commits together automatically when deciding to merge a PR (the only drawback is that the PR is marked as closed rather than merged because it is not the same commit anymore)

fabpot:
@TheCelavi Can you rebase on current master? Rebasing will fix the conflicts, but it will also remove your merge commit that we need to remove before being able to merge this PR. Thanks.

TheCelavi:
As soon as I get back home, I am on vacation, sorry...

TheCelavi:
Ok, that should do...

PR #1713, opened 2015-06-16
SCA, CS and exception message tweaks

SpacePossum:


lib/Twig/Node/CheckSecurity.php
```diff
@@ -41,9 +41,9 @@ public function compile(Twig_Compiler $compiler)
         }
 
         $compiler
-            ->write("\$tags = ")->repr(array_filter($tags))->raw(";\n")
-            ->write("\$filters = ")->repr(array_filter($filters))->raw(";\n")
-            ->write("\$functions = ")->repr(array_filter($functions))->raw(";\n\n")
+            ->write('$tags = ')->repr(array_filter($tags))->raw(";\n")
+            ->write('$filters = ')->repr(array_filter($filters))->raw(";\n")
+            ->write('functions = ')->repr(array_filter($functions))->raw(";\n\n")
```
stof: missing ``$``

fabpot:
I'm against making so many changes without a real benefit (especially the change from " to ' which I have rejected many times already).

lib/Twig/TokenStream.php
```diff
@@ -66,6 +68,9 @@ public function next()
     /**
      * Tests a token, sets the pointer to the next one and returns it or throws a syntax error.
      *
+     * @param int|string $primary
```
stof: how can it be a string ? It is passed to ``test()`` which expects an integer according to you phpdoc. One of them is wrong

SpacePossum:
>I'm against making so many changes without a real benefit (especially the change from " to ' which I have rejected many times already).

Sorry I missed that, how come this fixer ended up on the Symfony level?
"single_quote [symfony]
Convert double quotes to single quotes for simple strings." 

Is it OK if I make a new PR with the exception detail changes?

fabpot:
@SpacePossum Using single quotes for PHP files is fine, but CS do not apply to Twig and PHP-generated code.

stof:
Switching to strict comparisons in places where it is done here is a good idea though

fabpot:
@stof What's the benefit?

stof:
@fabpot consistency in the codebase, and avoiding unwanted conversions of variables (which is why the Symfony uses ``===`` as consistent convention, not ``==``).

and btw, given that fabbot runs on this repo with the same config that it runs on Symfony, it would be useful to update the Twig codebase to follow all rules followed in Symfony (and enforced by fabbot). Otherwise any contributor will then have fabbot failures asking for changes in files they modified because the existing code in these files does not match the fabbot expectations (and the usage of single quotes would deserve to be fixed for this reason too btw). Having the codebase fixed at once seems better than having each contributor adding such changes in their patches to make fabbot happy.
@keradus has not sent PRs to Twig to fix existing CS inconsistencies as he did for Symfony once the PHP-CS-Fixer was able to catch them, meaning we still have them in the code.

PR #1714, opened 2015-06-19
Adding a tip on the merge case with a variable

mikeSimonson:


doc/filters/merge.rst
```diff
@@ -40,6 +40,20 @@ overridden.
...
+.. tip::
+
+    If the values or keys you want to merge are variable, you need to enclose them into parenthesis :
+
+    .. code-block:: jinja
+        {% set mergedItems = {} %}
+        {% set items = { 'apple': 'fruit', 'orange': 'fruit' } %}
+
+        {% for key, value in items %} 
+            {% set mergedItems = mergedItems|merge({ (key): (value) }) %}
```
fabpot: Parenthesis are only needed for keys and as this is already documented when we talk about arrays, I don't see the need to repeat it here.

mikeSimonson: Ok. Care on a note about the array_merge particular behavior with numeric variable as key ?

fabpot: It's already implicit in the note below about the fact that `array_merge` is used behind the scene. But we can indeed add a sentence in the note about how that works. Can you do that?

mikeSimonson: done in #1715

PR #1715, opened 2015-06-22
Example of the weird behavior of array_merge

mikeSimonson:
Show how to keep a numerical index if needed.
related to #1714

### DIFF
--- a/doc/filters/merge.rst
+++ b/doc/filters/merge.rst
@@ -42,6 +42,25 @@ overridden.
         
 .. note::
 
-    Internally, Twig uses the PHP `array_merge`_ function.
+    Internally, Twig uses the PHP `array_merge`_ function, which means that it will loose any numerical index after a merge, unless the numerical index is concatenated with something to make a string.
+    
+    .. code-block:: jinja
+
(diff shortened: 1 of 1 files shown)


doc/filters/merge.rst
```diff
@@ -42,6 +42,25 @@ overridden.
         
 .. note::
 
-    Internally, Twig uses the PHP `array_merge`_ function.
+    Internally, Twig uses the PHP `array_merge`_ function, which means that it will loose any numerical index after a merge, unless the numerical index is concatenated with something to make a string.
```
stof: which means it is **not** a numerical index anymore at all, but a string index :smile:

mikeSimonson: right, how would you phrase that so that it doesn't sound to convoluted.

You can keep the index by concatenating it to something but you cannot keep it numeric ( in one step ) ?

You can keep the index value by concatenating it to something to transform it into a string, but you can't keep the numeric character of the index.

...

fabpot: As we already link to the array_merge documentation, I don't think we need to duplicate part of the behavior here.

PR #1720, opened 2015-06-29
Example about how to rename several blocks

pyguerder:
It was unclear to me how to rename more than 1 imported block. I had to look into the source code to find that, so I add it to the documentation.


fabpot:
Thank you @pyguerder.

PR #1721, opened 2015-07-01
Added bitwise shift operators

zorn-v:


stof:
I don't think these really belong to Twig core.

fabpot:
I agree with @stof, that does not belong to core.

zorn-v:
How about https://github.com/fabpot/Twig/commit/c2c01dedf864fbd0a3811ba76ad933b6cb0290c4#diff-0 ?
Bitwise operators without shifting is nonsense for me.

Anyway, can you merge it in extensions at least ?
https://github.com/twigphp/Twig-extensions/pull/141

stof:
@zorn-v the ``b-or`` filter is useful to build the binary flag for the ``json_encode`` options for instance

PR #1723, opened 2015-07-02
Update batch filter docs with arguments

alexeymalihin:
Changes based on https://github.com/twigphp/Twig/blob/1.x/test/Twig/Tests/Fixtures/filters/batch_float.test

### DIFF
--- a/doc/filters/batch.rst
+++ b/doc/filters/batch.rst
@@ -44,5 +44,5 @@ The above example will be rendered as:
 Arguments
 ---------
 
-* ``batch size``:   Number of items in one batch. If fractional number used - will be rounded up.
+* ``batch size``: Count in one batch. Fractional number will be rounded up.
 * ``placeholder``: Used to fill in missing items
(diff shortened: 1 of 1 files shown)


SpacePossum:
Hi,
I'm sure, but I think the limit of a line is 79 characters for rst files (sphinx).
Does the spacing between 'size``:   Number'  need to be multiple spaces?

alexeymalihin:
@SpacePossum, thanks, fixed.

SpacePossum:
:+1:   
(I meant to type `I'm not sure`. .. .didn't mean to sound like a know all ;) )

doc/filters/batch.rst
```diff
@@ -40,3 +40,9 @@ The above example will be rendered as:
             <td>No item</td>
         </tr>
     </table>
+
+Arguments
+---------
+
+* ``batch size``: Count in one batch. Fractional number will be rounded up.
```
fabpot: They keys here must be the ones you can use in a template, so `size` and `fill`. I'm going to fix that after merging.

fabpot:
Thank you @alexeymalihin.

doc/filters/batch.rst
fabpot: 7823014

PR #1724, opened 2015-07-02
fix batch filter with zero items

Tobion:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

also added a test for preserving keys behavior

### DIFF
--- a/lib/Twig/Extension/Core.php
+++ b/lib/Twig/Extension/Core.php
@@ -1488,7 +1488,7 @@ function twig_array_batch($items, $size, $fill = null)
 
     $result = array_chunk($items, $size, true);
 
-    if (null !== $fill) {
+    if (null !== $fill && !empty($result)) {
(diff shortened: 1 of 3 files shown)


lib/Twig/Extension/Core.php
```diff
@@ -1488,7 +1488,7 @@ function twig_array_batch($items, $size, $fill = null)
 
     $result = array_chunk($items, $size, true);
 
-    if (null !== $fill) {
+    if (null !== $fill && !empty($result)) {
         $last = count($result) - 1;
         if ($fillCount = $size - count($result[$last])) {
```
Tobion: accessing `$result[$last]` was the bug when the batch is empty.

fabpot:
Thank you @Tobion.

PR #1725, opened 2015-07-03
fixed sandbox disabling when using the include function

### LINKED ISSUE #1668: twig_include does not disable the sandbox properly after including a sandboxed template
The code disabling the sandbox in the twig_include function can only be reached for not found templates: https://github.com/twigphp/Twig/blob/67920147a81d5bd5bcd052b239ff85bae789f5a3/lib/Twig/Extension/Core.php#L1409-1434

fabpot:
closes #1668

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,6 @@
 * 1.18.3 (2015-XX-XX)
 
- * n/a
+ * fixed sandbox disabling when using the include function
 
 * 1.18.2 (2015-06-06)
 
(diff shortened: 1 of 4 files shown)


stof:
is it possible to add a test to avoid regressions ?

fabpot:
What do you mean, I added two new tests to cover the issue I've fixed.

stof:
hmm, missed them because the github UI loading the diff in chunks in AJAX was lagging when I looked at it previously

PR #1726, opened 2015-07-03
Twig_Template::getContext() removal

fabpot:
`Twig_Template::getContext()` is an implementation detail as it is only used on some PHP versions, for some cases. But even with the big warning in the phpdoc, people want to use it, which is a really bad idea. Removing it in 1.x would be kind of a "BC break", so I propose to remove it in 2.0.

### DIFF
--- a/lib/Twig/Node/Expression/Name.php
+++ b/lib/Twig/Node/Expression/Name.php
@@ -43,18 +43,31 @@ class Twig_Node_Expression_Name extends Twig_Node_Expression
                 ->raw(']')
             ;
         } else {
-            $compiler
-                ->raw('(isset($context[')
-                ->string($name)
-                ->raw(']) ? $context[')
-                ->string($name)
-                ->raw('] : ')
-            ;
-
             if ($this->getAttribute('ignore_strict_check') || !$compiler->getEnvironment()->isStrictVariables()) {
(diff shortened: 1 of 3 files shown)


fabpot:
The new way is even slightly better as we know the expcetion line number ahead of time (not that it matters as the code is mainly useful for dev envs).

stof:
couldn't the ``array_key_exists`` be added in the compiled template in 1.x too, to avoid calling ``getContext()`` for ``null`` variable ?

Btw, given that the method is final, it is already not an extension point. so we could add the new ``notFound`` method in 1.x, stop using ``getContext`` (no BC break as the logic in it cannot change) and deprecate the ``getContext`` method for anyone using it.

stof:
hmm, no. On PHP < 5.4, we use ``getContext`` for any access, not only for the case of not found, so we cannot stop using the method on 1.x

stof:
@fabpot I suggest tagging the method as ``@internal`` though. This way, IDEs like PhpStorm will warn you if you call the method in your code.

stof:
@fabpot I'm worndering whether using this notFound function for PHP 5.4+ on Twig 1.x could be done. It would simplify error reporting by avoiding the need to guess the line for errors.

PR #1728, opened 2015-07-05
removed some test hacks

fabpot:


test/Twig/Tests/Node/Expression/CallTest.php
```diff
@@ -91,12 +91,16 @@ public static function customStaticFunction($arg1, $arg2 = 'default', $arg3 = ar
...
+    private function getArguments($call, $args)
+    {
+        $m = new ReflectionMethod($call, 'getArguments');
+        $m->setAccessible(true);
+
+        return $m->invokeArgs($call, $args);
+    }
 }
 
 class Twig_Tests_Node_Expression_Call extends Twig_Node_Expression_Call
```
stof: could this class be removed entirely then and use ``Twig_Node_Expression_Call`` ?

fabpot: No, because it's an abstract class.

PR #1729, opened 2015-07-05
moved protected to private

### LINKED ISSUE #1645: [2.0] Switch internal methods to private rather than protected
Currently, Twig is using the protected visibility almost everywhere, which means that any internal refactoring is technically a BC break.
Many of these protected members are not meaningful extension points at all. So I think that Twig 2.0 might be a good time to switch them to private (following the same rules than in Symfony, and making it much easier to follow semver by reducing the surface subject to BC)

fabpot:
Moved some protected properties/methods from protected to private.

I've changed protected properties/methods mostly on internal classes but not where it could potentially break BC too badly with existing code.

Classes where we could change protected to private but where it could be a problem for existing code:

* Twig_ExpressionParser
* Twig_Environment
* Twig_Lexer
* Twig_Node
* Twig_Parser (done in second commit)
* Twig_Template

closes #1645

### DIFF
--- a/lib/Twig/Compiler.php
+++ b/lib/Twig/Compiler.php
@@ -17,14 +17,14 @@
  */
 class Twig_Compiler
 {
-    protected $lastLine;
-    protected $source;
-    protected $indentation;
-    protected $env;
-    protected $debugInfo;
-    protected $sourceOffset;
-    protected $sourceLine;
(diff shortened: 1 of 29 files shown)


SpacePossum:
:+1:

stof:
:+1:

PR #1731, opened 2015-07-06
Add example of creating a template from string

simshaun:
From https://github.com/twigphp/Twig/pull/1641, I thought this would be nice to have in the docs.

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -195,6 +195,13 @@ of strings bound to template names::
 This loader is very useful for unit testing. It can also be used for small
 projects where storing all templates in a single PHP file might make sense.
 
+``Twig_Loader_Array`` can also be used in conjunction with ``Twig_Environment``
+to create a template from a string::
+
+    $env = new \Twig_Environment(new \Twig_Loader_Array());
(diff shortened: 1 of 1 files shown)


doc/api.rst
```diff
@@ -195,6 +195,13 @@ of strings bound to template names::
 This loader is very useful for unit testing. It can also be used for small
 projects where storing all templates in a single PHP file might make sense.
 
+``Twig_Loader_Array`` can also be used in conjuction with ``Twig_Environment``
+to create a template from a string::
+
+    $env = new \Twig_Environment(new \Twig_Loader_Array());
+    $template = $env->createTemplate('Hello, {{ name }}');
+    echo $env->render($template, array('name' => 'Bob')); // Hello, Bob
```
Tobion: according to the phpdoc of the Twig environment this is not allowed. `render` expects a string, but `createTemplate` returns a Twig_Template. so if this is the recommended way of doing things, then the phpdocs must be updated.

simshaun: You're correct. I see there is further discussion about it on https://github.com/twigphp/Twig/pull/1706.

The example given there does work.

fabpot:
see 4e58c48

PR #1732, opened 2015-07-09
Missing quotes surrounding "foo"

chaoszcat:
If no quotes, this sample code will only print "I like and bar"

### DIFF
--- a/doc/filters/replace.rst
+++ b/doc/filters/replace.rst
@@ -6,7 +6,7 @@ The ``replace`` filter formats a given string by replacing the placeholders
 
 .. code-block:: jinja
 
-    {{ "I like %this% and %that%."|replace({'%this%': foo, '%that%': "bar"}) }}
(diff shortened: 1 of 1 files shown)


ninsuo:
I would add `{% set foo = "foo" %}` instead, just to show that we can use variables with `replace`

chaoszcat:
@ninsuo nice. That would be good too. But this patch is just a direct fix on what's not working :)

doc/filters/replace.rst
```diff
@@ -6,7 +6,7 @@ The ``replace`` filter formats a given string by replacing the placeholders
 
 .. code-block:: jinja
 
-    {{ "I like %this% and %that%."|replace({'%this%': foo, '%that%': "bar"}) }}
+    {{ "I like %this% and %that%."|replace({'%this%': "foo", '%that%': "bar"}) }}
```
fabpot: That's wrong. We do want a var here (see the explanation just below)

chaoszcat: Oops sorry didn't notice the purple comment below

PR #1741, opened 2015-07-20
Update copyright year

bocharsky-bw:


fabpot:
Good catch, thanks @bocharsky-bw.

PR #1743, opened 2015-07-25
Repair may occur under the windows "\ \" file separator

a939638621:
I just had a program running on Linux, but my web PHP was developed in windwos, and the following is bug:
```
Fatal error: Uncaught exception 'Twig_Error_Loader' with message 'The "/home/a939638621xan9t3g9y603j876s2x1/wwwroot\vendor\symfony\symfony\src\Symfony\Bundle\FrameworkBundle/Resources/views" directory does not exist.' in /home/a939638621xan9t3g9y603j876s2x1/wwwroot/vendor/twig/twig/lib/Twig/Loader/Filesystem.php:93 Stack trace: 
#0 /home/a939638621xan9t3g9y603j876s2x1/wwwroot/app/cache/prod/appProdProjectContainer.php(1080): Twig_Loader_Filesystem->addPath('/home/a93963862...', 'Framework') 
#1 /home/a939638621xan9t3g9y603j876s2x1/wwwroot/app/bootstrap.php.cache(2139): appProdProjectContainer->getTwig_LoaderService() 
#2 /home/a939638621xan9t3g9y603j876s2x1/wwwroot/app/cache/prod/appProdProjectContainer.php(1045): Symfony\Component\DependencyInjection\Container->get('twig.loader') 
#3 /home/a939638621xan9t3g9y603j876s2x1/wwwroot/app/bootstrap.php.cache(2139): appProdProjectContainer->getTwigService()
 #4 /home/a939638621xan9t3g9y603j876s2x1/wwwroot/app/cache/prod/appProdProjectContainer.php(860): Symfony\Component\DependencyIn in /home/a939638621xan9t3g9y603j876s2x1/wwwroot/vendor/twig/twig/lib/Twig/Loader/Filesystem.php on line 93
```
> Then find the source code and fix it!! So the merger request

PR #1744, opened 2015-07-25
Repair may occur under the windows "\ \" file separator

a939638621:
#1743  

Encoding has become the standard.

### DIFF
--- a/lib/Twig/Loader/Filesystem.php
+++ b/lib/Twig/Loader/Filesystem.php
@@ -89,6 +89,7 @@ class Twig_Loader_Filesystem implements Twig_LoaderInterface, Twig_ExistsLoaderI
         // invalidate the cache
         $this->cache = array();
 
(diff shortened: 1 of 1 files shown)


a939638621:
Why？？？ where it does not meet the standard encoding？？？？

ninsuo:
' != "

ninsuo:
By the way there are few chances your hack will be merged without any good reason to do it.

a939638621:
According to you said the removal of the cache and then uploaded to the Linux from the windowsNT, and did not solve the problem,
I still have to go to the source file,
If you don't believe, you can try it yourself!!
I was executing the following command to clear the cache!!

```bash
php app/console cache:clear
```

a939638621:
Or you own the code to write, a quick release of the new version of repair!! Sincere thanks

ninsuo:
Try: `php app/console cache:clear --env=prod`

PR #1745, opened 2015-07-29
capitalize_first filter: Makes string's first character uppercase

pazulx:
There is demand for filter that convert to uppercase only first character of a string. This is my first commit to open source project ever so if I'm doing something not `the right way` please tell me.

Example (in comparision to other string filters):

    {% set var = 'i love HTML' %}
    {{ dump(var|title) }} => output: "I Love Html" // all first letters of each words are converted to uppercase
    {{ dump(var|capitalize) }} => output: "I love html" // all other characters are converted to lowercase
    {{ dump(var|capitalize_first) }} => output: "I love HTML" // only first character is converted to uppercase

### DIFF
--- a/doc/filters/capitalize_first.rst
+++ b/doc/filters/capitalize_first.rst
@@ -1,0 +1,10 @@
+``capitalize first``
+==============
+
+The ``capitalize_first`` filter capitalizes a first character of a string.
+
+.. code-block:: jinja
+
+    {{ 'i love HTML'|capitalize_first }}
+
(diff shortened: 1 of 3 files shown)


iltar:
I'm not sure if this is the best solution, I can see a lot of people needing this, don't get me wrong. However, this is perfectly achievable with CSS already by using text-transform.

fabpot:
Closing this one as there is no need for it to be in core (we already have quite a few built-in features), and it can be easily added in a custom extension. Thanks for proposing.

PR #1748, opened 2015-08-04
Add writeln method to make syntax clean

c9s:
We currently write a lot of code below in compile method of Node classes:

```
$compiler->write("\$context[ ... ]\n");
```

And the double quotes are just for the "\n" charactor, and we have to escape all dollar signs, which makes the syntax not clean enough, however if we can write:

```
$compiler->writeln('$context[....]');
```

In the above example, we don't have to escape the dollar sign and it's better for readability.


c9s:
PR updated.  cc @fabpot

ninsuo:
:+1:

lyrixx:
I thinks this is a nice feature, so could you update existing code ?

c9s:
Hi @lyrixx, sure

c9s:
@lyrixx I simplified some `write()` method with `writeln()`, however I want to prevent code conflicts (there are many pending pull requests and branches), so I didn't change too much code in the current code base. but as you can see the expressions is now clearer.

lib/Twig/Node/Block.php
```diff
@@ -31,14 +31,14 @@ public function compile(Twig_Compiler $compiler)
...
-            ->write(sprintf("public function block_%s(\$context, array \$blocks = array())\n", $this->getAttribute('name')), "{\n")
+            ->writeln(sprintf('public function block_%s($context, array $blocks = array())', $this->getAttribute('name')), '{')
             ->indent()
         ;
 
         $compiler
             ->subcompile($this->getNode('body'))
             ->outdent()
-            ->write("}\n\n")
+            ->writeln('}')
```
SpacePossum: missing the second line break, should this be something like done here https://github.com/twigphp/Twig/pull/1748/files#diff-ffa527644d24f261a5759eec9fed03f0R75  ?

lib/Twig/Node/CheckSecurity.php
```diff
@@ -54,25 +54,25 @@ public function compile(Twig_Compiler $compiler)
             ->outdent()
             ->write(");\n")
```
SpacePossum: can be changed as well :)

lib/Twig/Node/Include.php
```diff
@@ -49,9 +49,9 @@ public function compile(Twig_Compiler $compiler)
         if ($this->getAttribute('ignore_missing')) {
             $compiler
                 ->outdent()
-                ->write("} catch (Twig_Error_Loader \$e) {\n")
+                ->writeln('} catch (Twig_Error_Loader $e) {')
                 ->indent()
-                ->write("// ignore missing template\n")
+                ->writeln('// ignore missing template')
                 ->outdent()
                 ->write("}\n\n")
```
SpacePossum: can be changed as well

lib/Twig/Node/Spaceless.php
```diff
@@ -32,7 +32,7 @@ public function compile(Twig_Compiler $compiler)
     {
         $compiler
             ->addDebugInfo($this)
-            ->write("ob_start();\n")
+            ->writeln('ob_start();')
             ->subcompile($this->getNode('body'))
             ->write("echo trim(preg_replace('/>\s+</', '><', ob_get_clean()));\n")
```
SpacePossum: and this one :)

c9s:
Hi @SpacePossum OK I will do this in recent days. :)

test/Twig/Tests/Node/ForTest.php
```diff
@@ -73,13 +73,13 @@ public function getTests()
...
         $tests[] = array($node, <<<EOF
 // line 1
-\$context['_parent'] = (array) \$context;
-\$context['_seq'] = twig_ensure_traversable({$this->getVariableGetter('values')});
-\$context['loop'] = array(
-  'parent' => \$context['_parent'],
-  'index0' => 0,
-  'index'  => 1,
-  'first'  => true,
+\$context["_parent"] = (array) \$context;
```
Tobion: semantically we want to keep the single quotes in these places. double quotes are only used when around variables.

fabpot:
Closing as there is no more activities here and comments were not addressed. Feel free to reopen if you have time to finish it though.

PR #1749, opened 2015-08-05
Suggest mbstring for multi byte character support

SpacePossum:


fabpot:
Not sure it adds anything useful. I mean, are you really ready those suggest entries?

SpacePossum:
I don't really read the suggestions when running composer, however I do read those when looking on packagist (for example https://packagist.org/packages/symfony/console). And I tend to check the `composer.json` of a project before adding it to a project.

fabpot:
I err on the side of closing this PR as a won't fix as we also support iconv, in which case you don't really need mbstring. Instead, I propose to add a note in the documentation to suggest installing one or the other with a preference for mbstring.

PR #1727, opened 2015-07-03
Fix the profiler node visitor signature for Twig 2.0

stof:

PR #1753, opened 2015-08-11
added a base node visitor to make compat between 1.x and 2.x possible

fabpot:
As suggested in #1644, that allows better compat between 1.x and 2.x


lib/Twig/BaseNodeVisitor.php
```diff
@@ -0,0 +1,38 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    final public function leaveNode(Twig_NodeInterface $node, Twig_Environment $env)
+    {
+        return $this->leaveNode($node, $env);
+    }
+
+    abstract public function enterNode(Twig_Node $node, Twig_Environment $env);
```
stof: this is broken. It must use a different name than the public method.

and it should be a protected method

lib/Twig/BaseNodeVisitor.php
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * @author Fabien Potencier <fabien@symfony.com>
+ */
+abstract class BaseNodeVisitor implements Twig_NodeVisitorInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    final public function enterNode(Twig_NodeInterface $node, Twig_Environment $env)
+    {
```
stof: this should throw an exception for custom implementations of the interface saying it is not supported instead of triggering a fatal error on the following typehint

stof:
@fabpot I suggest using this base class in core visitors too. this would avoid issues when introducing new visitors, like in #1727. What do you think ?

stof:
and it would have caught the fact that the code is invalid btw

fabpot:
@stof Should be fine now

lib/Twig/BaseNodeVisitor.php
```diff
@@ -0,0 +1,46 @@
...
+    final public function leaveNode(Twig_NodeInterface $node, Twig_Environment $env)
+    {
+        if (!$node instanceof Twig_Node) {
+            throw new LogicException('Twig_BaseNodeVisitor only supports Twig_Node instances.');
+        }
+
+        return $this->doLeaveNode($node, $env);
+    }
+
+    abstract protected function doEnterNode(Twig_Node $node, Twig_Environment $env);
```
stof: you need some phpdoc here, as you use ``{@inheritdoc}`` in child classes

stof:
you forgot the profiler node visitor

lib/Twig/BaseNodeVisitor.php
fabpot: right, done

stof:
:+1:

PR #1754, opened 2015-08-11
deprecated _self

fabpot:
It never makes sense to use the `_self` variable in a template except for the special `from` and `import` usage. So, I propose to deprecate it in 2.x and remove the unexcepted possible usages in 3.0.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
 * 1.20.0 (2015-XX-XX)
 
+ * deprecated the _self variable for usage outside of the from and import tags
  * added Twig_BaseNodeVisitor to ease the compatibility of node visitors 
    between 1.x and 2.x
 
--- a/doc/deprecated.rst
+++ b/doc/deprecated.rst
@@ -121,6 +121,10 @@ Globals
   or the extensions have been initialized is not possible anymore (but
   changing the value of an already registered global is possible).
 
+* As of Twig 1.x, the ``_self`` global variable is deprecated except for usage
+  in the ``from`` and the ``import`` tags. In Twig 2.0, ``_self`` is not
+  exposed anymore but still usable in the ``from`` and the ``import`` tags.
+
 Miscellaneous
 -------------
 
--- a/doc/templates.rst
+++ b/doc/templates.rst
@@ -124,7 +124,7 @@ Global Variables
 
 The following variables are always available in templates:
 
-* ``_self``: references the current template;
+* ``_self``: references the current template (deprecated since Twig 1.20);
 * ``_context``: references the current context;
 * ``_charset``: references the current charset.
 
--- a/test/Twig/Tests/TemplateTest.php
+++ b/test/Twig/Tests/TemplateTest.php
@@ -60,7 +60,7 @@ class Twig_Tests_TemplateTest extends PHPUnit_Framework_TestCase
             array('{{ array.a }}', 'Key "a" for array with keys "foo" does not exist in "%s" at line 1', false),
             array('{{ attribute(array, -10) }}', 'Key "-10" for array with keys "foo" does not exist in "%s" at line 1', false),
             array('{{ array_access.a }}', 'Method "a" for object "Twig_TemplateArrayAccessObject" does not exist in "%s" at line 1', false),
-            array('{% macro foo(obj) %}{{ obj.missing_method() }}{% endmacro %}{{ _self.foo(array_access) }}', 'Method "missing_method" for object "Twig_TemplateArrayAccessObject" does not exist in "%s" at line 1', false),
+            array('{% from _self import foo %}{% macro foo(obj) %}{{ obj.missing_method() }}{% endmacro %}{{ foo(array_access) }}', 'Method "missing_method" for object "Twig_TemplateArrayAccessObject" does not exist in "%s" at line 1', false),
             array('{{ magic_exception.test }}', 'An exception has been thrown during the rendering of a template ("Hey! Don\'t try to isset me!") in "%s" at line 1.', false),
             array('{{ object["a"] }}', 'Impossible to access a key "a" on an object of class "stdClass" that does not implement ArrayAccess interface in "%s" at line 1', false),
         );


fabpot:
By the way, the current docs even warn about possible misuses of the `_self` variable:

```
.. warning::

    When you define a macro in the template where you are going to use it, you
    might be tempted to call the macro directly via ``_self.input()`` instead
    of importing it; even if seems to work, this is just a side-effect of the
    current implementation and it won't work anymore in Twig 2.x.
```

And I found one such example in Symfony 2.7.

fabpot:
Second commit proposes to remove `_self` usage for macro imports as well and replace it with `current`: `from current import ...` or `import current as ...`. We could potentially have a BC break here if the `current` variable exists, but that seems highly improbable, so it's not taken care of here.

ninsuo:
I definitely agree.

Removing access to `_self` also removes access to `_self.env` and in most cases we'll not need to use sandboxes anymore to run untrusted code.

stof:
@ninsuo until macros are totally refactored, this would still be needed, because importing a macro actually brings the Twig_Template object in the context:

```twig
{% import 'any.twig' as macro %}

{{ dump(macro.env) %}
```

ninsuo:
Is there somebody currently working on macros? perhaps I can help. Would enjoy destroying this stuff definitely ^^

fabpot:
@ninsuo The difficult part is backward compatibility. I have some ideas but not sure this can be done for Twig 2.0.

ninsuo:
I was thinking about encapsulating the macro object in a proxy class, restricting access to macro methods. The change can made internally on `lib/Twig/Node/Import.php` and has no BC breaks.

fabpot:
@stof I've not tested, but I'm pretty sure you can get the env this way as a macro call is always automatically prefixed with `get`, so it would be converted to `macro.getenv()`, which does not exist.

ninsuo:
http://twigfiddle.com/85l2jm

stof:
hmm, I forgot about a detail of the implementation of macro calls

fabpot:
Just tested, and indeed, you can only access methods prefixed with `get`, so `{{ macro.environment }}` would work.

fabpot:
Anyway, back to the PR, everyone agrees that this is a good idea?

fabpot:
@ninsuo I would love to see an implementation of your idea. Can you do that in the coming days? I'm kind of in a hurry to get this out of the door.

ninsuo:
Yes. Everything that can protect against things like `{{ _self.env.registerUndefinedXXCallback("exec") }}` are good.

ninsuo:
Ok i'll handle this

fabpot:
@ninsuo Your example would not work as filters/functions/... are determined at compilation time, not at runtime.

ninsuo:
This works:

```
{{ _self.env.registerUndefinedFilterCallback("exec") }}
{{ _self.env.getFilter("ls -l") }}
```

lyrixx:
> It never makes sense to use the _self variable in a template

Symfony form theming can use this feature. It's easier when we need to change only few fields ...

fabpot:
@lyrixx Can you give a real-world example?

stof:
@fabpot is there any reason to have a ``getEnvironment`` getter in Twig_Template

@lyrixx the ``{% form_theme %}`` tag can implement a custom handling for ``_self`` though to preserve BC.

@fabpot http://symfony.com/doc/current/cookbook/form/form_customization.html#method-1-inside-the-same-template-as-the-form

fabpot:
I've just removed all "sensitive" comments for obvious reasons. We should act fast now. We can resume the discussion via email.

stof:
@lyrixx why not keeping the usage of ``_self`` in macro imports rather than renaming it to ``current`` in this case ? It would work the same way than today (we already have a special handling for it in the tag compilation). And it would avoid BC breaks for people having a ``current`` variable.

fabpot:
@stof I thought that the renaming would make the intent clear and not confuse people used to `_self` being a variable; people should read `from current import` or `import current as` as a language construct.

fabpot:
The `form_theme` tag can be made compatible with the removal of `_self`, like what we have for `from` and `import`.

stof:
@fabpot but then, what would be the way to do this in Symfony form theming ? Switching it to use ``current`` rather than ``_self`` would also be a BC break there.

Btw, giving that ``{% form_theme %}`` accepts any expression (as it can be a list of themes), it might be hard to implement BC for this.

fabpot:
So, the only remaining question is the renaming. I'm going to remove the second commit to be able to merge this one and create a new PR for the renaming.

fabpot:
@stof I'm sure it's trivial to implement, but anyway, let's split this PR.

fabpot:
see #1755

ninsuo:
@stof about `getEnvironment()` on templates:
```sh
$ grep -Ri 'getEnvironment' Twig/Node/
Twig/Node//Expression/Filter.php:        $filter = $compiler->getEnvironment()->getFilter($name);
Twig/Node//Expression/Function.php:        $function = $compiler->getEnvironment()->getFunction($name);
Twig/Node//Expression/Name.php:                if ($this->getAttribute('ignore_strict_check') || !$compiler->getEnvironment()->isStrictVariables()) {
Twig/Node//Expression/Test.php:        $test = $compiler->getEnvironment()->getTest($name);
Twig/Node//Module.php:            ->write('class '.$compiler->getEnvironment()->getTemplateClass($this->getAttribute('filename'), $this->getAttribute('index')))
Twig/Node//Module.php:            ->raw(sprintf(" extends %s\n", $compiler->getEnvironment()->getBaseTemplateClass()))
```

stof:
@ninsuo this is about getenvironment in compiler, not in template

ninsuo:
Oops, you're right, that's not compiled code

stof:
@fabpot I sent an email regarding the sensitive thing

fabpot:
see #1757 for deprecating `Twig_Template::getEnvironment()`

PR #1755, opened 2015-08-12
replaced _self by current for macro imports

fabpot:
A variation of this PR could be to keep `_self` in 2.0 (for BC reasons) and remove it in 3.0.

The reason to rename `_self` to `current` is that the only usage of `_self` is in `from` and `import` tags where it's not really a variable anymore, but just a language construct; and `from current import` or `import current as` seems better as language constructs.

I'm aware of one similar usage in Symfony, with the `form_theme` tag. We can do the same as here, supporting both `current` and `_self`, and remove usage of `_self` whenever we bumped the min version of Twig to 3.0.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
 * 1.20.0 (2015-XX-XX)
 
+ * added support for "from current import" and "import current as" (current replaces _self)
  * deprecated the _self variable for usage outside of the from and import tags
  * added Twig_BaseNodeVisitor to ease the compatibility of node visitors 
    between 1.x and 2.x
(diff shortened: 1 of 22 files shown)


Pierstoval:
:+1: ,

Seldaek:
I like _self because it looks like a magic variable / something special, so it stands out. I also like it because it's there and doesn't require any change. Not sure what renaming achieves except everyone having to replace stuff in their templates and adding potential confusion. At least according to the PR description I don't get the benefit.

Re: https://github.com/twigphp/Twig/pull/1754#issuecomment-130084334 I think if it worked and people do it so often that you need to have a warning in the docs, maybe it's a sign it should be made to work officially instead of removed? It's quite convenient to define a local macro and use it with _self.foo, I don't see why I should import a macro that's clearly defined in the same file.

Just my two cents, hopefully providing food for thought, but feel free to ignore me :p

Shine-neko:
+1 @Seldaek

aaronhall:
+1 @Seldaek

sstok:
+1 @Seldaek

hason:
@Seldaek The problem with local macros is solved in #1758.

PR #1757, opened 2015-08-12
deprecated Twig_Template::getEnvironment()

fabpot:


stof:
what about throwing a silenced deprecation, according to the Symfony way to handle them ?

as it is silenced, it will not hurt other projects not following this pattern (they just won't display it), unless they have a crappy error handler turning silenced errors into exceptions (which would break tons of libraries anyway)

doc/deprecated.rst
```diff
@@ -130,3 +130,7 @@ Miscellaneous
 
 * As of Twig 1.x, ``Twig_Environment::clearTemplateCache()`` is deprecated and
   will be removed in 2.0.
+
+* As of Twig 1.x, ``Twig_Template::getEnvironment()`` and
+  ``Twig_TemplateInterface::getEnvironment()`` are deprecated and will be
+  removed in 2.0.
```
sstok: The deprecation message says it's removed in 3.0.
`@deprecated since 1.20 (to be removed in 3.0)`

stof:
@fabpot any feedback on the warning ?

fabpot:
@stof I've just added the `trigger_error` call.

stof:
:+1: (we should make a pass on other deprecated methods in Twig btw)

PR #1758, opened 2015-08-12
Added support for directly call macros defined in the same template

### LINKED ISSUE #861: Consider auto-importing macros in the same file

hason:
Fixes #861 and #1755

### DIFF
--- a/lib/Twig/ExpressionParser.php
+++ b/lib/Twig/ExpressionParser.php
@@ -347,10 +347,25 @@ class Twig_ExpressionParser
                     return $node;
                 }
 
-                $args = $this->parseArguments(true);
-                $class = $this->getFunctionNodeClass($name, $line);
+                try {
+                    $class = $this->getFunctionNodeClass($name, $line);
+                } catch (Twig_Error_Syntax $e) {
+                    if (!$this->parser->hasMacro($name)) {
+                        throw $e;
+                    }
 
-                return new $class($name, $args, $line);
+                    $arguments = new Twig_Node_Expression_Array(array(), $line);
+                    foreach ($this->parseArguments() as $n) {
+                        $arguments->addElement($n);
+                    }
+
+                    $node = new Twig_Node_Expression_MethodCall(new Twig_Node_Expression_Name('_self', $line), 'get'.$name, $arguments, $line);
+                    $node->setAttribute('safe', true);
+
+                    return $node;
+                }
+
+                return new $class($name, $this->parseArguments(true), $line);
         }
     }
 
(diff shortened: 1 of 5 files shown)


stof:
the issue with this is that ``submit(foo)`` is not ambiguous: if there is no function with this name, it will call the macro in the same file. but if an extension registers this function, the template will magically break because the function will be used.
and this is not consistent with the way macros imported with ``{% from ... import ... %}`` work: these ones are winning over functions (making template behavior known as you can see from the template source that this refers to an imported macro rather than a function)

hason:
@stof good point, but similar problem is with two extensions which register the same functions.

stof:
@hason no. In this case the behavior of templates is consistent: it always calls the function. Template authors don't care where the function is defined. The fact that an extension can overwrite previous ones is a matter of the developer, not of the template author.
With your change, a template author will create a ``submit`` macro, and then use ``submit(foo)`` as they are used to be able to use macros of the template directly. But then, they will need to figure out why the template does a totally different behavior than what the author reads from the template (it is even possible that he does not get any error if the call is compatible with the function signature). Or even worse, the template could work and then break later when a dev creates a new function (again in a way which might not generate any error message depending on signatures).

hason:
@stof Well, what do you suggest? Should be a macro prefered?

stof:
@hason well, this would be consistent (and we could achieve it by registering them as imported symbols), but it would be a BC break.

fabpot:
This has been proposed before and rejected because of BC implications. Also, I very much prefer the explicit way for macros.

hason:
@fabpot This was rejected due to named arguments for macros. This PR solves only auto-importing macros.

lib/Twig/ExpressionParser.php
```diff
@@ -347,10 +347,25 @@ public function getFunctionNode($name, $line)
...
+                    if (!$this->parser->hasMacro($name)) {
+                        throw $e;
+                    }
+
+                    $arguments = new Twig_Node_Expression_Array(array(), $line);
+                    foreach ($this->parseArguments() as $n) {
+                        $arguments->addElement($n);
+                    }
+
+                    $node = new Twig_Node_Expression_MethodCall(new Twig_Node_Expression_Name('_self', $line), 'get'.$name, $arguments, $line);
```
stof: this will need a change because of #1766

fabpot:
implemented without BC breaks in #3019.

PR #1761, opened 2015-08-12
simplified code

fabpot:


ninsuo:
:+1:

fabpot:
ping @tucksaun

tucksaun:
:+1:

ext/twig/twig.c
```diff
@@ -620,23 +619,12 @@ static int twig_add_method_to_class(void *pDest APPLY_TSRMLS_DC, int num_args, v
...
 
-	if (strcmp("getenvironment", item) == 0) {
-		zend_class_entry **twig_template_ce;
-		if (zend_lookup_class("Twig_Template", strlen("Twig_Template"), &twig_template_ce TSRMLS_CC) == FAILURE) {
-			return 0;
-		}
-		if (instanceof_function(ce, *twig_template_ce TSRMLS_CC)) {
-			return 0;
-		}
-	}
```
stof: the issue here is that if someone uses Twig 1.x with the 2.x C extension (which is possible), it reintroduces the security issue.

tucksaun: Do we have any promise on this?
If so, I guess, this would need to be add to the test matrix.

stof: Well, given that the twig ext is global on the system (as any PHP ext) and that Twig automatically use it when available, it can be easy to have a case where you updated the ext on the server, but one of the projects deployed on it still runs Twig 1.x

PR #1763, opened 2015-08-13
Remove broken code

stof:
https://github.com/twigphp/Twig/commit/651613c32c315bbeb330b2a6354b34c288e9ea0a#diff-841d1051670799376e88c87114d2caf0 broke the code of ``Twig_Node_Module::compileLoadTemplate`` when passing a non-constant node (it triggers a fatal error)
Trait templates can only be constant expression nodes, so fixing the broken is not actually needed and it can be removed.
Given it is a protected method, I haven't changed the signature to typehint the node as a Twig_Node_Expression_Constant

### DIFF
--- a/lib/Twig/Node/Module.php
+++ b/lib/Twig/Node/Module.php
@@ -402,22 +402,7 @@ class Twig_Node_Module extends Twig_Node
                 ->raw(");\n")
             ;
         } else {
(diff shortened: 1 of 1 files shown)


fabpot:
Thank you @stof.

PR #1764, opened 2015-08-13
Twig_Filter can't be deprecated & preferred for 3.0

andrerom:
*Warning, before :coffee: has kicked in.*

Seems very strange that several classes are deprecated in 1.x and is reintroduced in 3.x with different content. why not just name space the future version instead of having to do this littel strange dance with naming conflicts?

Oh, and edit buttons from doc is broken.

### DIFF
--- a/doc/deprecated.rst
+++ b/doc/deprecated.rst
@@ -34,7 +34,6 @@ Filters
 
   * ``Twig_FilterInterface``
   * ``Twig_FilterCallableInterface``
-  * ``Twig_Filter``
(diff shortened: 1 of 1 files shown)


stof:
well, the 1.x Twig_Filter is deprecated. In 2.x, Twig_Filter is the same than Twig_SimpleFilter in 1.x and the signature is different

andrerom:
will most likely lead to more problems then it is worth imo, but just wanted to raise that.

PR #1765, opened 2015-08-13
Remove the reserved macro names

stof:
Some macro names are reserved because macros are compiled as ``get*`` methods in the template. but this name for compiled function is related to the old implementation of macros calls as ``GetAttr`` nodes.

https://github.com/twigphp/Twig/commit/e81c932e779352f5716290765ac9786950b6b473 and https://github.com/twigphp/Twig/commit/a4e10c14edfb5fa2b2996c7a10ef5348ea8115c5 changed this by calling macros through a dedicated node.
This allows to use any method name for the compiled method. Using a ``macro_`` prefix avoids conflict with any other method and so removes reserved names (and it is consistent with blocks).

I'm currently submitting this to Twig 2.x, but we could potentially consider it for inclusion in 1.21 already if we consider that the compiled method name is an internal implementation detail (I don't see any reason someone would depend on this name actually).
The only BC break is that you have to clear your cache when upgrading Twig to recompile templates. But this can be the case for any Twig upgrade anyway (autoReload is looking for changes in extensions themselves, but this change does not affect extensions but token parsers)


fabpot:
I would indeed do that for Twig 1.21 instead.

stof:
OK, closing and reopening against 1.x than

Issue #1217, opened 2013-10-03
Magical and incorrect macro call


hason:
Let us have templates:

``` jinja
{# index.twig #}
{% extends "parent.twig" %}

{% block content %}
{{ macros.hello("Jack") }}
{% endblock %}

{# parent.twig #}
{% macro hello(name) %}
{% endmacro %}

{% import _self as macros %}

{% block content "" %}
```

The `{{ macros.hello("Jack") }}` line in the template `index.html` is compiled to:

``` php
$this->getAttribute($this->getContext($context, 'macros'), 'hello', ...)
```

Properly, we should find out what is the variable `macros` and call the macro `hello` normally.

stof:
IMO, the issue is that we let this work in the first place. and this is because ``import`` is changing the context.

fabpot:
As of Twig 1.28, this code generates a deprecation and this code is not supported anymore in Twig 2.0.

Issue #1212, opened 2013-10-01
Macro importing from a template does not work with template inheritance


hason:
To reproduce the bug:

``` jinja
{# index.twig #}
{% extends "parent.twig" %}
{% block content %}
    {{ hello("Jack") }}
{% endblock %}

{# parent.twig #}
{% from "macros.twig" import hello %}
{% block content %}{% endblock %}

{# macros.twig #}
{% macro hello(name) %}
    Hello {{ name }}
{% endmacro %}
```

Parsing fails:

```
Twig_Error_Syntax: The function "hello" does not exist in "index.twig"
```

fabpot:
I think this one should be closed as #1217 should not work and will be fixed in 2.0.

fabpot:
It's indeed expected as macros imports are local to a template, not something we will change, so closing.

PR #1766, opened 2015-08-13
Remove the reserved macro names

stof:
Replaces #1765


lib/Twig/Parser.php
```diff
@@ -254,28 +253,17 @@ public function hasMacro($name)
...
     {
-        if ($this->isReservedMacroName($name)) {
-            throw new Twig_Error_Syntax(sprintf('"%s" cannot be used as a macro name as it is a reserved keyword', $name), $node->getLine(), $this->getFilename());
-        }
-
         $this->macros[$name] = $node;
     }
 
+    /**
+     * @deprecated since 1.21. Will be removed in 2.0. There is no reserved macro names anymore
```
fabpot: I would not even bother with that and remove the method altogether.

hason:
Great, this change I suggested several times #1310 ([commit](https://github.com/hason/Twig/commit/33aa7de9b3bd3cdfa5a0c8892500e82b26d055a2)). This is BC break (see #1217) and I vote :+1: for Twig 2.0 or Twig 3.0.

We really should resolve macros, especially #1212 and #1217.

stof:
arf, I hate the fact that we modify the context when importing macros, as the context propagates to other templates.

hason:
@fabpot What do you think? I think we should explicitly import all macros that are also imported in parent templates. Will this change in version 2 or 3?

fabpot:
We cannot really have this one in 1.x because of the BC break mentioned by @hason, even if the behavior in #1217 should not be relied on. So, I propose to merge it in master (sorry @stof, you will have to re-open yet a new one) and document the BC break in case some people relied on the current behavior.

fabpot:
By the way, using `__macro__` is probably "safer".

stof:
@fabpot why safer ? We are already using ``block_`` as prefix for blocks.

stof:
OK, closing and reopening it on master

Issue #1512, opened 2014-09-26
Twig_Markup objects cannot be compared by some operators


jaytaph:
When working with a Twig_Markup object, I can easily print this as it implements __toString().

```
{{ myvar }}
```

However, certain operators, like the "in" and "not in" which internally depend on the [twig_in_filter function](https://github.com/fabpot/Twig/blob/master/lib/Twig/Extension/Core.php#L893) , behave differently internally by checking the right operator against arrays or actual strings. This means Twig_Markup objects cannot be used with this operator:

```
{% if "foo" in myvar %}  
```

It's trivial to add some additional tests in the twig_in_filter function like [this one](https://github.com/fabpot/Twig/blob/master/lib/Twig/Extension/Core.php#L925), but more operators might be affected as well (like "starts with", and "ends with").

I'm not sure if this is the correct way, if so, i'm happy to create a PR with the additional string-cast.

fabpot:
Indeed, `Twig_Markup` objects are annoying and the more support we can add, the better. I will try to see if doing that across the board is realistic or not, starting with the in test.

fabpot:
fixed in #2952 for the in operator.

fabpot:
I've decided to fix only the "in" operator as no so many people are having this issue (this is the only report in the last 5 years).

PR #1768, opened 2015-08-15
changed join filter to return a safe string when all arguments are safe

### LINKED ISSUE #1420: Twig_Markup safe preservation within join()
When joining an array of safe Twig_Markup object's we're wondering if we can have it preserve the Twig_Markup when jointed?

```
$beetle = new Twig_Markup('<pine beetle>');
$lumber = new Twig_Markup('<lumber jack>');
$variables['danger'] = [$beetle, $lumber];
```

`{{ danger|join(', ') }}`

Creates
`&lt;pine beetle&gt;, &lt;lumber jack&gt;`
Would expect 
`<pine beetle>, <lumber jack>`

Understanding this may be a performance issue if you need to check each item but maybe there is a possibility here.  Thoughts?

FYI we are trying to turn on autoescape into Drupal 8, though we ran into this issue when dealing with a list of links that need to be marked safe and joined.

https://drupal.org/node/1825952

fabpot:
This is WIP to fix #1420 (first commit) and #1512 (second commit.

The first commit change the return value of the `join` filter to a Twig_Markup instance when all joined elements are "safe". That means that the returned value of the filter can now be a string or a Twig_Markup instance which is probably a BC break, especially because Twig_Markup instance cannot be used as a string everywhere in Twig (see #1512 for instance -- second commit).

The second commit makes it possible to use a Twig_Markup instance with the in test transparently to avoid WTF effects as such instances can be created by Twig directly.

Of course, if we decide to go down this road, this PR would need to:
- change some other filters/functions/whatever to make this new behavior consistent;
- try to change filters/functions/whatever to make Twig_Markup instances behave like strings as much as possible.

Is it worth it? Does it belong to 1.x or 2.x?

### DIFF
--- a/lib/Twig/Extension/Core.php
+++ b/lib/Twig/Extension/Core.php
@@ -935,9 +935,15 @@ function twig_sort_filter($array)
     return $array;
 }
 
-/* used internally */
+/**
+ * @internal
+ */
 function twig_in_filter($value, $compare)
 {
+    if ($value instanceof Twig_Markup) {
+        $value = (string) $value;
+    }
+
     if (is_array($compare)) {
         return in_array($value, $compare, is_object($value) || is_resource($value));
     } elseif (is_string($compare) && (is_string($value) || is_int($value) || is_float($value))) {
--- a/lib/Twig/Extension/Core.php
+++ b/lib/Twig/Extension/Core.php
@@ -172,7 +172,7 @@ class Twig_Extension_Core extends Twig_Extension
             new Twig_SimpleFilter('nl2br', 'nl2br', array('pre_escape' => 'html', 'is_safe' => array('html'))),
 
             // array helpers
-            new Twig_SimpleFilter('join', 'twig_join_filter'),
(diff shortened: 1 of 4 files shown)


lib/Twig/Extension/Core.php
```diff
@@ -1514,3 +1523,17 @@ function twig_array_batch($items, $size, $fill = null)
 
     return $result;
 }
+
+/**
+ * @internal
+ */
+function is_array_safe($values)
```
stof: should be prefixed by ``twig_`` to avoid conflicts

stof:
Given it is a BC break, it does not belong to 1.x IMO.

The issue of the Twig_Markup is that many places could be impacted by that. If we want to generalize their usage, it will make things more complex (and for all extension authors too as they would need to handle Twig_Markup too).

robfrawley:
@fabpot What is the status of this PR? Is it abandoned?

fabpot:
I've cherry-picked the second commit as it was a no-brainer. Given that dealing with `Twig_Markup` instances everywhere would make things more complex and no BC, I'm closing this one.

Issue #1767, opened 2015-08-13
Make sure deprecations are triggered in 1.x for any deprecated features


stof:
We need deprecations in many places in 1.x for removed things:

- Twig_Environment constructor without loader
- Twig_Filter, Twig_Test and Twig_Function usage when registering them
- custom implementations of deprecated interfaces being passed (not sure if this is needed though, or if we consider that people would be warned about this by the DebugClassLoader of Symfony when using it)
- ... (the list needs to be completed)

We should also warn about features deprecated in templates, but this would require #1756 first

fabpot:
see #1769

PR #1769, opened 2015-08-16
Deprecation notices

### LINKED ISSUE #1756: Provide a way to trigger deprecation warning in Twig
We should have a way to trigger deprecation warnings in compiled Twig code. The previous attempt in Symfony failed because many things are expressions in Twig, making it hard to bring a statement in.

A way to solve it may be to provide a function which could be used to trigger deprecations in expressions:

```php
function twig_deprecated($value, $message) {
    @trigger_error($message, E_USER_DEPRECATED);

    return $value;
}
```

A deprecated usage in an expression could then be wrapped into a call to this function. And we could have an optimization disabling deprecation calls in prod optionally (which would remove the wrapping node).
what do you think @fabpot ?

fabpot:
Added a way to warn our users of deprecation features. This framework can be used for core features, but also for functions/tests/filters created by third-party. I haven't added something for tags as it can be done easily in the token parser class.

Fixes #1756 and #1767


fabpot:
Forgot to mention that this is WIP, docs are not finished yet, just placeholders :)

stof:
> I haven't added something for tags as it can be done easily in the token parser class.

this would trigger it only during the template compilation, not at render time, so it may be hard.

lib/Twig/Environment.php
```diff
@@ -448,6 +450,8 @@ public function resolveTemplate($names)
      */
     public function clearTemplateCache()
     {
+        @trigger_error(sprintf('%s is deprecated and will be removed in Twig 2.0.', __METHOD__), E_USER_DEPRECATED);
```
stof: The Symfony convention for deprecation messages is to say ``The method __METHOD__ is deprecated since X.Y and will be removed in Z.0. Use ... instead.`` (the second sentence being removed if there is no alternative). What do you think about being consistent in Twig ?

lib/Twig/Extension/Core.php
```diff
@@ -298,17 +298,28 @@ public function parseNotTestExpression(Twig_Parser $parser, Twig_NodeInterface $
...
         if ($stream->test(Twig_Token::PUNCTUATION_TYPE, '(')) {
             $arguments = $parser->getExpressionParser()->parseArguments(true);
         }
 
-        return new $class($node, $name, $arguments, $parser->getCurrentToken()->getLine());
+        return new $class($node, $test->getName(), $arguments, $parser->getCurrentToken()->getLine());
     }
 
-    protected function getTestName(Twig_Parser $parser, $line)
+    protected function getTest(Twig_Parser $parser, $line)
```
stof: changing protected methods is technically a BC break

lib/Twig/Node/Expression/Name.php
```diff
@@ -30,11 +30,19 @@ public function compile(Twig_Compiler $compiler)
...
+                if ('_self' === $this->specialVars[$name]) {
+                    @trigger_error('Global variable "_self" is deprecated and will be removed in Twig 2.0.', E_USER_DEPRECATED);
+                }
+
                 $compiler->repr(true);
             } else {
                 $compiler->raw('array_key_exists(')->repr($name)->raw(', $context)');
             }
         } elseif ($this->isSpecial()) {
+            if ('_self' === $this->specialVars[$name]) {
```
stof: this is wrong. It should be ``if ('_self' === $name)`` (same above)

lib/Twig/SimpleFunction.php
```diff
@@ -87,4 +89,14 @@ public function isVariadic()
     {
         return $this->options['is_variadic'];
     }
+
+    public function isDeprecated()
+    {
+        return $this->options['deprecated'];
+    }
+
+    public function getAlternative()
```
stof: these methods must also be added in ``Twig_Function_*`` (and the test and filter equivalent), otherwise you will get fatal errors when the deprecated classes are used to register them)

lib/Twig/Test/IntegrationTestCase.php
```diff
@@ -38,10 +46,14 @@ public function getTests()
                 continue;
             }
 
+            $isLegacyTest = false !== strpos($file->getRealpath(), '.legacy.test');
+            if (($isLegacyTest && !$legacyOnly) || (!$isLegacyTest && $legacyOnly)) {
```
stof: ``if ($isLegacyTest xor $legacyOnly)``

stof:
You should also trigger deprecation notices at the beginning of files for deprecated classes, as done in Symfony

stof:
TokenParserBroker related methods should also be marked as deprecated as the feature is deprecated.

lib/Twig/Extension/Core.php
```diff
@@ -233,11 +233,11 @@ public function getTests()
             new Twig_SimpleTest('even', null, array('node_class' => 'Twig_Node_Expression_Test_Even')),
             new Twig_SimpleTest('odd', null, array('node_class' => 'Twig_Node_Expression_Test_Odd')),
             new Twig_SimpleTest('defined', null, array('node_class' => 'Twig_Node_Expression_Test_Defined')),
-            new Twig_SimpleTest('sameas', null, array('node_class' => 'Twig_Node_Expression_Test_Sameas')),
+            new Twig_SimpleTest('sameas', null, array('node_class' => 'Twig_Node_Expression_Test_Sameas', 'deprecated' => true, 'alternative' => 'same as')),
             new Twig_SimpleTest('same as', null, array('node_class' => 'Twig_Node_Expression_Test_Sameas')),
             new Twig_SimpleTest('none', null, array('node_class' => 'Twig_Node_Expression_Test_Null')),
             new Twig_SimpleTest('null', null, array('node_class' => 'Twig_Node_Expression_Test_Null')),
-            new Twig_SimpleTest('divisibleby', null, array('node_class' => 'Twig_Node_Expression_Test_Divisibleby')),
+            new Twig_SimpleTest('divisibleby', null, array('node_class' => 'Twig_Node_Expression_Test_Divisibleby', 'deprecated' => true, 'alternative' => 'divisibleby')),
```
sstok: Alternative is wrong, this should be `divisible by`

lib/Twig/Node/Expression/Name.php
fabpot: fixed

lib/Twig/Test/IntegrationTestCase.php
fabpot: fixed

lib/Twig/Extension/Core.php
fabpot: fixed

lib/Twig/Extension/Core.php
fabpot: I know, but nobody should rely on this and anyway, Twig is not Symfony. That's almost fixes in 2.0 as most methods/properties are now private.

lib/Twig/Environment.php
fabpot: I don't think we need to much details in Twig as all 1.x versions are basically interchangeable. Saying something was deprecated in 1.16 does not add anything useful.

lib/Twig/SimpleFunction.php
fabpot: I don't think it is worth it. People should first move away from these deprecated classes and then add the deprecation notice.

lib/Twig/Environment.php
fabpot: I've tweaked some messages to make them look more Symfony-like.

fabpot:
Deprecation for classes have been added.

I think this one is ready to be merged now.

lib/Twig/SimpleFunction.php
fabpot: I've of course also added some safeguard when calling the `isDeprecated()` method to avoid those fatal errors.

lib/Twig/Environment.php
stof: Well, the documentation also says ``as of 1.x``, which does not make sense as we always introduced the replacement in a particular version

stof:
@fabpot I think you should also trigger a deprecation warning when adding a new TokenParserBroker, either from an extension or directly in the environment

lib/Twig/Environment.php
```diff
@@ -117,6 +119,40 @@ public function __construct(Twig_LoaderInterface $loader = null, $options = arra
...
 
     /**
+     * Returns deprecations for templates contained in a directory.
+     *
+     * @param string $dir The directory where templates are stored
+     * @param string $ext Limit the loaded templates by extension
+     *
+     * @return An array of deprecations
+     */
+    public function getDeprecations($dir, $ext = '.twig')
```
stof: I don't think this really belongs to the Twig environment. It should be built on top of it IMO (as done for the twig linting command of Symfony).
I suggest putting this in an dedicated utility class wrapping the environment

lib/Twig/Environment.php
```diff
@@ -117,6 +119,40 @@ public function __construct(Twig_LoaderInterface $loader = null, $options = arra
...
+    {
+        $deprecations = array();
+        $previous = set_error_handler(function ($type, $msg) use (&$deprecations) {
+            if (E_USER_DEPRECATED === $type) {
+                $deprecations[] = $msg;
+            }
+        });
+
+        foreach (new RecursiveIteratorIterator(new RecursiveDirectoryIterator($dir), RecursiveIteratorIterator::LEAVES_ONLY) as $file) {
+            if (!preg_match('{'.preg_quote($ext).'$}', $file)) {
```
stof: why not using a RegexIterator for that ?

and excaping the extension to build the regex should be done outside the loop

fabpot:
@stof where? we already trigger a notice when a `TokenParserBroker` is added `Twig_Env::initExtension()`. And just loading the class also triggers a warning.

lib/Twig/Environment.php
```diff
@@ -117,6 +119,40 @@ public function __construct(Twig_LoaderInterface $loader = null, $options = arra
...
+            }
+
+            try {
+                $this->compileSource(file_get_contents($file), (string) $file);
+            } catch (Twig_Error_Syntax $e) {
+                // ignore templates containing syntax errors
+            }
+        }
+
+        set_error_handler($previous);
```
stof: this should use ``restore_error_handler`` instead to restore it, instead of readding the previous one on top of the stack

lib/Twig/Environment.php
```diff
@@ -117,6 +119,40 @@ public function __construct(Twig_LoaderInterface $loader = null, $options = arra
     }
 
     /**
+     * Returns deprecations for templates contained in a directory.
+     *
+     * @param string $dir The directory where templates are stored
+     * @param string $ext Limit the loaded templates by extension
+     *
+     * @return An array of deprecations
```
stof: ``@return array`` is needed, otherwise the phpdoc is invalid and IDEs cannot use it (or even better ``@return string[]``)

lib/Twig/Environment.php
fabpot: fixed

lib/Twig/Environment.php
fabpot: fixed

lib/Twig/Environment.php

lib/Twig/Environment.php
fabpot: Right, moved to Twig_Deprecations (for the lack of a better name)

fabpot:
The `Twig_Deprecations::collect()` method implementation is naive and does not really work in real-life. I'm going to rework it.

lib/Twig/Environment.php
```diff
@@ -1067,11 +1079,11 @@ public function addGlobal($name, $value)
                 $this->globals = $this->initGlobals();
             }
 
-            /* This condition must be uncommented in Twig 2.0
             if (!array_key_exists($name, $this->globals)) {
-                throw new LogicException(sprintf('Unable to add global "%s" as the runtime or the extensions have already been initialized.', $name));
+                // The deprecation notice must be turned into the following exception in Twig 2.0
+                @trigger_error(sprintf('Registering global variable "%s" at runtime or when the extensions have already been initialized is deprecated.', $name), E_USER_DEPRECATED);
+                //throw new LogicException(sprintf('Unable to add global "%s" as the runtime or the extensions have already been initialized.', $name));
```
Taluu: I don't really see the point here, to use a deprecated error notice instead of an exception and then suggest to actually use the same exception (unless I read it wrong) in 2.0 ?

fabpot:
New version of the Twig_Deprecations class should be better now.

lib/Twig/Environment.php
stof: @Taluu the deprecation does not prevent it from working in 1.x. We are deprecating this possibility but not breaking it.
2.0 will forbid it, which will allow to simplify some code (as we will have the list of globals known without allowing to add new ones)

lib/Twig/Deprecations.php
```diff
@@ -0,0 +1,109 @@
...
+    /**
+     * Returns deprecations for passed templates.
+     *
+     * @param Traversable $iterator An iterator of templates (where keys are template names and values the contents of the template)
+     *
+     * @return array() An array of deprecations
+     */
+    public function collect(Traversable $iterator)
+    {
+        set_error_handler(function ($type, $msg) use (&$deprecations) {
```
sstok: `$deprecations` is not defined yet (it's defined below).
Also closures don't work in PHP 5.2 (but is that really a problem?)

lib/Twig/Environment.php
Taluu: Ah, my bad, didn't see that this condition was commented (see removed line 1072), I thought that previously this exception was already thrown, that is what was bothering me.

lib/Twig/Deprecations.php
```diff
@@ -0,0 +1,109 @@
...
+ * (c) Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+/**
+ * @author Fabien Potencier <fabien@symfony.com>
+ */
+class Twig_Deprecations
```
stof: What about naming it ``Twig_Util_DeprecationsCollector``, or ``Twig_Tool_DeprecationsCollector`` ? We could then add more utility tools in this "namespace", for instance extracting the linting logic out of the command of the Symfony bridge to make it available to anyone.

sstok: Interesting idea :+1:  :)

lib/Twig/Deprecations.php
```diff
@@ -0,0 +1,109 @@
...
+}
+
+/**
+ * @internal
+ */
+class Twig_TemplateDirIterator implements Iterator
+{
+    private $iterator;
+
+    public function __construct(Traversable $iterator)
```
stof: the code actually requires an Iterator here, not a Traversable. It calls Iterator methods on it.

Btw, a solution to simplify it would be to extend IteratorIterator (which itself supports any Traversable btw, not only iterators):

```php
class Twig_TemplateDirIterator extends IteratorIterator
{
    public function current()
    {
        return file_get_contents(parent::current());
    }

    public function key()
    {
        return (string) parent::key();
    }
}

File: lib/Twig/Deprecations.php [hunk 15]
```diff
@@ -0,0 +1,109 @@
...
+        set_error_handler(function ($type, $msg) use (&$deprecations) {
+            if (E_USER_DEPRECATED === $type) {
+                $deprecations[] = $msg;
+            }
+        });
+
+        $deprecations = array();
+        foreach ($iterator as $name => $contents) {
+            try {
+                $this->twig->parse($this->twig->tokenize($contents, $name));
```
stof: shouldn't we compile it too ? Deprecation warnings might be triggered at compile time rather than at parsing time

fabpot: We don't have any compile-time deprecations and I don't see how we could have one that we could not trigger before. The only exception I can see is `getEnvironment()`, but as `_self` is deprecated, I think it's not worth adding the complexity here. Am I missing something?

lib/Twig/Deprecations.php
fabpot: Done

lib/Twig/Util/DeprecationCollector.php
```diff
@@ -0,0 +1,110 @@
...
+}
+
+/**
+ * @internal
+ */
+class Twig_TemplateDirIterator implements Iterator
+{
+    private $iterator;
+
+    public function __construct(Traversable $iterator)
```
stof: my comments on this class have been hidden by github because of the file renaming, but they are still valid

lib/Twig/Deprecations.php
fabpot: Good catch, fixed.

PHP 5.2, hmmm... fixed

lib/Twig/Deprecations.php
fabpot: fixed

lib/Twig/Util/DeprecationCollector.php
```diff
@@ -0,0 +1,95 @@
...
+            try {
+                $this->twig->parse($this->twig->tokenize($contents, $name));
+            } catch (Twig_Error_Syntax $e) {
+                // ignore templates containing syntax errors
+            }
+        }
+
+        restore_error_handler();
+
+        return $this->deprecations;
```
stof: you should reset the array of deprecation in the object property here to avoid keeping a useless reference to the array

fabpot: Not sure I'm following here. You want me to add `$this->deprecations = array();`?

stof: yes, so that the utility class does not keep state in memory longer than necessary. this way, the array of deprecation would not be referenced anymore by the collector once it is returned to the calling code

fabpot:
Ready?

lib/Twig/Util/DeprecationCollector.php
```diff
@@ -0,0 +1,82 @@
...
+
+        foreach ($iterator as $name => $contents) {
+            try {
+                $this->twig->parse($this->twig->tokenize($contents, $name));
+            } catch (Twig_Error_Syntax $e) {
+                // ignore templates containing syntax errors
+            }
+        }
+
+        restore_error_handler();
```
Taluu: I think you should also get the old error_handler and restore it at the end of the treatment, so that the old behavior is restored instead of a blank one

stof:
:+1:

doc/recipes.rst
```diff
@@ -1,6 +1,64 @@
...
+catches deprecation notices, and return them.
+
+.. tip::
+
+    If your templates are not stored on the filesystem, use the ``collect()``
+    method instead which takes an ``Iterator``; the iterator must return
+    template names as keys and template contents as values (as done by
+    ``Twig_Util_TemplateDirIterator``).
+
+However, this code won't find all deprecations (like using deprecated some Twig
```
sstok: `like using some deprecated Twig` instead of `like using deprecated some Twig`

doc/advanced.rst
```diff
@@ -267,6 +267,23 @@ arguments, but after the environment and the context. For instance, a call to
 ``'foo'|a_path_b()`` will result in the following arguments to be passed to
 the filter: ``('a', 'b', 'foo')``.
 
+Deprecated Filters
+~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 1.21
+    Support for deprecated filters was added in Twig 1.21.
```
sstok: Not sure if this is a problem, but normally you have an empty line after the block start.
*From what I know the first contains options like `:lineno:` but I'm not sure if Sphixdoc will accept tex directly after the block start :)*

```
.. versionadded:: 1.21

    Support for deprecated filters was added in Twig 1.21.
```

stof: AFAIK, versionadded is special and does not work if you put an empty line before the directive body

doc/recipes.rst
```diff
@@ -1,6 +1,64 @@
...
+
+.. versionadded:: 1.21
+    This works as of Twig 1.21.
+
+Deprecated features generate deprecation notices (via a call to the
+``trigger_error()`` PHP function). By default, they are silenced and never
+displayed nor logged.
+
+To easily remove all deprecated feature usages from your templates, write and
+run a script along the lines of the following::
```
hhamon: Extra colon at the end.

doc/recipes.rst
```diff
@@ -1,6 +1,64 @@
...
+.. tip::
+
+    If your templates are not stored on the filesystem, use the ``collect()``
+    method instead which takes an ``Iterator``; the iterator must return
+    template names as keys and template contents as values (as done by
+    ``Twig_Util_TemplateDirIterator``).
+
+However, this code won't find all deprecations (like using deprecated some Twig
+classes). To catch all notices, register a custom error handler like the one
+below::
```
hhamon: Extra colon at the end of the line.

PR #1772, opened 2015-08-17
Optimize the retrieval of reserved macro names

stof:
The list of reserved names does not need to be reset for each template compilation.


ninsuo:
:+1:

fabpot:
Thank you @stof.

PR #1773, opened 2015-08-18
moved more methods/properties to private

fabpot:
I'd like to make even more properties/methods private in 2.0. We can always put them back to protected if that makes sense.

### DIFF
--- a/lib/Twig/Environment.php
+++ b/lib/Twig/Environment.php
@@ -18,32 +18,32 @@ class Twig_Environment
 {
     const VERSION = '2.0.0-DEV';
 
-    protected $charset;
-    protected $loader;
-    protected $debug;
-    protected $autoReload;
-    protected $cache;
-    protected $lexer;
-    protected $parser;
-    protected $compiler;
-    protected $baseTemplateClass;
-    protected $extensions;
-    protected $parsers;
-    protected $visitors;
-    protected $filters;
-    protected $tests;
-    protected $functions;
-    protected $globals;
-    protected $runtimeInitialized;
-    protected $extensionInitialized;
-    protected $loadedTemplates;
-    protected $strictVariables;
-    protected $unaryOperators;
-    protected $binaryOperators;
-    protected $templateClassPrefix = '__TwigTemplate_';
-    protected $functionCallbacks;
-    protected $filterCallbacks;
-    protected $staging;
+    private $charset;
+    private $loader;
+    private $debug;
+    private $autoReload;
+    private $cache;
+    private $lexer;
+    private $parser;
+    private $compiler;
+    private $baseTemplateClass;
+    private $extensions;
+    private $parsers;
+    private $visitors;
+    private $filters;
+    private $tests;
+    private $functions;
+    private $globals;
+    private $runtimeInitialized;
+    private $extensionInitialized;
+    private $loadedTemplates;
+    private $strictVariables;
+    private $unaryOperators;
+    private $binaryOperators;
+    private $templateClassPrefix = '__TwigTemplate_';
+    private $functionCallbacks;
+    private $filterCallbacks;
+    private $staging;
 
     /**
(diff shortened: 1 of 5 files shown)


lib/Twig/Loader/Filesystem.php
```diff
@@ -19,9 +19,9 @@ class Twig_Loader_Filesystem implements Twig_LoaderInterface
     /** Identifier of the main namespace. */
     const MAIN_NAMESPACE = '__main__';
 
-    protected $paths = array();
-    protected $cache = array();
-    protected $errorCache = array();
+    private $paths = array();
+    private $cache = array();
+    private $errorCache = array();
```
fabpot: That breaks Symfony which expects those properties to be available, but that's messy. We need a better abstraction.

stof: @fabpot the question is why the Symfony bundle notation is implemented as a child of the Twig_Loader_Filesystem (I guess the reason is because Twig_Loader_Chain was not available at this time and was added only later when wanting to support Twig_Loader_Filesystem in Silex alongside the Twig_Loader_Array).
We could implement the loading of ``Bundle:Folder:template`` as a separate standalone loader and rely on the loader chain instead (this could be considered as a BC break in Symfony though, and so be harder to apply in 2.3+).

stof: btw, I think that the Symfony loader is broken (for the bundle notation) for recent versions of Twig when using a Twig_Loader_Chain or multiple templating engines in symfony (i.e. anything using the ``exists`` method to find the right loader supporting the template) as it extends ``findTemplate`` to implement the logic but Twig_Loader_Filesystem was refactored so that findTemplate uses exists rather than the opposite

lib/Twig/Environment.php
```diff
@@ -1168,7 +1168,7 @@ protected function initExtension(Twig_ExtensionInterface $extension)
         }
     }
 
-    protected function writeCacheFile($file, $content)
+    private function writeCacheFile($file, $content)
```
stof: we have a recipe in the doc overwriting this method. You should either keep this as an extension point or remove the documentation about using this method as an extension point

twoixter:
@fabpot Sorry to comment on a PR so old, but now we're in the process of upgrading major versions from our App (PHP 7.x, Twig 2.x, etc) and stumbled upon this breaking change for us.

We inherited from `Twig_Loader_Filesystem` to customise the template finding logic. For example, to allow to find `template.en.twig` or `template.es.twig` based on the current locale. Another customisation is choosing from different directories depending on the partner (we use white label templating) so that it's inherited preferring to load `templates/partner_name/template.twig` if it exists instead of `templates/base/template.twig` (in this use case, we can't use template namespaces).

This allow for us to use the regular `{% include "template.twig" %}` without worrying with extra logic for all includes like `{% include "template."~locale~".twig" %}` or whatever complex logic to select from a set of potential template files.

Making properties private is fine and a good practice, but we use the methods `normalizeName`, `parseName`, `validateName` and `isAbsolutePath` that I find them more like "helper" methods and can't imagine a good reason not to be `protected` unless for preventing inheritance of the whole `Twig_Loader_Filesystem` class.

Again, your commit was more than two years ago and I'm sorry to comment so late. But I'll be grateful if this makes at least rethink again the need to over-protect methods. I'm beginning to get acquainted with the 2.x branch so it might be some alternative method other than creating a new Class copy-pasting the whole code from `Twig_Loader_Filesystem`.

One final note is that we use Twig standalone, out of frameworks like Symphony, Drupal, etc, if that makes more sense.

nicolas-grekas:
Please create new issue

twoixter:
Thank you @nicolas-grekas ! Done as issue #2609 

I can also make a Pull Request myself, I just want to be sure first if the changes made sense.

PR #1775, opened 2015-08-19
Fix the deprecation warning for functions and tests

stof:
Otherwise, when a function does not define any filter, the message looks like that:

```
Using an instance of "Twig_Environment" for function "nelmio_js_logger" is deprecated. Use Twig_SimpleFunction instead. 
```

This happens because the notice about the undefined variable gets silenced too, and ``get_class(null)`` returns the current class.


Tobion:
:+1:

fabpot:
Thank you @stof.

PR #1776, opened 2015-08-19
Add gitattributes w/ export-ignore

CelticParser:
This will allow for a `production` ready version of Twig when archived/exported for release(s) and still allowing a development version within the repository.

The `export-ignore` excludes un-needed files and directories when a branch is archived for release these files are included when the repo is cloned/forked.

This would be very useful because `releases` are intended for `production use` and when users are using a package (via composer or zip), they most likely are not interested in downloading the source code as a whole into production, thus saving disk space and meeting security protocol.

  - e.g. /test folder, or your .travis.yml file, etc.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,13 @@
+# Ignore un-needed files and directories for export of production releases
+# Keeps lib directory
+#
+# See: http://git-scm.com/book/en/v2/Customizing-Git-Git-Attributes#Exporting-Your-Repository
+
+doc export-ignore
+ext export-ignore
+test export-ignore
+.editorconfig export-ignore
+.gitattributes export-ignore
+.gitignore export-ignore
+.travis.yml export-ignore
+phpunit.xml.dist export-ignore


sstok:
This has been rejected before #942

barryvdh:
Still +1 on this. If people want tests/docs, they can still use `--prefer-source`

Tobion:
:+1: as I also said in https://github.com/symfony/symfony/issues/6605#issuecomment-60306001

fabpot:
fabpot/Twig#2

.gitattributes
```diff
@@ -0,0 +1,13 @@
...
+# See: http://git-scm.com/book/en/v2/Customizing-Git-Git-Attributes#Exporting-Your-Repository
+
+doc export-ignore
+ext export-ignore
+test export-ignore
+.editorconfig export-ignore
+.gitattributes export-ignore
+.gitignore export-ignore
+.travis.yml export-ignore
+phpunit.xml.dist export-ignore
```
Tobion: missing newline

sstok:
Can we at least consider ignoring the docs? When was the last time anyone actually needed this with the source. You can already watch them rendered on-line :)

SpacePossum:
the docs are out right?
https://github.com/twigphp/Twig/pull/1776#discussion-diff-39025045R6

sstok:
@SpacePossum yes ;) but if this change is rejected as a whole we can at least consider ignoring the docs.

barryvdh:
@fabpot Can we revisit this? I don't think this has the same problems as the Symfony, like people relying on the tests, right?

fabpot:
I've re-read the conversation on Symfony and Composer about this issue and I haven't changed my mind.

Closing it now.

PR #1777, opened 2015-08-20
A possible way to add custom logic to process context parameters

krocos:
From my point of view it can be useful to process many similar parameters before they pass to the template for rendering

For example:
```php
class FormContext implements ContextParameterInterface
{
    /**
     * @var FormInteface
     */
    private $form;

    /**
     * @param FormInteface $form
     */
    public function __construct(FormInteface $form)
    {
        $this->form = $form;
    }
    
    /**
     * @return mixed The prepared template context parameter
     */
    public function prepare()
    {
        $form->add('submit', 'submit');
        /* or any other manipulations here */
        
        return $form->createView();
    }
}

class SomeController
{
    public function someAction()
    {
        /** @var FormInteface $form */
        $form = /* create form */;
        
        return $this->render('someTempalte.html.twig', [
            'form' => new FormContext($form), /* or call some service here */
        ]);
    }
}
```

### DIFF
--- a/lib/Twig/ContextParameterInterface.php
+++ b/lib/Twig/ContextParameterInterface.php
@@ -1,12 +1,12 @@
 <?php
 
 /**
- * For implementing when an additional logic is needed before
- * pass the template context parameter to the template.
+ * To be implemented when an additional logic is required before
+ * passing the context parameter to the template.
  *
  * @author Vladimir Balin <krocos@mail.ru>
  */
-interface ContextParameterInterface
+interface Twig_ContextParameterInterface
 {
     /**
      * @return mixed The prepared template context parameter
--- a/lib/Twig/ContextParameterInterface.php
+++ b/lib/Twig/ContextParameterInterface.php
@@ -1,0 +1,15 @@
+<?php
+
+/**
+ * For implementing when an additional logic is needed before
+ * pass the template context parameter to the template
+ *
+ * @author Vladimir Balin <krocos@mail.ru>
+ */
+interface ContextParameterInterface
+{
+    /**
+     * @return mixed The prepared template context parameter
+     */
+    public function prepare();
+}
--- a/lib/Twig/ContextParameterInterface.php
+++ b/lib/Twig/ContextParameterInterface.php
@@ -2,7 +2,7 @@
 
 /**
  * For implementing when an additional logic is needed before
- * pass the template context parameter to the template
+ * pass the template context parameter to the template.
  *
  * @author Vladimir Balin <krocos@mail.ru>
  */
--- a/lib/Twig/Template.php
+++ b/lib/Twig/Template.php
@@ -311,7 +311,7 @@ abstract class Twig_Template implements Twig_TemplateInterface
      */
     public function display(array $context, array $blocks = array())
     {
-        $this->prepareContextParameters($context);
+        $context = $this->prepareContextParameters($context);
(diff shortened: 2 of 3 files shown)


krocos:
It can help here [Symfony #11818](https://github.com/symfony/symfony/issues/11818) too

ninsuo:
:+1: nice feature
Missing docs+tests but wait for other's feedback before spending time on it.

lib/Twig/ContextParameterInterface.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+/**
+ * For implementing when an additional logic is needed before
+ * pass the template context parameter to the template.
+ *
+ * @author Vladimir Balin <krocos@mail.ru>
+ */
+interface ContextParameterInterface
```
stof: the class name should be ``Twig_ContextParameterInterface``. Your interface is currently not autoloadable and polutes the global namespace outside Twig, making conflicts likely

lib/Twig/Template.php
```diff
@@ -568,4 +569,18 @@ protected function getAttribute($object, $item, array $arguments = array(), $typ
 
         return $ret;
     }
+
+    /**
+     * Prepare template context parameters if need.
+     *
+     * @param array $context
+     */
+    private function prepareContextParameters(&$context)
```
stof: please don't use references. This is a bad idea as it often makes performance worse because of reference mismatches (and there is a mismatch here). See http://jpauli.github.io/2014/06/27/references-mismatch.html for the explanation

krocos:
And about tests. What conditions I need to test? Test that the context parameter which implements Twig_ContextParameterInterface passed to the display function will be prepared correctly and passed to the mergeGlobals function already prepared?

ninsuo:
... reference mismatches. I use references very often and you just killed me. dude, I code since 1997 and this is the proof that we learn everyday.

ninsuo:
yeah @krocos, just a simple test to check that your code is working well. I'm not telling that your code doesn't work, but this test will protect your code against future regressions.

krocos:
I'm not sure I understand what's wrong with my tests

sstok:
I wonder if this is the right approach, the `FormContext` example still requires you to create a context manually. The https://github.com/symfony/symfony/issues/11818 is about calling the createView() automatically but now you are creating a context which is still a manual process.

Allowing to use transformer(s) on the complete set variables would make this possible, similar to whats done in FOSRestBundle.

krocos:
The context variables preprocessing it's just the one possible way. But once you create a context you can use it anywhere and you can change the logic that lies behind the context. Because of that I can talk about the help (*not the way out, just help*) in the [symfony/symfony#11818](https://github.com/symfony/symfony/issues/11818) issue. *You create a context just once and use it every time you want and you can control the logic*.

stof:
@krocos the Symfony discussion was about simplifying what symfony users have to do. ``new FormContext($form)`` is not simpler than ``$form->createView()``

krocos:
Yes of of course, but what if you want add more logic to do? Not just `createView()`. And you need to use that logic many times... I understand that it's another way and you can just create the service to apply any logic needed *in Symfony*... And the pull request is not about Symfony and about the Twig template engine :)

fabpot:
I'm closing this PR as Twig must receive a context that is usable directly. Pre-processing can happen in any other layer before the context is sent to Twig.

PR #1778, opened 2015-08-20
Add missing deprecation triggers

stof:
Assetic is extending the ``Twig_Function`` class directly currently, and this was not triggering any deprecation warning


fabpot:
Thank you @stof.

PR #1779, opened 2015-08-20
Callable typehint

Tobion:
Add callable typehint which are available since php 5.4 so are safe to use in twig 2.0.

### DIFF
--- a/lib/Twig/Function.php
+++ b/lib/Twig/Function.php
@@ -13,6 +13,8 @@
  * Represents a template function.
  *
  * @author Fabien Potencier <fabien@symfony.com>
+ *
+ * @see http://twig.sensiolabs.org/doc/templates.html#functions
  */
 class Twig_Function
 {
@@ -21,7 +23,14 @@ class Twig_Function
     private $options;
     private $arguments = array();
 
-    public function __construct($name, $callable, array $options = array())
+    /**
+     * Creates a template function.
+     *
+     * @param string        $name     Name of this function
+     * @param callable|null $callable A callable implementing the function. If null, you need to overwrite the "node_class" option to customize compilation.
+     * @param array         $options  Options array
+     */
+    public function __construct($name, callable $callable = null, array $options = array())
     {
         $this->name = $name;
         $this->callable = $callable;
@@ -42,6 +51,11 @@ class Twig_Function
         return $this->name;
     }
 
+    /**
+     * Returns the callable to execute for this function.
+     *
+     * @return callable|null
+     */
     public function getCallable()
     {
         return $this->callable;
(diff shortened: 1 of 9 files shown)


Tobion:
Ref. similar symfony PR: symfony/symfony#14330

lib/Twig/Function.php
```diff
@@ -21,7 +21,7 @@ class Twig_Function
     private $options;
     private $arguments = array();
 
-    public function __construct($name, $callable, array $options = array())
+    public function __construct($name, callable $callable, array $options = array())
```
stof: this should be nullable IMO, to preserve the case of functions compiled in a special way through a special node class and not matching an actual PHP callable (see the Symfony form function)

Tobion: just for Function or also Test and Filter?

stof: all of them IMO, as all of them allow to specify custom node classes

lib/Twig/Node/Expression/Call.php
```diff
@@ -83,7 +83,7 @@ protected function compileArguments(Twig_Compiler $compiler)
         $compiler->raw(')');
     }
 
-    protected function getArguments($callable, $arguments)
+    protected function getArguments(callable $callable, $arguments)
```
stof: this should also allow ``null`` once function, filter and tests allow it

lib/Twig/Extension/Escaper.php
```diff
@@ -77,7 +84,7 @@ public function getDefaultStrategy($filename)
     {
         // disable string callables to avoid calling a function named html or js,
         // or any other upcoming escaping strategy
-        if (!is_string($this->defaultStrategy) && is_callable($this->defaultStrategy)) {
```
Tobion: I removed that check because it hides problems. If somebody passed an array that is not callable, it will just return the array as-is (which is wrong). Now it will raise an error properly.

stof: @Tobion I suggest making this change in 1.x.

and we could even validate it in the setter

Tobion:
@stof I made the callables nullable, but I still have alot of test failures. Do you understand what is going on?

stof:
@Tobion there are a bunch of cases where the callable seems invalid: https://travis-ci.org/twigphp/Twig/jobs/76491001

lib/Twig/Extension/Escaper.php
stof: there is another valid strategy which is not a string or callable: ``false``

lib/Twig/Extension/Escaper.php
```diff
@@ -55,7 +62,7 @@ public function getFilters()
      * The strategy can be a valid PHP callback that takes the template
      * "filename" as an argument and returns the strategy to use.
      *
-     * @param mixed $defaultStrategy An escaping strategy
+     * @param string|callable $defaultStrategy An escaping strategy
```
stof: ``false`` is also a valid value (no escaping)

Tobion:
I don't see where the errors come from.

lib/Twig/Extension/Escaper.php
Tobion: extracted into #1780

Tobion:
@stof @fabpot ready

fabpot:
Thank you @Tobion.

PR #1780, opened 2015-08-21
fix escaping strategy that is an array but not a callable

Tobion:
https://github.com/twigphp/Twig/pull/1779#discussion-diff-37534003

### DIFF
--- a/lib/Twig/Extension/Escaper.php
+++ b/lib/Twig/Extension/Escaper.php
@@ -91,7 +91,7 @@ class Twig_Extension_Escaper extends Twig_Extension
     {
         // disable string callables to avoid calling a function named html or js,
         // or any other upcoming escaping strategy
-        if (is_array($this->defaultStrategy)) {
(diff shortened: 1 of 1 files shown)


lib/Twig/Extension/Escaper.php
```diff
@@ -78,13 +85,13 @@ public function setDefaultStrategy($defaultStrategy)
...
      *
-     * @return string The default strategy to use for the template
+     * @return string|false The default strategy to use for the template
      */
     public function getDefaultStrategy($filename)
     {
         // disable string callables to avoid calling a function named html or js,
         // or any other upcoming escaping strategy
-        if (!is_string($this->defaultStrategy) && is_callable($this->defaultStrategy)) {
+        if (is_array($this->defaultStrategy)) {
```
stof: this is wrong. You can have other callables than arrays (invokables objects are allowed)

Tobion: fixed, forgot about that.

fabpot:
Thank you @Tobion.

PR #1781, opened 2015-08-21
move property initialization to the declaration when possible

Tobion:
This normalization is what I did in symfony in symfony/symfony#9487

I switched to short array syntax only in properties since I was touching that anyway and it's a good first step in case we want to go this route for everything in the future.

### DIFF
--- a/lib/Twig/Extension/Profiler.php
+++ b/lib/Twig/Extension/Profiler.php
@@ -11,11 +11,11 @@
 
 class Twig_Extension_Profiler extends Twig_Extension
 {
-    private $actives;
+    private $actives = [];
 
     public function __construct(Twig_Profiler_Profile $profile)
     {
-        $this->actives = array($profile);
+        $this->actives[] = $profile;
     }
 
     public function enter(Twig_Profiler_Profile $profile)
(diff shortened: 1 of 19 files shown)


lib/Twig/NodeVisitor/Optimizer.php
```diff
@@ -28,10 +28,9 @@ class Twig_NodeVisitor_Optimizer extends Twig_BaseNodeVisitor
     // obsolete, does not do anything
     const OPTIMIZE_VAR_ACCESS = 8;
 
-    private $loops = array();
-    private $loopsTargets = array();
+    private $loops = [];
+    private $loopsTargets = [];
     private $optimizers;
-    private $prependedNodes = array();
```
Tobion: unsed

stof:
the move of the initialization should be done in the 1.x branch whenever possible IMO, to make it easier to merge 1.x into 2.x during the time both will be maintained. and the same argument is true about not switching to short array syntax yet

lib/Twig/Extension/Profiler.php
```diff
@@ -11,11 +11,11 @@
 
 class Twig_Extension_Profiler extends Twig_Extension
 {
-    private $actives;
+    private $actives = [];
```
stof: given we initialize it with a non-empty array, I don't think this is worth it. It means doing an extra array mutation (you create the empty array first and then you alter it)

Tobion: the other way would add an extra step as well: first it's initialized as null and then changed to an array.

Tobion:
Bug fixes are seldom in twig and have low impact (i.e. in 1.x). I don't think the short array syntax will make it problematic.

fabpot:
Can you revert the short array syntax change? I prefer consistency here.

Tobion:
Done in #1786

PR #1784, opened 2015-08-22
optimized the filesystem loader

fabpot:
* added a cache for templates that do not exist
* sped up Filesystem::exists() by avoiding the creation of exceptions

same as d2b97b49f240efb453e75558f00e17ccb1d915c5 that was reverted because of a BC break
and same as what is in 2.0 (but in a BC way).

### DIFF
--- a/lib/Twig/Loader/Filesystem.php
+++ b/lib/Twig/Loader/Filesystem.php
@@ -21,6 +21,7 @@ class Twig_Loader_Filesystem implements Twig_LoaderInterface, Twig_ExistsLoaderI
 
     protected $paths = array();
     protected $cache = array();
+    protected $errorCache = array();
 
     /**
      * Constructor.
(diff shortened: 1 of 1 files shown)


fabpot:
This one is related to #1785.

This is the only I found to keep BC in all Twig versions. The code is rather ugly but I don't have a better idea. See symfony/symfony#15535 as well and #1773 for some discussions on the same topic.

lib/Twig/Loader/Filesystem.php
```diff
@@ -168,18 +167,33 @@ public function isFresh($name, $time)
 
     protected function findTemplate($name)
     {
+        $throw = func_num_args() > 1 ? func_get_arg(1) : true;
```
fabpot: We cannot add an additional argument without breaking BC here. The argument is added in 2.0 though.

fabpot:
ping @stof

PR #1785, opened 2015-08-22
Faster filesystem loader master

fabpot:
We need to keep calling `findTemplate()` for all methods to keep BC.

### DIFF
--- a/lib/Twig/Loader/Filesystem.php
+++ b/lib/Twig/Loader/Filesystem.php
@@ -144,7 +144,38 @@ class Twig_Loader_Filesystem implements Twig_LoaderInterface
      */
     public function exists($name)
     {
+        return $this->doFindTemplate($this->normalizeName($name));
+    }
+
+    /**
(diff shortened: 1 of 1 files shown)


lib/Twig/Loader/Filesystem.php
```diff
@@ -150,9 +150,44 @@ public function exists($name)
             return true;
         }
 
-        if (isset($this->errorCache[$name])) {
+        try {
+            return false !== $this->findTemplate($name, false);
```
fabpot: We need to keep calling `findTemplate` for BC.

lib/Twig/Loader/Filesystem.php
```diff
@@ -150,9 +150,44 @@ public function exists($name)
...
+        } catch (Twig_Error_Loader $exception) {
```
fabpot: The catch here is for implementation which do not respect the `$throw` argument.

PR #1788, opened 2015-08-22
Master

MohammedThaer:
From the CHANGELOG:
> removed Twig_Autoloader (use Composer instead)

I don't use Twing with Composer I use it with WordPress what should I do?


ninsuo:
You can use this code as workaround (`$sourceDirectory` being Twig's source root folder)

```php
    /**
     * In Twig 2.x, there are no more custom Twig autoloader as Twig
     * should be managed using Composer. But of course, twigfiddle
     * cannot handle it using Composer as it need to support several
     * Twig versions.
     *
     * This code is gently provided by @sarciszewski on GitHub:
     * https://github.com/twigphp/Twig/issues/1646#issuecomment-78349076
     *
     * @param string $sourceDirectory
     */
    protected function registerAutoloader($sourceDirectory)
    {
        spl_autoload_register(function ($class) use ($sourceDirectory) {
            $prefix = 'Twig';
            $base_dir = $sourceDirectory.'/lib/Twig/';

            $len = strlen($prefix);
            if (strncmp($prefix, $class, $len) !== 0) {
                return;
            }

            $relative_class = substr($class, $len);
            $file = $base_dir.str_replace('_', '/', $relative_class).'.php';
            if (file_exists($file)) {
                require $file;
            }
        });
    }
```

GromNaN:
You can use composer with WordPress as precisely described here: https://roots.io/using-composer-with-wordpress/

MohammedThaer:
@ninsuo Thank you, it works!, I just removed the last slash from
`$base_dir = $sourceDirectory.'/lib/Twig/';` and made it `$base_dir = $sourceDirectory.'/lib/Twig';`

@GromNaN Thank you, I don't want to use WordPress with Composer, I use Twig in my free/premium plugins not to create a web applications, so I prefer to make my plugins lighter.

PR #1789, opened 2015-08-23
added a note about TwigFiddle

fabpot:
Not sure where to mention it, the IDE section seems ok.

ping @ninsuo

### DIFF
--- a/doc/templates.rst
+++ b/doc/templates.rst
@@ -60,6 +60,9 @@ Many IDEs support syntax highlighting and auto-completion for Twig:
 * *Emacs* via `web-mode.el`_
 * *Atom* via the `PHP-twig for atom`_
 
+Also, `TwigFiddle`_ is an online service that allows you to execute Twig templates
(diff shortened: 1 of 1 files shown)


SpacePossum:
:+1: 

(maybe  http://php.net/manual/en/pcre.pattern.php -> https://secure.php.net/manual/en/pcre.pattern.php)

ninsuo:
Yes, this looks good.

Also, as I developed twigfiddle to reproduce issues when asking/answering questions on Stackoverflow, it may be a good idea to notice it in development.rst, something like:

Support is given through Stack Overflow.
If possible, try to reproduce your issue on twigfiddle before asking your question, this will help the community quickly target and answer it.

But that's maybe too much, as Stackoverflow can be new for lots of twig users, using Stackoverflow+twigfiddle may be confusing/discouraging.

Thanks for the link anyway!

fabpot:
Done: http://twig.sensiolabs.org/development

PR #1796, opened 2015-08-26
OPcache support fix

iKwinto:
According this recipe http://twig.sensiolabs.org/doc/recipes.html#refreshing-modified-templates-when-opcache-or-apc-is-enabled when using OPcache:

```
class Twig_Environment_APC extends Twig_Environment
{
    protected function writeCacheFile($file, $content)
    {
        parent::writeCacheFile($file, $content);

        // Compile cached file into bytecode cache
        if (extension_loaded('Zend OPcache') && ini_get('opcache.enable')) {
            opcache_compile_file($file);
        } elseif (extension_loaded('apc') && ini_get('apc.enabled')) {
            apc_compile_file($file);
        }
    }
}
```

it takes an error:

```
Cannot redeclare class __TwigTemplate_12e7606b10ee267fa4174f660f86451aed936cc5680300a6442092f337c910cf
```

This is due to fact that the function ```opcache_compile_file($file)``` executes the file. But it should not do so in accordance with the documentation: http://php.net/manual/en/function.opcache-compile-file.php

There is a bug report: https://bugs.php.net/bug.php?id=66066 (since 2013 :-1:).

This patch prevents reexecution of the compiled template.


stof:
a better solution would be to change the documentation to use ``opcache_invalidate($file)`` here (we don't need to force the compilation as it will be done anyway during the next loading of the file after the invalidation)

iKwinto:
Agree with stof, it`s better sollution.
Just tested — it fixes that error.

fabpot:
Thank you @iKwinto.

PR #1797, opened 2015-08-31
fix escaping strategy detection based on filename

Tobion:
Fixes symfony/symfony#15095

It previously excluded any fileextension (not just `.twig`) which is pretty wrong.

### DIFF
--- a/test/Twig/Tests/FileExtensionEscapingStrategyTest.php
+++ b/test/Twig/Tests/FileExtensionEscapingStrategyTest.php
@@ -16,7 +16,7 @@ class Twig_Tests_FileExtensionEscapingStrategyTest extends PHPUnit_Framework_Tes
      */
     public function testGuess($strategy, $filename)
     {
-        $this->assertEquals($strategy, Twig_FileExtensionEscapingStrategy::guess($filename));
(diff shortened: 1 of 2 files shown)


test/Twig/Tests/FileExtensionEscapingStrategyTest.php
```diff
@@ -28,12 +28,13 @@ public function getGuessData()
             array('html', 'foo'),
             array('html', 'foo.bar.twig'),
             array('html', 'foo.txt/foo'),
-            array('html', 'foo.txt/foo.js/'),
```
Tobion: The problem is that `var_dump(pathinfo('foo.txt/foo.js/', PATHINFO_EXTENSION));` returns `string(2) "js"`. But this test is kinda strange anyway, as you cannot use a directory name here anyway. I don't see how this would be possible in twig.

Tobion: i fixed it because otherwise it would not be consistent that a possible twig suffix is not cut for directories

stof:
:+1:

fabpot:
Thank you @Tobion.

PR #1803, opened 2015-09-03
Drop support for PHP < 5.5

SpacePossum:
I think this condition can be removed.

### DIFF
--- a/test/Twig/Tests/Fixtures/filters/date_immutable.test
+++ b/test/Twig/Tests/Fixtures/filters/date_immutable.test
@@ -1,7 +1,5 @@
 --TEST--
 "date" filter
---CONDITION--
-version_compare(phpversion(), '5.5.0', '>=')
 --TEMPLATE--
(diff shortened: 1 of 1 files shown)


SpacePossum:
btw. this hack can be removed as well:
https://github.com/twigphp/Twig/blob/master/lib/Twig/Extension/Core.php#L471

I just don't know how.. ;)

Tobion:
Why close? Seems correct.

SpacePossum:
I dropped it because it seems the test seems related to DateTimeImmutable objects as well:
https://github.com/twigphp/Twig/blob/master/lib/Twig/Extension/Core.php#L471 
so maybe it would be more correct to update the test (and the code in Core) with those objects rather then dropping the constraint?

Tobion:
I see

Tobion:
Fixed in #1819

SpacePossum:
:+1:  nice!

PR #1805, opened 2015-09-08
[2.x] More flexible integration test.

SpacePossum:
The current test requires that the extending class implements the `getExtensions` method,
however the class doesn't need an extra extension to run, so I made it optional.

I've added a method to get `filters` so making tests for custom filters is more easy (no need to create an `Extension` class). Some goes for custom `functions` and `tests`.

I'll comment the other changes

### DIFF
--- a/lib/Twig/Test/IntegrationTestCase.php
+++ b/lib/Twig/Test/IntegrationTestCase.php
@@ -17,11 +17,45 @@
  */
 abstract class Twig_Test_IntegrationTestCase extends PHPUnit_Framework_TestCase
 {
-    abstract protected function getExtensions();
+    /**
+     * @return string
+     */
     abstract protected function getFixturesDir();
 
     /**
-     * @dataProvider getTests
+     * @return Twig_ExtensionInterface[]
+     */
+    protected function getExtensions()
+    {
+        return array();
+    }
+
+    /**
+     * @return Twig_SimpleFilter[]
+     */
+    protected function getFilters()
+    {
+        return array();
+    }
+
+    /**
+     * @return Twig_SimpleFunction[]
+     */
+    protected function getFunctions()
+    {
+        return array();
+    }
+
+    /**
+     * @return Twig_SimpleTest[]
+     */
+    protected function getTests()
+    {
+        return array();
+    }
+
+    /**
+     * @dataProvider getUnitTests
      */
     public function testIntegration($file, $message, $condition, $templates, $exception, $outputs)
     {
@@ -37,7 +71,7 @@ abstract class Twig_Test_IntegrationTestCase extends PHPUnit_Framework_TestCase
         $this->testIntegration($file, $message, $condition, $templates, $exception, $outputs);
     }
 
-    public function getTests($name, $legacyTests = false)
+    public function getUnitTests($name, $legacyTests = false)
     {
         $fixturesDir = realpath($this->getFixturesDir());
         $tests = array();
@@ -48,6 +82,11 @@ abstract class Twig_Test_IntegrationTestCase extends PHPUnit_Framework_TestCase
             }
 
             if ($legacyTests xor false !== strpos($file->getRealpath(), '.legacy.test')) {
+                if (empty($tests)) {
+                    // add a dummy test to avoid a PHPUnit message
+                    $tests[] = array('not', '-', '', array(), '', array());
+                }
+
                 continue;
             }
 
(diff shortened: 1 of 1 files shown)


lib/Twig/Test/IntegrationTestCase.php
```diff
@@ -17,27 +17,52 @@
...
+    {
+        return array();
     }
 
     /**
-     * @dataProvider getLegacyTests
-     * @group legacy
+     * @return Twig_SimpleFunction[]
      */
-    public function testLegacyIntegration($file, $message, $condition, $templates, $exception, $outputs)
```
SpacePossum: This can go after 9561582e2f629f661852a44e033f4d832e619aa5
(unless we want to keep this end point for others, however I don't think it is used much)

lib/Twig/Test/IntegrationTestCase.php
```diff
@@ -72,19 +93,9 @@ public function getTests($name, $legacyTests = false)
             $tests[] = array(str_replace($fixturesDir.'/', '', $file), $message, $condition, $templates, $exception, $outputs);
         }
 
-        if (!$tests) {
-            // add a dummy test to avoid a PHPUnit message
-            return array(array('not', '-', '', array(), '', array()));
-        }
```
SpacePossum: This should go because if there are no tests, it should fail in the same way unit tests fail if there are no test cases
```
(1) Warning
No tests found in class "FooBarTest".
```

lib/Twig/Test/IntegrationTestCase.php
stof: this should be kept IMO. We don't have legacy tests in 2.0.0 because we are cleaning all legacy layers, but we will need it again in 2.x when deprecating stuff again

SpacePossum: I see, I didn't know that. 
Could the legacy code testing be moved to its own class, this would lead to a cleaner `IntegrationTestCase` (and history of that file)? Or would that take away the whole point of the test?

Tobion: But there will be legacy tests in the future again. So this should not be removed.

lib/Twig/Test/IntegrationTestCase.php
```diff
@@ -17,27 +17,52 @@
...
     {
-        $this->testIntegration($file, $message, $condition, $templates, $exception, $outputs);
+        return array();
     }
 
-    public function getTests($name, $legacyTests = false)
+    /**
+     * @return Twig_SimpleTest[]
+     */
+    protected function getCustomTests()
```
Tobion: the naming should be consistent. you could rename the test provider method.

lib/Twig/Test/IntegrationTestCase.php
SpacePossum: I see the point of having legacy test, I'm just wondering if `IntegrationTestCase` is the best place to handle those. 
If the `IntegrationTestCase` is meant as end point for others to reuse as base for fixtures tests  (outside of the Twig codebase) then I think the legacy support can better be done in another test.

lib/Twig/Test/IntegrationTestCase.php
SpacePossum: thanks, will update :)

lib/Twig/Test/IntegrationTestCase.php
stof: well, to me, it is, because it makes it very easy to handle things. If you handle it in a separate testcase, it means you have to register it twice and separate the storage of legacy and non-legacy fixtures.

The support of legacy integration tests is also part of the extension. Libraries extending Twig are also allowed to deprecate part of their own API while keeping it tested

SpacePossum: Thanks for explaining, it makes sense to me :) I've moved the legacy tests logic back.

lib/Twig/Test/IntegrationTestCase.php
```diff
@@ -37,7 +71,7 @@ public function testLegacyIntegration($file, $message, $condition, $templates, $
         $this->testIntegration($file, $message, $condition, $templates, $exception, $outputs);
     }
 
-    public function getTests($name, $legacyTests = false)
+    public function getUnitTests($name, $legacyTests = false)
```
stof: why renaming it ?

lib/Twig/Test/IntegrationTestCase.php
```diff
@@ -48,6 +82,11 @@ public function getTests($name, $legacyTests = false)
             }
 
             if ($legacyTests xor false !== strpos($file->getRealpath(), '.legacy.test')) {
+                if (empty($tests)) {
+                    // add a dummy test to avoid a PHPUnit message
+                    $tests[] = array('not', '-', '', array(), '', array());
```
stof: this should be added only at the end of the loop. We don't want to add tons of dummy tests

lib/Twig/Test/IntegrationTestCase.php
```diff
@@ -17,11 +17,45 @@
...
+     */
+    protected function getFunctions()
+    {
+        return array();
+    }
+
+    /**
+     * @return Twig_SimpleTest[]
+     */
+    protected function getTests()
```
stof: reusing an existing method name for a different purpose is a very bad idea. It makes it very hard to support both Twig 1.x and 2.x at the same time as the method has a different meaning

stof:
This could even be implemented in 1.x. There is nothing preventing to implement it in a BC way

lib/Twig/Test/IntegrationTestCase.php
SpacePossum: it is an attempted to follow this suggestion: https://github.com/twigphp/Twig/pull/1805#discussion_r38908773

PR #1806, opened 2015-09-08
More flexible integration test

SpacePossum:
replaces https://github.com/twigphp/Twig/pull/1805 as a 1.x version

### DIFF
--- a/lib/Twig/Test/IntegrationTestCase.php
+++ b/lib/Twig/Test/IntegrationTestCase.php
@@ -17,8 +17,42 @@
  */
 abstract class Twig_Test_IntegrationTestCase extends PHPUnit_Framework_TestCase
 {
-    abstract protected function getExtensions();
+    /**
+     * @return string
+     */
     abstract protected function getFixturesDir();
+
+    /**
+     * @return Twig_ExtensionInterface[]
+     */
+    protected function getExtensions()
+    {
+        return array();
+    }
+
+    /**
+     * @return Twig_SimpleFilter[]
+     */
+    protected function getFilters()
+    {
+        return array();
+    }
+
+    /**
+     * @return Twig_SimpleFunction[]
+     */
+    protected function getFunctions()
+    {
+        return array();
+    }
+
+    /**
+     * @return Twig_SimpleTest[]
+     */
+    protected function getTwigTests()
+    {
+        return array();
+    }
 
     /**
      * @dataProvider getTests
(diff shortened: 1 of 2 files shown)


lib/Twig/Test/IntegrationTestCase.php
```diff
@@ -17,10 +17,44 @@
...
+     */
+    protected function getFilters()
+    {
+        return array();
+    }
+
+    /**
+     * @return Twig_SimpleFunction[]
+     */
+    protected function getFunctions()
```
Tobion: how about getTwigFunctions to be consistent?

test/Twig/Tests/LegacyIntegrationTest.php
```diff
@@ -22,6 +22,15 @@ public function getFixturesDir()
...
     }
+
+    public function getTests($name, $legacyTests = false)
+    {
+        if (!$legacyTests) {
+            return array(array('not', '-', '', array(), '', array()));
+        }
+
+        return parent::getTests($name, true);
+    }
```
SpacePossum: The test will no fail again if there are no fixtures to test (a minor BC which is reverted by this). 
The exception on that rule is now placed in the test that needs that exception and no longer in the parent.

lib/Twig/Test/IntegrationTestCase.php
```diff
@@ -17,10 +17,44 @@
...
-    abstract protected function getExtensions();
+    /**
+     * @return string
+     */
     abstract protected function getFixturesDir();
 
     /**
+     * @return Twig_ExtensionInterface[]
+     */
+    protected function getExtensions()
```
SpacePossum: this cannot be renamed since it would be a BC break

lib/Twig/Test/IntegrationTestCase.php
```diff
@@ -17,10 +17,44 @@
...
+    {
+        return array();
+    }
+
+    /**
+     * @return Twig_SimpleFilter[]
+     */
+    protected function getTwigFilters()
```
fabpot: Why not just `getFilters`?

fabpot:
Do you any use case in mind? I don't see when this could be interesting for end users.

lib/Twig/Test/IntegrationTestCase.php
Tobion: because `getTests` is already taken and thus it would be inconsistent

SpacePossum:
Yes :)
I use this in my projects to test custom `filters`, so I create a `.test` file, set the fixture directory and I'm all set. That is why I added the option to just register custom filters and not only `extensions` (and allow not setting `extensions`).

The second thing I try to fix here; is that it used to be that this test would fail if it won't find any `.test` files, but now the test passes (even for non-legacy items).

So yes, I'm trying to fix two things in one PR once more, so I'm happy to revert/split up whatever needed.

lib/Twig/Test/IntegrationTestCase.php
SpacePossum: I'm happy to name it whatever you (Stof, Tobian and Fabien) think is best

fabpot:
Thank you @SpacePossum.

SpacePossum:
thank you @fabpot :+1:

PR #1807, opened 2015-09-09
Inline original source code in compiled templates

nicolas-grekas:
For introspection purposes (e.g. instead of https://github.com/symfony/symfony/pull/15653) it is sometimes useful to be able to retrieve the original template source code.

I propose to inline it commented after the compiled class declaration.
This creates no memory nor cpu overhead while still allowing to get the source with very simple parsing logic.

### DIFF
--- a/lib/Twig/Environment.php
+++ b/lib/Twig/Environment.php
@@ -573,7 +573,13 @@ class Twig_Environment
     public function compileSource($source, $name = null)
     {
         try {
-            return $this->compile($this->parse($this->tokenize($source, $name)));
+            $compiled = $this->compile($this->parse($this->tokenize($source, $name)), $source);
+
+            if (isset($source[0])) {
+                $compiled .= '// '.str_replace(array("\r\n", "\r", "\n"), "\n// ", $source)."\n";
+            }
+
+            return $compiled;
         } catch (Twig_Error $e) {
             $e->setTemplateFile($name);
             throw $e;
--- a/test/Twig/Tests/EnvironmentTest.php
+++ b/test/Twig/Tests/EnvironmentTest.php
@@ -125,6 +125,14 @@ class Twig_Tests_EnvironmentTest extends PHPUnit_Framework_TestCase
         }
     }
 
+    public function testCompileSource()
+    {
+        $twig = new Twig_Environment($this->getMock('Twig_LoaderInterface'));
+
+        $expected = file_get_contents(__DIR__.'/Fixtures/environment-compile.php');
+        $this->assertSame($expected, $twig->compileSource("{{ foo }}\n{{ bar }}\n", 'index'));
+    }
+
     public function testExtensionsAreNotInitializedWhenRenderingACompiledTemplate()
     {
         $options = array('cache' => sys_get_temp_dir().'/twig', 'auto_reload' => false, 'debug' => false);
--- a/test/Twig/Tests/Fixtures/environment-compile.php
+++ b/test/Twig/Tests/Fixtures/environment-compile.php
@@ -1,0 +1,45 @@
+<?php
+
+/* index */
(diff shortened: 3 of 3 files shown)


lib/Twig/Environment.php
```diff
@@ -551,13 +551,20 @@ public function setCompiler(Twig_Compiler $compiler)
...
      *
      * @return string The compiled PHP source code
      */
-    public function compile(Twig_Node $node)
+    public function compile(Twig_Node $node, $source = null)
     {
-        return $this->getCompiler()->compile($node)->getSource();
+        $compiled = $this->getCompiler()->compile($node)->getSource();
+
+        if (isset($source[0])) {
```
Tobion: this check looks strange misuse

nicolas-grekas: faster version for `0 < strlen($source)`

lib/Twig/Environment.php
```diff
@@ -551,13 +551,20 @@ public function setCompiler(Twig_Compiler $compiler)
...
      * Compiles a node and returns the PHP code.
      *
-     * @param Twig_Node $node A Twig_Node instance
+     * @param Twig_Node   $node   A Twig_Node instance
+     * @param string|null $source The template source code
      *
      * @return string The compiled PHP source code
      */
-    public function compile(Twig_Node $node)
+    public function compile(Twig_Node $node, $source = null)
```
stof: should the signature of this method really be changed, or should it be added only in ``compileSource`` ?

stof:
This should probably be implemented in Twig 1.x rather than only Twig 2 (otherwise Symfony will still need to maintain a separate implementation for Twig 1)

lib/Twig/Environment.php
```diff
@@ -551,13 +551,20 @@ public function setCompiler(Twig_Compiler $compiler)
...
     {
-        return $this->getCompiler()->compile($node)->getSource();
+        $compiled = $this->getCompiler()->compile($node)->getSource();
+
+        if (isset($source[0])) {
+            $compiled .= '// '.str_replace(array("\r\n", "\r", "\n"), "\n// ", $source)."\n";
```
fabpot: Having a very long one liner is very problematic in some text editor that do not deal very well with wide texts.

nicolas-grekas: You misread, this is not a very long line, but many commented lines, same amount as the template source code

fabpot: indeed, can you update the test to have multiple lines then?

lib/Twig/Environment.php
nicolas-grekas: signature change reverted

lib/Twig/Environment.php
nicolas-grekas: test updated

nicolas-grekas:
Comments addressed, should be merged in 1.x yes

lib/Twig/Environment.php
```diff
@@ -573,7 +573,13 @@ public function compile(Twig_Node $node)
     public function compileSource($source, $name = null)
     {
         try {
-            return $this->compile($this->parse($this->tokenize($source, $name)));
+            $compiled = $this->compile($this->parse($this->tokenize($source, $name)), $source);
+
+            if (isset($source[0])) {
+                $compiled .= '// '.str_replace(array("\r\n", "\r", "\n"), "\n// ", $source)."\n";
```
fabpot: Can we add the source only when debug is true?

nicolas-grekas: yes we can :) but why? getDebugInfo does not dependent on debug, and currently, the debug flag has no impact on the compile stage

fabpot:
Thank you @nicolas-grekas.

Issue #1811, opened 2015-09-10
private runtimeInitialized property in Twig_Environment is a barrier to overriding loadTemplate()


Cottser:
This issue is for 2.x. Coming from https://www.drupal.org/node/2555243.

Drupal 8 extends `Twig_Environment`, and because the `runtimeInitialized` property is now private in 2.x our override of `loadTemplate()` no longer works.

I think we need to either add a getter or make it protected. Once there is a clear path forward I'm happy to submit a PR.

This looks to be the only barrier to our upgrading to Twig 2.x as we near RC1 that is outside of our control. This was changed in #1773.

fabpot:
#1812 should fix your issue and allow Drupal to stop copy/pasting some Twig code, which is always a good thing.

Cottser:
@fabpot first of all, thanks for all the help with this.

This is still proving to be an issue in Drupal 8, the cache class implementation does not allow us the protection from the race condition where load() == TRUE but then the storage goes away.

See around https://www.drupal.org/node/2568171#comment-10337391 for some comments.

In short I think we still need to override loadTemplate() in Drupal 8 to add the eval() if the storage fails.

If you can think of another way that loadTemplate() and Twig_CacheInterface could be implemented to support this of course that would be great but if it's not a common use case then just allowing us to override loadTemplate() might be for the best.

Cottser:
I may submit a PR to add another method to the cache interface under the assumption that what we're trying to do is a somewhat common use case. Stay tuned.

fabpot:
@Cottser twigphp/Twig#1840 is now merged. Does it cover everything now? Or do you still have issue on the Drupal side?

Cottser:
@fabpot looks great now and I think we can now be fully Twig 2.x compatible, that's what I'm focussing on today at DrupalCon :)

Thanks!

Cottser:
@fabpot just so you know you can see at the top of https://www.drupal.org/node/2555243 what our roadmap looks like for getting ready for 2.x at this point. If you feel like it's the right time to cut 1.22.2 that would be very helpful for Drupal.

At https://www.drupal.org/node/2568181#comment-10355611 there are a couple patches which are passing on our newer test infrastructure (old test infrastructure is pretty blocked up ATM). One patch is for 1.x and for 2.x, the only difference between the two is the actual Twig library update and the 2.x patch effectively has the first patch from https://www.drupal.org/node/2571817#comment-10351435 applied.

Issue #1573, opened 2014-12-01
Cache Interface


baptistedonaux:
My issue is very ambitious but I think that it can interest any people.

I would like store the cached template in an other service that Filesystem. Actually, to enable the cache into Twig, we can only put the absolute path and store the cache on a Filesystem. In my case, I would like implement a interface and get my template since an other source.

My issue is a little crazy but I think that other people can be interested to make a renderer most faster per example.

PR #1812, opened 2015-09-11
Extracted the filesystem cache to its own class

fabpot:
Twig caches the compiled PHP classes on the filesystem. This is **always** the best strategy as it allows Twig to automatically benefit from PHP opcache/APC. But overriding how the classes are stored on the filesystem is difficult with the current way, so this PR proposes to extract this logic to its own class (that should allow Drupal to stop copy/pasting some Twig code and accessing private code - see #1811).

This PR also unifies the no-cache feature by extracting it to its own class as well.

BC is kept and the `false` and `$dir` caching strategies are still supported.

The `Twig_Cache_Interface` can be used to create other cache classes, but that's not documented and should only be used with extreme care (performance-wise). It means that Twig itself will **never** ship with other implementations like Memcache, Redis, whatever storage is hype nowadays. This should fix #1421 (this PR builds on top of it), #1573, #1415, #741, #728.


lib/Twig/Cache/Null.php
```diff
@@ -0,0 +1,58 @@
...
+ * @author Fabien Potencier <fabien@symfony.com>
+ */
+class Twig_Cache_Null implements Twig_CacheInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getCacheKey($className, $prefix)
+    {
+        return false;
```
jibran: Interface says it returns string so should it be `''` instead?

fabpot: fixed

lib/Twig/Environment.php
```diff
@@ -213,24 +216,37 @@ public function isStrictVariables()
     }
 
     /**
-     * Gets the cache directory or false if cache is disabled.
+     * Gets the current cache implementation.
      *
-     * @return string|false
+     * @return Twig_CacheInterface|string|false A Twig_CacheInterface implementation,
```
Taluu: the return string or false looks wrong IMO, as these are converted to `Twig_CacheInterface` anyway in the only entry point for this property (which is `setCache`)

lib/Twig/Environment.php
```diff
@@ -326,15 +338,13 @@ public function loadTemplate($name, $index = null)
...
-            if (false === $cache = $this->getCacheFilename($name)) {
-                eval('?>'.$this->compileSource($this->getLoader()->getSource($name), $name));
-            } else {
-                if (!is_file($cache) || ($this->isAutoReload() && !$this->isTemplateFresh($name, filemtime($cache)))) {
-                    $this->writeCacheFile($cache, $this->compileSource($this->getLoader()->getSource($name), $name));
-                }
+            $key = $this->cache->getCacheKey($cls, $this->templateClassPrefix);
 
-                require_once $cache;
+            if (!$this->cache->has($key) || ($this->isAutoReload() && !$this->isTemplateFresh($name, $this->cache->getTimestamp($key)))) {
```
Taluu: how about moving the auto reload property into the cache, which lets it decide if it should autoreload or not in the `has` method ? And also moving the `isTemplateFresh` into the cache too ?

lib/Twig/Environment.php
fabpot: we return originalCache here, not cache

Taluu: Ah yes, didn't see it was to kept bc as it is also set in the `setCache` method. My bad.

doc/deprecated.rst
```diff
@@ -143,6 +143,9 @@ Miscellaneous
 * As of Twig 1.x, ``Twig_Environment::clearTemplateCache()`` is deprecated and
   will be removed in 2.0.
 
+* As of Twig 1.x, ``Twig_Environment::clearCacheFiles()`` is deprecated and
```
stof: I would say ``as of Twig 1.22``. Deprecations are always added in a specific release. ``As of 1.x`` does not mean anything

lib/Twig/Cache/Filesystem.php
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getCacheKey($className, $prefix)
+    {
+        $class = substr($className, strlen($prefix));
+
+        return $this->directory.'/'.$class[0].'/'.$class[1].'/'.$class.'.php';
```
stof: what if ``$class`` is shorter than 2 chars ? this code should be safe about what it receives

lib/Twig/Cache/Filesystem.php
```diff
@@ -0,0 +1,104 @@
...
+                    throw new RuntimeException(sprintf('Unable to create the cache directory (%s).', $dir));
+                }
+            }
+        } elseif (!is_writable($dir)) {
+            throw new RuntimeException(sprintf('Unable to write in the cache directory (%s).', $dir));
+        }
+
+        $tmpFile = tempnam($dir, basename($key));
+        if (false !== @file_put_contents($tmpFile, $content)) {
+            // rename does not work on Win32 before 5.2.6
```
stof: this can be simplified: the min requirement is PHP 5.2.7

lib/Twig/Cache/Filesystem.php
```diff
@@ -0,0 +1,104 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function getTimestamp($key)
+    {
+        return filemtime($key);
+    }
+
+    /**
+     * @deprecated since 1.22 (to be removed in 2.0)
```
stof: the deprecation warning should be triggered

lib/Twig/CacheInterface.php
```diff
@@ -0,0 +1,65 @@
...
+ */
+interface Twig_CacheInterface
+{
+    /**
+     * Returns the cache key for the given template class name.
+     *
+     * @param string $className The template class name
+     * @param string $prefix    A template class prefix
+     *
+     * @return string|false
```
stof: what can it return false ?

lib/Twig/Environment.php
```diff
@@ -213,24 +216,37 @@ public function isStrictVariables()
...
-     * @param string|false $cache The absolute path to the compiled templates,
-     *                            or false to disable cache
+     * @param Twig_CacheInterface|string|false $cache A Twig_CacheInterface implementation,
+     *                                                an absolute path to the compiled templates,
+     *                                                or false to disable cache
      */
     public function setCache($cache)
     {
-        $this->cache = $cache ? $cache : false;
+        $this->originalCache = $cache;
```
stof: this should be done only after validating it. Otherwise, someone calling ``setCache`` and catching the exception would be able to set an invalid value in this property.

lib/Twig/Environment.php
```diff
@@ -242,13 +258,9 @@ public function setCache($cache)
      */
     public function getCacheFilename($name)
```
stof: should this method be deprecated too ? there is nothing forcing to use a filename as cache key (if the cache is not a filesystem one, it is very unlikely)

lib/Twig/Environment.php
stof: @Taluu this would force to make the cache aware of the loader and of the extensions, which is not a good idea.

lib/Twig/Cache/Filesystem.php
stof: the method should also be marked as ``@internal`` as it is added in this class only to be able to implement the BC layer but people should not use it.

lib/Twig/Environment.php
```diff
@@ -1280,28 +1288,10 @@ protected function initExtension(Twig_ExtensionInterface $extension)
 
     protected function writeCacheFile($file, $content)
```
Tobion: This method is currently not called anymore which is not BC. So I guess we still need to call it from loadTemplate but disable the deprecation in this case (via additional optional argument).

stof: this is quite hard, because we should not call it when the Null cache is used (otherwise it is not BC either)

stof: hmm, we could call it only when overwritten though, and trigger a deprecation warning in this case (otherwise people overwriting it won't get the deprecation warning)

lib/Twig/Cache/Filesystem.php
fabpot: That can never happen.

lib/Twig/Cache/Filesystem.php
fabpot: see #1815

lib/Twig/Cache/Filesystem.php
fabpot: `@internal` added.

I have not added a deprecation notice as there is already one is the caller. As the method is now marked as `@internal`, I think that's enough.

lib/Twig/CacheInterface.php
fabpot: not possible anymore

lib/Twig/Environment.php
fabpot: Indeed, done.

lib/Twig/Environment.php

lib/Twig/Environment.php
stof: Actually, this method was an extension point previously for people extending the class, as it allowed to change the filename in the cache. This PR breaks BC for this extension point (extension point based on inheritance are a huge pain regarding BC)

lib/Twig/Cache/Filesystem.php
Tobion: maybe not in the current use-case. but the class should be working in different scenarios as well, i.e. standalone.

fabpot: I disagree, this class is only meant to be used with Twig, it's not a generic cache class.

lib/Twig/Environment.php
fabpot: I've implemented something to keep BC here. Tell me what you think about it.

lib/Twig/Environment.php
fabpot: I've done the same strategy as for the write method.

fabpot:
I think I've covered all comments. Feedback welcome.

lib/Twig/Environment.php
```diff
@@ -112,6 +117,23 @@ public function __construct(Twig_LoaderInterface $loader = null, $options = arra
         $this->addExtension(new Twig_Extension_Escaper($options['autoescape']));
         $this->addExtension(new Twig_Extension_Optimizer($options['optimizations']));
         $this->staging = new Twig_Extension_Staging();
+
+        // For BC
+        if (is_string($this->originalCache)) {
+            $r = new ReflectionMethod($this, 'writeCacheFile');
+            if ($r->getDeclaringClass() === get_class($this)) {
```
stof: you should rather check ``!== __CLASS__`` here, to account for different level of inheritance

fabpot: good idea, done.

stof:
some tests are trigger notices in non-legacy cases

fabpot:
@stof fixed

Tobion:
:+1:

test/Twig/Tests/EnvironmentTest.php
```diff
@@ -150,15 +150,17 @@ public function testCompileSourceInlinesSource()
 
     public function testExtensionsAreNotInitializedWhenRenderingACompiledTemplate()
     {
-        $options = array('cache' => sys_get_temp_dir().'/twig', 'auto_reload' => false, 'debug' => false);
+        $options = array(
+            'cache' => $cache = new Twig_Cache_Filesystem($dir = sys_get_temp_dir().'/twig'),
```
Tobion: the cache instantiation could be on a separate line. it's too confusing.

fabpot: fixed

fabpot:
Thank you @andrewtch.

lib/Twig/Environment.php
```diff
@@ -326,15 +363,21 @@ public function loadTemplate($name, $index = null)
...
         if (!class_exists($cls, false)) {
-            if (false === $cache = $this->getCacheFilename($name)) {
-                eval('?>'.$this->compileSource($this->getLoader()->getSource($name), $name));
+            if ($this->bcGetCacheFilename) {
+                $key = $this->getCacheFilename($name);
             } else {
-                if (!is_file($cache) || ($this->isAutoReload() && !$this->isTemplateFresh($name, filemtime($cache)))) {
-                    $this->writeCacheFile($cache, $this->compileSource($this->getLoader()->getSource($name), $name));
-                }
+                $key = $this->cache->generateKey($cls, $this->templateClassPrefix);
```
Cottser: This change from `getCacheFilename($name)` to `$this->cache->generateKey($cls, $this->templateClassPrefix)` is proving to be a bit of a problem in the end goal of trying *not* to override loadTemplate() in Drupal 8. We use `$name` as part of the cache key.

fabpot: Understood. I was not happy to leak such an implementation detail (the class prefix) and #1822 tries to fix your issue and provide a much better UX. Tell me if that works for you.

andrewtch:
@fabpot , you're welcome. Too bad I won't get SC badge for that )

PR #1813, opened 2015-09-11
Add $template->getSource()

nicolas-grekas:
Now that we have the source inlined, we should provide a way to get it. This also make it possible to feature-test if the source is available for backward/forward compat.

### DIFF
--- a/lib/Twig/Template.php
+++ b/lib/Twig/Template.php
@@ -305,6 +305,33 @@ abstract class Twig_Template implements Twig_TemplateInterface
     }
 
     /**
+     * Returns the template source code.
+     *
+     * @return string|null The template source code or null if it is not available
+     */
+    public function getSource()
+    {
+        $reflector = new ReflectionClass($this);
+        $file = $reflector->getFileName();
+
+        if (!file_exists($file)) {
+            return;
+        }
+
+        $source = file($file, FILE_IGNORE_NEW_LINES);
+        array_splice($source, 0, $reflector->getEndLine());
+
+        $i = 0;
+        while (isset($source[$i]) && '/* */' === substr_replace($source[$i], '', 3, -2)) {
+            $source[$i] = str_replace('*//*', '*/', substr($source[$i], 3, -2));
+            ++$i;
+        }
+        array_splice($source, $i);
+
+        return implode("\n", $source);
+    }
+
+    /**
      * {@inheritdoc}
      */
     public function display(array $context, array $blocks = array())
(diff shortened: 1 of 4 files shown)


lib/Twig/Template.php
```diff
@@ -305,6 +305,33 @@ public function getBlocks()
     }
 
     /**
+     * Returns the template source code.
+     *
+     * @return string|false The template source code or false if it can't be retrieved
```
stof: I would rather return ``null`` than ``false``. Using a boolean for a not available value is not right semantically. this is what ``null`` is for.

nicolas-grekas: fixed

stof:
:+1: (except the change requested by fabbot)

lib/Twig/Template.php
```diff
@@ -305,6 +305,33 @@ public function getBlocks()
     }
 
     /**
+     * Returns the template source code.
+     *
+     * @return string|null The template source code or null if it can't be retrieved
```
Tobion: Why not throw an exception instead of returning null? I mean the null case is an error condition. And IMO multi-values return types should not be used for that.

nicolas-grekas: It's not an error condition nor a logic mistake, just a case where the source code is not available. I could return an empty string if you prefer, but adding an exception will just force people to add try/catch boilerplate where a simple if would be enough

lib/Twig/Template.php
```diff
@@ -305,6 +305,33 @@ public function getBlocks()
...
+    {
+        $reflector = new \ReflectionClass($this);
+        $file = $reflector->getFileName();
+
+        if (!file_exists($file)) {
+            return;
+        }
+
+        $source = file($file);
+        array_splice($source, 0, $reflector->getEndLine());
```
Tobion: I think using `array_slice` would be easier and more straight-forward.

nicolas-grekas: but would consume more memory, requiring a temporary duplication of the array...

lib/Twig/Template.php
```diff
@@ -305,6 +305,33 @@ public function getBlocks()
     }
 
     /**
+     * Returns the template source code.
+     *
+     * @return string|null The template source code or null if it can't be retrieved
+     */
+    public function getSource()
+    {
+        $reflector = new \ReflectionClass($this);
```
fabpot: Not compatible with PHP 5.2 :)

nicolas-grekas: Fixed

fabpot:
Thank you @nicolas-grekas.

PR #1817, opened 2015-09-11
Fix inlining of sources with CRLF

nicolas-grekas:


fabpot:
Thank you @nicolas-grekas.

lib/Twig/Environment.php
```diff
@@ -641,7 +641,7 @@ public function compileSource($source, $name = null)
             $compiled = $this->compile($this->parse($this->tokenize($source, $name)), $source);
 
             if (isset($source[0])) {
-                $compiled .= '/* '.str_replace(array("\r\n", "\r", "\n"), "*/\n/* ", str_replace('*/', '*//*', $source))."*/\n";
+                $compiled .= '/* '.strtr($source, array('*/' => '*//*', "\r\n" => "*/\n/* ", "\r" => "*/\n/* ", "\n" => "*/\n/* "))."*/\n";
```
Tobion: Why not keep CRLF in the code? You replace with with LF only. This changes the source code line endings. So it does not actually return the real source code.

nicolas-grekas: because generated templates are LF-only files and also because twig already makes no difference and replaces CRLF/CR by LF
at least that's my reasoning :)

Tobion: ok but then we should use str_replace which makes it way faster: #1818

PR #1818, opened 2015-09-11
use str_replace instead of strtr to escape the source in the template

Tobion:
strtr is really slow when used this way.

Benchmark https://3v4l.org/PXWDg/perf#tabs vs https://3v4l.org/iuq1Z/perf#tabs

strtr is only slightly faster when used with a byte-by-byte replacement as done in https://github.com/twigphp/Twig/blob/1.x/lib/Twig/Lexer.php#L263

### DIFF
--- a/lib/Twig/Loader/Filesystem.php
+++ b/lib/Twig/Loader/Filesystem.php
@@ -233,7 +233,7 @@ class Twig_Loader_Filesystem implements Twig_LoaderInterface, Twig_ExistsLoaderI
 
     protected function normalizeName($name)
     {
-        return preg_replace('#/{2,}#', '/', strtr((string) $name, '\\', '/'));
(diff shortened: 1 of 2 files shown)


lib/Twig/Loader/Filesystem.php
```diff
@@ -233,7 +233,7 @@ protected function parseName($name, $default = self::MAIN_NAMESPACE)
 
     protected function normalizeName($name)
     {
-        return preg_replace('#/{2,}#', '/', strtr((string) $name, '\\', '/'));
+        return preg_replace('#/{2,}#', '/', str_replace('\\', '/', (string) $name));
```
Tobion: ref symfony/symfony#15614

nicolas-grekas:
:+1:

fabpot:
Thank you @Tobion.

PR #1820, opened 2015-09-11
no need to call setTimezone on a current date

Tobion:
Avoiding unnecessary method calls.

- When retrieving the current time, there is no point changing the timezone from default to the specified timezone, e.g. if you just use twigs `date()` function.
- When dealing with timestamps, the initial timezone is irrelevant as well since unix timstamps are always in GMT.


fabpot:
Thank you @Tobion.

PR #1822, opened 2015-09-12
changed the arguments of generateKey

fabpot:
We now pass the template name and the template class, which seems much better from a UX standpoint.

Passing the prefix was really just a hack and an implementation leak to be able to determine the variable part of the class name and generate sub-directories for templates. Now, we generate the cache key by taking the last characters instead of the first ones, to avoid the need for the class prefix.

That should also make Drupal happy :)

### DIFF
--- a/lib/Twig/Cache/Filesystem.php
+++ b/lib/Twig/Cache/Filesystem.php
@@ -29,11 +29,11 @@ class Twig_Cache_Filesystem implements Twig_CacheInterface
     /**
      * {@inheritdoc}
      */
-    public function generateKey($className, $prefix)
+    public function generateKey($name, $className)
     {
-        $class = substr($className, strlen($prefix));
+        $hash = hash('sha256', $className);
 
-        return $this->directory.'/'.$class[0].'/'.$class[1].'/'.$class.'.php';
+        return $this->directory.'/'.$hash[0].'/'.$hash[1].'/'.$hash.'.php';
     }
 
     /**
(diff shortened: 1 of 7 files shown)


fabpot:
I've also deprecated `Twig_Environment::getTemplateClassPrefix()` as this is really just an implementation detail people should not rely on.

lib/Twig/Cache/Filesystem.php
```diff
@@ -29,11 +29,11 @@ public function __construct($directory)
...
      * {@inheritdoc}
      */
-    public function generateKey($className, $prefix)
+    public function generateKey($name, $className)
     {
-        $class = substr($className, strlen($prefix));
+        $len = strlen($className);
 
-        return $this->directory.'/'.$class[0].'/'.$class[1].'/'.$class.'.php';
+        return $this->directory.'/'.$className[$len - 1].'/'.$className[$len - 2].'/'.$className.'.php';
```
Tobion: The class name might contain the index with a `_x` suffix. In this case the directory structure is not randomized evenly anymore. A simple solution would be to just change ´getTemplateClass` to add the index between prefix and hash.
But this still ties to Cache implementation to the class generation, i.e. the cache is basically aware how the class name is generated.

Tobion: An architectural better solution would be to hash the class name again and then use two chars from the hash. This would also solve the problem when `strlen($className) < 2`.

fabpot: Good idea, code updated accordingly.

Cottser:
Thanks very much @fabpot, will take this for a spin shortly.

Cottser:
Code looks good to me and this unblocked my work, thanks again. Remaining steps for me are:

1. Test race condition or reach out to others who can (https://www.drupal.org/node/2429659), I'm hoping it no longer needs special handling.
2. Test with Twig 2.x.

PR #1823, opened 2015-09-12
rewrote the recipes about APC and opcache to avoid deprecation notices

fabpot:


fabpot:
I'm wondering if we could not add tis feature directly into Twig? Probably by setting an option on the `File_Cache_Filesystem`. What do you think?

SpacePossum:
It would be great to have this directly in Twig, it makes OP caching way easier and less error-prone :+1: 

 (Especially because this is in the recipe part of the documentation. I feel it is important enough to put it somewhere more prominent, since if you don't know about the instructions (/recipe) and switch on APC some pretty cryptic errors might occur).

fabpot:
The second commit integrates the code directly into Twig.

SpacePossum:
:+1:  very nice to have this in, thanks!

lib/Twig/Cache/Filesystem.php
```diff
@@ -73,6 +78,15 @@ public function write($key, $content)
         if (false !== @file_put_contents($tmpFile, $content) && @rename($tmpFile, $key)) {
             @chmod($key, 0666 & ~umask());
 
+            if (self::FORCE_BYTECODE_INVALIDATION == ($this->options & self::FORCE_BYTECODE_INVALIDATION)) {
+                // Compile cached file into bytecode cache
+                if (function_exists('opcache_invalidate') && ini_get('opcache.enable')) {
+                    opcache_invalidate($key);
```
Tobion: I think the second parameter (force) should be set to true as we know the file changed and otherwise the option FORCE_BYTECODE_INVALIDATION would be misleading.

Tobion: #1827

lib/Twig/Cache/Filesystem.php
```diff
@@ -73,6 +78,15 @@ public function write($key, $content)
         if (false !== @file_put_contents($tmpFile, $content) && @rename($tmpFile, $key)) {
             @chmod($key, 0666 & ~umask());
 
+            if (self::FORCE_BYTECODE_INVALIDATION == ($this->options & self::FORCE_BYTECODE_INVALIDATION)) {
```
stof: should be ``===`` to respect coding standards (using strict comparison everywhere it is possible rather than letting PHP convert types)

stof:
should we add an option in the Twig_Environment to control the cache instance it creates or no ?

fabpot:
I don't think we need yet another option in the env. If you want the new feature, just explicitly create the cache instance.

martijn80:
I've read the docs at : https://twig.symfony.com/doc/1.x/recipes.html#refreshing-modified-templates-when-opcache-or-apc-is-enabled

But when I want to use this option in a symfony 2.8 project with twig 1.33 I have no clue where to put the code from the recipe. 

So I tried adding it to config.yml like this, without success ("twig.cache". Expected scalar, but got array."). 

What am I overseeing?

```
twig:
    cache:
        directory: '%kernel.cache_dir%/twig'
        options: '@=constant("Twig\\Cache\\Filesystem::FORCE_BYTECODE_INVALIDATION")'
```

stof:
TwigBundle does not have a way to pass a Twig Cache instance in the option instead of a path. It currently rely on Twig itself to do the instantiation.

so for now, it would require using a compiler pass to change the service configuration:

```php
    public function process(ContainerBuilder $container)
    {
        $def = $container->getDefinition('twig');

        $options = $def->getArgument(1);

        if (!is_string($options['cache'])) {
            return;
        }

        $cacheDef = new Definition(Filesystem::class, array($options['cache'], Filesystem::FORCE_BYTECODE_INVALIDATION);

        $options['cache'] = $cacheDef;

        $def->replaceArgument(1, $options);
    }
```

stof:
I suggest you to open a feature request on Symfony to ask for support for this feature though

martijn80:
Hi stof, thanks for your reply and code example. There was a parenthes missing, but it helped me a lot! This complete example worked for me : 

```php
<?php
namespace Your\Namespace\DependencyInjection;

use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
use Symfony\Component\DependencyInjection\ContainerBuilder;
use Symfony\Component\DependencyInjection\Definition;

class TwigCacheCompilerPass implements CompilerPassInterface
{

    public function process(ContainerBuilder $container)
    {
        $def     = $container->getDefinition('twig');
        $options = $def->getArgument(1);

        if (!is_string($options['cache']))
        {
            return;
        }

        $cacheDef = new Definition(
            \Twig_Cache_Filesystem::class,
            array(
                $options['cache'],
                \Twig_Cache_Filesystem::FORCE_BYTECODE_INVALIDATION,
            )
        );

        $options['cache'] = $cacheDef;

        $def->replaceArgument(1, $options);
    }
}
```

Good idea to request for this feature in Symfony (I didnt realise this was the Twig repo, I thought this was the TwigBundle repo).

PR #1824, opened 2015-09-12
added enabled extension names in the generated template class names

### LINKED ISSUE #1742: share cache between templates with/without the profiler
Currently the cached Twig templates that had the profiler enabled cannot be used when the profiler is _not_ enabled. Is there any plan / idea (or reasoning) behind the fact that the template cache cannot be shared between the two environments?

The main difference between the internal cache is:
```php
{
    $__internal_some_random_hash = $this->env->getExtension("profiler");
    __internal_some_random_hash->enter($__internal_some_random_hash_prof = new Twig_Profiler_Profile($this->getTemplateName(), "template", "some_template.twig"));
    // ...the actual parsed template
    $__internal_some_random_hash->leave($__internal_some_random_hash_prof);
    }
```
and ofcourse the ```getDebugInfo()``` is different ;)

In our situation we utilize various on demand profiling and by reusing the cache seems logical doesnt it? Whats the idea behind this?

fabpot:
The generated template class names now take into account enabled extensions.

fixes #1742

### DIFF
--- a/lib/Twig/Environment.php
+++ b/lib/Twig/Environment.php
@@ -298,7 +298,9 @@ class Twig_Environment
      */
     public function getTemplateClass($name, $index = null)
     {
-        return $this->templateClassPrefix.hash('sha256', $this->getLoader()->getCacheKey($name)).(null === $index ? '' : '_'.$index);
+        $key = $this->getLoader()->getCacheKey($name).'__'.implode('__', array_keys($this->extensions));
+
+        return $this->templateClassPrefix.hash('sha256', $key).(null === $index ? '' : '_'.$index);
     }
 
     /**
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
 * 1.22.0 (2015-XX-XX)
 
+ * changed template cache names to take into account enabled extensions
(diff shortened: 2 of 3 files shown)


lib/Twig/Environment.php
```diff
@@ -298,7 +298,14 @@ public function getCacheFilename($name)
      */
     public function getTemplateClass($name, $index = null)
     {
-        return $this->templateClassPrefix.hash('sha256', $this->getLoader()->getCacheKey($name)).(null === $index ? '' : '_'.$index);
+        $key = $this->getLoader()->getCacheKey($name);
+
+        // add extensions in the template class name
+        foreach ($this->extensions as $name => $_) {
+            $key .= '__'.$name;
```
Tobion: implode(array_keys()) would look cleaner I think.

Tobion:
The cache could also depend on the actual configuration of the extensions, e.g. configs, enabled visitors etc. But I guess at least covering the enabled extensions is better than nothing and should indeed fix the issue referenced. :+1:

fruitl00p:
Awesome! :+1: thanks alot!

stof:
@fabpot should we also change the cache key depending on the availability of the C extension ?

fabpot:
@stof  see #1844

PR #1827, opened 2015-09-13
No need to check modification time when opcache_invalidate

Tobion:
- We know the file is new
- No need to check if opcache/apc is enabled as these methods will just return false in this case


fabpot:
Thank you @Tobion.

PR #1828, opened 2015-09-14
Prevent potential doc comments in escaped inlined source

nicolas-grekas:
Hopefully the last patch of this multi-patches feature...


fabpot:
Thank you @nicolas-grekas.

PR #1830, opened 2015-09-14
String-cast in template_from_string

### LINKED ISSUE #1829: [1.22.0] template_from_string() and object
After upgrading from 1.21.2 to 1.22.0 I get the following error:

``Fatal error: Cannot use object of type Foo as array in D:\htdocs\baubinar\vendor\twig\twig\lib\Twig\Environment.php on line 650``

The template:
``{{ include(template_from_string(foo))|raw }}``

Foo has implemented ``__toString`` and after downgrading to 1.21.2 everything works fine again.

Tobion:
Fixes #1829 caused by https://github.com/twigphp/Twig/pull/1807/files#diff-5e190f538c5c328c7473a5cfd3b7cd95R578

As twig allows many types in twig templates, e.g. array or traversable, we can do the same for string.
But we only take this approach for twig exposed function/filters etc, i.e. template code, but not core code.

### DIFF
--- a/lib/Twig/Extension/StringLoader.php
+++ b/lib/Twig/Extension/StringLoader.php
@@ -37,11 +37,11 @@ class Twig_Extension_StringLoader extends Twig_Extension
  * </pre>
  *
  * @param Twig_Environment $env      A Twig_Environment instance
- * @param string           $template A template as a string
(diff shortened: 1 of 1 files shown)


fabpot:
Thank you @Tobion.

PR #1831, opened 2015-09-15
Add test to check if a block exists

### LINKED ISSUE #1821: Add a new test to check if a block exists
Suggested names:
- `block name`
- `block`
- `defined block`

``` twig
{% if "content" is block name %}
<div>{{ block("content") }}</div>
{% endif %}
```

hason:
Fixes #1821

### DIFF
--- a/doc/tests/block_name.rst
+++ b/doc/tests/block_name.rst
@@ -1,0 +1,13 @@
+``block name``
+==============
+
+.. versionadded:: 1.28
+    The ``block name`` test was added in Twig 1.28.
+
+``block name`` checks if a block is defined in the current context of the template.
+
+.. code-block:: jinja
+
+    {% if 'title' is block name %}
+        <title>{{ block('title') }}<title>
+    {% endif %}
--- a/lib/Twig/Template.php
+++ b/lib/Twig/Template.php
@@ -284,6 +284,8 @@ abstract class Twig_Template implements Twig_TemplateInterface
      *
      * @return bool true if the block exists, false otherwise
      *
+     * @see blockExists
+     *
      * @internal
      */
     public function hasBlock($name)
@@ -625,5 +627,38 @@ abstract class Twig_Template implements Twig_TemplateInterface
         }
 
         return $ret;
+    }
+
+    /**
+     * Returns whether a block exists or not in the current context of the template.
+     *
+     * This method checks blocks defined in the current template
+     * or defined in "used" traits or defined in parent templates.
+     *
+     * @param string $name    The block name
+     * @param array  $context The context
+     * @param array  $blocks  The current set of blocks
+     *
+     * @return bool true if the block exists, false otherwise
+     *
+     * @see hasBlock
+     *
+     * @internal
+     */
+    protected function blockExists($name, array $context, array $blocks = array())
+    {
+        if (isset($blocks[$name])) {
+            return $blocks[$name][0] instanceof self;
+        }
+
+        if (isset($this->blocks[$name])) {
+            return true;
+        }
+
+        if (false !== $parent = $this->getParent($context)) {
+            return $parent->blockExists($name, $context);
+        }
+
+        return false;
     }
 }
--- a/doc/tests/index.rst
+++ b/doc/tests/index.rst
@@ -4,6 +4,7 @@ Tests
 .. toctree::
     :maxdepth: 1
 
+    block_name
     constant
     defined
     divisibleby
--- a/lib/Twig/Extension/Core.php
+++ b/lib/Twig/Extension/Core.php
@@ -222,6 +222,7 @@ class Twig_Extension_Core extends Twig_Extension
             new Twig_SimpleTest('constant', null, array('node_class' => 'Twig_Node_Expression_Test_Constant')),
             new Twig_SimpleTest('empty', 'twig_test_empty'),
             new Twig_SimpleTest('iterable', 'twig_test_iterable'),
+            new Twig_SimpleTest('block name', null, array('node_class' => 'Twig_Node_Expression_Test_BlockName')),
         );
     }
 
--- a/lib/Twig/Node/Expression/Test/BlockName.php
+++ b/lib/Twig/Node/Expression/Test/BlockName.php
@@ -1,0 +1,33 @@
+<?php
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) 2016 Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+/**
+ * Checks whether a block is defined.
+ *
+ * <pre>
+ *  {% if 'title' is block name %}
+ *    <title>{{ block('title') }}</title>
+ *  {% endif %}
+ * </pre>
+ *
+ * @author Martin Hasoň <martin.hason@gmail.com>
+ */
+class Twig_Node_Expression_Test_BlockName extends Twig_Node_Expression_Test
+{
+    public function compile(Twig_Compiler $compiler)
+    {
+        $compiler
+            ->raw('$this->blockExists(')
+            ->subcompile($this->getNode('node'))
+            ->raw(', $context, $blocks)')
+        ;
+    }
+}
--- a/test/Twig/Tests/Fixtures/tests/block_name.test
+++ b/test/Twig/Tests/Fixtures/tests/block_name.test
@@ -1,0 +1,38 @@
+--TEST--
+"block name" test
+--TEMPLATE--
+{% extends 'parent' %}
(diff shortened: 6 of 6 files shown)


lib/Twig/Template.php
```diff
@@ -126,6 +126,34 @@ public function displayParentBlock($name, array $context, array $blocks = array(
...
+     * @param string $name      The block name to test
+     * @param array  $context   The context
+     * @param array  $blocks    The current set of blocks
+     * @param bool   $useBlocks Whether to use the current set of blocks
+     *
+     * @return bool
+     */
+    public function existsBlock($name, array $context, array $blocks = array(), $useBlocks = true)
+    {
+        $name = (string) $name;
```
stof: should we really cast here ? IMO, we should expect a string here all the time. The casting could be done in the compiled code of the ``Twig_Node_Expression_Test_Block``. It would avoid casting again when calling the method on the parent

lib/Twig/Template.php
```diff
@@ -126,6 +126,34 @@ public function displayParentBlock($name, array $context, array $blocks = array(
...
+
+        if ($useBlocks && isset($blocks[$name])) {
+            return $blocks[$name][0] instanceof self;
+        } elseif (isset($this->blocks[$name])) {
+            return $this->blocks[$name][0] instanceof self;
+        } elseif (false !== $parent = $this->getParent($context)) {
+            return $parent->existsBlock($name, $context, array_merge($this->blocks, $blocks), false);
```
stof: merging blocks is useless: you won't use the blocks anyway. So you just make extra work

lib/Twig/Template.php
```diff
@@ -126,6 +126,34 @@ public function displayParentBlock($name, array $context, array $blocks = array(
...
+     *
+     * @return bool
+     */
+    public function existsBlock($name, array $context, array $blocks = array(), $useBlocks = true)
+    {
+        $name = (string) $name;
+
+        if ($useBlocks && isset($blocks[$name])) {
+            return $blocks[$name][0] instanceof self;
+        } elseif (isset($this->blocks[$name])) {
```
stof: as the previous if returns, this should simply be a ``if``, not a ``elseif``

lib/Twig/Template.php
```diff
@@ -126,6 +126,34 @@ public function displayParentBlock($name, array $context, array $blocks = array(
...
+            return $this->blocks[$name][0] instanceof self;
```
stof: this can simply be ``return true``. We control exactly what goes in ``$this->blocks``, as this is only coming from Twig code (we need the validation for the ``$blocks`` array as this is the input).

lib/Twig/Template.php
```diff
@@ -126,6 +126,34 @@ public function displayParentBlock($name, array $context, array $blocks = array(
...
+     * directly.
+     *
+     * @param string $name      The block name to test
+     * @param array  $context   The context
+     * @param array  $blocks    The current set of blocks
+     * @param bool   $useBlocks Whether to use the current set of blocks
+     *
+     * @return bool
+     */
+    public function existsBlock($name, array $context, array $blocks = array(), $useBlocks = true)
```
stof: why passing the context here ? You never use it

lib/Twig/Template.php
```diff
@@ -126,6 +126,34 @@ public function displayParentBlock($name, array $context, array $blocks = array(
     }
 
     /**
+     * Checks if a block exists.
```
stof: ``Checks whether ...``

lib/Twig/Template.php
stof: and I suggest putting the method before ``displayParentBlock``. It makes sense to keep all the block rendering functions grouped together.

stof: sorry, the context is used to get the parent

lib/Twig/Template.php
```diff
@@ -126,6 +126,34 @@ public function displayParentBlock($name, array $context, array $blocks = array(
...
+     *
+     * This method is for internal use only and should never be called
+     * directly.
```
stof: the long description should also describe the difference between this method and ``hasBlock``

doc/tests/block.rst
```diff
@@ -0,0 +1,13 @@
+``block``
+========
```
fabpot: missing `=`

lib/Twig/Template.php
```diff
@@ -244,6 +244,38 @@ public function hasBlock($name)
...
+     * This method does return blocks defined in the current template
+     * or defined in "used" traits or defined in parent templates.
+     *
+     * @param string $name    The block name
+     * @param array  $context The context
+     * @param array  $blocks  The current set of blocks
+     *
+     * @return bool true if the block exists, false otherwise
+     */
+    public function existsBlock($name, array $context, array $blocks = array())
```
fabpot: What about making it protected?

fabpot: Can you also add an `@internal` tag like I've just done in #1835 for other internal methods?

lib/Twig/Template.php
```diff
@@ -593,4 +595,37 @@ protected function getAttribute($object, $item, array $arguments = array(), $typ
 
         return $ret;
     }
+
+    /**
+     * Returns whether a block exists or not in the current context of the template.
+     *
+     * This method does return blocks defined in the current template
```
stof: it does not return blocks

lib/Twig/Node/Expression/Test/Block.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) 2011 Fabien Potencier
```
stof: should be ``2015``. this class was not there in 2011.

lib/Twig/Node/Expression/Test/Block.php
```diff
@@ -0,0 +1,33 @@
...
+ * </pre>
+ *
+ * @author Martin Hasoň <martin.hason@gmail.com>
+ */
+class Twig_Node_Expression_Test_Block extends Twig_Node_Expression_Test
+{
+    public function compile(Twig_Compiler $compiler)
+    {
+        $compiler
+            ->raw('$this->existsBlock((string) ')
```
fabpot: string cast should be removed

xabbuh:
Imho `blockExists()` reads more natural than `existsBlock()`.

doc/tests/block.rst
```diff
@@ -0,0 +1,13 @@
...
+=========
+
+.. versionadded:: 1.23
+    The ``block`` test was added in Twig 1.23.
+
+``block`` checks if a block is defined in the current context of the template.
+
+.. code-block:: jinja
+
+    {% if 'title' is block %}
```
fabpot: Now that I'm reading it in a template, using a test looks wrong to me: `'title'` is a string, not a block.

I was thinking about using a function instead like `has_block('title')`:

```jinja
{% if has_block('title') %}
    <title>{{ block('title') }}<title>
{% endif %}
```

What do you think?

hason: I prefer a test instead of a function. In #1821 I suggested ``{% if 'title' is block name %}``, for example.

hason: Additionally the ``has_block`` should check the existence of a block in the current template as same as [Twig_Template::hasBlock](https://github.com/twigphp/Twig/blob/1.x/lib/Twig/Template.php#L253) and not the existence of a block in the context.

The trouble with ``is block`` is similar to ``is constant``. More correct would be:
```twig
{% if post.status is same as(constant('Post::PUBLISHED')) %}
```

stof: @hason ``Twig_Template::hasBlock`` does not take inheritance into account, which precisely defeats the use case for this test: if it does not take inheritance into account, there is no reason to do this check at runtime; the developer can know it when writing the template and one of the branches of your ``if`` will always be dead.

hason: @stof Sure. I just don't like the confusing name of the function ``has_block``, that gives the impression: "if the template has a block". That's why I prefer rather the test ``is block``.

stof: I prefer ``is block name`` rather than ``is block``

hason:
@fabpot What do you think?

doc/tests/block_name.rst
```diff
@@ -0,0 +1,13 @@
...
+==============
+
+.. versionadded:: 1.28
+    The ``block name`` test was added in Twig 1.28.
+
+``block name`` checks if a block is defined in the current context of the template.
+
+.. code-block:: jinja
+
+    {% if 'title' is block name %}
```
fabpot: I still find this very weird to read and not easy to understand. I don't have any better proposal for now.

fabpot:
Closes in favor of #2239

PR #1833, opened 2015-09-16
fixed implementation of Null cache

### LINKED ISSUE #1832: Twig_Null_Cache is oddly implemented
https://github.com/twigphp/Twig/blob/1.x/lib/Twig/Cache/Null.php uses write() to load the template and load() then just does nothing.

This is kinda problematic if you think about implementing a generic chain of cache backends. There you ideally want to iterate over all of them on write() and just dump the content.

What about adding a hashmap in write() and using that to load the compiled filein load()

fabpot:
fixes #1832

### DIFF
--- a/lib/Twig/Cache/Null.php
+++ b/lib/Twig/Cache/Null.php
@@ -16,12 +16,14 @@
  */
 class Twig_Cache_Null implements Twig_CacheInterface
 {
+    private $buffer = array();
+
     /**
      * {@inheritdoc}
      */
     public function generateKey($name, $className)
     {
-        return '';
+        return $className;
     }
 
     /**
@@ -37,7 +39,7 @@ class Twig_Cache_Null implements Twig_CacheInterface
      */
     public function write($key, $content)
     {
-        eval('?>'.$content);
+        $this->buffer[$key] = $content;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


lib/Twig/Cache/Null.php
```diff
@@ -45,6 +47,9 @@ public function write($key, $content)
      */
     public function load($key)
     {
+        eval('?>'.$this->buffer[$key]);
```
dawehner: Does twig provide any protection against loading it twice already? In case yes, this looks basically how I expect it to look like. Thank you a lot for the very quick response!

lib/Twig/Cache/Null.php
```diff
@@ -16,12 +16,14 @@
  */
 class Twig_Cache_Null implements Twig_CacheInterface
 {
+    private $buffer = array();
```
dawehner: Do you use now short array syntax, as you require PHP 5.5?

fabpot: Nope, this is 1.x, still supporting older versions of PHP. But anyway, I don't plan to migrate to the new short array syntax, mainly because I'm going to maintain both 1.x and 2.x in parallel, so ease of merge is a must.

lib/Twig/Cache/Null.php
fabpot: Twig does prevent loading the same template twice (same being a template with the same class name). So, I think this code is safe, except if you are doing very weird things I suppose.

stof:
:+1:

LionsAd:
This looks good to me.

PR #1835, opened 2015-09-17
added @internal when appropriate

fabpot:


lib/Twig/Node/ForLoop.php
```diff
@@ -13,6 +13,8 @@
  * Internal node used by the for node.
  *
  * @author Fabien Potencier <fabien@symfony.com>
+ *
+ * @internal
```
stof: should this really be marked as ``@internal`` ? Node visitors may need to be aware of it.

fabpot: That's an implementation detail, but ok, let's remove the tag for now.

PR #1837, opened 2015-09-19
Fix race condition in `Environment::loadTemplate` (master)

### LINKED ISSUE #1836: Race condition in `Environment::loadTemplate`
The `Environment::loadTemplate` currently is subject of a race condition. It can be reproduced with the following steps:

* Set a breakpoint in `Environment::loadTemplate` at the top of the `!class_exists($cls, false)` branch.
* Step through until the `$this->cache->load($key);` call.
* Remove the on-disk cache (`rm -r app/cache/twig')
* Continue the debugger

This will result in the following fatal error:

    Compile Error: Twig_Cache_Filesystem::load(): Failed opening required '[...]/app/cache/dev/twig/4/5/451[...].php' (include_path='[...]')

The problem is that between `$this->cache->has($key)` and `$this->cache->load($key)` anything can happen.

Proposed solution: Remove `has()` and instead try to `load()` the class directly from the cache. If the class still does not exist after that, compile the template and write it to the cache and then `eval()` the generated source directly.

znerol:
Fixes #1836

### DIFF
--- a/lib/Twig/Cache/Filesystem.php
+++ b/lib/Twig/Cache/Filesystem.php
@@ -90,6 +90,6 @@ class Twig_Cache_Filesystem implements Twig_CacheInterface
      */
     public function getTimestamp($key)
     {
-        return @filemtime($key);
+        return (int) @filemtime($key);
     }
 }
--- a/lib/Twig/Cache/Filesystem.php
+++ b/lib/Twig/Cache/Filesystem.php
@@ -46,7 +46,7 @@ class Twig_Cache_Filesystem implements Twig_CacheInterface
      */
     public function load($key)
     {
-        return false !== (@include_once $key);
+        @include_once $key;
     }
 
     /**
--- a/lib/Twig/Cache/Filesystem.php
+++ b/lib/Twig/Cache/Filesystem.php
@@ -44,17 +44,9 @@ class Twig_Cache_Filesystem implements Twig_CacheInterface
     /**
      * {@inheritdoc}
      */
-    public function has($key)
-    {
-        return is_file($key);
-    }
-
-    /**
-     * {@inheritdoc}
-     */
     public function load($key)
     {
-        require_once $key;
+        return false !== (@include_once $key);
     }
 
     /**
--- a/lib/Twig/Cache/Filesystem.php
+++ b/lib/Twig/Cache/Filesystem.php
@@ -90,6 +90,6 @@ class Twig_Cache_Filesystem implements Twig_CacheInterface
      */
     public function getTimestamp($key)
     {
-        return filemtime($key);
+        return @filemtime($key);
     }
 }
--- a/lib/Twig/Cache/Null.php
+++ b/lib/Twig/Cache/Null.php
@@ -29,7 +29,6 @@ class Twig_Cache_Null implements Twig_CacheInterface
      */
     public function write($key, $content)
     {
-        eval('?>'.$content);
     }
 
     /**
@@ -37,7 +36,6 @@ class Twig_Cache_Null implements Twig_CacheInterface
      */
     public function load($key)
     {
-        return false;
     }
 
     /**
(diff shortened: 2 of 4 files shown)


lib/Twig/Cache/Null.php
```diff
@@ -52,7 +44,7 @@ public function load($key)
      */
     public function getTimestamp($key)
     {
-        // never called as has() always returns false
+        // never called as load() always returns false
```
Tobion: load() returns nothing, so this makes no sense

znerol: Right. The comment is obsolete anyway, because with `has()` being removed, we need to check the timestamp before `load()`.

lib/Twig/Cache/Filesystem.php
```diff
@@ -44,17 +44,9 @@ public function generateKey($name, $className)
...
-        return is_file($key);
-    }
-
-    /**
-     * {@inheritdoc}
-     */
     public function load($key)
     {
-        require_once $key;
+        @include_once $key;
```
stof: silencing errors is a bad idea IMO

lib/Twig/Environment.php
```diff
@@ -314,11 +314,15 @@ public function loadTemplate($name, $index = null)
...
-                $this->cache->write($key, $this->compileSource($this->getLoader()->getSource($name), $name));
+            if (!$this->isAutoReload() || $this->isTemplateFresh($name, $this->cache->getTimestamp($key))) {
+                $this->cache->load($key);
             }
 
-            $this->cache->load($key);
+            if (!class_exists($cls, false)) {
+                $content = $this->compileSource($this->getLoader()->getSource($name), $name);
+                $this->cache->write($key, $content);
+                eval('?>'.$content);
```
stof: The file should be loaded rather than using eval, to allow benefiting from opcode caching.

znerol: Regrettably the `dump` and `load` approach is also subject to a race condition.

lib/Twig/Cache/Filesystem.php
znerol: Checking the existence of a file followed by attempting to load it is prone to a race condition.

sstok:
What is realistically the change of this ever happening? If you are clearing the cache you shouldn't be having an active web application at that moment as things are most likely to break.

znerol:
@sstok People are running into this issue with Drupal 8. Thu current solution with Twig 1.x is to override `loadTemplate()`, but that will not be possible anymore in Twig 2.x due to changed visibility of instance variables.

lib/Twig/Cache/Filesystem.php
```diff
@@ -98,6 +90,6 @@ public function write($key, $content)
      */
     public function getTimestamp($key)
     {
-        return filemtime($key);
+        return @filemtime($key);
```
Tobion: in the error case this returns false which is not allowed according to the phpdoc of `getTimestamp`

Tobion: so probably need to cast it to int sa Null cache also returns 0

Tobion: please add a test for this and `load`

lib/Twig/Environment.php
Tobion: I agree with @znerol that this is the correct approach. And it should not have a performance impact as this eval is only happening if the cache doesn't exist. And this this case (a new generated class), the opcode caching wouldn't happen anyway.

Tobion:
This PR should be based on 1.x branch and needs rebase. 1.x is not merged in master yet and has some conflicting changes.

znerol:
@Tobion: This will remove the `CacheInterface::has()` method. Is that okay BC-wise?

znerol:
1.x is in #1840

lib/Twig/Cache/Filesystem.php
znerol: Noted that the filesystem cache is only covered by unrelated tests, and that the auto reload feature (timestamp) is not covered at all. Opened issues #1841 and #1842

fabpot:
closing in favor of #1840

PR #1838, opened 2015-09-20
Fix instance variable declaration in Filesystem cache

znerol:


Tobion:
`$this->options` is not defined. So it should actually be renamed.

znerol:
Right.

znerol:
Opened a new one #1839, closing this.

PR #1840, opened 2015-09-20
Fix race condition in `Environment::loadTemplate` (1.x)

### LINKED ISSUE #1836: Race condition in `Environment::loadTemplate`

znerol:
Fixes #1836 (1.x branch)

### DIFF
--- a/lib/Twig/Cache/Filesystem.php
+++ b/lib/Twig/Cache/Filesystem.php
@@ -44,17 +44,9 @@ class Twig_Cache_Filesystem implements Twig_CacheInterface
     /**
      * {@inheritdoc}
      */
-    public function has($key)
-    {
-        return is_file($key);
-    }
-
-    /**
-     * {@inheritdoc}
-     */
     public function load($key)
     {
-        require_once $key;
+        @include_once $key;
     }
 
     /**
@@ -98,6 +90,6 @@ class Twig_Cache_Filesystem implements Twig_CacheInterface
      */
     public function getTimestamp($key)
     {
-        return filemtime($key);
+        return (int) @filemtime($key);
     }
 }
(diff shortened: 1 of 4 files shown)


LionsAd:
This looks pretty great to me.

Probably has() would need to be retained as deprecated function though - due to BC.

fabpot:
`has()` has just been added, so no need to keep BC here. I will review this PR tomorrow or later this week, during DrupalCon :) See you there.

stof:
@fabpot BC is needed. The Twig_CacheInterface has already been released.

fabpot:
No need to keep BC when the interface has been released a week ago. Let's not add overhead for no good reasons.

fabpot:
Thanks @znerol for working on this issue.

drzraf:
After looking at
https://github.com/twigphp/Twig/pull/2043
I wonder:
* isn't this commit reintroducing the race-condition that the file_exists() was expected to fix?
* why is @ needed? If PHP die on buggy generated code? it would be best to let the error flow out as much as PHP permits

PR #1843, opened 2015-09-21
Non-existing files are not cached by statcache

Tobion:
Reverts #1583

Based on the docs this cannot happen because the stat cache is only active for existing files.

> You should also note that PHP doesn't cache information about non-existent files. So, if you call file_exists() on a file that doesn't exist, it will return FALSE until you create the file. If you create the file, it will return TRUE even if you then delete the file. However unlink() clears the cache automatically.

http://php.net/manual/en/function.clearstatcache.php

Also the arugments you pass to clearstatcache(false, $filename) make no sense according the documentation and are not effective.

> filename
Clear the realpath and the stat cache for a specific filename only; only used if clear_realpath_cache is TRUE.

Also doctrine cachee does't use this approach either: https://github.com/doctrine/cache/blob/master/lib/Doctrine/Common/Cache/FileCache.php#L185 And nobody reported problems with this as far as I can see.

### DIFF
--- a/lib/Twig/Cache/Filesystem.php
+++ b/lib/Twig/Cache/Filesystem.php
@@ -64,11 +64,8 @@ class Twig_Cache_Filesystem implements Twig_CacheInterface
     {
         $dir = dirname($key);
         if (!is_dir($dir)) {
-            if (false === @mkdir($dir, 0777, true)) {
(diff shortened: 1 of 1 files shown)


fabpot:
Thank you @Tobion.

PR #1844, opened 2015-09-21
changed template cache names to take into account the Twig C extension

fabpot:


stof:
:+1:

CHANGELOG
```diff
@@ -1,5 +1,6 @@
 * 1.22.2 (2015-XX-XX)
 
+ * changed template cache names to take into account the Twig C extension
```
Tobion: should be for 1.22.3 now

Tobion:
:+1:  apart from the comment. I'd like to propose another PR on top of this when merged.

stof:
@fabpot this broke the tests when the C extension is enabled

ninsuo:
I don't understand why revoking the cache if the C extension is enabled.
- there are no difference in the generated template?
- shouldn't we take into account the `disable_c_ext`?

Tobion:
The tests are going to be fixed by #1852

Issue #1848, opened 2015-09-24
Twig compilation cache - attempting to stat non-existent file?


domhnallw:
Hi folks:
I've discovered I'm having inconsistent template loading issues with Twig that seem to have happened after version 1.21.2.

We're currently using the following versions, pulled from composer this morning:
 * Twig 1.22.2
 * Twig-extensions 1.3

Server is Ubuntu 14.04 64-bit, running PHP 5.5.29 on Apache 2.4.16

Basically, the problem I'm having is that some templates compile and render just fine, while some just don't. One of my colleagues here just tested his prior to updating from 1.21.2 - worked fine before, and now it doesn't. The messages we get look like this:

```
Got error 'PHP message: PHP Fatal error:  Uncaught exception 'ErrorException' with message 'filemtime(): stat failed for /var/www/html/myapp/compilation_cache/c/0/c0c28c5396c42cf220b237b791a9a8305de55491371ddf034ed61fd60cf4c158.php' in /var/www/html/myapp/vendor/twig/twig/lib/Twig/Cache/Filesystem.php:90
```

In this case, the compiled template file mentioned in the error message does not exist (though the source template definitely does, and loads and renders just fine with Twig 1.21.2). At the point that this error occurs, two other templates have been loaded and rendered without issue.

I can delete the cache folder (`/myapp/compilation_cache`) and Twig can recreate it and the first two templates get compiled, saved into this folder, and ultimately render on-screen just fine, so the file/folder permissions seem to be fine. It still stumbles on this third template (and several others), which are very very plain - no `import`s, no `include`s, no `macro`s, no `use`s, or anything like that. Yet. :)

I don't know if I'm doing something stupid or not but since this problem only occurs in very recent version(s), so I thought I'd log it as an issue to see if it's happening to anyone else.

fabpot:
ping @Tobion Isn't it related to the removal of `clearcache`?

domhnallw:
I can add that before updating Twig, that there was a compiled copy of the now "broken" template in the compilation cache folder which rendered fine, but I did spot from the error logs that the derived hash that Twig was using to try to find the previously-compiled file had changed (because the file it was looking for now didn't exist - and yes, this is before I deleted the cache folder :) ), which I presume is related to "changed template cache names to take into account enabled extensions" in the changelog for 1.22.0?

Tobion:
@fabpot I don't think it's related to #1843. Otherwise the error would be different.

[...]

PR #1850, opened 2015-09-24
use is_file instead of error suppression operator

fsw:
I am not 100% sure if changing this was intended in #1840 and related to race condition or accidental.

We were using custom error handler and ignored this part of php doc:

> If you have set a custom error handler function with set_error_handler() then it will still get called, but this custom error handler can (and should) call error_reporting() which will return 0 when the call that triggered the error was preceded by an @.

So we hit issues after this changed.
Of course we fixed our error handler but afaik "@" should be avoided when it does not have a good reason.

So, unless there is a reason, please pull this in coz i want to be famous.

### DIFF
--- a/lib/Twig/Cache/Filesystem.php
+++ b/lib/Twig/Cache/Filesystem.php
@@ -87,6 +87,6 @@ class Twig_Cache_Filesystem implements Twig_CacheInterface
      */
     public function getTimestamp($key)
     {
-        return (int) @filemtime($key);
+        return is_file($key) ? filemtime($key) : 0;
     }
 }


Tobion:
You must keep the error supression and int cast because you cannot be sure that the file still exists when you call `filemtime` (the probability that this occurs is just lower).

But in general I think adding this check is the best thing we can do.
In dev env with auto_reload: true symfony would otherwise log a silenced error for each un-cached template as I explained in https://github.com/twigphp/Twig/issues/1848#issuecomment-142912912

The only problem I see is that with auto_reload: false, this call is skipped. So if somebody would use a wrong error handler he, would only realize this in production mode as he would run into the same problem with `load() { @include_once $key; }`.

stof:
the issue with doing 2 checks is that it is sensitive to race conditions

fsw:
Oh, thanks guys, now i understand that this was the key part of fixing the race condition issue.
I think adding this check just for people with wrong error handler would not be good as PHP docs clearly explain this issue. It was our bad.
I will just let myself paste error message we got here so maybe someone with similar issue can find this:
> Warning: filemtime(): stat failed for "<CachedTemplatePath>" at lib/Twig/Cache/Filesystem.php:90
cheers.

Tobion:
@stof The implemenentation proposed here is not sensitive the race conditions.

stof:
@Tobion it is: if the file got deleted between the is_file and filemtime calls, the error would not be silenced anymore (and it would return the wrong type of value).

Tobion:
See what I read as first response.

> You must keep the error supression and int cast because you cannot be sure that the file still exists when you call filemtime (the probability that this occurs is just lower).

PR #1852, opened 2015-09-24
fix template class name generation to prevent possible collisions

Tobion:
- fix template class name generation to prevent possible collisions
- save one inode per file
- fix tests as they are broken with the cext since #1844

### DIFF
--- a/lib/Twig/Cache/Filesystem.php
+++ b/lib/Twig/Cache/Filesystem.php
@@ -38,7 +38,7 @@ class Twig_Cache_Filesystem implements Twig_CacheInterface
     {
         $hash = hash('sha256', $className);
 
-        return $this->directory.'/'.$hash[0].'/'.$hash[1].'/'.$hash.'.php';
+        return $this->directory.'/'.$hash[0].$hash[1].'/'.$hash.'.php';
     }
 
     /**
--- a/lib/Twig/Environment.php
+++ b/lib/Twig/Environment.php
@@ -297,18 +297,16 @@ class Twig_Environment
      *  * The currently enabled extensions;
      *  * Whether the Twig C extension is available or not.
      *
-     * @param string $name  The name for which to calculate the template class name
-     * @param int    $index The index if it is an embedded template
+     * @param string   $name  The name for which to calculate the template class name
+     * @param int|null $index The index if it is an embedded template
      *
      * @return string The template class name
      */
     public function getTemplateClass($name, $index = null)
     {
-        $key = $this->getLoader()->getCacheKey($name).'__'.implode('__', array_keys($this->extensions));
-
-        if (function_exists('twig_template_get_attributes')) {
-            $key .= '__cext';
-        }
+        $key = $this->getLoader()->getCacheKey($name);
+        $key .= json_encode(array_keys($this->extensions));
+        $key .= function_exists('twig_template_get_attributes');
 
         return $this->templateClassPrefix.hash('sha256', $key).(null === $index ? '' : '_'.$index);
     }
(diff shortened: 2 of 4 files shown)


lib/Twig/Cache/Filesystem.php
```diff
@@ -36,9 +36,8 @@ public function __construct($directory, $options = 0)
      */
     public function generateKey($name, $className)
     {
-        $hash = hash('sha256', $className);
-
-        return $this->directory.'/'.$hash[0].'/'.$hash[1].'/'.$hash.'.php';
+        // We assume the class name is suffixed by a hash (see Twig_Environment::getTemplateClass) that we can use to balance the directory tree.
+        return $this->directory.'/'.substr($className, -2).'/'.$className.'.php';
```
Tobion: using one directory instead of two saves 1 node per file. the file distribution stays the same. even though the math in #1684 is wrong, I agree that this diversity is enough.

lib/Twig/Environment.php
```diff
@@ -297,20 +297,20 @@ public function getCacheFilename($name)
...
+     * @param int|null $index The index if it is an embedded template
      *
      * @return string The template class name
      */
     public function getTemplateClass($name, $index = null)
     {
-        $key = $this->getLoader()->getCacheKey($name).'__'.implode('__', array_keys($this->extensions));
-
-        if (function_exists('twig_template_get_attributes')) {
-            $key .= '__cext';
```
Tobion: adding it optionally can create collision, e.g. an added twig extension whose name is "cext" would create the same key.

lib/Twig/Environment.php
```diff
@@ -297,20 +297,20 @@ public function getCacheFilename($name)
...
-        }
+        $key = $this->getLoader()->getCacheKey($name);
+        $key .= '_'.count($this->extensions);
```
Tobion: we need the count to prevent collisions, e.g. 1 extension "foo_bar" vs two extensions "foo" + "bar"

jvasseur: You could still get conflicts with something like "foo_bar" + "baz" vs "foo" + "bar_baz".

lib/Twig/Cache/Filesystem.php
fabpot: It was done with a hash because of your comment here: https://github.com/twigphp/Twig/pull/1822#discussion_r39337173

:-1: for this change. The `getTemplateClass()` is public and can be changed in any way.

Tobion: I know. I'm just trying to make it as fast as possible by default and avoid hashing the hash for every template on each request. Maybe we should remove the directory balancing at all by default. Todays filesystems do not have a limitation on number of files per directory. And also most users, won't have thousands of templates anyway where it could matter. And if people really have this need they can change the filesystem cache themselves. No need to hurt everybody by default.

fabpot: This was added because some of our customers had this problem.

Tobion: Yeah but IMO performance-relevant things that only affect a small subset of people should be opt-in and not opt-out.

fabpot: No, Twig should work for everyone and it is the case right now.

lib/Twig/Environment.php
sstok: What about an `&` sign? this is illegal for an extension-name right :)

stof: no it is not. the extension name can be any string (even unicode names are valid)

Tobion: fixed

lib/Twig/Cache/Filesystem.php
Tobion: reverted even though I don't fully agree. It can never work for everyone as it depends on the filesystem and the number of templates you have. So whatever workaround you add it can always be the wrong one depending on who is using it.

fabpot:
Thank you @Tobion.

PR #1857, opened 2015-09-29
Remove deprecated return types from the PHPDocs.

SpacePossum:
There are deprecation warnings added to the environment (for example https://github.com/twigphp/Twig/blob/1.x/lib/Twig/Environment.php#L1287).

This PR aims to remove the deprecated return types from the PHPDocs as well. It will be useful for developers, but for for IDE's/SCA as well.


SpacePossum:
I don't get the failing tests, are the PHPDocs added when the templates are compiled?

stof:
@SpacePossum tests are broken in the master branch already

fabpot:
Thank you @SpacePossum.

PR #1860, opened 2015-09-30
Document join combined with range operator

Rvanlaak:
... because of the syntax difference as shown in #1859

### DIFF
--- a/doc/filters/join.rst
+++ b/doc/filters/join.rst
@@ -17,6 +17,18 @@ define it with the optional first parameter:
     {{ [1, 2, 3]|join('|') }}
     {# outputs 1|2|3 #}
 
+Combining the range operator with the join filter allows to output that same
+range of elements without having to define all of them:
+
+.. code-block:: jinja
+
+    {{ (1..3)|join('|') }}
+    {# outputs 1|2|3 #}
+    
+    {% set variable = 6 %}
+    {{ (1..variable)|join('|') }}
+    {# outputs 1|2|3|4|5|6 #}
+
 Arguments
 ---------
 


stof:
The syntax is not different at all. ``[1, 2, 3]`` is the syntax to define an array inline. ``1..3`` gives you an array. the syntax to apply the filter is always the same.

doc/filters/join.rst
```diff
@@ -17,6 +17,18 @@ define it with the optional first parameter:
...
+range of elements without having to define all of them:
+
+.. code-block:: jinja
+
+    {{ (1..3)|join('|') }}
+    {# outputs 1|2|3 #}
+    
+    {% set variable = 6 %}
+    {{ (1..variable)|join('|') }}
+    {# outputs 1|2|3|4|5|6 #}
```
fabpot: I don't understand what this is about. The above example `{{ [1, 2, 3]|join }}` could be anything else like `{{ [22, 42, 54]|join }}` and then the added example would not make sense anymore, right? If the fact the 1, 2, 3 is a sequence is disturbing, let's change it to some more random numbers. The docs try to document one filter/function at a time and only mention other ones when there is a relation, and that's not the case here.

Rvanlaak: This PR tries to explain that `|` has higher precedence as `..`, so the docblock should make that clear. Otherwise this PR could be closed :+1:

fabpot: Got it now, so this should be added where we document `..` (templates.rst). What about #1869 instead?

Rvanlaak:
close this one in favor of #1869

doc/filters/join.rst
Rvanlaak: :+1:

PR #1863, opened 2015-10-06
Checking sandbox for ArrayAccess

### LINKED ISSUE #106: Sandbox array bug & fix
# Situation: 

An unexpected behavior regarding arrays, in Twig_Template::getAttribute() if a referenced object is an array then security checking is skipped. It is allright, but in addition to checking with is_array() Twig also checks for objects that implement ArrayAccess and if so then is considered an array and security checking is again skipped. 
# Issue:

It doesn't seem a good behavior, because there are a lot of objects that implement ArrayAccess and a security policy must be enforced on them. For example I found about this behavior while feeding the template with some active record objects and the base AR class implements ArrayAccess to allow access for model attributes using array like syntax. But still the security policy needs to be enforced especially since AR objects are tightly linked to the database and I don't want users to have any ideas... 
# Resolution:

I think the best way is to differentiate between native arrays (check with is_array()) and objects that implement ArrayAccess. If it is later, then security policy must be enforced.

CarsonF:
Fixes #106

Allows this to be prevented in sandbox:

``` twig
{{ app.secret_key }} {# app is \Pimple #}
```

### DIFF
--- a/ext/twig/twig.c
+++ b/ext/twig/twig.c
@@ -710,6 +710,7 @@ PHP_FUNCTION(twig_template_get_attributes)
 	char *class_name = NULL;
 	zval *tmp_class;
 	char *type_name;
+	zval *zmethod;
 
 	if (zend_parse_parameters(ZEND_NUM_ARGS() TSRMLS_CC, "ozz|asbb", &template, &object, &zitem, &arguments, &type, &type_len, &isDefinedTest, &ignoreStrictCheck) == FAILURE) {
 		return;
@@ -739,6 +740,10 @@ PHP_FUNCTION(twig_template_get_attributes)
 				return true;
 			}
 
+			if ($object instanceof ArrayAccess && $this->env->hasExtension('sandbox')) {
+				$this->env->getExtension('sandbox')->checkMethodAllowed($object, 'offsetGet');
+			}
+
 			return $object[$arrayItem];
 		}
 */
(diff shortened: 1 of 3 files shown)


stof:
@jpauli can you review the changes in the C extension ?

CarsonF:
Thinking about it more I'm realizing that this doesn't allow for distinction on which array item. i.e., All `Pimple` services/parameters can be allowed or none.

I'm thinking this:
```php
->checkMethodAllowed($object, 'offsetGet');
```
should be replaced with this:
```php
->checkPropertyAllowed($object, $arrayItem);
```

@stof thoughts?

stof:
@fabpot what do you think about this ?

test/Twig/Tests/EnvironmentTest.php
```diff
@@ -180,13 +180,12 @@ public function testExtensionsAreNotInitializedWhenRenderingACompiledTemplate()
 
     public function testAutoReloadCacheMiss()
     {
-        $template_name = __FUNCTION__;
-        $template_content = __FUNCTION__;
+        $templateName = __FUNCTION__;
+        $templateContent = __FUNCTION__;
```
OskarStark: renaming those variables should be done in a separate PR, it just increases the diff ... :(

OskarStark: ah sorry, maybe those lines come from the commit of @fabpot , which is in this PR?

CarsonF: I didn't rename them, @fabpot did. I'll rebase

OskarStark: :+1:  thank you @CarsonF

CarsonF:
Silex is EoL and container shouldn't be used in Twig anyways.

PR #1864, opened 2015-10-07
[Performance] Added a cache for extension freshness test

superdav42:
Freshness of each extension was been checked everytime a template was loaded.
Projects with many templates and extension benefit noticeably with this cache.
Only environments with debug or autoload set are affected.

### DIFF
--- a/lib/Twig/Environment.php
+++ b/lib/Twig/Environment.php
@@ -48,6 +48,7 @@ class Twig_Environment
     private $originalCache;
     private $bcWriteCacheFile = false;
     private $bcGetCacheFilename = false;
+    private $lastModifiedExtension = 0;
 
     /**
      * Constructor.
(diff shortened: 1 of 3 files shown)


lib/Twig/Environment.php
```diff
@@ -454,14 +455,19 @@ public function createTemplate($template)
...
     {
-        foreach ($this->extensions as $extension) {
-            $r = new ReflectionObject($extension);
-            if (filemtime($r->getFileName()) > $time) {
-                return false;
+        if(null === $this->extensionsFresh) {
+            foreach ($this->extensions as $extension) {
+                $r = new ReflectionObject($extension);
+                if (filemtime($r->getFileName()) > $time) {
+                    $this->extensionsFresh = false;
```
stof: your cache is wrong. The freshness depends on the time received as argument. Each template will have a different time when checking for freshness (as it is the time at which it was compiled the last time)

lib/Twig/Environment.php
```diff
@@ -36,6 +36,7 @@ class Twig_Environment
     protected $globals;
     protected $runtimeInitialized = false;
     protected $extensionInitialized = false;
+    protected $extensionsFresh;
```
stof: new properties should be private (they are all becoming private in 2.x)

superdav42:
@fabpot Updated cache to store time of most recently modified extension.

fabpot:
Thanks @superdav42. I've created #1870 which gets one step further and do everything when adding an extension. That should be even faster.

Issue #1854, opened 2015-09-28
Twig v1.22.2 ContextErrorException: Warning: filemtime(): stat failed for 


tuanquynh0508:
Hi !

I am using Symfony 2.3 with Twig 1.*, everything not working after I run "composer update". System error: 

ContextErrorException: Warning: filemtime(): stat failed for 
/home/nntuan/projects/dashboard/app/cache/dev_tuan/twig/7/1/71baaa9514a151dcfa15e2d0da9c1fa2c5ebb9718d44d3f294ef0e433986bbea.php in /home/nntuan/projects/dashboard/vendor/twig/twig/lib/Twig/Cache/Filesystem.php line 90

I not found file "71baaa9514a151dcfa15e2d0da9c1fa2c5ebb9718d44d3f294ef0e433986bbea.php" in folder: /home/nntuan/projects/dashboard/app/cache/dev_tuan/twig/7/1/

Help me please!.

Tobion:
Duplicate of #1848. Please use the search next time.

tuanquynh0508:
Thanks you!.

Tobion:
Please close this ticket then.

PR #1866, opened 2015-10-09
Check if cached file exists

torzech:
Without this modification:

```
Array
(
    [code] => 2
    [message] => filemtime(): stat failed for /Users/torzech/code/[---cut---]/app/cache/6/9/69b8b36648330d23ccfb6e8b7542e0dbdd518a910ab986d602639e1b855de522.php
    [file] => /Users/torzech/code/[---cut---]/libs/twig/twig/lib/Twig/Cache/Filesystem.php
    [line] => 91
)
```

### DIFF
--- a/lib/Twig/Cache/Filesystem.php
+++ b/lib/Twig/Cache/Filesystem.php
@@ -88,7 +88,7 @@ class Twig_Cache_Filesystem implements Twig_CacheInterface
     public function getTimestamp($key)
     {
         if (\file_exists($key)) {
-            return (int)@filemtime($key);
+            return (int) @filemtime($key);
         }
 
         return 0;
--- a/lib/Twig/Cache/Filesystem.php
+++ b/lib/Twig/Cache/Filesystem.php
@@ -87,6 +87,10 @@ class Twig_Cache_Filesystem implements Twig_CacheInterface
      */
     public function getTimestamp($key)
     {
-        return (int) @filemtime($key);
+        if (\file_exists($key)) {
+            return (int)@filemtime($key);
+        }
+
+        return 0;
     }
 }


SpacePossum:
ref. https://github.com/twigphp/Twig/issues/1848
https://github.com/twigphp/Twig/issues/1854
(and a bunch more https://github.com/twigphp/Twig/issues?utf8=%E2%9C%93&q=filemtime)

cc @Tobion

stof:
@fabpot I think I know why so many people report this: the Symfony Debug component is still *logging* silenced errors by default in the dev env

Tobion:
@stof yes that is what I said in #1848

> The only "annoying" thing I noticed is, when you remove the cache and then run your app, all those non-existing template caches are reported as "silenced errors" from the symfony profiler whenever the template cache doesn't exist yet.

lib/Twig/Cache/Filesystem.php
```diff
@@ -87,6 +87,10 @@ public function write($key, $content)
      */
     public function getTimestamp($key)
     {
-        return (int) @filemtime($key);
+        if (\file_exists($key)) {
```
GromNaN: The backslash is not necessary before a global function.

fabpot: It should indeed be removed for 5.2 compat.

fabpot:
So, the fix should be done in Symfony instead, right?

Tobion:
@fabpot no symfony cannot do anything about it. IMO this change here is valid (apart from the CS) as it avoids alot of warning spams.

fabpot:
Thank you @torzech.

PR #1870, opened 2015-10-13
Improved performance of template freshness checks

fabpot:
Same as #1864 but with with even less perf impact.

### DIFF
--- a/lib/Twig/Environment.php
+++ b/lib/Twig/Environment.php
@@ -455,15 +455,6 @@ class Twig_Environment
      */
     public function isTemplateFresh($name, $time)
     {
-        if(0 === $this->lastModifiedExtension) {
-            foreach ($this->extensions as $extension) {
-                $r = new ReflectionObject($extension);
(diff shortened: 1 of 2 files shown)


fabpot:
Thank you @superdav42.

lib/Twig/Environment.php
```diff
@@ -767,6 +761,11 @@ public function addExtension(Twig_ExtensionInterface $extension)
             throw new LogicException(sprintf('Unable to register extension "%s" as extensions have already been initialized.', $extension->getName()));
         }
 
+        $r = new ReflectionObject($extension);
+        if (($extensionTime = filemtime($r->getFileName())) > $this->lastModifiedExtension) {
+            $this->lastModifiedExtension = $extensionTime;
+        }
```
stof: this add a check even when auto_reload is false. Are you sure this has less perf impact ? To me, it has more impact.

Resetting the value should not have been an issue anyway: it is forbidden to add more extension after you started rendering templates

GromNaN:
This change broke the backward compatibility. I've submitted a patch to fix it.
https://github.com/twigphp/Twig/pull/1874

PR #1872, opened 2015-10-13
Can we use empty() php function for twig_test_empty() to allow "0", "0.0" as well?

igritsay:
Because the fact that {% if var is empty %} not same as <?php if (empty($var)) ?> is confusing...


stof:
this is a big BC break

fabpot:
This was done on purpose. Personally, I try to avoid using `empty` in PHP as much as possible, the semantic seems very wrong.

PR #1874, opened 2015-10-13
Fix adding mock extension to Twig environment

GromNaN:
When the extension class is not defined in a file but in eval'd code we cannot assume `ReflectionObject::getFileName()` will return a valid file path.
https://bugs.php.net/bug.php?id=63901

This PR adds a test case that produce the following error without the fix:

```
1) Twig_Tests_EnvironmentTest::testAddMockExtension
filemtime(): stat failed for phar:///usr/local/php5-5.5.14-20140628-105310/bin/phpunit/phpunit-mock-objects/Framework/MockObject/Generator.php(335) : eval()'d code

/xxx/Twig/vendor/symfony/phpunit-bridge/DeprecationErrorHandler.php:40
/xxx/Twig/lib/Twig/Environment.php:769
/xxx/Twig/test/Twig/Tests/EnvironmentTest.php:304
```

I encountered this issue with mocked twig extensions in SonataAdminBundle test suite: 
https://github.com/sonata-project/SonataAdminBundle/blob/master/Tests/Controller/HelperControllerTest.php#L254

| Q | A |
| --- | --- |
| Bug fix? | yes |
| New feature? | no |
| BC breaks? | no |
| Deprecations? | no |
| Tests pass? | yes |
| Fixed tickets | #1870 |
| License | MIT |
| Doc PR | n/a |

### DIFF
--- a/lib/Twig/Environment.php
+++ b/lib/Twig/Environment.php
@@ -462,7 +462,7 @@ class Twig_Environment
         if (0 === $this->lastModifiedExtension) {
             foreach ($this->extensions as $extension) {
                 $r = new ReflectionObject($extension);
-                if (($extensionTime = filemtime($r->getFileName())) > $this->lastModifiedExtension) {
+                if (file_exists($r->getFileName()) && ($extensionTime = filemtime($r->getFileName())) > $this->lastModifiedExtension) {
                     $this->lastModifiedExtension = $extensionTime;
                 }
             }
--- a/test/Twig/Tests/EnvironmentTest.php
+++ b/test/Twig/Tests/EnvironmentTest.php
@@ -292,6 +292,22 @@ class Twig_Tests_EnvironmentTest extends PHPUnit_Framework_TestCase
         $this->assertCount(2, $twig->getNodeVisitors());
     }
 
+    public function testAddMockExtension()
+    {
+        $extension = $this->getMock('Twig_ExtensionInterface');
+        $extension->expects($this->once())
+            ->method('getName')
+            ->will($this->returnValue('mock'));
+
+        $loader = new Twig_Loader_Array(array('page' => 'hey'));
+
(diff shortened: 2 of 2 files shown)


OskarStark:
any news on this PR?

fabpot:
I've reverted the culprit commit. Can you rebase as the test also make sense with the old implementation?

GromNaN:
Thanks for the revert https://github.com/twigphp/Twig/commit/50e68e68f09a7e324443e3e9f028c3f05f46763f.
I've rebased and amended my commit.

fabpot:
I don't see the change on Twig_Environment anymore.

GromNaN:
Do you think I should replace this line (https://github.com/twigphp/Twig/commit/50e68e68f09a7e324443e3e9f028c3f05f46763f#diff-5e190f538c5c328c7473a5cfd3b7cd95R465)
```php
if (($extensionTime = filemtime($r->getFileName())) > $this->lastModifiedExtension) {
```
By
```php
if (file_exists($r->getFileName()) && ($extensionTime = filemtime($r->getFileName())) > $this->lastModifiedExtension) {
```

fabpot:
I thought it was the bug you were trying to fix, so yes (this is probably a very edgy case but apprently you hit it previously).

GromNaN:
I hit the bug only by adding a mock of `ExtensionInterface`. That's why the test pass with your commit.
I've moves my fix to the method `isTemplateFresh` as the bug would appear if a mock extension is added to a twig environment to actually render a template.

stof:
the difference is that the previous PR made the bug appear on registration of the extension, while it appears only when loading a template with auto_reload without the optimization the PR was implementing

GromNaN:
@stof I could just call `isTemplateFresh`.

fabpot:
Thank you @GromNaN.

PR #1878, opened 2015-10-14
Allow returning of instance instead of arrays for small/simple custom extensions

DaGhostman:
Allow the methods `\Twig_Environment::getFilters`, `\Twig_Environment::getFunctions`, `\Twig_Environment::getTests`, `\Twig_Environment::getNodeVisitors` and `\Twig_Environment::getTokenParsers` to return the respective object instance instead of surrounding it in an array.

This is more elegant and intuitive in cases where an extensions is exposing only one filter, function and etc.

### DIFF
--- a/lib/Twig/ExtensionInterface.php
+++ b/lib/Twig/ExtensionInterface.php
@@ -28,35 +28,35 @@ interface Twig_ExtensionInterface
     /**
      * Returns the token parser instances to add to the existing list.
      *
-     * @return Twig_TokenParserInterface[]
+     * @return Twig_TokenParserInterface[]|Twig_TokenParserInterface
      */
     public function getTokenParsers();
 
     /**
      * Returns the node visitor instances to add to the existing list.
      *
-     * @return Twig_NodeVisitorInterface[] An array of Twig_NodeVisitorInterface instances
+     * @return Twig_NodeVisitorInterface[]|Twig_NodeVisitorInterface An array of Twig_NodeVisitorInterface instances
      */
     public function getNodeVisitors();
 
     /**
      * Returns a list of filters to add to the existing list.
      *
-     * @return Twig_SimpleFilter[]
+     * @return Twig_SimpleFilter[]|Twig_SimpleFilter
      */
     public function getFilters();
 
     /**
      * Returns a list of tests to add to the existing list.
      *
-     * @return Twig_SimpleTest[]
+     * @return Twig_SimpleTest[]|Twig_SimpleTest
      */
     public function getTests();
 
     /**
      * Returns a list of functions to add to the existing list.
      *
-     * @return Twig_SimpleFunction[]
+     * @return Twig_SimpleFunction[]|Twig_SimpleFunction
      */
     public function getFunctions();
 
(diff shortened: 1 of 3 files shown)


lib/Twig/ExtensionInterface.php
```diff
@@ -28,35 +28,35 @@ public function initRuntime(Twig_Environment $environment);
...
-     * @return Twig_TokenParserInterface[]
+     * @return Twig_TokenParserInterface[]|Twig_TokenParserInterface
      */
     public function getTokenParsers();
 
     /**
      * Returns the node visitor instances to add to the existing list.
      *
-     * @return Twig_NodeVisitorInterface[] An array of Twig_NodeVisitorInterface instances
+     * @return Twig_NodeVisitorInterface[]|Twig_NodeVisitorInterface An array of Twig_NodeVisitorInterface instances
```
stof: changing the signature of an interface is actually a BC break for consumers of the interface.

and I know of some usages of these methods outside of the Twig_Environment, so this cannot be done in 1.x

Tobion:
:-1: We  usually don't want magic and multi-valued return types.

stof:
having ``getFunctions`` return a single ``Twig_SimpleFunction`` object rather than an array does not seem intuitive to me actually. The method is still named with a plural.

I'm also -1 on this.

DaGhostman:
@stof I will to some extent agree with the change on the interface, signature however the functionality is still preserved imo. If you could provide an example of the BC break maybe?

And as for the plural naming, I do believe that it is better for plural named method to return 1 or more instead of non-plural to return more than one instance.

test/Twig/Tests/EnvironmentTest.php
```diff
@@ -312,37 +312,27 @@ class Twig_Tests_EnvironmentTest_Extension extends Twig_Extension
...
-        );
+        return new Twig_Tests_EnvironmentTest_NodeVisitor();
     }
 
     public function getFilters()
     {
-        return array(
-            new Twig_SimpleFilter('foo_filter', 'foo_filter'),
-        );
+        return new Twig_SimpleFilter('foo_filter', 'foo_filter');
```
ninsuo: There are not much differences between this line and the actual behaviour:

```jinja
return [new Twig_SimpleFilter('foo_filter', 'foo_filter')];
```

stof:
@DaGhostman if some code calls ``$extension->getFilters()``, it currently assumes it receives an array, because this is what the interface defines. after your change, it can also receive other things => the code does not work anymore. This is what a BC break is (old code does not work anymore without change).

Regarding plural naming, your methods are not returning 1 or more elements. They are returning an element or a list of elements. there is a huge different between an element or a list containing 1 element. Plural makes sense when returning a list, not when returning an element.

ninsuo:
A simple BC break example:

Imagine my users can create they own twig code, and to help them I want to show all filters they can use somewhere. In my code, I *actually* have:

```php
        $availableFilters = [];
        foreach ($this->getExtensions() as $extension) {
                $availableFilters[] = $filter->getName();
            }
        }            
```

If I upgrade my Twig version after your code have been merged, I'll end up with an exception.

DaGhostman:
@ninsuo core functionality is preserved atm. when `\Twig_Environment::getFilters()` is called it checks if filters have been initialized (`if (!$this->extensionInitialized)` check). And upon call `initExtensions()` is called which does the change.

And I do believe similar use cases are cover in `EnvironmentTest::testAddExtension` test

fabpot:
Like others, I think it is counter-intuitive, error-prone, and not needed.

PR #1886, opened 2015-10-23
deprecated Twig_ExtensionInterface::initRuntime()

fabpot:
I'm working on splitting Twig extensions into 2 different phases: compilation and runtime. The goal is to avoid having to load the runtime environment of an extension when compiling templates. That also opens the way to be able to lazy-load Twig extensions.

While working on the split, I realized that the `initRuntime()` method on `Twig_ExtensionInterface` is not needed anymore. It was added at a time `needs_environment` did not exist and was a way to keep the environment around for custom filters/tests/functions. But nowadays, that's not needed anymore. I did a quick search on Github, and most of the implementation I found just store the environment in a local property, which is not needed anymore. So, I propose to deprecate it in 1.x and remove it in 2.0.

### DIFF
--- a/doc/deprecated.rst
+++ b/doc/deprecated.rst
@@ -26,6 +26,13 @@ Extensions
 * As of Twig 1.x, the ability to remove an extension is deprecated and the
   ``Twig_Environment::removeExtension()`` method will be removed in 2.0.
 
+* As of Twig 1.23, the ``Twig_ExtensionInterface::initRuntime()`` method is
+  deprecated. You have two options to avoid the deprecation notice: if you
+  implement this method to store the environment for your custom filters,
+  functions, or tests, use the ``needs_environment`` option instead; if you
+  have more complex needs, explicitly implement
+  ``Twig_Extension_InitRuntimeInterface`` (not recommended).
+
 PEAR
 ----
 
--- a/lib/Twig/Extension/InitRuntimeInterface.php
+++ b/lib/Twig/Extension/InitRuntimeInterface.php
@@ -1,0 +1,22 @@
+<?php
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+/**
+ * Declares the deprecated Twig_Extension::initRuntime() method.
+ *
+ * Explicitely implement this interface if you really need to implement the
+ * deprecated initRuntime() method in your extensions.
+ *
+ * @author Fabien Potencier <fabien@symfony.com>
+ */
+interface Twig_Extension_InitRuntimeInterface
+{
+}
--- a/test/Twig/Tests/EnvironmentTest.php
+++ b/test/Twig/Tests/EnvironmentTest.php
@@ -308,6 +308,40 @@ class Twig_Tests_EnvironmentTest extends PHPUnit_Framework_TestCase
         $this->assertTrue($twig->isTemplateFresh('page', time()));
     }
 
+    public function testInitRuntimeWithAnExtensionUsingInitRuntimeNoDeprecation()
+    {
+        $twig = new Twig_Environment($this->getMock('Twig_LoaderInterface'));
+        $twig->addExtension(new Twig_Tests_EnvironmentTest_ExtensionWithoutDeprecationInitRuntime());
+
+        $twig->initRuntime();
+    }
+
+    /**
+     * @requires PHP 5.3
+     */
+    public function testInitRuntimeWithAnExtensionUsingInitRuntimeDeprecation()
+    {
+        $twig = new Twig_Environment($this->getMock('Twig_LoaderInterface'));
+        $twig->addExtension(new Twig_Tests_EnvironmentTest_ExtensionWithDeprecationInitRuntime());
+
+        $this->deprecations = array();
+        set_error_handler(array($this, 'handleError'));
+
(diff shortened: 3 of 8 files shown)


fabpot:
And of course, I've just had a look at Symfony, and we are using this method in the Twig form extension. I need to see why and how to fix that.

fabpot:
Ok, it's probably to avoid a circular reference, we can probably find another way.

test/Twig/Tests/EnvironmentTest.php
```diff
@@ -54,11 +54,11 @@ public function testGlobals()
         $globals = $twig->getGlobals();
         $this->assertEquals('bar', $globals['foo']);
 
-        // globals can be modified after runtime init
+        // globals can be modified after a template has been loaded
         $twig = new Twig_Environment($this->getMock('Twig_LoaderInterface'));
         $twig->addGlobal('foo', 'foo');
         $twig->getGlobals();
-        $twig->initRuntime();
+        $twig->loadTemplate('index');
```
fabpot: This is more semantically correct. Calling `initRuntime` is just a side-effect of loading a template.

stof:
@fabpot I suggest finding the solution in Symfony before merging this

fabpot:
To keep BC, I've added a new interface: `Twig_Extension_InitRuntimeInterface` which can be implemented if really needed (see changes for more information).

doc/deprecated.rst
```diff
@@ -26,6 +26,13 @@ Extensions
 * As of Twig 1.x, the ability to remove an extension is deprecated and the
   ``Twig_Environment::removeExtension()`` method will be removed in 2.0.
 
+* As of Twig 1.x, the ``Twig_ExtensionInterface::initRuntime()`` is deprecated.
```
stof: Please say ``As of Twig 1.23``. ``As of 1.x`` does not make sense as ``1.x`` is not a version

lib/Twig/Extension/InitRuntimeInterface.php
```diff
@@ -0,0 +1,30 @@
...
+ * (c) Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+/**
+ * Declares the deprecated Twig_Extension::initRuntime() method.
+ *
+ * Implement explicitely this interface if you really need to implement the
```
xabbuh: Explicitly implement this [...]

doc/deprecated.rst
xabbuh: [...] the `Twig_ExtensionInterface::initRuntime()` method is [...]

fabpot: fixed typo
@stof we are not using that anywhere, that should be done in a separate PR.

lib/Twig/Extension/InitRuntimeInterface.php
fabpot: fixed

lib/Twig/Extension/InitRuntimeInterface.php
```diff
@@ -0,0 +1,30 @@
...
+interface Twig_Extension_InitRuntimeInterface
+{
+    /**
+     * Initializes the runtime environment.
+     *
+     * This is where you can load some file that contains filter functions for instance.
+     *
+     * @param Twig_Environment $environment The current Twig_Environment instance
+     */
+    public function initRuntime(Twig_Environment $environment);
```
stof: This interface can be implemented at the same time than Twig_ExtensionInterface only in PHP 5.3.9+, as it defines the same method (remember our issues with the Validator component in Symfony 2.5).
So this makes it unsuited for the Twig 1.x lowest bound (and for usage in Symfony 2.3): https://3v4l.org/PQSTH

fabpot: So, what about just making this interface a simple marker?

doc/deprecated.rst
stof: Well, using the exact version for the new one already makes it simpler as we know which version it is.

lib/Twig/Extension/InitRuntimeInterface.php
fabpot: done

stof: I suggest adding a test covering the usage of this interface in the Twig testsuite

fabpot: done

test/Twig/Tests/EnvironmentTest.php
```diff
@@ -308,6 +308,33 @@ public function testAddMockExtension()
...
+    public function testInitRuntimeWithAnExtensionUsingInitRuntimeDeprecation()
+    {
+        $twig = new Twig_Environment($this->getMock('Twig_LoaderInterface'));
+        $twig->addExtension(new Twig_Tests_EnvironmentTest_ExtensionWithDeprecationInitRuntime());
+
+        $deprecations = array();
+        set_error_handler(function ($type, $msg) use (&$deprecations) {
+            if (E_USER_DEPRECATED === $type) {
+                $deprecations[] = $msg;
+            }
```
stof: you should call the PHPUnit error handler for any other error, to avoid loosing them

fabpot: Instead, I've added an assertion on the deprecation message.

fabpot:
Not sure why it does not work on 5.2.

PR #1889, opened 2015-10-24
marked some Twig_Environment methods as being internal

fabpot:
This PR marks some `Twig_Environment` methods as being internals:
- `getFunctions()`, `getFilters()`, `getTests()`, `getFunction()`, `getFilter()`, `getTest()`, `getTokenParsers()`, `getTags()`, `getNodeVisitors()`, `getUnaryOperators()`, `getBinaryOperators()` because I don't see how they can be used in any useful way and anyway, `getFunctions()` and `getFilters()` do not return all possible functions or filters. Moreover, I think that all this logic could be refactored and moved into its own class for 2.0 (could be done by deprecating those existing methods in 1.x if we agree to merge this PR).
- `getGlobals()` because it should only be called in a runtime context, not in a compilation one (so limiting its usage to internal use only allows for better control of usage).
- `initGlobals()`, `initExtensions()`, `initExtension()` because they are private in Twig 2.0.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,9 @@
 * 1.24.1 (2016-XX-XX)
 
- * n/a
+ * marked the following methods as being internals on Twig_Environment: 
+   getFunctions(), getFilters(), getTests(), getFunction(), getFilter(), getTest(),
+   getTokenParsers(), getTags(), getNodeVisitors(), getUnaryOperators(), getBinaryOperators(),
+   getFunctions(), getFilters(), getGlobals(), initGlobals(), initExtensions(), and initExtension()
 
 * 1.24.0 (2016-01-25)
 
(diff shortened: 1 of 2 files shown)


stof:
Assetic relies on ``getFunction`` currently. However, it might be possible to avoid it thanks to changes I did when adding support for Twig 2.x. I will look into it after my vacations (which start today for 1 week)

fabpot:
@stof any news on this one?

stof:
I haven't worked on Assetic again. I will need to do it (I forgot about this discussion)

stof:
@fabpot I did the change in Assetic (was even easier than I though actually). See the PR linked above

fabpot:
Thanks @stof

VincentLanglet:
Hi @fabpot, @stof.

I'm working own a custom twig lexer in order to lint twig files and I need to access the unary operator and the binary operator of the TwigEnvironnement, in order to do a similar logic than the Twig Lexer
https://github.com/twigphp/Twig/blob/3.x/src/Lexer.php#L457-L458

Those methods are internal, and are just doing 
```
$this->extensionSet->getFoo();
```
Would it be ok to remove `@internal` on those methods or to provide another way to get those ?

stof:
Please open a dedicated discussion instead of commenting on a PR that was merged almost 7 years ago.

PR #1890, opened 2015-10-24
deprecated Twig_Environment::computeAlternatives()

fabpot:


lib/Twig/Error/Syntax.php
```diff
@@ -17,4 +17,38 @@
...
+        }
+
+        $this->rawMessage .= sprintf('. Did you mean "%s"?', implode('", "', $alternatives));
+        $this->updateRepr();
+    }
+
+    /**
+     * @internal
+     *
+     * To be merged with the addAlternatives() method in 2.0.
```
stof: shouldn't this be ``addMessageSuggestions`` ?

lib/Twig/Error/Syntax.php
```diff
@@ -17,4 +17,38 @@
...
+     * @param string $name  The original name of the item that does not exist
+     * @param array  $items An array of possible items
+     */
+    public function addMessageSuggestions($name, array $items)
+    {
+        if (!$alternatives = self::computeAlternatives($name, $items)) {
+            return;
+        }
+
+        $this->rawMessage .= sprintf('. Did you mean "%s"?', implode('", "', $alternatives));
```
stof: should we really add the dot here ? It means that the message without alternatives does not have a final dot

lib/Twig/Error/Syntax.php
fabpot: fixed

lib/Twig/Error/Syntax.php
fabpot: fixed

fabpot: I've also added some new tests as it was not covered.

stof:
:+1:

PR #1903, opened 2015-11-02
Fix BC in NodeTestCase

remicollet:
For context see https://github.com/twigphp/Twig-extensions/issues/155

### DIFF
--- a/lib/Twig/Test/NodeTestCase.php
+++ b/lib/Twig/Test/NodeTestCase.php
@@ -15,17 +15,21 @@ abstract class Twig_Test_NodeTestCase extends PHPUnit_Framework_TestCase
     /**
      * @dataProvider getTests
      */
-    public function testCompile($node, $source, $environment = null)
+    public function testCompile($node, $source, $environment = null, $regex = false)
     {
-        $this->assertNodeCompilation($source, $node, $environment);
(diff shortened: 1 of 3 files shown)


remicollet:
Sorry, PR open against master... should probably be cherry-picked in 1.x

lib/Twig/Test/NodeTestCase.php
```diff
@@ -15,17 +15,21 @@
     /**
      * @dataProvider getTests
      */
-    public function testCompile($node, $source, $environment = null)
+    public function testCompile($node, $source, $environment = null, $regex = false)
```
stof: the name is wrong. this is not a regex

stof:
making the cache per user should be submitted as a separate PR. It seems unrelated to the BC break fix

lib/Twig/Test/NodeTestCase.php
remicollet: what name do you want ?

stof: ``pattern`` may work. Or maybe ``isFormat`` (as PHPUnit calls it a format)

remicollet:
Rewite and rebase for 1.x on PR #1905

PR #1904, opened 2015-11-02
make cache per user (on linux)

remicollet:
Just because I hate having failed test when I run test once on 1 account, then once on another account ;)


fabpot:
Thank you @remicollet.

PR #1905, opened 2015-11-02
fix BC, revert to pre 1.23 behavior, and make use of pattern an option

remicollet:
For context see twigphp/Twig-extensions#155


stof:
:+1:

fabpot:
Thank you @remicollet.

PR #1902, opened 2015-11-02
fix deprecation message

xabbuh:
see https://github.com/twigphp/Twig/commit/e3a325f249a5ff4cdbbabbec7271c13dee2d7ac5#commitcomment-14125098

PR #1906, opened 2015-11-02
Fix "name" variable undefined in deprecated method

lokhman:
Variable "name" is not defined in deprecated method "initRuntime". Patch should fix it.


stof:
duplicate of #1902

stof:
ah no, it is not the same place.

but the way used in #1902 to fix it is better as there is no placeholder in the message

stof:
:+1:

lokhman:
**stof** Yeah, didn't noticed that `name` is not used in the message at all. Removed `sprintf` completely.

fabpot:
I prefer to add the extension name in the message (see #1907)

lokhman:
**fabpot** :+1:

PR #1913, opened 2015-11-10
added runtime extensions

fabpot:


lib/Twig/Environment.php
```diff
@@ -758,6 +759,16 @@ public function hasExtension($name)
     }
 
     /**
+     * Registers an extension runtime loader.
+     *
+     * Loaders have priority over the "standard" getExtension().
+     */
+    public function registerExtensionRuntimeLoader(Twig_RuntimeExtensionLoaderInterface $callable)
```
stof: the variable name is weird. It is not a callable

stof:
how would this work to lazy-load extensions ?

fabpot:
Just done dome renamings.

fabpot:
@stof the goal is to lazy-load the runtime not the extension. An extension is just about declaring which functions/tests/tags/... are available, and they must not depend on any other dependencies. So, the dependencies are part of the runtime, and those can be lazy-loaded when using Symfony or Pimple very easily with some conventions. See the referenced Symfony PR for the implementation of the Form runtime.

sstok:
I like it :) but it is missing some tests.

stof:
@fabpot how does it work for the definition of the callable in Twig_SimpleFilter ? Your Symfony PR would not work on Twig 2, where the callable is typehinted, as it is not callable anymore. And you are already loosing the support of named arguments in 1.x as the callable is not callable anymore.

Please write tests for this feature in Twig, to see whether it works fine

stof:
and shouldn't we have a way for extension to tell the compiler whether they use a runtime or no instead of asking loaders to try to provide one ? This would avoid the lookup in all loaders for each extension access, as most extensions won't use it. This could make things much faster.

fabpot:
I know about the callable type hints we have in Twig 2.0, but I'm just going to remove them.

stof:
@fabpot this will not solve the fact that we need the callable to access parameters through reflection

fabpot:
To speed things up, I'm going to get the runtime only once per template by doing so when initializing the template and then reference a property, that should be enough.

fabpot:
I've just updated the PR which now is able to deal with named arguments. I still need to add some more tests and perhaps throw an exception when the runtime is different from the extension and the `runtime_class` is not defined.

fabpot:
I've just pushed a totally new approach, which is way simpler.

First, no deprecations, no changes, just a new way of registering runtime implementations for Twig elements (functions/filters/tests).

When defining a Twig element, you can now use a class different from the extension: ``array('ImplementationClass', 'method')`` or ``ImplementationClass::method'``. Of course, Twig has no idea on how to create an instance of ``ImplementationClass``, so that's the goal of the runtime loader, which most of the time is going to use a dependency injection container to do its work.

Sounds much easier to me, easier to understand, and fixes all the problems about callables as ``array('ImplementationClass', 'method')`` or ``ImplementationClass::method'`` are considered callables by PHP even if the methods are non-static.

WDYT @stof?

fabpot:
The only "complexity" is how we are going to make it work with Silex and Symfony as we don't have a mechanism to retrieve a service by its class name :)

lib/Twig/Node/Expression/Call.php
```diff
@@ -10,18 +10,29 @@
...
 abstract class Twig_Node_Expression_Call extends Twig_Node_Expression
 {
+    private $reflector;
+
     protected function compileCallable(Twig_Compiler $compiler)
     {
         $closingParenthesis = false;
         if ($this->hasAttribute('callable') && $callable = $this->getAttribute('callable')) {
-            if (is_string($callable)) {
+            if (is_string($callable) && false === strpos($callable, '::')) {
```
fabpot: I added that as some Twig tests define Twig elements as functions that do not exist. Not sure we want to support that, but that's why I've added this here and the code to correctly dump static methods below.

hason:
@fabpot
> we don't have a mechanism to retrieve a service by its class name

I prepared PR https://github.com/symfony/symfony/pull/18268 for Symfony and created the https://github.com/symfonette/class-named-services library.

mnapoli:
> The only "complexity" is how we are going to make it work with Silex and Symfony as we don't have a mechanism to retrieve a service by its class name :)

Is there anything preventing to use service names (instead of class names)?

All that would be very useful in many frameworks (working just like the way we can register controllers using service names).

mnapoli:
Also would it make sense to use container-interop's `ContainerInterface`? Or if you don't want to use that standard, would PSR-11 be a good candidate for that once/if it passes? I'm trying to see if that could be a good use case for it.

derrabus:
If I understood correctly, you're using the FQCN for lookups against this runtime loader. From the point of view of a Symfony developer, it would feel more natural and convenient if I could use service names instead of FQCN, for several reasons:
* With FQCN, I need a mechanism that resolves the FQCN to a service ID.
* Using service IDs, I can rename/move a class without having to change the loader.
* With service IDs, I can have multiple services using the same class with different configurations.
* Furthermore, service IDs allow me to use aliases to change the implementation depending on configuration/environment.

From Twig's POV, it should not really matter if the string that is used to look up a certain service is a FQCN or an arbitrary service ID or am I missing something?

lib/Twig/Node/Expression/Call.php
```diff
@@ -10,18 +10,29 @@
  */
 abstract class Twig_Node_Expression_Call extends Twig_Node_Expression
 {
+    private $reflector;
```
stof: do we really need to make it stateful ? Attributes are mutable, but you don't reset this property when mutating the callable

lib/Twig/Node/Expression/Call.php
```diff
@@ -10,18 +10,29 @@
...
             } else {
-                $type = ucfirst($this->getAttribute('type'));
-                $compiler->raw(sprintf('call_user_func_array($this->env->get%s(\'%s\')->getCallable(), array', $type, $this->getAttribute('name')));
-                $closingParenthesis = true;
+                list($r, $callable) = $this->reflectCallable($callable);
+                if ($r instanceof ReflectionMethod && is_string($callable[0])) {
+                    if ($r->isStatic()) {
+                        $compiler->raw(sprintf('%s::%s', $callable[0], $callable[1]));
+                    } else {
+                        $compiler->raw(sprintf('$this->env->getRuntime(\'%s\')->%s', $callable[0], $callable[1]));
```
stof: at this point, do you have a guarantee that ``$callable[0]`` does not contain a quote and does not end with a ``\`` ? If no, some escaping is necessary.

fabpot: Yes, as `$callable[0]` is a class name at this point, but I will use the proper compiler method to avoid any surprise if we decide to change the class name the to something else.

doc/advanced.rst
```diff
@@ -136,6 +136,10 @@ Creating a filter is as simple as associating a name with a PHP callable::
     // or a simple PHP function
     $filter = new Twig_SimpleFilter('rot13', 'str_rot13');
 
+    // or a class static method
+    $filter = new Twig_SimpleFilter('rot13', array('SomeClass', 'rot13Filter'));
```
harikt: It looks you only need the static example one below. The above one is already added on 143. Or did I miss something else ?

fabpot:
@derrabus @mnapoli One thing you might have missed is that getting an extension by name has been deprecated very recently (#2148). Extensions are now referenced by their class names only. So, re-adding a name for the runtime looks like going backward at this point.

One "simple" way to get names out of class names would be to provide a map between the class name and the service name or just a default convention on the service name (a prefix + the class name in underscore form or something similar).

doc/advanced.rst
fabpot: This is indeed the exact same notation but one would be a static method while the other one is not. I've added some notes to make it (I hope) more clear.

lib/Twig/Node/Expression/Call.php
fabpot: They are mutable, but if done after the compilation, that won't change anything, and this property is only used when compilation starts. It's stateful as we need it twice. The only edge case would be a node that is re-used after a compilation, changed, and recompiled, not sure it that's a reasonable thing to do anyway.

doc/advanced.rst
```diff
@@ -765,6 +765,99 @@ The ``getTests()`` method lets you add new test functions::
...
     }
 
+Definition vs Runtime
+~~~~~~~~~~~~~~~~~~~~~
+
+Twig filters, functions, and tests runtime implementations can be defined as
+any valid PHP callable:
+
+* **functions/closures/static methods**: Simple to implement and fast (used by
+  all Twig core extensions); but it is hard for the runtime to depend on
```
fabpot: right

stof:
@fabpot what we could imagine in Symfony is forcing twig runtimes to be tagged with a specific tag, allowing to build a map of class names to service ids for runtimes.

fabpot:
@stof Exactly, that's what I proposed when I wrote "provide a map between the class name and the service name", convention can be for Silex for instance.

fabpot:
I've started to implement the Symfony counterpart of this PR, see the linked PR, but mostly symfony/symfony#20092

mnapoli:
> @derrabus @mnapoli One thing you might have missed is that getting an extension by name has been deprecated very recently (#2148). Extensions are now referenced by their class names only.

Right but if I come back on what you said:

> When defining a Twig element, you can now use a class different from the extension: array('ImplementationClass', 'method') or ImplementationClass::method'. Of course, Twig has no idea on how to create an instance of ImplementationClass, so that's the goal of the runtime loader, which most of the time is going to use a dependency injection container to do its work.

So there's a difference between referencing an extension (which is now done by class name), and referencing the callable for an element (functions/filters/tests).

For example a problem we are experiencing in my team is that we have a Twig extension for the whole app that provides everything and anything (lots of filters, functions, etc.). We started with dependency injection in the `TwigExtension` class but ended up with a clogged constructor and lots of services initialized for nothing. We could split the class in multiple extensions but all the services would still always be loaded.

With your change we could have a single `TwigExtension` declaring all the elements, yet have the implementations in separate classes (group in topics that make sense). Each of these classes could be a proper service (use dependency injection fully). With the new "loader" those function/filters/… implementations would only be instantiated when the function/filter/… is actually used, which is more efficient.

So lazy-loading the Twig extension is one thing, but lazy-loading the implementation of the elements (function/filter/…) is another problem and is very interesting. And indeed there's no need for extension names (so #2148 is not conflicting), but there's a need for being able to use services instead of class names inside the callables (`new \Twig_SimpleFilter('price', ['app.price_formatter', 'formatPrice']),`).

I hope that makes sense.

fabpot:
@mnapoli I think we agree on (almost) everything. The goal of this PR is indeed to lazy-load the runtimes, not the extensions which are now only about how to expand Twig (without the implementation). Using the class name is the best way to not consider how the runtimes are going to be loaded. Using a service name is exactly the same as giving a name to the extension. They already have a unique identifier, the class name.

Anyway, have a look at my PRs on Symfony, everything works quite well as is.

PR #1919, opened 2015-11-13
ANY_CALL vs ArrayAccess issue

kakserpom:
Template->getAttribute(): ANY_CALL fails if object implements ArrayAccess and offsetExists() returns false.

### DIFF
--- a/lib/Twig/Template.php
+++ b/lib/Twig/Template.php
@@ -469,7 +469,7 @@ abstract class Twig_Template implements Twig_TemplateInterface
                 return $object[$arrayItem];
             }
 
-            if (self::ARRAY_CALL === $type || !is_object($object)) {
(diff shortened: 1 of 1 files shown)


stof:
this is wrong. It forbids calling getter methods on objects which also implement ArrayAccess

stof:
btw, it is clear that you haven't tried to run tests with your change, as many are failing because of the bug you are trying to introduce here

kakserpom:
> this is wrong. It forbids calling getter methods on objects which also implement ArrayAccess
Could it be the case? If object implements ArrayAccess, why call methods?

stof:
@kakserpom because it is perfectly valid to implement more things than ArrayAccess. And this is used out there. See https://github.com/symfony/symfony/blob/2.8/src/Symfony/Component/Form/FormView.php for instance, which also provides some public properties and methods

kakserpom:
Thank you. disableStrictVariables() helped my situation.

PR #1925, opened 2015-11-22
Added a unambiguous index1 and revindex1 to enhance compatibility ...

babs:
with other potential implementations
Thiw way index0 and index1 are properly defined and avoid the index ambiguity.


babs:
Will update test accordingly first

PR #1926, opened 2015-11-22
Added a unambiguous index1 and revindex1 to enhance compatibility ...

babs:
with other potential implementations.
This way index0 and index1 are properly defined and avoid the index ambiguity.
[Code, test and doc.]


fabpot:
which "other" implementation is using 'index1'?

babs:
None currently, that's why I used the word potential. I'd like Pebble and jTwig to do so but they won't until Twig does :/

fabpot:
I don't understand why the current way is ambiguous.

babs:
Because the default index is 1 indexed which pretty unusual in computing.
For example, jTwig first implementation was 0 indexed like Pebble.

babs:
Don't get me wrong, I agree this doesn't bring new things to Twig itself, but more to its ecosystem.

fabpot:
I'm going to close this one for the following reasons:

* Jinja and Twig are using the exact same names for years;
* You cannot guess that this variable exists, so you probably learned about it via the documentation where things are explained and the 0 variable is defined;
* I don't like having two variables that does the exact same thing as people will start asking questions (like me) about why the previous one is ambiguous and when to use one or the other;
* Just renaming the old one would be a huge BC break.

Thanks for reporting this, but I don't think it is worth a fix, sorry.

PR #1927, opened 2015-11-23
Added skipping empty paths

KpuTuK:
```php
<?php
class Foo {
protected $twig;
public function __construct($otherDir = null) {
        $this->twig = new \Twig_Environment(
            new \Twig_Loader_Filesystem([
                $otherDir,
                dirname(__DIR__).'/Templates'
            ])
        );
    }
}

$foo = new Foo();
$foo->.....
$foo2 = new Foo('call directory');
$foo2->...

before:  //The "" directory does not exist.
after: ....
```

### DIFF
--- a/lib/Twig/Loader/Filesystem.php
+++ b/lib/Twig/Loader/Filesystem.php
@@ -89,7 +89,9 @@ class Twig_Loader_Filesystem implements Twig_LoaderInterface, Twig_ExistsLoaderI
     {
         // invalidate the cache
         $this->cache = $this->errorCache = array();
-
+        if ($path === null) {
+            return;
+        }
         if (!is_dir($path)) {
             throw new Twig_Error_Loader(sprintf('The "%s" directory does not exist.', $path));
         }


Tobion:
:-1: It's not the responsibility of twig to skip empty path but your responsibility to call the method right.

lib/Twig/Loader/Filesystem.php
```diff
@@ -89,7 +89,9 @@ public function addPath($path, $namespace = self::MAIN_NAMESPACE)
     {
         // invalidate the cache
         $this->cache = $this->errorCache = array();
-
+        if ($path === null) {
```
Tobion: $path cannot be null according to the phpdoc

lib/Twig/Loader/Filesystem.php
```diff
@@ -89,7 +89,9 @@ public function addPath($path, $namespace = self::MAIN_NAMESPACE)
     {
         // invalidate the cache
         $this->cache = $this->errorCache = array();
```
Tobion: now the cache is cleared even if the method doesn't do anything. so you can see that the fix doesn't make sense in twig.

PR #1930, opened 2015-12-01
Fix Parser context push on stack

molaux:
Hello,

With php7RC8, in Twig_Parser, parse method starts by saving current context in
copying self properties on stack. It sounds like get_object_vars returns now
references instead of copies of properties (at least for arrays). Consequently,
when reinitializing Parser for subparse line 92 to 99, the preserved context on
stack was also erased.

As a result, when calling {{ parent() }} after a {%embed%}{%endembed%}, the restored blockStack being empty, Twig throws a Twig_Error_Syntax('Calling "parent" outside a block is forbidden.'), despite the fact it is really in a block that extends another. 
When calling {{ parent() }} before  the embed tag, all works properly. (It seems that the empty context, resulting of subparses, is not a problem as far as you don't need to access it)

### DIFF
--- a/lib/Twig/Parser.php
+++ b/lib/Twig/Parser.php
@@ -63,16 +63,14 @@ class Twig_Parser implements Twig_ParserInterface
     public function parse(Twig_TokenStream $stream, $test = null, $dropNeedle = false)
     {
         // push all variables into the stack to keep the current state of the parser
-        $vars = get_object_vars($this);
-        unset($vars['stack'], $vars['env'], $vars['handlers'], $vars['visitors'], $vars['expressionParser'], $vars['reservedMacroNames']);
-
-        // copy all references (that still points to a "$this property")
-        $varsCopy = array();
-        foreach ($vars as $key => $value) {
-            $varsCopy[$key] = $value;
+        // using get_object_vars() instead of foreach would lead to https://bugs.php.net/71336
+        $vars = array();
+        foreach ($this as $k => $v) {
+            $vars[$k] = $v;
         }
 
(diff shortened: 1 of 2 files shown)


stof:
This should be reported as a bug in PHP 7 IMO

molaux:
Yes, but to be honest, at the present time, I am unable to reproduce the behavior of get_object_vars in a simple test case, outside Twig. No idea what is the real cause for this. But for me, in Twig_Parser around line 90, this simple test doesn't display the same thing
```php
var_dump($vars['blockStack']);
$this->blockStack = array();
var_dump($vars['blockStack']);
```

stof:
@molaux please also write a test in the testsuite reproducing the issue, to prevent regressions (we have a folder with integration tests which is suited for that)

molaux:
@stof just enhanced an existing test.

fabpot:
ping @nicolas-grekas

nicolas-grekas:
Issue confirmed and reported as https://bugs.php.net/71336

nicolas-grekas:
I suggest this path as workaround:
```diff
-        $vars = get_object_vars($this);
+        // using get_object_vars() instead of foreach would lead to https://bugs.php.net/71336
+        $vars = array();
+        foreach ($this as $k => $v) {
+            $vars[$k] = $v;
+        }
```

fabpot:
@molaux Can you update the patch with @nicolas-grekas suggestion?

molaux:
Done

fabpot:
Thank you @molaux.

PR #1933, opened 2015-12-04
Activated template cache check with isAutoReload (auto_reload = false)

volchnik:
Noticed that twig crashes with auto_reload disabled while trying to render templates from cache that are not cached yet. Here is some workaround.


volchnik:
Tests didn't pass. May be I misunderstand meaning of auto_reload flag.
> recompile the template whenever the source code changes

PR #1934, opened 2015-12-05
Changed initialization and default value of auto_reload setting

volchnik:
Fixing error during template rendering when:
* debug set to false
* cache directory defined and empty
* auto_reload not defined

Making auto_reload always enabled during debug is enabled.

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -95,9 +95,9 @@ The following options are available:
 * ``auto_reload`` *boolean*
 
   When developing with Twig, it's useful to recompile the
-  template whenever the source code changes. If you don't provide a value for
-  the ``auto_reload`` option, it will be determined automatically based on the
-  ``debug`` value.
+  template whenever the source code changes. If ``debug`` is set to ``true`` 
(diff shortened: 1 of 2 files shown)


stof:
Why should it be always enabled when debug is enabled ? I don't understand the reasonning. People should be able to configure them separately

volchnik:
Well, switching off undefined auto_reload when debug is false is also bad idea (especially for novice users)
May be the best idea is just change auto_reload default to true.

volchnik:
Besides changing auto_reload default, reasoning for always enable auto_reload in debug mode:
enabled debug mode must be pessimistic for cache pruning.

stof:
@volchnik the issue with your idea is that it means that setting auto_reload to false does not respect the config set by the user explicitly

volchnik:
There will be some coupling between auto_reload and debug settings, ignoring user config in one case (debug=true&auto_reload=false). Thought this case will be useful for novice, and not really break legacy configurations.

stof:
@volchnik but why should this coupling happen ?

volchnik:
That case seems to be useless even harmful for those who mess with configuration

PR #1936, opened 2015-12-08
Ensure the template passed to createTemplate is always a string

### LINKED ISSUE #1892: createTemplate with empty string
As discussed in #1706 and https://github.com/twigphp/Twig/pull/1641#issuecomment-76732213 the way to render strings is as follows:

````php
$string = 'Hello, {{ name }}';
$template = $env->createTemplate($string);
echo $template->render(array('name' => 'Bob')); // Hello, Bob
````

A common use case for this is customization based on user input. However if you render user input it could happen that the template is empty which leads to errors if you then call render on that template.

I am seeing the error as follows:

````
Template "__string_template__da47b0fa50ef004786efc7745172ce3402223a0f96b70d8f9d0bbbc2f70589a7" is not defined.
````

I would assume even if you pass an empty string as template everything should work, if I use a normal template the content could be empty as well. If someone confirms I might be able to look into this.

peterrehm:
If you pass `null` to createTemplate you end up with an error like Template "__string_template__6d8f7c778df0b84ad07aa37e18f237bd84beee3d268093cb88941253f539b3a2" is not defined.

In general `null` does not make sense as template, but as in my use case it can happen when you try to validate dynamic templates, e.g. based on input from the form component. 

Either we throw an Exception which states that null as a template is not allowed or we cast it as propesed by this PR.

- fixes #1892

### DIFF
--- a/lib/Twig/Environment.php
+++ b/lib/Twig/Environment.php
@@ -426,7 +426,8 @@ class Twig_Environment
     public function createTemplate($template)
     {
         $name = sprintf('__string_template__%s', hash('sha256', uniqid(mt_rand(), true), false));
-
+        $template = (string) $template;
+        
         $loader = new Twig_Loader_Chain(array(
             new Twig_Loader_Array(array($name => $template)),
             $current = $this->getLoader(),


stof:
Passing null here is forbidden. It does not respect the method signature (in a PHP 7+ library, we would have a scalar typehint to enforce it). There is many place in the code where we consider that the phpdoc defines the contract as we don't have scalar typehints.

IMO, you need to fix you code

peterrehm:
I have fixed it in my code as I wrapped the call, but I think it makes sense to have either option (exception or cast).

fabpot:
I agree with @stof here. If we start going down this road, we would have to add so many checks everywhere. Moreover, adding `(string) ` means that you can now pass an object implementing `__toString()` for instance.

Issue #1914, opened 2015-11-10
Twig_TemplateInterface is missing


emodric:
`Twig_TemplateInterface` is missing from master branch. It seems to me it was removed by a recent merge of 1.x branch.

Deprecation docs for 1.x branch (both in [doc/deprecated.rst](https://github.com/twigphp/Twig/blob/1.x/doc/deprecated.rst#interfaces) and in [interface itself](https://github.com/twigphp/Twig/blob/1.x/lib/Twig/TemplateInterface.php#L17)) say this interface will be removed in 3.0 and that in 2.0 will only be blanked out.

stof:
This is a mistake in the doc. All these interfaces are gone in 2.0

emodric:
Ah, alright then.

To migrate from the interface, is it enough to use Twig_Template instead on 1.x?

stof:
Yes (using a different implementation of the interface which does not extend Twig_Template does not work properly anyway, which is why the interface is deprecated)

emodric:
Thanks @stof!

Closing the issue and PR!

PR #1941, opened 2015-12-17
Reintroduce Twig_TemplateInterface, since it should be removed only in 3.0

emodric:
As mentioned in https://github.com/twigphp/Twig/issues/1914, docs say this interface should be removed in 3.0, and not in 2.0


emodric:
Closing after @stof 's comments on the issue.

PR #1942, opened 2015-12-17
DisplayCallGraphTemplate: a visual way to debug templates and blocks

dunglas:
A picture is worth a thousand words:

capture d ecran 2015-12-17 a 16 27 55

To enable it:

```php
$twig = new Twig_Environment(null, array('base_template_class' => 'Twig_Template_DisplayCallGraphTemplate');
```

Using Symfony:

```yaml
# app/config/config_dev.yml

twig:
    base_template_class: Twig_Template_DisplayCallGraphTemplate
```

The color scheme and the font are the same than in the Symfony Profiler. 

/cc @xavierleune @cssyren

### DIFF
--- a/lib/Twig/Template/DisplayCallGraphTemplate.php
+++ b/lib/Twig/Template/DisplayCallGraphTemplate.php
@@ -16,10 +16,10 @@
  */
 abstract class Twig_Template_DisplayCallGraphTemplate extends Twig_Template
 {
-    private $templateStart = '<div style="border: 1px solid rgba(240, 181, 24, 0.3); margin: 5px;"><span style="background-color: rgba(240, 181, 24, 0.3); color: black; font-family: monospace;">Template "%s"</span>';
-    private $templateEnd = '</div>';
-    private $blockStart = '<div style="border: 1px solid rgba(100, 189, 99, 0.2); margin: 5px;"><span style="background-color: rgba(100, 189, 99, 0.2); color: black; font-family: monospace;">Block "%s"</span>';
-    private $blockEnd = '</div>';
+    protected $templateStart = '<div style="border: 1px solid rgba(240, 181, 24, 0.3); margin: 5px;"><span style="background-color: rgba(240, 181, 24, 0.3); color: black; font-family: monospace;">Template "%s"</span>';
+    protected $templateEnd = '</div>';
+    protected $blockStart = '<div style="border: 1px solid rgba(100, 189, 99, 0.2); margin: 5px;"><span style="background-color: rgba(100, 189, 99, 0.2); color: black; font-family: monospace;">Block "%s"</span>';
+    protected $blockEnd = '</div>';
 
     /**
      * @var string[]
--- a/lib/Twig/Template/DisplayCallGraphTemplate.php
+++ b/lib/Twig/Template/DisplayCallGraphTemplate.php
@@ -1,0 +1,107 @@
+<?php
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) 2010 Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+/**
+ * Displays the call graph (templates names and block names) in a visual way directly in the generated HTML.
+ *
+ * @author Kévin Dunglas <dunglas@gmail.com>
+ */
+abstract class Twig_Template_DisplayCallGraphTemplate extends Twig_Template
+{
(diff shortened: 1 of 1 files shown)


xavierleune:
:+1:

stof:
This would break any UI where the template is included somewhere where HTML is invalid (or not using HTML as the output), so I'm not a fan of it. This is even worse for blocks, as blocks could happen inside attribute values for instance.

Koc:
How to configure blacklists?

dunglas:
@stof it's why the `isEnabled()` method, `blacklist` and `start`/`end` properties are `protected` (and because we cannot use `static::` in PHP 5.2). You can customize this class according to your project need. And the user must enable it manually.

What we have in mind is extending this template in the project and display the call graph only when a specific `$_GET` parameter exists. It should limit such strange behaviors.

Btw, your concern are similar for most of the tools outputing HTML including xdebug.

@Koc override this template in your own project:

```php
namespace AppBundle\Twig;

abstract class DisplayCallGraphTemplate extends \Twig_Template_DisplayCallGraphTemplate
{
    public function __construct(\Twig_Environment $env)
    {
        parent::__construct($env);

       $this->templateBlackList[] = '@MyBundle';
       $this->blockBlackList[] = 'myBlock';
       $this->blockStart = '<blockstart name="%s">'; // For instance, here we are in an XML Document
    }

  // Override this method for a more complex exclusion logic for templates
  // protected function isTemplateEnabled() {}

  // Same for blocks
  // protected function isBlockEnabled($name) {}
}
```

```yaml
# app/config/config_dev.yml

twig:
    base_template_class: AppBundle\Twig\DisplayCallGraphTemplate
```

oxman:
:+1:

dunglas:
If this patch is merged I'll add a button in the Symfony web profiler toolbar to show and hide block and template names in a click.

lib/Twig/Template/DisplayCallGraphTemplate.php
```diff
@@ -0,0 +1,107 @@
...
+
+    /**
+     * @var string[]
+     */
+    protected $templateBlackList = array(
+        '@WebProfiler',
+        '@Security',
+        '@Doctrine',
+        '@Swiftmailer',
+        '@Debug',
```
Tobion: twig does not know anything about those namespaces

dunglas:
@Tobion good catch thanks. Removed these references, they belong to the child class in Symfony.

fabpot:
I'm not sure it should be part of Twig core as this only works for HTML templates and nothing prevents this from being decoupled from Twig.

dunglas:
@fabpot is https://github.com/twigphp/Twig-extensions a better place or should it be in another namespace?

fabpot:
Indeed, Twig-extensions is probably a better place.

dunglas:
Twig-extensions is not included in the Symfony SE anymore.

My final goal was to integrate this with the Symfony web debug toolbar (add a button to display blocks - will work fine because the profiler is only displayed on HTML pages).

Is it ok to add Twig-extensions to the `suggest` section of the Twig Bridge and Web Profiler Bundle and add back Twig-extensions in the `require` section to Symfony SE?

lib/Twig/Template/DisplayCallGraphTemplate.php
```diff
@@ -0,0 +1,99 @@
...
+/**
+ * Displays the call graph (templates names and block names) in a visual way directly in the generated HTML.
+ *
+ * @author Kévin Dunglas <dunglas@gmail.com>
+ */
+abstract class Twig_Template_DisplayCallGraphTemplate extends Twig_Template
+{
+    protected $templateStart = '<div style="border: 1px solid rgba(240, 181, 24, 0.3); margin: 5px;"><span style="background-color: rgba(240, 181, 24, 0.3); color: black; font-family: monospace;">Template "%s"</span>';
+    protected $templateEnd = '</div>';
+    protected $blockStart = '<div style="border: 1px solid rgba(100, 189, 99, 0.2); margin: 5px;"><span style="background-color: rgba(100, 189, 99, 0.2); color: black; font-family: monospace;">Block "%s"</span>';
```
Tobion: IMO extensibility to change all those values is better implemented using protected getters. so to change it, people can overwrite the method. then we do not need to deal with setters and properties and it's easier to maintain bc.

stof: agreed

oxman:
Any new ?

fabpot:
Closing as this won't be included in Twig for the reasons I've explained before.

dunglas:
I'll reopen the PR against twig-extensions.

PR #1945, opened 2015-12-21
skip read-only dir tests on windows and cleanup tmp dirs correctly

Tobion:
- fix tests on windows (skipping them as read-only dirs are not possible on windows, only read-only files; read-only attribute on folders (`exec('attrib +R '`) does not prevent files to be created inside them)
- correctly cleanup tmp directories at the end of tests

### DIFF
--- a/test/Twig/Tests/FileCachingTest.php
+++ b/test/Twig/Tests/FileCachingTest.php
@@ -11,11 +11,10 @@
 
 class Twig_Tests_FileCachingTest extends PHPUnit_Framework_TestCase
 {
-    protected $fileName;
-    protected $env;
-    protected $tmpDir;
+    private $env;
+    private $tmpDir;
 
-    public function setUp()
+    protected function setUp()
     {
         $this->tmpDir = sys_get_temp_dir().'/TwigTests';
         if (!file_exists($this->tmpDir)) {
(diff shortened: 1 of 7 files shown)


test/Twig/Tests/FileCachingTest.php
```diff
@@ -29,13 +28,19 @@ public function setUp()
...
     }
 
-    public function tearDown()
+    protected function tearDown()
     {
-        if ($this->fileName) {
-            unlink($this->fileName);
+        $iterator = new RecursiveIteratorIterator(new RecursiveDirectoryIterator($this->tmpDir), RecursiveIteratorIterator::CHILD_FIRST);
+        foreach ($iterator as $filename => $fileInfo) {
+            if (!$iterator->isDot()) {
```
stof: this can be filtered by the iterator itself by passing the ``SKIIP_DOTS`` flag

fabpot:
Tests are broken on PHP 5.2.

Tobion:
@fabpot Fixed. I'll create a PR for master to remove the PHP 5.2 compatibility code when this is merged.

fabpot:
Thank you @Tobion.

fabpot:
Merged now. Can submit a PR on master?

Tobion:
You need to merge 1.x into master first.

fabpot:
indeed, merged now

Tobion:
See #1947

PR #1950, opened 2015-12-30
Fix sandbox for methods

hason:
Fixes:
- Configure sandbox for method `__call`.
- Call enabled method with name same as property.

### DIFF
--- a/src/Template.php
+++ b/src/Template.php
@@ -15,6 +15,8 @@ namespace Twig;
 use Twig\Error\Error;
 use Twig\Error\LoaderError;
 use Twig\Error\RuntimeError;
+use Twig\Extension\SandboxExtension;
+use Twig\Sandbox\SecurityError;
 
 /**
  * Default base class for compiled templates.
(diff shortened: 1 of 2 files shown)


lib/Twig/Template.php
```diff
@@ -573,6 +576,10 @@ protected function getAttribute($object, $item, array $arguments = array(), $typ
             $method = (string) $item;
             $call = true;
         } else {
+            if (isset($propertySandboxException)) {
+                throw $propertySandboxException;
```
stof: shouldn't this go after the `$isDefinedTest` condition ?

lib/Twig/Template.php
```diff
@@ -523,18 +523,21 @@ protected function getAttribute($object, $item, array $arguments = array(), $typ
...
-                }
+                    if ($this->env->hasExtension('sandbox')) {
+                        $this->env->getExtension('sandbox')->checkPropertyAllowed($object, $item);
+                    }
 
-                return $object->$item;
+                    return $object->$item;
+                }
             }
+        } catch (Twig_Sandbox_SecurityError $propertySandboxException) {
```
linaori: Is there any particular case why the try is surrounding such a big piece of code?

lokera07:
Hola

lib/Twig/Template.php
hason: Thanks! Fixed.

fabpot:
Closing as obsolete: current Twig already resolves an unallowed sandbox property to a getter or `__call()` fallback in and re-throws the property error when no allowed method exists, so this fix is no longer needed.

PR #1955, opened 2016-01-07
Let PHP convert Twig_Markup to JSON instead of pre-filtering it

stof:
See http://blog.blackfire.io/blackfire-on-blackfire-twig-and-json.html for the reason to do this.
The bump of the requirement to PHP 5.5+ in Twig 2 makes it possible to rely on ``JsonSerializable``


fabpot:
Great!

romainneutron:
\o/

Issue #1957, opened 2016-01-11
Current v1.23.2 release broken due to a typo


ratza:
Hi! I had the luck to update my dependencies 15 minutes after the release of v1.23.2.

The latest version has been released with a typo. The line 581 in `lib/Twig/ExpressionParser.php` reads:
```php
if (!is_bool($functon->getDeprecatedVersion())) {
```

The variable name of the variable should read `$function`.

Not sure if a pull request is needed as the release should be removed or marked as broken, and re-released with the small fix.

Thanks for the awesome work!

SpacePossum:
duplicates https://github.com/twigphp/Twig/issues/1956

Issue #1956, opened 2016-01-11
Twig 1.23.2 and typo in ExpresionParser.php


flylord:
You have typo on 581. line. It should be $function but you have $functon

            if (!is_bool($functon->getDeprecatedVersion())) {
                $message .= sprintf(' since version %s', $function->getDeprecatedVersion());
            }

chandon:
+1 ?

PHP Notice:  Undefined variable: functon in /home/visiint/symfony/vendor/twig/twig/lib/Twig/ExpressionParser.php on line 581

PR #1958, opened 2016-01-11
Fxi typo

SpacePossum:
https://github.com/twigphp/Twig/issues/1957
and
https://github.com/twigphp/Twig/issues/1956


fabpot:
Thank you @SpacePossum.

PR #1964, opened 2016-01-16
undeprecated _self

### LINKED ISSUE #1932: [Suggestion] Add a way to access the name of the current template
``_self`` accessing the current Twig_Template instance is deprecated (and removed in 2.0). This causes headaches in several places needing an easy way to reference the current template (and this is why we kept a special handling of it in from and import tags).

However, most places relying on ``_self`` today don't actually need the template instance. The template name would be enough as they pass it to functions resolving the template anyway. Would you accept a way to get the name of the current template rather than the instance (there is no security implication with just the name). This could even allow to remove ``_self`` entirely as the from and import tag would also accept it.

See https://github.com/symfony/symfony-demo/issues/242#issuecomment-153388234 and https://github.com/symfony/symfony/issues/16804 for instance

fabpot:
Un-deprecated `_self`. As mentioned in #1932, `_self` is sometimes used to get the current template name, which is a legitimate usage and almost always the user intent. Moreover, when `_self` is used in a template, it is mostly passed to Twig functions that are able to deal with Twig templates or Twig template names indifferently.

fixes #1932

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
 * 1.23.2 (2015-XX-XX)
 
+ * undeprecated _self (should only be used to get the template name, not the template instance)
  * added versions in deprecated messages
  * made file cache tolerant for trailing (back)slashes on directory configuration
  * deprecated unused Twig_Node_Expression_ExtensionReference class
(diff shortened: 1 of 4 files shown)


fabpot:
see #1965 for the 2.0 counterpart.

Tobion:
:+1:

CHANGELOG
```diff
@@ -1,5 +1,6 @@
 * 1.23.2 (2015-XX-XX)
```
xabbuh: This looks wrong (the last patch release was `1.23.3`).

PR #1965, opened 2016-01-16
reintroduced _self (returns the current template name)

### LINKED ISSUE #1932: [Suggestion] Add a way to access the name of the current template

fabpot:
fixes #1932, see #1964 fro the 1.x counterpart.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,9 +1,9 @@
-* 2.0.0 (201X-XX-XX)
+* 2.0.0 (2016-XX-XX)
 
  * sped up the json_encode filter
  * removed reserved macro names; all names can be used as macro
  * removed Twig_Template::getEnvironment()
- * removed _self variable (except for usage in from and import tags)
+ * changed _self variable to return the current template name
  * made the loader a required argument of Twig_Environment constructor
  * removed Twig_Environment::clearTemplateCache()
  * removed Twig_Autoloader (use Composer instead)
--- a/doc/templates.rst
+++ b/doc/templates.rst
@@ -127,6 +127,7 @@ Global Variables
 
 The following variables are always available in templates:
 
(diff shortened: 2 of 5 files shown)


CHANGELOG
```diff
@@ -1,9 +1,9 @@
-* 2.0.0 (201X-XX-XX)
+* 2.0.0 (2016-XX-XX)
 
  * sped up the json_encode filter
  * removed reserved macro names; all names can be used as macro
  * removed Twig_Template::getEnvironment()
- * removed _self variable (except for usage in from and import tags)
+ * changed _self variable to returns the current template name
```
Tobion: typo "return"

doc/templates.rst
```diff
@@ -127,6 +127,7 @@ Global Variables
 
 The following variables are always available in templates:
 
+* ``_self``: reference the current template name;
```
Tobion: "references" as below

CHANGELOG
fabpot: fixed

doc/templates.rst
fabpot: fixed

PR #1966, opened 2016-01-17
Use strict comparison for charset

JhonnyL:
There is no reason not to use strict comparison here. Also some minor cs fixes.


Tobion:
:+1:

PR #1969, opened 2016-01-23
add hasser support to getAttribute()

### LINKED ISSUE #1750: Template getAttribute() does not support hasser
Using Twig with Symfony, I expected a support for hasser (just like in Symfony PropertyAccess component). Instead, Twig only supports getter and isser.

xabbuh:
This resolves #1750.

### DIFF
--- a/doc/templates.rst
+++ b/doc/templates.rst
@@ -110,6 +110,7 @@ is set, Twig will throw an error (see :ref:`environment options<environment_opti
       (even if ``bar`` is the constructor - use ``__construct()`` instead);
     * if not, and if ``foo`` is an object, check that ``getBar`` is a valid method;
     * if not, and if ``foo`` is an object, check that ``isBar`` is a valid method;
+    * if not, and if ``foo`` is an object, check that ``hasBar`` is a valid method;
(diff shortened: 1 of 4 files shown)


fabpot:
Can you also update the docs?

Tobion:
Could be considered a BC break which happens when you have a magic getter and a hasser on your object. It will now choose the other one.

xabbuh:
Add a line to the docs. @Tobion is right about the fact that this can be considered a BC break (it was never stated before that hasser methods do work too so this hardly is a bugfix).

fabpot:
What about adding this to Twig 2.0 only?

xabbuh:
We could also swap the order of the checks for the hasser method and the magic `__call()` method, but merging into 2.0 would be okay for me too.

Tobion:
+1 in 2.0. Does this have an performance impact? If I remember right, hasser support was proposed before already (although I could not find it anymore).

xabbuh:
I don't see any performance impact (the available class methods are cached anyway and `isset()` is in O(1)).

fabpot:
Thank you @xabbuh.

PR #1972, opened 2016-01-25
Null coalescing op

### LINKED ISSUE #1949: Adding a Null-coalescing operator
I frequently find myself needing to use [null-coalescing](https://en.wikipedia.org/wiki/Null_coalescing_operator) in my Twig templates (i.e. given several variables, use the first one that is set and not null).

The ternary operators and `default` filter are sometimes useful, but both have the downside of becoming unmanageably verbose when there are more than one or two possible variables in play.

For example:
```thing1|default(thing2|default(thing3|default(thing4|default('Oy vey.'))))```
or worse...
```(thing1 is defined and thing1 is not null ? thing1 : (thing2 is defined and thing2 is not null ? thing2 : (thing3 is defined and thing3 is not null ? thing3 : ('This is ridiculous.') ) ) )```

It's in Twig's [raison d'etre](http://twig.sensiolabs.org/) to be "Concise"... so why not:
`thing1 ?? thing2 ?? thing3 ?? thing4 ?? 'Nice.'`

PHP offers a null-coalescing operator natively as of [version 7](http://php.net/manual/en/language.operators.comparison.php), but it is straightforward to implement a shim for earlier versions.

I added a proof-of-concept in a CraftCMS plugin, here: https://bitbucket.org/topshelfcraft/null-coalescing-twig-operator/src/5f86e32aa53d6a49f26ebc36926c404bd145221b/qq/twigoperators/QqNullCoalesce.php

I would like to clean up my POC, combine the shim and native versions into a single class, and contribute it to Twig core. Would this be a welcome addition?

fabpot:
fixes #1949

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
 * 1.23.4 (2016-XX-XX)
 
+ * adding support for the ?? operator
  * fixed the defined test when used on a constant, a map, or a sequence
  * undeprecated _self (should only be used to get the template name, not the template instance)
  * fixed parsing on PHP7
--- a/doc/templates.rst
+++ b/doc/templates.rst
@@ -808,6 +808,13 @@ The following operators don't fit into any of the other categories:
       {{ foo ?: 'no' }} is the same as {{ foo ? foo : 'no' }}
       {{ foo ? 'yes' }} is the same as {{ foo ? 'yes' : '' }}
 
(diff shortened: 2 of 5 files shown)


peterrehm:
Awesome feature, great work!

lukeholder:
:+1:

SpacePossum:
:+1: thanks :D

PR #1974, opened 2016-01-27
refactored extensions management

fabpot:


lib/Twig/ExtensionSet.php
```diff
@@ -0,0 +1,495 @@
...
+            $this->initExtensions();
+        }
+
+        return $this->functions;
+    }
+
+    /**
+     * Get a function by name.
+     *
+     * Subclasses may override this method and load functions differently;
```
sstok: This class is marked final, so it can't be extended. Is comment still valid?

lib/Twig/ExtensionSet.php
```diff
@@ -0,0 +1,495 @@
...
+     * so no list of functions is available.
+     *
+     * @param string $name function name
+     *
+     * @return Twig_Function|false A Twig_Function instance or false if the function does not exist
+     */
+    public function getFunction($name)
```
sstok: I wonder if it would be an idea to add a matched result to the functions property.

Getting function `math_add` matches pattern 'math_*', and then the 'math_add' is added to functions array with the result (callback). So there is no need to do a look-up all every time.

lib/Twig/ExtensionSet.php
```diff
@@ -0,0 +1,495 @@
...
+        if (!$this->initialized) {
+            $this->initExtensions();
+        }
+
+        return $this->binaryOperators;
+    }
+
+    private function initExtensions()
+    {
+        if ($this->initialized) {
```
Tobion: each call to `initExtensions` already checks this property before calling the method. So this condition here seems useless and removing it would spare one check per call.

sstok: :+1:

lib/Twig/ExtensionSet.php
```diff
@@ -0,0 +1,495 @@
...
+
+        foreach ($this->functions as $pattern => $function) {
+            $pattern = str_replace('\\*', '(.*?)', preg_quote($pattern, '#'), $count);
+
+            if ($count) {
+                if (preg_match('#^'.$pattern.'$#', $name, $matches)) {
+                    array_shift($matches);
+                    $function->setArguments($matches);
+
+                    return $function;
```
iltar: These ifs could be merged; `if ($count && ...)`

lib/Twig/ExtensionSet.php
```diff
@@ -0,0 +1,495 @@
...
+
+        foreach ($this->filters as $pattern => $filter) {
+            $pattern = str_replace('\\*', '(.*?)', preg_quote($pattern, '#'), $count);
+
+            if ($count) {
+                if (preg_match('#^'.$pattern.'$#', $name, $matches)) {
+                    array_shift($matches);
+                    $filter->setArguments($matches);
+
+                    return $filter;
```
iltar: ifs here can be merged too

lib/Twig/ExtensionSet.php
fabpot: This only happens when compiling a template, so not sure this is worth it.

lib/Twig/ExtensionSet.php
fabpot: indeed

lib/Twig/ExtensionSet.php
fabpot: indeed

lib/Twig/ExtensionSet.php
fabpot: done

lib/Twig/ExtensionSet.php
fabpot: done

fabpot:
I'm going to merge this one as is. Initially, I wanted to removed all proxy methods from Twig_Environment, but that would be a big BC (even if they are all marked internal now). We will probably deprecated them in 2.1 and removed them in 3.0.

PR #1978, opened 2016-01-29
[DOC] Fix documentation, recipes section

iamluc:
Fix incorrect display on http://twig.sensiolabs.org/doc/recipes.html#loading-a-template-from-a-string


fabpot:
Thank you @iamluc.

Issue #1498, opened 2014-09-07
suggestion add .gitattributes to limit data "over head" for production


jfusion:
hello This is not really a bug but more of a request.

Could you add .gitattributes to allow for limiting the package size ? so i don't have to go clear up after each checkout?

EG:
doc/ export-ignore
ext/ export-ignore
test/ export-ignore
.gitattributes export-ignore
.editorconfig  export-ignore
.gitignore export-ignore
.travis.yml export-ignore
CHANGELOG export-ignore
README.rst export-ignore
phpunit.xml.dist export-ignore

You may know better tan me if this is correct or not. but if checkout out for production this would limit a given package size

I might add that this only refer to the stable releases made using tags (as far as i know)

PR #1981, opened 2016-02-01
[+]: ".gitattributes" <- prevent installing special files via composer

voku:
[Review on Reviewable](https://reviewable.io/reviews/twigphp/twig/1981)


barryvdh:
Duplicates of #942 #1498 and #1776
The last one is still open, so this can be closed

PR #1998, opened 2016-03-04
[DX] "Unexpected" vs. "Unknown" -tag exception message.

SpacePossum:
Templates like this:

```twig
{% block my-test %}
{% endblock %}
```

Would result in:
>Twig_Error_Syntax: Unknown "endblock" tag in "index.twig" at line

This PR changes that to:
>Twig_Error_Syntax: Unexpected "endblock" tag in "index.twig" at line

I think this is better because:
* more accurate message, the `endblock` is known to system (and well typed by the developer); it is however misplaced in the template
* having the test file in the repo will help developers to figure out what is going wrong more easy, since a lot of the time this exception is a result of trying to use an expression as block name

### DIFF
--- a/lib/Twig/Parser.php
+++ b/lib/Twig/Parser.php
@@ -175,8 +175,14 @@ class Twig_Parser implements Twig_ParserInterface
                                 $e->appendMessage(sprintf(' (expecting closing tag for the "%s" tag defined near line %s).', $test[0]->getTag(), $lineno));
                             }
                         } else {
-                            $e = new Twig_Error_Syntax(sprintf('Unknown "%s" tag.', $token->getValue()), $token->getLine(), $this->getFilename());
-                            $e->addSuggestions($token->getValue(), array_keys($this->env->getTags()));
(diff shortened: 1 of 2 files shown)


SpacePossum:
ping ;)

xabbuh:
Is it not possible to get a full list of all known tags at this stage to cover even more cases?

SpacePossum:
Would be great, however while `Twig_TokenParser`'s do state what tags they can handle by the `getTag` method there is no `getEndTag` kind of method. There is also no way to check if it is logical that a `TokenParser` has end tags since there is nothing to distinguish block-kind parsers and no-block kinds.
Happy to introduce this on 2.x line, however I'm not sure if it is worth it?

iltar:
I've encountered this issue a few times as well, it's quite confusing if it's not a pre-defined tag.

SpacePossum:
Is there anything I can/should do here?

SpacePossum:
any interest @fabpot ?

SpacePossum:
nevermind

PR #2009, opened 2016-03-30
Clarify exception message in Twig_Template::getAttribute

chalasr:
When a non existing property/method is called for an object, the exception message is:

>  Method "property()" for object "[class]" does not exist in [template]"

But in fact the `Twig_Template::getAttribute()` method check for (at least) 3 more methods (`isProperty()`, `getProperty()` and `__call__()`).

It could be more adapted and easier to be debugged if the message would be:

> Neither the property "property" nor one of the methods "property]()", "getProperty()"/"isPropertyt()" or "__call()" exist and have public access in class "[class]" in [template]

This message is mostly inspired from the one used in the PropertyAccess component.

BTW I think that it would be great to replace all the checks listed above by a `PropertyAccessor::isReadable([property])`.
I can work on a PR if it's not overkill.

### DIFF
--- a/lib/Twig/Template.php
+++ b/lib/Twig/Template.php
@@ -581,7 +581,7 @@ abstract class Twig_Template implements Twig_TemplateInterface
                 return;
             }
 
-            throw new Twig_Error_Runtime(sprintf('Method "%s" for object "%s" does not exist', $item, get_class($object)), -1, $this->getTemplateName());
+            throw new Twig_Error_Runtime(sprintf('Neither the property "%1$s" nor one of the methods "%1$s()", "get%1$s()"/"is%1$s()" or "__call()" exist and have public access in class "%2$s"', $item, get_class($object)), -1, $this->getTemplateName());
         }
 
         if ($isDefinedTest) {
--- a/ext/twig/twig.c
+++ b/ext/twig/twig.c
@@ -1040,7 +1040,7 @@ PHP_FUNCTION(twig_template_get_attributes)
 				efree(item);
 				return;
 			}
-			TWIG_RUNTIME_ERROR(template TSRMLS_CC, "Method \"%s\" for object \"%s\" does not exist", item, TWIG_GET_CLASS_NAME(object TSRMLS_CC));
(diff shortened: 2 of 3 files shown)


fabpot:
Sounds like a good idea to me. The C extension must also be updated.

chalasr:
@fabpot Thank's for your quick reply.

I'm trying to update the C extension, but I'm facing difficulties to use `ucfirst`.

Is there something specific to do in order to use this php function ? I tried to call `php_ucfirst` insted, unsuccessfully. 
Or should I write a `TWIG_UCFIRST` in pure C ?

stof:
``869 additions, 856 deletions`` in the C code looks wrong to me (and prevents reviewing btw, as github does not display the diff)

lib/Twig/Template.php
```diff
@@ -581,7 +581,7 @@ protected function getAttribute($object, $item, array $arguments = array(), $typ
                 return;
             }
 
-            throw new Twig_Error_Runtime(sprintf('Method "%s" for object "%s" does not exist', $item, get_class($object)), -1, $this->getTemplateName());
+            throw new Twig_Error_Runtime(sprintf('Neither the property "%1$s" nor one of the methods "%1$s()", "get%2$s()"/"is%2$s()" or "__call()" exist and have public access in class "%3$s"', $item, ucfirst($item), get_class($object)), -1, $this->getTemplateName());
```
stof: just use ``'get'.$item`` as used when calling the method. PHP method names are case insensitive anyway

stof: this will help you for the C extension btw, as you won't need to implement ucfirst anymore

chalasr:
@stof 
Use of `ucfirst` was more a matter of convention, I removed it.
The too large diff for the C extension was coming from my editor that unexpectedly replaced tabs by spaces.
Many thanks for your help.

@fabpot Now, it seems to be working well with and without the C extension.

stof:
:+1:

SpacePossum:
:+1:

xabbuh:
:+1:

fabpot:
Thank you @chalasr.

PR #2016, opened 2016-04-17
Promote the include() function instead of the "include" tag

javiereguiluz:
The `{% include %}` tag is pseudo-officially deprecated in favor of the `include()` function. Let's update the documentation to promote the function instead of the tag.


fabpot:
Thank you @javiereguiluz.

PR #2019, opened 2016-04-19
Update Filesystem.php

gustavonovaes:
Add compatibility with Windows


stof:
PHP 5.3+ properly handles using ``/`` as directory separator in paths, even on windows. Are you really still running PHP 5.2 ?

gustavonovaes:
Not really, I'm using 5.6.

I thought this would be my problem. I did some testing after the change and before the pull request, but it seems not solved. My bad.

PR #2033, opened 2016-05-10
Adding support for SORT_LOCALE_STRING flag

cferthorney:
Adds support for SORT_LOCALE_STRING flag on the `asort()` used within twig_sort_filter.

This is proposed to help support correct alphabetical sorted within different locales.  Inspired by http://stackoverflow.com/questions/10273165/how-to-sort-an-array-considering-localization

### DIFF
--- a/lib/Twig/Extension/Core.php
+++ b/lib/Twig/Extension/Core.php
@@ -953,7 +953,7 @@ function twig_sort_filter($array)
         throw new Twig_Error_Runtime(sprintf('The sort filter only works with arrays or "Traversable", got "%s".', gettype($array)));
     }
 
-    asort($array);
+    asort($array, SORT_LOCALE_STRING);
 
     return $array;
 }


stof:
This does not add support for it. It forces using it (which is a BC break btw)

cferthorney:
Should have thought of the BC issue, my bad.  I'll try writing a twig filter for it and use that as a twig extension.

Thanks for taking the time to review the PR though.
Dave

SpacePossum:
There is another way; making the sort order flag an optional argument of the filter which defaults to the current. That way there is no BC and you can still have what you want. The question is, do people want this in core?

cferthorney:
The optional argument sounds good, although as you say is it something people want in core? Locales and internationalisation are such a mine field. I'll work on a separate filter for now and also adjust this PR with optional argument as people can then choose. If enough people want it in core, it should then be workable.

Dave

fabpot:
I'm not sure we want this in core. So, I'm closing for now.

PR #2040, opened 2016-05-13
handle Throwable

fprochazka:
Related #2041


stof:
:+1:

nicolas-grekas:
:+1:

fabpot:
Thank you @fprochazka.

CriseX:
Doesn't this pull request imply that all previous versions of PHP other than PHP7 are now unsupported.

I don't know exactly how the php parser works in this regard, however, intuitively above is true. As long as the Throwable class is not attempted to be loaded unless an exception that inherits said class is thrown the code in this pull request remains backwards compatible, however, if that is not the case then this could break compatibility with all other versions of PHP.

xabbuh:
@CriseX The added lines don't affect execution on older PHP versions.

PR #2043, opened 2016-05-17
Fix load cache

MaCroTux:
In PHP7 @ do not works and it is resource expensive, switching to if statement.

### DIFF
--- a/lib/Twig/Cache/Filesystem.php
+++ b/lib/Twig/Cache/Filesystem.php
@@ -46,7 +46,7 @@ class Twig_Cache_Filesystem implements Twig_CacheInterface
      */
     public function load($key)
     {
-        if (file_exists($key)) include_once $key;
+        if (file_exists($key)) @include_once $key;
     }
 
     /**
--- a/lib/Twig/Cache/Filesystem.php
+++ b/lib/Twig/Cache/Filesystem.php
@@ -46,7 +46,7 @@ class Twig_Cache_Filesystem implements Twig_CacheInterface
      */
     public function load($key)
     {
-        @include_once $key;
+        if (file_exists($key)) include_once $key;
     }
 
     /**


fabpot:
Can you tell us why `@` does not work on PHP7?

lib/Twig/Cache/Filesystem.php
```diff
@@ -46,7 +46,7 @@ public function generateKey($name, $className)
      */
     public function load($key)
     {
-        @include_once $key;
+        if (file_exists($key)) include_once $key;
```
Tobion: this is not safe to race conditions. so the `@` would need to be kept anyway

kidandcat:
Be aware of using error control operator in statements before include() like this:

<?PHP

(@include("file.php"))
OR die("Could not find file.php!");

?>

This cause, that error reporting level is set to zero also for the included file. So if there are some errors in the included file, they will be not displayed.

MaCroTux:
Yes, it works, but @ is captured for exception handler and fill my error logs.

lib/Twig/Cache/Filesystem.php
kidandcat: Could you explain me how would a race condition be affected by that if?

fabpot: Very simple, another process can remove the file between the `file_exists()` call and the `include_once()` one.

kidandcat: The time lapsus between the check and the include is in the nano seconds range.

If you are able to reproduce that condition even once, tell me, I will stop using my computer.

fabpot: You can stop now :)

Take any high-loaded website like symfony.com, and try to clear the cache live. Bam!

Anyway, it looks like your error handler is misbehaving here.

MaCroTux: https://github.com/twigphp/Twig/pull/2043/commits/65a0bd0ff0038352ae170310c8375b65ecd7a28d

kidandcat: Ok, so, I'm a javascript developer, I don't have @, so have a good day, I need to go find a if(exists)openfile substitute for all my projects.

Meanwhile I will begin with the safe and wellcoded try{ openfile }catch(e){}, but don't worry about unknown catched errors, they never happens.

stof:
> It is far better to test for the condition that you know will cause an error before preceding to run the code.

That's true, but only if you can do it without race conditions, which is *not* the case here

lib/Twig/Cache/Filesystem.php
```diff
@@ -46,7 +46,7 @@ public function generateKey($name, $className)
      */
     public function load($key)
     {
-        @include_once $key;
+        if (file_exists($key)) @include_once $key;
```
Tobion: please use multiline if with braces

Tobion:
I'm :+1: to merge this as it's consistent with #1866 when auto_reload = false.

lib/Twig/Cache/Filesystem.php
SpacePossum: this should've been reported by fabbot (https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.11/Symfony/CS/Tests/Fixer/PSR2/BracesFixerTest.php#L166), I wander how far it is behind on the fixer (or if there is an issue with the fixer).

fabpot:
Thank you @MaCroTux.

apapsch:
> Yes, it works, but @ is captured for exception handler and fill my error logs.

Similar issue:
This bug makes Twig unusable for me because of an error handler which dies after it captured the error. This is not constrained to PHP 7.

PR #2047, opened 2016-05-27
Fix codeblock for recipe

barryvdh:
See http://twig.sensiolabs.org/doc/recipes.html#loading-a-template-from-a-string
That is malformed currently.


stof:
this is wrong, as the default language is PHP, and this codeblock is not a PHP one.

stof:
Actually, it looks like the fix done in https://github.com/twigphp/Twig/pull/1978 is not yet deployed.

@fabpot how often are you updating the online version of the Twig doc ?

barryvdh:
Oh sorry. If it's not up to date, that would explain the probleem indeed.

PR #2052, opened 2016-06-02
fix a case where the autoescaping does not work as expected

uwej711:
The Twig_NodeVisitor_Escaper collects a list of blocks for all templates that it visits. If you define the same block (i.e. with the same name) in txt and html templates this results sometimes in the html block not being escapes.

This is illustrated in the added test.

To fix it, I propose to reset the list of the blocks for each module.

Alternatively we need to make clear that blocks should not share names between text and html templates.

### DIFF
--- a/lib/Twig/NodeVisitor/Escaper.php
+++ b/lib/Twig/NodeVisitor/Escaper.php
@@ -58,6 +58,7 @@ class Twig_NodeVisitor_Escaper extends Twig_BaseNodeVisitor
         if ($node instanceof Twig_Node_Module) {
             $this->defaultStrategy = false;
             $this->safeVars = array();
+            $this->blocks = array();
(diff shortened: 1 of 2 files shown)


stof:
``doLeaveNode`` should also reset it then

uwej711:
Yes, maybe resetting there too or as an alternative only there is also possible I think, as with safeVars.

uwej711:
Just to know what to do in my current project where this issue came up: should I rename all my blocks in txt twig templates to be sure or is this (or another fix for this) to be merged and released soon?

Kind regards
Uwe

stof:
Looks good to me. :+1:

fabpot:
Thank you @uwej711.

uwej711:
You're welcome!

Issue #2053, opened 2016-06-03
Error at Twig/Node/Expression/Call.php [ 243 ]


icexdev:
Call to undefined method ReflectionFunction::getDeclaringClass()

PR #2054, opened 2016-06-03
Fix static callable.

SpacePossum:
fixes https://github.com/twigphp/Twig/issues/2053

### DIFF
--- a/lib/Twig/Node/Expression/Call.php
+++ b/lib/Twig/Node/Expression/Call.php
@@ -239,12 +239,12 @@ abstract class Twig_Node_Expression_Call extends Twig_Node_Expression
             if ($argument && $argument->isArray() && $argument->isDefaultValueAvailable() && array() === $argument->getDefaultValue()) {
                 array_pop($parameters);
             } else {
-                $callableName = $r->name;
-                if ($r->getDeclaringClass()) {
-                    $callableName = $r->getDeclaringClass()->name.'::'.$callableName;
-                }
-
-                throw new LogicException(sprintf('The last parameter of "%s" for %s "%s" must be an array with default value, eg. "array $arg = array()".', $callableName, $this->getAttribute('type'), $this->getAttribute('name')));
+                throw new LogicException(sprintf(
+                    'The last parameter of "%s" for %s "%s" must be an array with default value, eg. "array $arg = array()".',
+                    $r instanceof ReflectionMethod ? $r->getDeclaringClass()->name.'::'.$r->name : $r->name,
+                    $this->getAttribute('type'),
+                    $this->getAttribute('name')
+                ));
             }
         }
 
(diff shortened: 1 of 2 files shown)


fabpot:
Can you add some tests to avoid any regressions?

SpacePossum:
I've added the tests that walk through all possible paths of the changed code.
I reverted a smaller change I did earlier as it was not needed.

lib/Twig/Node/Expression/Call.php
```diff
@@ -239,9 +239,10 @@ private function getCallableParameters($callable, $isVariadic)
             if ($argument && $argument->isArray() && $argument->isDefaultValueAvailable() && array() === $argument->getDefaultValue()) {
                 array_pop($parameters);
             } else {
-                $callableName = $r->name;
-                if ($r->getDeclaringClass()) {
-                    $callableName = $r->getDeclaringClass()->name.'::'.$callableName;
+                if ($r instanceof ReflectionMethod) {
+                    $callableName = $r->class.'::'.$r->name;
+                } else {
+                    $callableName = $r->name;
```
fabpot: This should be moved back where it was before. This was done on purpose so that the `$callableName` is always defined (for HHVM/Hack).

SpacePossum: not sure if I follow, `$callableName` will always be defined here. Is there something special with HHVM/Hack related to the previous ` $r->getDeclaringClass()` or something?

stof: @fabpot it is always defined in the new logic

lib/Twig/Node/Expression/Call.php
```diff
@@ -239,9 +239,10 @@ private function getCallableParameters($callable, $isVariadic)
             if ($argument && $argument->isArray() && $argument->isDefaultValueAvailable() && array() === $argument->getDefaultValue()) {
                 array_pop($parameters);
             } else {
-                $callableName = $r->name;
-                if ($r->getDeclaringClass()) {
-                    $callableName = $r->getDeclaringClass()->name.'::'.$callableName;
+                if ($r instanceof ReflectionMethod) {
+                    $callableName = $r->class.'::'.$r->name;
```
stof: this changes the class being displayed

lib/Twig/Node/Expression/Call.php
fabpot: In any case, there is no need to change this part of the code.

SpacePossum: yes there is, if `$r` is not a `ReflectionMethod` calling `getDeclaringClass` on it will result in the error reported

SpacePossum:
ping @fabpot I think this one ready, please let me know if you anything changed :)

fabpot:
Thank you @SpacePossum.

PR #2057, opened 2016-06-07
Ext for Twig 2

jpauli:


ext/twig/php_twig.h
```diff
@@ -2,34 +2,54 @@
    +----------------------------------------------------------------------+
    | Twig Extension                                                       |
    +----------------------------------------------------------------------+
-   | Copyright (c) 2011 Derick Rethans                                    |
+   | Copyright (c) 2011 Sensiolabs                                        |
```
stof: the year may need to be updated

ext/twig/php_twig.h
```diff
@@ -2,34 +2,54 @@
...
    | Author: Derick Rethans <derick@derickrethans.nl>                     |
+   |         Julien PAULI <jpauli@php.net>                                |
    +----------------------------------------------------------------------+
  */
 
 #ifndef PHP_TWIG_H
 #define PHP_TWIG_H
 
-#define PHP_TWIG_VERSION "2.0.0-DEV"
+#define PHP_TWIG_VERSION "2.0.0"
```
stof: this change is wrong. It should keep the ``-DEV`` suffix outside of releases, to be consistent with the way the Twig version number is handled

ext/twig/php_twig.h
```diff
@@ -2,34 +2,54 @@
...
 
 #include "php.h"
 
 extern zend_module_entry twig_module_entry;
-#define phpext_twig_ptr &twig_module_entry
-#ifndef PHP_WIN32
+zend_module_entry *get_module(void);
+
+#ifdef COMPILE_DL_TWIG
 zend_module_entry *get_module(void);
```
stof: is it expected to duplicate the line available just above in case ``COMPILE_DL_TWIG`` is defined ?

ext/twig/php_twig.h
tucksaun: and the name too: Sensio**L**abs :)

test/Twig/Tests/TemplateTest.php
```diff
@@ -492,12 +492,12 @@ class Twig_TemplateArrayAccessObject implements ArrayAccess
 
     public function offsetExists($name)
     {
-        return array_key_exists($name, $this->attributes);
+        return array_key_exists((string)$name, $this->attributes);
```
stof: why changing these tests ?

stof:
@jpauli is this new extension faster than the old one ? If yes, can you share some comparison (worth it would be no ext vs old ext vs new ext)

stof:
@jpauli a few weeks ago, you told me about an implementation of the extension which would not expose a ``twig_get_attribute`` function but hook into the class directly, allowing to keep the same template compilation. Have you abandoned this idea or is it scheduled for a follow-up PR ?

ext/twig/php_twig.h
derickr: Also, you can't just change Copyrights in files.

PR #2058, opened 2016-06-13
Minor rst syntax tweaks

javiereguiluz:
It looks like you can't show `<code>` inside italics with RST. This is how the current docs look when rendered:

rst_syntax_issue


fabpot:
Thank you @javiereguiluz.

PR #2062, opened 2016-06-21
[DOC] add named macro end-tag documentation

JakeFr:
Hello!
Just discover endmacro tag supports name, just like block.
But it is not mentionned in the doc.
So here's a PR


fabpot:
Thank you @JakeFr.

PR #2064, opened 2016-06-22
Allow tests to be chained

pierredup:
This allows multiple tests to be chained, where the original value is used for the tests, and you don't have to explicitly set the value. A common case is to check if a variable is defined and not empty.
So this makes the following possible:

```twig
{% if somevar is defined and not empty %}
   ... do something
{% endif %}
```

Normal cases still works where the value is defined for each test

```twig
{% if somevar is defined and somevar is not empty %}
```

And multiple tests can be chained together

```twig
{% if myvar is even and divisible by(3) and not same as(144) %}
```

### DIFF
--- a/lib/Twig/Extension/Core.php
+++ b/lib/Twig/Extension/Core.php
@@ -295,7 +295,21 @@ class Twig_Extension_Core extends Twig_Extension
             $arguments = $parser->getExpressionParser()->parseArguments(true);
         }
 
-        return new $class($node, $name, $arguments, $parser->getCurrentToken()->getLine());
+        if (!$stream->test(Twig_Token::OPERATOR_TYPE, array('and', 'or'))) {
+            return new $class($node, $name, $arguments, $parser->getCurrentToken()->getLine());
+        }
+
+        $instance = new $class(clone $node, $name, $arguments, $parser->getCurrentToken()->getLine());
+
+        while ($stream->test(Twig_Token::OPERATOR_TYPE, array('and', 'or'))) {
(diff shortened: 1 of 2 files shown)


stof:
this makes things quite complex, because ``and`` can now have 2 different meanings depending of what is used after it (and it can be ambiguous, as ``divisible`` is a valid variable name too)

pierredup:
I'm checking specific for valid test name, and continue as normal if there is no valid test.
E.G the following should both work

```twig
{% if somevar is defined and divisible by(3) %}
{% if somevar is defined and divisible is not empty %}
```

This should still make it clear when the `divisible` test is used vs a variable called `divisible`.

And the `and` (and `or`) works the same as before, where it checks against both tests. The only difference in this case, is that it re-uses the original value specified

stof:
OK, bad example here, as ``divisible`` is not a test name. Replace it with any single-word test name (for instance ``defined``)

sstok:
Would it work if the tests are wrapped in parenthesis?
`{% if somevar is (defined and not empty) %}`

fabpot:
I'm 👎 for this new feature. That does not make a big difference and adds another layer of complexity.

pierredup:
It doesn't make a big different when you have short variable names, but with longer names it becomes more difficult to write and read, E.G

```twig
{% if policy.product_line.start_date is defined and policy.product_line.start_date is not empty %}

OR

{% if policy.product_line.object_id is not empty and policy.product_line.object_id is same as(128)
```

stof:
Well, you can use a local variable to shorten the condition in this case

fabpot:
Closing as I'm still not convinced this should be added to core.

PR #2065, opened 2016-06-24
Update macro.rst

dubaaron:
Small grammar fix: subject-verb agreement "Macros differ" vs "Macros differs"


fabpot:
Good catch, thanks @dubaaron.

PR #2067, opened 2016-06-24
Update parent.rst

dubaaron:
Add "see also" to "use" tag, and example of how it works in that context.

### DIFF
--- a/doc/functions/parent.rst
+++ b/doc/functions/parent.rst
@@ -17,4 +17,22 @@ parent block when overriding a block by using the ``parent`` function:
 The ``parent()`` call will return the content of the ``sidebar`` block as
 defined in the ``base.html`` template.
 
-.. seealso:: :doc:`extends<../tags/extends>`, :doc:`block<../functions/block>`, :doc:`block<../tags/block>`
+``parent()`` can also be used in conjunction with the :doc:`block<../tags/use>` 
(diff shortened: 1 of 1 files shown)


doc/functions/parent.rst
```diff
@@ -17,4 +17,22 @@ parent block when overriding a block by using the ``parent`` function:
 The ``parent()`` call will return the content of the ``sidebar`` block as
 defined in the ``base.html`` template.
 
-.. seealso:: :doc:`extends<../tags/extends>`, :doc:`block<../functions/block>`, :doc:`block<../tags/block>`
+``parent()`` can also be used in conjunction with the :doc:`block<../tags/use>` 
+tag, in which case it seems to pull from the most recently defined parent block:
```
SpacePossum: nice addition! 
I would try to frase without the `seems to` part, as to me it looks like there is some uncertainty in the behavior of Twig while there isn't.

dubaaron: Agreed.  I only wrote "seems to" because there is some uncertainty on my part, and I couldn't find documentation to resolve my uncertainty.  If you know more than me and know that there is no uncertainty in that behavior, then please, remove "seems to".   I wrote "seems to" because that is what I observed, without having more strongly sourced knowledge.

fabpot:
I don't like adding this snippet here as it really belongs to the description of the `use` tag, which already covers this part. So, closing.

Issue #2059, opened 2016-06-13
Template-dependent template loading


wimleers:
We're trying to use `{% extends "image.html.twig" %}` in more places in Drupal 8 core.

Today, Drupal 8's Twig templates sadly still duplicate most of the original template. Using `extends` allows for less repetition, and for bug fixes in base themes to propagated to custom sites, without having to change code.

Unfortunately, we're running into a significant problem, which appears to be an architectural limitation in Twig.

Drupal.org issue: https://www.drupal.org/node/2387069. Below, I've explained it in a way that hopefully removes the need to have any Drupal knowledge, having Twig knowledge is sufficient.

## Drupal's needs.

Drupal's theme system has inheritance. In Drupal 8 core by default, the Bartik theme extends the Classy theme, which extends the Stable theme. Classy and Stable are called "base themes".

This means that extending works:

- *from current theme to the first ancestor theme* when rendering `SOMETHING.html.twig`: Bartik's would extend Classy's which would extend Stable's. Crucially, Classy's may be removed (then Bartik's extends Stable's) and then added again (which then goes back to the original behavior), and we'd expect this to work
- *within the current theme* when rendering a variant: `SOMETHING--variant.html.twig` would extend the current theme's SOMETHING.html.twig, and if that doesn't exist, it'd go and look at the ancestor.

It also means we specifically do *not* want to specify the full path to the base theme's template (nor the namespaced path — we also have registered the necessary namespaces). In other words: `{% extends "path/to/base_theme/SOMETHING.html.twig %}` or `{% extends "@base_theme/SOMETHING.html.twig" %}` is unacceptable.

## Concrete example

1. `core/themes/stable/templates/foo.html.twig`
2. `core/themes/classy/templates/foo.html.twig` contains `{% extends "foo.html.twig" %}` (to add a specific class) and expects Stable's foo.html.twig to be extended
3. `core/themes/bartik/templates/foo.html.twig` contains `{% extends "foo.html.twig" %}` (to override a Twig block) and expects Classy's `foo.html.twig` to be extended — or, really, any of its ancestor themes, because it couldn't care less if Classy's `foo.html.twig` was removed

Notes:

- in point 3, again, we specifically do NOT want to specify `{% extends "@classy/foo.html.twig" %}` because we don't want to care. We just want the parent theme's template — according to the theme registry — to be extended. Think e.g. bug fixes that are added later.
- in point 1, the "root template" could just as well not live in a theme, but in a module. That's also a valid use case.

[...]

PR #2069, opened 2016-06-28
Don't set a fake parent when parsing `{% embed "foo.html.twig." %}`

wimleers:
Context: #2059.

### DIFF
--- a/lib/Twig/TokenParser/Embed.php
+++ b/lib/Twig/TokenParser/Embed.php
@@ -22,18 +22,14 @@ class Twig_TokenParser_Embed extends Twig_TokenParser_Include
 
         list($variables, $only, $ignoreMissing) = $this->parseArguments();
 
-        // inject a fake parent to make the parent() function work
         $stream->injectTokens(array(
             new Twig_Token(Twig_Token::BLOCK_START_TYPE, '', $token->getLine()),
             new Twig_Token(Twig_Token::NAME_TYPE, 'extends', $token->getLine()),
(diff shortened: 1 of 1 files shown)


lib/Twig/TokenParser/Embed.php
```diff
@@ -22,19 +22,15 @@ public function parse(Twig_Token $token)
 
         list($variables, $only, $ignoreMissing) = $this->parseArguments();
 
-        // inject a fake parent to make the parent() function work
         $stream->injectTokens(array(
             new Twig_Token(Twig_Token::BLOCK_START_TYPE, '', $token->getLine()),
             new Twig_Token(Twig_Token::NAME_TYPE, 'extends', $token->getLine()),
-            new Twig_Token(Twig_Token::STRING_TYPE, '__parent__', $token->getLine()),
+            new Twig_Token(Twig_Token::STRING_TYPE, $parent->getAttribute('value'), $token->getLine()),
```
stof: this is broken, because it works only for a ConstantNode, while we support any expression

stof:
the fact that tests are passing tells me that we are missing some tests here (similar to the ones we have for ``extends``)

wimleers:
+1

fabpot:
@wimleers I can try to work on a fix tonight if you cannot find a way to fix it.

wimleers:
@fabpot Feel free to do so — I don't know the Twig codebase well enough.

fabpot:
In #2124, I've made something that works for when the parent is indeed a constant.

Issue #1430, opened 2014-07-04
Report accidently not substituted variables


lathspell:
(Note: The same feature request was reported as #1418 but closed just because some commentor was satisfied not because my feature request was implemented. I improved the example though.)

There seems to be no kind of warning or exception if a rendered template contains variables that are accidentally not substituted.

The "strict_variables" setting from http://twig.sensiolabs.org/doc/api.html#environment-options strict_variables does not help here as it just prevents the opposide case where 
my code would try to substitude an undefined variable like "last_name". It does not warn if I just miss one variable.

Example Template "template.twig":

```
Hello {{ firstname}} {{ lastname }}
```

Example Code:

```
require_once 'Twig/Autoloader.php';
Twig_Autoloader::register();
$twig = new Twig_Environment(new Twig_Loader_Filesystem('.'));
$out = $twig->render("template.twig", array("firstname" => "Tim"));
print_r($out);
```

Expected Output:

```
Exception: Variable "lastname" was not substituted!
```

Actual Output:

```
"Hello Tim "
```

PR #2071, opened 2016-07-02
Notice log show

gymadarasz:
This request related to issue #1430 (and some of more). I think I understand what is this guy (@lathspell) would, the problem is not that the `strict_variables` option doesn't work, I think he need some "Report" instead of an exception. Typical use case when the back-end (php) developer working with an other front-end (html) site builder. The sitebuilder can not add any more ("un-substituted") variable to the template because the back-end developer switched on the `strict_variables` (typically in dev-environment) and the front-end colleague doesn't see nothing html/css/{{ variables }} etc of his work.. (only the Exception Page) but if the `strict_variables` switched Off the back-end colleague doesn't see nothing because there is not exception or any error message only an empty string (so nothing) replaced into the variable placeholder. I wish to open this feature/change request and I attach a change commit to see how I imaging to resolve this "conflict" with some extra development environment arguments for `strict_variables` option. I added (and practically used on debugging) two more option witches related to `strict_variables` and these options working only if the `strict_variables` is switched on (so only in dev-environment):

if `strict_variables` is ON and `strict_variables_notice_log` is ON also, the notices of "undefined" variables added to the `error.log` by `trigger_error` (like to native PHP style) and doesn't throw exception.

if `strict_variables` is ON and `strict_variables_notice_show` is ON also, the notices of "undefined" variables assigned to the template (like as Twig variables) and doesn't throw exception.

if `strict_variables` is ON and both of `strict_variables_notice_log` and `strict_variables_notice_show` also ON, the error going to `error.log` and notice message into Twig template too.

if `strict_variables` is Off or both of `strict_variables_notice_log` and `strict_variables_notice_show` also Off (as default) then nothing to changed everything is working originally.
- so the developers can choose some of more option how they want to use Twig and they dev-environments, potentially differents environment for back-end and front-end colleagues depends on how is they development process (e.g typically template-first or back-end-logic first...)

### DIFF
--- a/lib/Twig/Template.php
+++ b/lib/Twig/Template.php
@@ -437,9 +437,21 @@ abstract class Twig_Template implements Twig_TemplateInterface
         if (!array_key_exists($item, $context)) {
             if ($ignoreStrictCheck || !$this->env->isStrictVariables()) {
                 return;
-            }
+            } else {
+                $errorMessage = sprintf('Variable "%s" does not exist', $item);
+                $templateName = $this->getTemplateName();
+                $lineno = -1;
 
-            throw new Twig_Error_Runtime(sprintf('Variable "%s" does not exist', $item), -1, $this->getTemplateName());
+                if ($log = $this->env->isStrictVariablesNoticeLog() || $show = $this->env->isStrictVariablesNoticeShow()) {
+                    $error = new Twig_Error($errorMessage, $lineno, $templateName);
+                    if ($log) {
+                        trigger_error($error->getMessage(), E_USER_NOTICE);
+                        $show = $this->env->isStrictVariablesNoticeShow();
+                    }
+                    return $show ? 'Notice: ' . $error->getMessage() : null;
+                }
+            }
+            throw new Twig_Error_Runtime($errorMessage, $lineno, $templateName);
         }
 
         return $context[$item];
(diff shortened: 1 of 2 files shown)


lib/Twig/Template.php
```diff
@@ -437,9 +437,22 @@ protected function displayWithErrorHandling(array $context, array $blocks = arra
         if (!array_key_exists($item, $context)) {
             if ($ignoreStrictCheck || !$this->env->isStrictVariables()) {
                 return;
-            }
+            } else {
+                $errorMessage = sprintf('Variable "%s" does not exist', $item);
+                $templateName = $this->getTemplateName();
+                $lineno = -1;
+
+                if ($log = $this->env->isStrictVariablesNoticeLog() || $show = $this->env->isStrictVariablesNoticeShow()) {
```
fabpot: Without parenthesis around the assignments, it does not do what you expect it to.

lib/Twig/Template.php
```diff
@@ -437,9 +437,22 @@ protected function displayWithErrorHandling(array $context, array $blocks = arra
...
+                    $error = new Twig_Error($errorMessage, $lineno, $templateName);
+                    if ($log) {
+                        trigger_error($error->getMessage(), E_USER_NOTICE);
```
fabpot: We never trigger errors like this in Twig or any other projects of mine for that matter.

lib/Twig/Template.php
```diff
@@ -437,9 +437,22 @@ protected function displayWithErrorHandling(array $context, array $blocks = arra
...
+                        $show = $this->env->isStrictVariablesNoticeShow();
```
fabpot: should not be needed.

fabpot:
This implementation cannot be merged as the method does not exist anymore in Twig 2.0, so the feature would be lost.

PR #2075, opened 2016-07-08
[Doc] Hint for replacing strings in variables

MalteWunsch:


stof:
this has nothing specific to the ``replace`` filter. The replace filter expects a hashmap as argument, and using dynamic keys in a hashmap is already documented in the Twig documentation. Duplicating it (with an explanation which is not clear about what happens btw) is a bad idea IMO.

fabpot:
I agree with @stof

MalteWunsch:
@stof Nothing specific here, agreed, I just hadn't found it in the documentation. You're pointing to http://twig.sensiolabs.org/doc/templates.html#literals ?

As I think the issue is still relevant in this context, would you agree to PR adding this link?

PR #2090, opened 2016-08-01
Do not mix indention levels

greg0ire:
Not sure if this was intentional, but the "if" line had an indention of 3 spaces,
while the "true" line has an indention of 4 spaces.


fabpot:
Thank you @greg0ire.

PR #2091, opened 2016-08-04
Update split.rst

ahmadmayahi:
The correct result for `{% set foo = "one,two,three,four,five"|split(',', 3) %}` is:
`['one', 'tow', 'three', 'four,five']`


jvasseur:
It's not, the second parameter of split is the maximum number of elements in the result array.

fabpot:
Current docs are correct. See http://twigfiddle.com/84dzn5

PR #2094, opened 2016-08-10
Exception on object w/o __toString method

WietseWind:
Prevent an ugly 50X error; throw an Syntax Exception when someone tries to display (`{{ xxx }}`) an object that's missing the __toString method.


fabpot:
Can you explain when you have this nasty 500? Because reading the code, it can only happen in a custom escaper, in which case, it's up to the it to reject the value correctly. I don't think it can happen with the built-in escaping strategies.

WietseWind:
@fabpot I checked and I cannot reproduce, although I'm sure we had some troubles with certain objects. Let's forget about this right now ;)

PR #2105, opened 2016-08-29
Floor - use cast and not intval

SpacePossum:
Use cast in fav. of the `inval`, because;
- its faster
- cannot be overloaded
- less chars


lib/Twig/Node/Expression/Binary/FloorDiv.php
```diff
@@ -12,9 +12,9 @@ class Twig_Node_Expression_Binary_FloorDiv extends Twig_Node_Expression_Binary
 {
     public function compile(Twig_Compiler $compiler)
     {
-        $compiler->raw('intval(floor(');
+        $compiler->raw('(int)floor(');
```
fabpot: I would have added a space after the cast.

fabpot:
Thank you @SpacePossum.

PR #2106, opened 2016-08-30
Revert "feature #1757 deprecated Twig_Template::getEnvironment()"

nicolas-grekas:
This reverts commit cfb0593028d164376abb8489b4534404616625f9, reversing
changes made to 1eb5fb26ef5847a5e8c11e24241bccfb54f17ddd.

WDYT? removing getEnvironment led to #1813, and now I'd like to get the filename (when using the Filesystem loader), but I can't.
If OK, I think we should also revert #1813 on master, and maybe also #1807 on 1.x

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,6 @@
-* 1.24.3 (2016-XX-XX)
+* 1.25.0 (2016-XX-XX)
 
- * n/a
+ * reverted deprecation of Twig_Template::getEnvironment()
 
 * 1.24.2 (2016-09-01)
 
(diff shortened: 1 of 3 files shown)


stof:
How is ``->getSource`` related to the deprecation of ``getEnvironment`` ?

nicolas-grekas:
Because when you have the env, you can do `$template->getEnvironment()->getLoader()->getSource($template->getTemplateName())` to get the source.
With this is mind, adding getSource could be seen as a workaround for the removal of getEnvironment.

ninsuo:
👌

fabpot:
I'm :-1:. This is hackish and I don't want to support this. I've fixed the `getSource` issue more elegantly in #2129 and we need to expose some more things, let's do it. But getting the environment to be able to get the loader really looks bad to me.

PR #2109, opened 2016-09-02
Add 'type' tag to avoid calling getAttribute()

superdav42:
This is a proposal for a new tag which could enable some significant optimizations.

``` twig
{% type obj = 'SomeClass' %}
```

With this tag added to a template `{{ obj.id }}` could compile to `$context['obj']->getId()` provided `getId()` was a public accessible method. You can see the unit tests for more examples.

Unfortunately as it's handled by the parser the type of each variable will need to be set in each template. I would recommend its use with variables that are used many times in one template such as loops or templates included in loops.

Twig_Template::getAttribute() is a rather expensive method as many people have experienced. Essentially this will allow us to move the work of getAttribute to compile time instead of render time. I have been this tag in a large project and it has made a significant improvement on performance, especially with variables in large for loops.

### DIFF
--- a/lib/Twig/Environment.php
+++ b/lib/Twig/Environment.php
@@ -34,6 +34,7 @@ class Twig_Environment
     private $tests;
     private $functions;
     private $globals;
+    private $globalSymbols;
     private $runtimeInitialized = false;
     private $extensionInitialized = false;
     private $loadedTemplates;
@@ -483,7 +484,7 @@ class Twig_Environment
     public function getParser()
     {
         if (null === $this->parser) {
-            $this->parser = new Twig_Parser($this);
+            $this->parser = new Twig_Parser($this, $this->globalSymbols);
         }
 
         return $this->parser;
@@ -1028,6 +1029,18 @@ class Twig_Environment
         } else {
             $this->staging->addGlobal($name, $value);
         }
+    }
+
+    /**
+     * Sets the expected class of a global var
+     * Setting this allows the parser to do extra optimizations.
+     *
+     * @param string $name The global variable name
+     * @param string $type The class name
+     */
+    public function addGlobalType($name, $type)
+    {
+        $this->globalSymbols['types'][$name] = array('name' => $type);
     }
 
     /**
--- a/lib/Twig/ExpressionParser.php
+++ b/lib/Twig/ExpressionParser.php
@@ -394,6 +394,49 @@ class Twig_ExpressionParser
 
                 return $node;
             }
+
+            if ($node instanceof Twig_Node_Expression_Name && null !== ($symbol = $this->parser->getImportedSymbol('types', $node->getAttribute('name'))) && false === $this->parser->getEnvironment()->hasExtension('sandbox')) {
+                $class = $symbol['name'];
+
+                if (0 === strcasecmp($class, 'array')) {
+                    $type = Twig_Template::ARRAY_CALL;
+
+                    return new Twig_Node_Expression_GetAttr($node, $arg, $arguments, $type, $lineno);
+                }
+
+                if (!class_exists($class)) {
+                    throw new Twig_Error_Syntax(sprintf('Class "%s" set for %s was not found.', $class, $node->getAttribute('name')), $token->getLine(), $this->parser->getFilename());
+                }
+
+                $name = $arg->getAttribute('value');
+
+                if (Twig_Template::METHOD_CALL !== $type) {
+                    $vars = get_class_vars($class);
+                    if (isset($vars[$name])) {
+                        if ($this->parser->getEnvironment()->hasExtension('sandbox')) {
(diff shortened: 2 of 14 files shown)


stof:
I don't like the fact that this stores types globally.

IMO, this should be implemented without any change to the parser. Optimizing the attribute access should be done by a node visitor, not by the expression parser (btw, this means it can be developed outside the core initially).

Note that I already thought about building such extension. But it is hard to get it right, because many things could invalidate the known types (think about the fact that blocks can be referenced from different places, with a different context for instance)

superdav42:
We certainly don't have to include support for adding global types. The idea was the type of global vars could be set with 
```
$env->addGlobalType('app', 'Symfony\\Bundle\\FrameworkBundle\\Templating\\GlobalVariables');
``` 
inside a extension or similar. This is to avoid having to add this to every template.
```
{% type app = 'Symfony\\Bundle\\FrameworkBundle\\Templating\\GlobalVariables' %}
```

I played with using a node visitor to implement this but ultimately it turned out to be very complicated to get working. Using the parser was must simpler because it already had a working symbol table. If it's something we want to incorporate into the core then the parser seems the most logical place but if we want to keep it outside the core a node visitor would be more appropriate.

stof:
For global vars, the environment already has the value, so it could guess it

stof:
And using the parser is not simpler than usign a node visitor. It is simpler because your code ignores all possible ways to change the value of the variable, making the code unsafe for usage.
even if we incorporate it in the core, it should still be a node visitor (other optimization are also implemented as node visitors)

superdav42:
I'd be willing to rewrite this as a node visitor if we think it would be useful and valuable. 

I did want to clarify what you meant by "many things could invalidate the known types". The way it is designed the type tag specifies what types are expected by the template. Much like type hinting for functions in php. So an exception is thrown if a variable of the wrong type is passed to the template, it is entirely up to the developer to ensure the correct types are used.

stof:
@superdav42 but what about any other assignment in templates ? Compiling by assuming that the type will be correct and then getting a PHP fatal error in the compiled code is not good.

superdav42:
@stof For other assignments I assume you'd only use the type tag to declare types for variables that would never have a different type passed in. Perhaps a use case would help. Say you're listing 100 comment objects on a page with this template:
```twig
{% for comment in comments %}
  {% type comment = 'Acme\\CommentBundle\\Entity\\Comment' %}
  <div>
    <h4>{{ comment.title }} by {{ comment.authorName }}</h4>
    <p> {{ comment.body|raw }}
    </p>
  </div>
{% endfor %}
```
Here 3*100 getAttribute() calls would be converted to a straight method call. In this case the var comments is always expected to be an array of Acme\CommentBundle\Entity\Comment objects. Anything else passed in would have a runtime exception. We could avoid this with a try...catch or a check for method_exists in the compiled code and fallback to getAttribute() if you think it's necessary.

stof:
@superdav42 the hard thing is that your type declaration should apply only until the next place being able to alter the value of the variable. Otherwise, things will break.

try/catch is not an option as it cannot be used inside expressions. It would require having statements.
And adding checks in the compiled code to account for the fact that the variable might have changed since the type declaration (if you don't invalidate it when the variable could have changed) would mean that you would remove your optimization, as you would redo part of the logic of ``getAttribute`` (and btw, ``getAttribute`` received some optimization recently)

superdav42:
@stof can you give an example of where an variable could change type in a template? It seems like a rather obscure use case unless I'm not understanding it correctly.

stof:
@superdav42 at any block boundary, the types become unknown (actually the whole context), as blocks can be rendered from elsewhere. And there are other places with such an impact.

superdav42:
Ok, I've worked on changing this to use a NodeVisitor in a separate repository: 
https://github.com/superdav42/TwigOptimizations
It uses a much different approach that doesn't require specifying types inside the template. 
It records the type used when the template is first rendered, then uses a shutdown function to recompile each template to avoid using getAttribute when possible but adds a fallback if a different type is used in another render. 
Feedback on this would be appreciated.

iltar:
I don't really like having to put things in my template to tell twig what type it is, this is something complex that is not supposed to be in a template. The problem (I think) is that it needs to be compiled in something you only know at run-time and I don't really see how you can know this upfront. Even when I tell twig my var is of a specific type, I can't guarantee this is always the type.

fabpot:
I'm still not convinced either that it's something we need in Twig. It looks wrong to me (and I worked several times on something similar but gave up as I was never satisfied with the end user experience). 

As I've worked on optimizing the performance of method calls and array accesses a lot recently (array accesses are now not calling get_attribute() anymore when possible and method calls are going to be twice as fast in the next 2.x version), I'm going to close this pull request for now.

PR #2120, opened 2016-09-19
ignore template filename in string representation

xabbuh:
Adding the template filename to the node attributes changes the node's string representation. This breaks the behaviour of the `hasElement()` of the `Twig_Node_Expression_Array` class.

I am not sure if this is really the best way to fix it or if we shouldn't rather introduce a different property to store the filename instead.


fabpot:
see #2122

Issue #1939, opened 2015-12-10
Evalutate `needs_environment` and `needs_context` in compiled template, not during compile


SpacePossum:
I ran into a little issue when using different implementations for the same filter and using caching.

When a template gets compiled the filters are evaluated to see if the filters need the context (and/or environment). This causes issues if you reuse the template (using caching) with a different filter (implementation) which has different requirements related to the need for context.

This is a simple test to show the issue;

`test.php`
```php
<?php
require_once __DIR__.'/lib/Twig/Autoloader.php';
Twig_Autoloader::register();
$loader = new Twig_Loader_Filesystem(__DIR__);

$twig = new Twig_Environment($loader,array('cache' => __DIR__.'/cache',));

class Filter1 extends \Twig_SimpleFilter {
    public function __construct()
    {
        parent::__construct('testFilter', function ($value) {
            return '1';
        });
    }
}

$twig->addFilter(new Filter1());

echo $twig->render('test.twig', array('dummy' => 'a'));

$twig = new Twig_Environment($loader,array('cache' => __DIR__.'/cache',));

class Filter2 extends \Twig_SimpleFilter {
    public function __construct()
    {
        parent::__construct('testFilter', function ($contect, $value) {
            return '2';
        }, array('needs_context' => true));
    }
}

$twig->addFilter(new Filter2());

echo $twig->render('test.twig', array('dummy' => 'a'));
```

`test.twig`
```twig
{{ dummy|testFilter }}
```

Ends with 
```
1PHP Warning:  Missing argument 2 for Filter2::{closure}() in /home/possum/work/Twig/test.php on line 34
```

Before I start working on this, does anyone have a good idea how to solve this without reworking https://github.com/twigphp/Twig/blob/1.x/lib/Twig/Node/Expression/Call.php#L23 to much?

stof:
This cannot be done at runtime: these flags are changing the way the template is compiled, as the call needs more arguments. Moving such check at runtime would require to use much less efficient code at runtime when calling a template, which would have an impact on performance.

If you have 2 environments configured differently, sharing the cache folder is likely to produce weird things, as the cache depends on the configuration. And IMO, such weird setup does not warrant making Twig filters and functions slower

stof:
btw, your use case would also mean that auto-escaping checks for filters should be done at runtime (in case they don't have the same safety), and this would simply be impossible (it would basically require doing most of the compilation at runtime, which you can already do if you don't care about performance: disable the cache).

[...]

PR #2121, opened 2016-09-19
[WIP/POC] Caching fails when using array loader with filter

SpacePossum:
I haven't found the root cause yet

### DIFF
--- a/test/Twig/Tests/Loader/ArrayTest.php
+++ b/test/Twig/Tests/Loader/ArrayTest.php
@@ -11,6 +11,28 @@
 
 class Twig_Tests_Loader_ArrayTest extends PHPUnit_Framework_TestCase
 {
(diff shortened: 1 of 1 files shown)


SpacePossum:
The only way to stop Twig from 'caching' the compiled template is to make sure each time the template is compiled the class name is unique. Even if the deprecation of flushing the cache was undone it will stay broken as the templates are already 'cached' as loaded classes. 

Since the loaders do not know if caching is enabled the environment should generate new class names when caching is off.

On a side note, using different names for the templates doesn't help either as the `Twig_Loader_Array` bases the cache key on the _content_ of the template only

SpacePossum:
oh I remember, if you dynamically load filters and/or tests the caching will break the setup and it is all most impossible to solve (https://github.com/twigphp/Twig/issues/1939)

PR #2122, opened 2016-09-19
changed the way we store node filenames

fabpot:


lib/Twig/Node.php
```diff
@@ -229,10 +231,18 @@ public function getIterator()
...
     }
 
-    public function getFilename()
+    public function setFilename($filename)
     {
-        if ($this->hasAttribute('module_filename')) {
-            return $this->getAttribute('module_filename');
+        $this->filename = $filename;
+        foreach ($this->nodes as $node) {
+            if (null !== $node) {
```
stof: can it really be null ? This would break the contract

fabpot: `parent` is such an example on the main Module node. I agree that this is not great, and probably something I will fix in 2.0.

fabpot: Actually, there are more than just `parent` :( I will address that in another PR.

lib/Twig/Node.php
```diff
@@ -229,10 +231,18 @@ public function getIterator()
...
+                $node->setFileName($filename);
```
xabbuh: typo: `setFilename()`

xabbuh:
👍 much better solution

lib/Twig/Node.php
fabpot: see #2123

PR #2123, opened 2016-09-19
deprecated the ability to store non Node instances in Node::$nodes

fabpot:


lib/Twig/Node.php
```diff
@@ -37,6 +37,11 @@ class Twig_Node implements Twig_NodeInterface
      */
     public function __construct(array $nodes = array(), array $attributes = array(), $lineno = 0, $tag = null)
     {
+        foreach ($nodes as $name => $node) {
+            if (!$node instanceof Twig_NodeInterface) {
```
stof: shouldn't you check for Twig_Node as the interface is deprecated ?

stof: The ``getNode`` method already says it returns a ``Twig_Node``, not a ``Twig_NodeInterface``

PR #2125, opened 2016-09-20
fixed Twig cache sharing when used with different versions of PHP

### LINKED ISSUE #2013: Force recompile of cached twig files when PHP version changes
Many cached PHP template files contain code that is specific to the PHP version you are using. This causes fatal errors for example in test sites, which have multiple different PHP versions installed.

Right now `Twig_Cache_Filesystem` has this code:

```
    public function generateKey($name, $className)
    {
        $hash = hash('sha256', $className);
        return $this->directory.$hash[0].$hash[1].'/'.$hash.'.php';
    }
```

Should something like `PHP_VERSION` be added to the key, maybe inside the `$hash`?

fabpot:
closes #2013

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
 * 1.25.0 (2016-XX-XX)
 
+ * fixed Twig cache sharing when used with different versions of PHP
  * removed embed parent workaround for simple use cases
  * deprecated the ability to store non Node instances in Node::$nodes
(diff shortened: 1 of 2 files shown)


mahagr:
Looks good, though there may be Twig extensions out there that optimize code for other PHP versions (PHP7 or takes advantage of some bug fix in older ones).

I think it would be better to create key from `major.minor` version instead as it would give better control for Twig extensions to take advantage of new features in PHP and is forward compatible as well.

stof:
I agree about distinguishing by PHP minor version

PR #2127, opened 2016-09-20
removed usage of realpath in cache key

### LINKED ISSUE #1994: Avoid using realpath in Twig_Loader_Filesystem
As of twig 1.18.0 the findTemplate method from the class Twig_Loader_Filesystem uses internally realpath function to cache the actual path to the target template. However the result of findTemplate method is used as the cache key for given template. This cause an error of mismatching the cache filenames when application is built on separate server and then is deployed to backend. In such a case absolute path on backend and build server is different and actual cache generating will be done the second time after deploying when template will be needed for the first time.

fabpot:
closes #1994

### DIFF
--- a/lib/Twig/Loader/Filesystem.php
+++ b/lib/Twig/Loader/Filesystem.php
@@ -198,11 +198,7 @@ class Twig_Loader_Filesystem implements Twig_LoaderInterface, Twig_ExistsLoaderI
 
         foreach ($this->paths[$namespace] as $path) {
             if (is_file($path.'/'.$shortname)) {
-                if (false !== $realpath = realpath($path.'/'.$shortname)) {
-                    return $this->cache[$name] = $realpath;
-                }
-
-                return $this->cache[$name] = $path.'/'.$shortname;
+                return $this->cache[$name] = $this->normalizePath($path.'/'.$shortname);
             }
         }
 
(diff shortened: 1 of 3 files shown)


lib/Twig/Loader/Filesystem.php
```diff
@@ -257,4 +253,19 @@ protected function validateName($name)
...
     }
+
+    private function normalizePath($path)
+    {
+        $parts = explode('/', str_replace('\\', '/', $path));
+        $new = array();
+        foreach ($parts as $i => $part) {
+            if ('..' === $part) {
+                array_pop($new);
+            } elseif ('.' !== $part && ('' !== $part || $i == 0)) {
```
stloyd: Should be `$i === 0`.

lib/Twig/Loader/Filesystem.php
```diff
@@ -257,4 +253,19 @@ protected function validateName($name)
...
         }
     }
+
+    private function normalizePath($path)
+    {
+        $parts = explode('/', str_replace('\\', '/', $path));
+        $new = array();
+        foreach ($parts as $i => $part) {
+            if ('..' === $part) {
+                array_pop($new);
```
stof: you should check first that ``$new`` is not empty here

fabpot: It cannot be empty because of the call to `validateName`

lib/Twig/Loader/Filesystem.php
fabpot: done

PR #2128, opened 2016-09-20
added support for {{@ ... }}, {%@ ... %}, and {#@ ... #} notation to keep blocks verbatim

### LINKED ISSUE #2032: Convenient tag escaping feature
Hello guys! **It is feature request** and **question**.

As we know batch of modern frameworks use the same ```{{}}``` and ```{%%}``` tags as Twig (Of course it is Twig's and Django templates merit :) ). But it makes usage of Twig with these js frameworks inconvenient. For example i can't use semi-pure angularjs or react with my favorite template engine in Symfony projects. And ```verbatim``` or ```{{ '{{' }} {{ '}}'}}``` is a wild thing. 

Today I found that Blade resolved this issue by ampersand prepending ```@{{ fs }} --> out {{ fs }}```. Looking into Twig source and api didn't help much to understand if it is possible to implement by extending or it is only possible by source code update. 

**The question is**: Is there such implementation I don't know about or is it possible to implement it fast by extending Twig core? Or are there plans to implement it in Twig core? Thanks!

fabpot:
As some JS frameworks use the same block notation as Twig (mainly ``{{ }}``), it is quite inconvenient to mix them with Twig.

This PR adds a new way to escape a Twig block by adding ``@`` just after the opening delimiter (a bit like Blade: https://laravel.com/docs/5.3/blade#blade-and-javascript-frameworks).

closes #2032

What do you think @romainneutron?

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
 * 1.25.0 (2016-XX-XX)
 
+ * added support for {{@ ... }}, {%@ ... %}, and {#@ ... #} notation to keep blocks verbatim
  * changed the way we store template source in template classes
  * removed usage of realpath in cache keys
  * fixed Twig cache sharing when used with different versions of PHP
--- a/doc/templates.rst
+++ b/doc/templates.rst
@@ -480,19 +480,22 @@ Escaping
 --------
 
 It is sometimes desirable or even necessary to have Twig ignore parts it would
-otherwise handle as variables or blocks. For example if the default syntax is
-used and you want to use ``{{`` as raw string in the template and not start a
-variable you have to use a trick.
+otherwise handle as variables or blocks. For example when using a JavaScript
+framework that also uses ``{{`` ``}}`` delimiters.
 
-The easiest way is to output the variable delimiter (``{{``) by using a variable
-expression:
+The easiest way to tell Twig to not "evaluate" some blocks is to add ``@``
+after the opening delimiter (``{{``, ``{%``, or ``{#``):
 
 .. code-block:: jinja
 
-    {{ '{{' }}
+    Hello {{@ firstname }}!
 
 For bigger sections it makes sense to mark a block
 :doc:`verbatim<tags/verbatim>`.
+
+.. versionadded:: 1.25
+    Support for the ``@`` notation was added in Twig 1.25. In previous
+    versions, you can wrap the opening delimiter like this: ``{{ '{{' }}``
 
 Macros
 ------
--- a/lib/Twig/Lexer.php
+++ b/lib/Twig/Lexer.php
@@ -33,6 +33,8 @@ class Twig_Lexer implements Twig_LexerInterface
     protected $positions;
     protected $currentVarBlockLine;
 
+    private $verbatimReplacements;
+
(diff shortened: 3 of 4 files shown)


SpacePossum:
so if I have something like this in a template ATM:
```twig
{{ mail }}@{{ domain }}
```

I've to change my template to escape the new added escape `@`?

stof:
This is indeed a BC break (and a hard to detect one, as it will not trigger any error, but change the output).

I think we could put the ``@`` inside the Twig markup instead, as ``{{@... }}`` and ``{%@... %}`` are invalid in Twig currently (``@`` cannot be used in a Twig identifier). So ``{%@ .... %}`` would output ``{% ... %}``.
This would still be a potential BC break for comments, as they can start with a ``@`` (but this will be true for any char)

fabpot:
I've done the changes, it's now ``{{@``. How does that sound? Is ``@`` the right char?

SpacePossum:
I wonder if this would help, JS template systems like https://mustache.github.io/ or Handlebars would not benefit from this change it I think, but maybe angularjs does (cc @romainneutron)?

Nevertheless it is still a BC break so I would expect it to be on master.2.x, it would be less of an issue if not the comment `{# #}` was effected and AFAIK that one does not need to change for JS frameworks (but I would like a frontenders view on that one).

fabpot:
Which BC break? The fact that existing `{#@ something #}` comments would be transformed? Seems really small to me and not really a problem. Why do you think Mustache would not benefit from this?

SpacePossum:
I was thinking the other way around (IDE/SCA), so no problems, don't mind me.
If the only BC is `{#@` getting a special meaning, I'm all :+1: for this one.
Sorry about the noice :)

hason:
I prefer smaller character than `@` and as pair, e.g.
- `{{< expr >}}`

stof:
@SpacePossum this is indeed the only BC break.

Regarding the choice of a char, I don't know what would be the best one. for BC, we just need something which is not allowed in Twig identifiers and not a unary operator

fabpot:
I would try to avoid enforcing a pair as in a template, you can potentially have many things to escape. As for the char, I don't really care, I'm not a big fan of `@`, but as mentioned by @stof, we have a limited number of choices and Laravel already use this one.

willdurand:
Could it be possible to escape the second opening curly brace? `{\{ js }}`?

beberlei:
Aren't we reinforcing a bad practice here to mix one template language in another?

isyndicate:
Why not just wrap the non-twig content with standard {% verbatim %}{% endverbatim %} for those [hopefully rare] cases where front-end templating languages have to be mixed? Code readability with mixed escape characters is not going to be ideal.

fabpot:
To be clear, I'm not even sure this is needed. I don't have the need myself, but as Blade supports this and because we had an issue about the same topic, I thought we might add something simpler than the `{% verbatim %}` tag. Anyway, looks like this is not that popular, so we can also just close this :)

fabpot:
Ok, closing for now.

fabpot:
What about adding a recipe instead? see #2140

PR #2129, opened 2016-09-20
changed the way we store template source in template classes

### LINKED ISSUE #2011: Issues with Twig_Template::getSource()
How to reproduce:

  * Install the Symfony Demo app (`$ symfony demo`)
  * Browse any page (e.g. `/en/blog/`)
  * Click on the `Show code` button
  * **ERROR** Twig code is empty
  * Reload the page and click again on the `Show code` button
  * **FIXED** Twig code is displayed

@stof thinks this is caused *"because Twig does not load the cache file but evals the code when the cache does not exist yet"*. More details: https://github.com/symfony/symfony-demo/issues/224

fabpot:
In #1807 and #1813, we added a way to get the original template source code from a template. This PR greatly simplifies the approach by removing all the black magic :)

This should also fix #2011

ping @nicolas-grekas

### DIFF
--- a/lib/Twig/Template.php
+++ b/lib/Twig/Template.php
@@ -52,6 +52,13 @@ abstract class Twig_Template implements Twig_TemplateInterface
     abstract public function getDebugInfo();
 
     /**
+     * Returns the template source code.
+     *
+     * @return string The template source code
+     */
+    abstract public function getSource();
+
+    /**
      * @deprecated since 1.20 (to be removed in 2.0)
      */
     public function getEnvironment()
@@ -327,33 +334,6 @@ abstract class Twig_Template implements Twig_TemplateInterface
     public function getBlocks()
     {
         return $this->blocks;
-    }
-
-    /**
-     * Returns the template source code.
-     *
-     * @return string|null The template source code or null if it is not available
-     */
-    public function getSource()
-    {
-        $reflector = new ReflectionClass($this);
-        $file = $reflector->getFileName();
-
-        if (!file_exists($file)) {
-            return;
-        }
-
-        $source = file($file, FILE_IGNORE_NEW_LINES);
-        array_splice($source, 0, $reflector->getEndLine());
-
-        $i = 0;
-        while (isset($source[$i]) && '/* */' === substr_replace($source[$i], '', 3, -2)) {
-            $source[$i] = str_replace('*//* ', '*/', substr($source[$i], 3, -2));
-            ++$i;
-        }
-        array_splice($source, $i);
-
-        return implode("\n", $source);
     }
 
     /**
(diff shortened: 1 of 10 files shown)


lib/Twig/Template.php
```diff
@@ -321,33 +321,6 @@ public function getBlocks()
     }
 
     /**
-     * Returns the template source code.
-     *
-     * @return string|null The template source code or null if it is not available
-     */
-    public function getSource()
```
stof: ``getSource`` should still be defined in this class (to be overwritten by child classes), to make it part of the ``Twig_Template`` API (and to make it part of the blacklisted method names when generating methods dynamically in compiled templates).

In Twig 1.x, we should make it concrete (returning an empty string). In Twig 2.x, it can become an abstract method

lib/Twig/TokenStream.php
```diff
@@ -21,16 +21,20 @@ class Twig_TokenStream
     protected $current = 0;
     protected $filename;
 
+    private $source;
+
     /**
      * Constructor.
      *
      * @param array  $tokens   An array of tokens
      * @param string $filename The name of the filename which tokens are associated with
```
stof: should be changed to ``string|null`` here too

lib/Twig/Template.php
fabpot: Not sure why we need to make it concrete in Twig 1.x.

lib/Twig/TokenStream.php
fabpot: fixed

lib/Twig/Template.php
fabpot: I mean, I understand the BC concern, but I don't think we need to care about it in this case. Nobody is expected to replace `Twig_Node_Module` in such a drastic way. Am I missing something?

fabpot: By the way, I added `getDebugInfo()`, which was not there (see #2132).

stof: Do we know whether some people are doing crazy things by implementing a child class of Twig_Template by hand ? We probably don't care about supporting it though

fabpot: To be on the safe side, I've made it concrete in 1.x and abstract in 2.0

emodric: > Do we know whether some people are doing crazy things by implementing a child class of Twig_Template by hand ? We probably don't care about supporting it though

To reply to @stof's question: Yes, eZ Platform does it in their code to support integration with legacy version which used custom templating engine.

https://github.com/ezsystems/LegacyBridge/blob/master/mvc/Templating/Twig/Template.php

stof: @fabpot ``getDebugInfo`` should also be concrete in 1.x then

mvrhov:
@fabpot Do we need the source in production ready templates? E.g when not in development mode.
IMO this pollutes the OPcache string table.

fabpot:
@mvrhov see #2164

PR #2134, opened 2016-09-21
Expanded definition of empty

servel333:
Based on the [source](https://github.com/twigphp/Twig/blob/f61bc838a97c26e2952af09070d1c4847eeedbe7/lib/Twig/Extension/Core.php#L1405), defined 'empty'. The comment in the code block is too subtle and I didn't notice it until I got to the edit page, and it also does not fully describe the definition of empty based on what the code does.

### DIFF
--- a/doc/tests/empty.rst
+++ b/doc/tests/empty.rst
@@ -4,9 +4,9 @@
 ``empty`` checks if a variable is
  - an empty string
  - an empty array
+ - an empty hash
  - exactly ``false``
  - exactly ``null``
- - an instance of Countable_ and has a count_ of zero
 
 .. code-block:: jinja
 
--- a/doc/tests/empty.rst
+++ b/doc/tests/empty.rst
@@ -1,12 +1,7 @@
 ``empty``
 =========
 
-``empty`` checks if a variable is
- - an empty string
- - an empty array
- - an empty hash
- - exactly ``false``
- - exactly ``null``
+``empty`` checks if a variable is an empty string, an empty array, an empty hash, exactly ``false``, exactly ``null``:
 
 .. code-block:: jinja
 
--- a/doc/tests/empty.rst
+++ b/doc/tests/empty.rst
@@ -1,7 +1,12 @@
 ``empty``
 =========
 
-``empty`` checks if a variable is empty:
+``empty`` checks if a variable is
+ - an empty string
+ - an empty array
+ - exactly ``false``
+ - exactly ``null``
+ - an instance of Countable_ and has a count_ of zero
 
 .. code-block:: jinja
 
@@ -9,3 +14,7 @@
     {% if foo is empty %}
         ...
     {% endif %}
+
+.. _Countable: http://php.net/manual/en/class.countable.php
+.. _count: http://php.net/manual/en/function.count.php
+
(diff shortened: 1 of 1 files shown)


doc/tests/empty.rst
```diff
@@ -1,11 +1,20 @@
 ``empty``
 =========
 
-``empty`` checks if a variable is empty:
+``empty`` checks if a variable is
+ - an empty string
+ - an empty array
+ - exactly ``false``
+ - exactly ``null``
+ - an instance of Countable_ and has a count_ of zero
```
fabpot: This last sentence should be converted into something more Twig-related like: `an empty array or an empty hash`

doc/tests/empty.rst
```diff
@@ -1,11 +1,20 @@
...
 .. code-block:: jinja
 
     {# evaluates to true if the foo variable is null, false, an empty array, or an empty string #}
     {% if foo is empty %}
         ...
     {% endif %}
+
+.. _Countable: http://php.net/manual/en/class.countable.php
+.. _count: http://php.net/manual/en/function.count.php
+
```
fabpot: extra blank line can be removed

doc/tests/empty.rst
servel333: I tried to find a list of what in PHP built-ins implements Countable to be more specific, but I couldn't find a definitive list. It's also worth noting that even though PHP `count` works with a PHP `array`, the way the code in Twig is written it will not catch an empty array with that code, it later checks specifically for equality with an empty array.

I don't mind changing this, but I'm unsure of the correct phrasing that would be useful for Twig users.

fabpot: My proposal does not work?

servel333: An empty array is already mentioned, so I guess just remove "an instance of Countable_ and has a count_ of zero" and add "an empty hash"?

fabpot: Ah sorry, indeed, an empty is already mentioned.

doc/tests/empty.rst
```diff
@@ -1,11 +1,19 @@
 ``empty``
 =========
 
-``empty`` checks if a variable is empty:
+``empty`` checks if a variable is
+ - an empty string
+ - an empty array
+ - an empty hash
+ - exactly ``false``
+ - exactly ``null``
```
fabpot: Last small change is to put everything back on one line to keep things consistent with the other parts of the docs: ``empty`` checks if a variable is an empty string, an empty array, an empty hash, ...

servel333: Makes sense. Done.

servel333: Before the bulleted list clarified that "an instance of Countable_ and has a count_ of zero" go together, but removing that makes CSV format work.

fabpot:
Thank you @servel333.

doc/tests/empty.rst
```diff
@@ -1,11 +1,14 @@
 ``empty``
 =========
 
-``empty`` checks if a variable is empty:
+``empty`` checks if a variable is an empty string, an empty array, an empty hash, exactly ``false``, exactly ``null``:
 
 .. code-block:: jinja
 
     {# evaluates to true if the foo variable is null, false, an empty array, or an empty string #}
```
stof: this comment misses the empty hash

fabpot: I have removed this comment in bba5e1a as it's now redundant

PR #2139, opened 2016-09-24
Improved the error message when a child template defines contents outside parent blocks

javiereguiluz:
### Context

When using template inheritance, you cannot put template contents outside Twig blocks (except some special tags (form_theme), macros, etc.) Sadly, making this mistake is very common when learning Twig.

### Problem

The error message associated to this issue is cryptic and "impossible to understand" for newcomers:

```
A template that extends another one cannot have a body.
```

I recently delivered a Symfony training and nobody was able to understand what the error was. They didn't have the slightest clue. I explained it to them ... and a few days latter they had the same problem and again they couldn't remember what this error was related to.

### Solution

Most of the Twig/Symfony errors are very helpful. Let's improve this one to make it useful too.

Two years ago I tried to fix this error message (see #1396). Let's see if this new PR is accepted.

### DIFF
--- a/lib/Twig/Parser.php
+++ b/lib/Twig/Parser.php
@@ -373,7 +373,7 @@ class Twig_Parser implements Twig_ParserInterface
             (!$node instanceof Twig_Node_Text && !$node instanceof Twig_Node_BlockReference && $node instanceof Twig_NodeOutputInterface)
         ) {
             if (false !== strpos((string) $node, chr(0xEF).chr(0xBB).chr(0xBF))) {
-                throw new Twig_Error_Syntax('The file of a template that extends another one cannot start with a byte order mark (BOM); it must be removed', $node->getLine(), $this->getFilename());
+                throw new Twig_Error_Syntax('A template that extends another one cannot start with a byte order mark (BOM); it must be removed', $node->getLine(), $this->getFilename());
             }
 
             throw new Twig_Error_Syntax('A template that extends another one cannot include contents outside Twig blocks. Did you forget to put the contents inside a {% block %} tag.', $node->getLine(), $this->getFilename());
--- a/lib/Twig/Parser.php
+++ b/lib/Twig/Parser.php
@@ -373,10 +373,10 @@ class Twig_Parser implements Twig_ParserInterface
             (!$node instanceof Twig_Node_Text && !$node instanceof Twig_Node_BlockReference && $node instanceof Twig_NodeOutputInterface)
         ) {
             if (false !== strpos((string) $node, chr(0xEF).chr(0xBB).chr(0xBF))) {
-                throw new Twig_Error_Syntax('A template that extends another one cannot have a body but a byte order mark (BOM) has been detected; it must be removed.', $node->getLine(), $this->getFilename());
+                throw new Twig_Error_Syntax('The file of a template that extends another one cannot start with a byte order mark (BOM); it must be removed', $node->getLine(), $this->getFilename());
             }
 
-            throw new Twig_Error_Syntax('A template that extends another one cannot have a body.', $node->getLine(), $this->getFilename());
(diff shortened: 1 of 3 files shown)


lib/Twig/Parser.php
```diff
@@ -373,10 +373,10 @@ protected function filterBodyNodes(Twig_NodeInterface $node)
             (!$node instanceof Twig_Node_Text && !$node instanceof Twig_Node_BlockReference && $node instanceof Twig_NodeOutputInterface)
         ) {
             if (false !== strpos((string) $node, chr(0xEF).chr(0xBB).chr(0xBF))) {
-                throw new Twig_Error_Syntax('A template that extends another one cannot have a body but a byte order mark (BOM) has been detected; it must be removed.', $node->getLine(), $this->getFilename());
+                throw new Twig_Error_Syntax('The file of a template that extends another one cannot start with a byte order mark (BOM); it must be removed', $node->getLine(), $this->getFilename());
```
SpacePossum: A template is not always a file, I think the message is just as clear without 'The file of a '

lib/Twig/Parser.php
```diff
@@ -373,10 +373,10 @@ protected function filterBodyNodes(Twig_NodeInterface $node)
...
             }
 
-            throw new Twig_Error_Syntax('A template that extends another one cannot have a body.', $node->getLine(), $this->getFilename());
+            throw new Twig_Error_Syntax('A template that extends another one cannot include contents outside Twig blocks. Did you forget to put the contents inside a {% block %} tag.', $node->getLine(), $this->getFilename());
```
SpacePossum: missing question mark at the end?

lib/Twig/Parser.php
javiereguiluz: You are right. I always forget about that. Fixed. Thanks.

lib/Twig/Parser.php
javiereguiluz: You are right. But if I make that change, I get an error because of this test:

```php
$this->assertSame('.', substr($message, strlen($message) - 1), $message, 'Exception message must end with a dot.');
```

SpacePossum: I see and sorry I've put that safe-guard in there once... hmm....

fabpot: Can we change the test to allow sentences to finish by a dot or a question mark?

fabpot: I'm going to make the change in another PR.

lib/Twig/Parser.php
```diff
@@ -373,10 +373,10 @@ protected function filterBodyNodes(Twig_NodeInterface $node)
             (!$node instanceof Twig_Node_Text && !$node instanceof Twig_Node_BlockReference && $node instanceof Twig_NodeOutputInterface)
         ) {
             if (false !== strpos((string) $node, chr(0xEF).chr(0xBB).chr(0xBF))) {
-                throw new Twig_Error_Syntax('A template that extends another one cannot have a body but a byte order mark (BOM) has been detected; it must be removed.', $node->getLine(), $this->getFilename());
+                throw new Twig_Error_Syntax('A template that extends another one cannot start with a byte order mark (BOM); it must be removed', $node->getLine(), $this->getFilename());
             }
 
-            throw new Twig_Error_Syntax('A template that extends another one cannot have a body.', $node->getLine(), $this->getFilename());
+            throw new Twig_Error_Syntax('A template that extends another one cannot include contents outside Twig blocks. Did you forget to put the contents inside a {% block %} tag.', $node->getLine(), $this->getFilename());
```
fabpot: My comment on the other PR is still true. It's not just blocks, but also macros.

lib/Twig/Parser.php
fabpot: see #2143

lib/Twig/Parser.php
fabpot: One can argue that a macro is a block.

fabpot:
Thank you @javiereguiluz.

PR #2140, opened 2016-09-24
added a recipe about how to use Twig and AngularJS together

fabpot:


Koc:
You can add also https://docs.angularjs.org/api/ng/directive/ngBind as one of the variants

doc/recipes.rst
```diff
@@ -515,4 +515,40 @@ From PHP, it's also possible to load a template stored in a string via
...
+
+Still, if you want to use AngularJS and Twig in the same template, there are
+two ways to make it work depending on the amount of AngularJS you need to
+include in your templates:
+
+* Escaping the AngularJS delimiters by wrapping AngularJS sections with the
+  ``{% verbatim %}`` tag or by escaping each delimiter via ``{{ '{{' }}`` and
+  ``{{ '}}' }}``;
+
+* Changing the delimiters of one of the template engine (depending of which
```
javiereguiluz: `of one of the template engine` -> `of one of the template engines`

fabpot:
@Koc Ok, can you propose something? I must admit that I'm not an Angular user, so, I'm not sure how to best describe its usage.

stof:
I would add a note explaining that changing the delimiter is a hard change, as it requires changing all templates in the project (including templates coming from third parties). For instance, Symfony does not let you configure this, as it would break everything.

PR #2141, opened 2016-09-24
added a missing deprecation notice

fabpot:


lib/Twig/Node.php
```diff
@@ -213,6 +213,10 @@ public function getNode($name)
      */
     public function setNode($name, $node = null)
     {
+        if (!$node instanceof Twig_NodeInterface) {
```
stof: the check should be on ``Twig_Node`` instead, as the interface is deprecated and people must extend the base class

PR #2142, opened 2016-09-24
removed code that deal with potential null nodes (not possible anymore)

fabpot:


lib/Twig/NodeTraverser.php
```diff
@@ -70,10 +70,6 @@ public function traverse(Twig_Node $node)
 
     private function traverseForVisitor(Twig_NodeVisitorInterface $visitor, Twig_Node $node = null)
     {
-        if (null === $node) {
```
stof: you forgot to change the method signature

test/Twig/Tests/NodeVisitor/OptimizerTest.php
```diff
@@ -89,10 +89,6 @@ public function getTestsForForOptimizer()
 
     public function checkForConfiguration(Twig_Node $node = null, $target, $withLoop)
     {
-        if (null === $node) {
```
stof: same here about the method signature

lib/Twig/NodeTraverser.php
fabpot: fixed

test/Twig/Tests/NodeVisitor/OptimizerTest.php
fabpot: done

PR #2144, opened 2016-09-24
[WIP] Take into account passed options when generating cache key (closes #1604)

### LINKED ISSUE #1604: Cached template name should depends on enabled extensions and options
Use case: we can disable/enable debug mode on Symfony-based site depends on cookie (for debugging on production sites). And the problem is: twig regenerate templates which uses extensions available only for debug env. So all users without debug cookie got fatal error.

Koc:
Cache depends on options like optimization or debug mode. If we generate different cache files depends on php version or enabled extensions - we can take into account options with wich env was created.

### DIFF
--- a/lib/Twig/Environment.php
+++ b/lib/Twig/Environment.php
@@ -52,6 +52,7 @@ class Twig_Environment
     private $legacyExtensionNames = array();
     private $runtimeLoaders = array();
     private $runtimes = array();
+    private $optionsHash;
 
     /**
      * Constructor.
@@ -117,6 +118,8 @@ class Twig_Environment
         $this->strictVariables = (bool) $options['strict_variables'];
         $this->setCache($options['cache']);
 
+        $this->updateOptionsHash();
+
         $this->addExtension(new Twig_Extension_Core());
         $this->addExtension(new Twig_Extension_Escaper($options['autoescape']));
         $this->addExtension(new Twig_Extension_Optimizer($options['optimizations']));
@@ -158,6 +161,7 @@ class Twig_Environment
     public function setBaseTemplateClass($class)
     {
         $this->baseTemplateClass = $class;
+        $this->updateOptionsHash();
     }
 
     /**
@@ -166,6 +170,7 @@ class Twig_Environment
     public function enableDebug()
     {
         $this->debug = true;
+        $this->updateOptionsHash();
     }
 
     /**
@@ -174,6 +179,7 @@ class Twig_Environment
     public function disableDebug()
     {
         $this->debug = false;
+        $this->updateOptionsHash();
     }
 
     /**
@@ -218,6 +224,7 @@ class Twig_Environment
     public function enableStrictVariables()
     {
         $this->strictVariables = true;
+        $this->updateOptionsHash();
     }
 
     /**
@@ -226,6 +233,7 @@ class Twig_Environment
     public function disableStrictVariables()
     {
         $this->strictVariables = false;
+        $this->updateOptionsHash();
     }
 
     /**
@@ -303,7 +311,10 @@ class Twig_Environment
      *
      *  * The cache key for the given template;
      *  * The currently enabled extensions;
-     *  * Whether the Twig C extension is available or not.
+     *  * Whether the Twig C extension is available or not;
+     *  * PHP version;
+     *  * Twig version;
+     *  * Options with what environment was created.
      *
      * @param string   $name  The name for which to calculate the template class name
      * @param int|null $index The index if it is an embedded template
@@ -312,10 +323,7 @@ class Twig_Environment
      */
     public function getTemplateClass($name, $index = null)
     {
-        $key = $this->getLoader()->getCacheKey($name);
-        $key .= json_encode(array_keys($this->extensions));
-        $key .= function_exists('twig_template_get_attributes');
-        $key .= ':'.PHP_MAJOR_VERSION.':'.PHP_MINOR_VERSION;
+        $key = $this->getLoader()->getCacheKey($name).$this->optionsHash;
 
         return $this->templateClassPrefix.hash('sha256', $key).(null === $index ? '' : '_'.$index);
     }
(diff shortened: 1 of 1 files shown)


lib/Twig/Environment.php
```diff
@@ -44,6 +44,7 @@ class Twig_Environment
     protected $functionCallbacks = array();
     protected $filterCallbacks = array();
     protected $staging;
+    protected $optionsHash;
```
stof: should be private

lib/Twig/Environment.php
```diff
@@ -107,6 +108,8 @@ public function __construct(Twig_LoaderInterface $loader = null, $options = arra
             'optimizations' => -1,
         ), $options);
 
+        // cache can be object so remove it from options hash
+        $this->optionsHash = serialize(array_diff_key($options, array('cache' => null)));
```
stof: this implementation is wrong. All the settings impacting the compilation are mutable, so the hash should be recomputed on mutation

lib/Twig/Environment.php
Koc: heh, it was created as private but after I've changed visibility

lib/Twig/Environment.php
Koc: I haven't know this, should be updated, indeed.

Koc: There is no property for an `optimization` option, for example. But template differs when change this option. I see two approaches:

a. add `options` property with all options and use it instead of `debug`/`autoReload` 
b. add property for optimization or create getter `Twig_Extension_Optimizer::getOptimizattion` and call it every time

fabpot:
I'm still not convinced about this change. Like I was back then. For instance, even if templates are different depending on the optimizations, I don't see why we need to take this into account as whatever the value, it will work on any machine (PHP version being already taken into account). I don't want to vary the cache on everything. We just want to be sure that cache created on a machine for the prod env can be moved to another easily. In dev, whenever you change the options, you need to clear the cache or enable auto_reload/...

Koc:
@fabpot but you have added cache key dependency on enabled extensions... Enabled extensions is so similar to configuration.

Our use case: we have "secret cookie" wich enabled debug on production. After this templates recompiles with some functions and templates cache produces 500 error for all people without "secret cookie". So we need rebuild twig cache again.

stof:
@Koc no. An extension being available or no changes whether a template can be compiled or throw a syntax error. Optimizations being ON or OFF are not changing the behavior of the template (just the speed).

Koc:
@stof, @fabpot the PR was updated, is it acceptable now?

lib/Twig/Environment.php
```diff
@@ -313,6 +315,10 @@ public function getTemplateClass($name, $index = null)
         $key .= json_encode(array_keys($this->extensions));
         $key .= function_exists('twig_template_get_attributes');
         $key .= ':'.PHP_MAJOR_VERSION.':'.PHP_MINOR_VERSION;
+        $key .= ':debug:'.($this->debug ? '1' : '0');
+        $key .= ':charset:'.$this->charset;
```
stof: AFAIK, this is useless: all Twig filters needing the charset are getting it from the environment at runtime. They cannot get it injected at compile-time

lib/Twig/Environment.php
```diff
@@ -313,6 +315,10 @@ public function getTemplateClass($name, $index = null)
...
+        $key .= ':base_template_class:'.$this->baseTemplateClass;
+        $key .= ':strict_variables:'.($this->strictVariables ? '1' : '0');
```
fabpot: Instead of computing this all the time, what about storing the value in the class constructor and just reusing it here? What about adding Twig's version?

stof: the issue is that most of these flags are mutable

Koc: +1 for Twig version. @stof is right, we should recompute this string on calls like `setDebug`/`setAutoReload` etc. String concatenation is not slow imho.

lib/Twig/Environment.php
Koc: @fabpot WDYT about charset? Remove it?

lib/Twig/Environment.php
fabpot: I know that they are mutable and can be changed any time, but honestly, that's not a supported use case. You are supposed to configure first, then use it to compiler/render templates. We can make that explicit on those methods if needed.

lib/Twig/Environment.php
fabpot: Yes, remove

Koc:
The PR was updated

Koc:
And rebased

lib/Twig/Environment.php
```diff
@@ -1490,4 +1500,18 @@ protected function writeCacheFile($file, $content)
     {
         $this->cache->write($file, $content);
     }
+
+    /**
+     * Generates options hash used for template class name.
+     */
```
fabpot: the comment can be removed

lib/Twig/Environment.php
```diff
@@ -314,8 +325,7 @@ public function getTemplateClass($name, $index = null)
     {
         $key = $this->getLoader()->getCacheKey($name);
         $key .= json_encode(array_keys($this->extensions));
```
fabpot: Why not moving extensions as well in the hash? That would avoid the `json_encode` call.

stof: now that keys are class name, we can. json_encode was used before because no char was forbidden in extension names (we even had a test using a snowman as name at some point to test that), and so we could not implode the array with a separator.

fabpot: @Koc Can you make the change?

Koc: Of course, but can you describe a little bit more? `implode(array_keys(...))` and update hash on each `addExtension` call?

Koc:
The PR was updated

fabpot:
Thank you @Koc.

Issue #2147, opened 2016-09-26
[RFC] FQCN extension names?


ro0NL:
Hi,

Any thoughts on referencing extensions by FQCN? Similar as symfony form types are doing..

```php
$env->addExtension(new Twig_Extension_Core());
$env->getExtension(Twig_Extension_Core::class);
```

Imo. the `getName` is causing namespace collisions that can only be solved by creative naming yourself (where we already have a unique class name).

Also see https://github.com/symfony/symfony/pull/20053, this causes symfony to provide `asset` and `intl_bridge` whereas twig-extensions provides `intl`.

fabpot:
Indeed, that's probably a good idea. I will have a look at this.

ro0NL:
Awesome :)

fabpot:
see #2148

PR #2148, opened 2016-09-27
deprecated Twig_ExtensionInterface::getName()

fabpot:
see #2147

### DIFF
--- a/doc/functions/date.rst
+++ b/doc/functions/date.rst
@@ -41,6 +41,9 @@ If no argument is passed, the function returns the current date:
     .. code-block:: php
 
         $twig = new Twig_Environment($loader);
+        $twig->getExtension('Twig_Extension_Core')->setTimezone('Europe/Paris');
+
+        // before Twig 1.26
         $twig->getExtension('core')->setTimezone('Europe/Paris');
 
 Arguments
--- a/lib/Twig/Environment.php
+++ b/lib/Twig/Environment.php
@@ -16,7 +16,7 @@
  */
 class Twig_Environment
 {
-    const VERSION = '1.25.1-DEV';
+    const VERSION = '1.26.0-DEV';
 
     protected $charset;
     protected $loader;
@@ -49,6 +49,7 @@ class Twig_Environment
     private $bcWriteCacheFile = false;
     private $bcGetCacheFilename = false;
     private $lastModifiedExtension = 0;
+    private $legacyExtensionNames = array();
 
     /**
      * Constructor.
@@ -771,29 +772,41 @@ class Twig_Environment
     /**
      * Returns true if the given extension is registered.
      *
-     * @param string $name The extension name
+     * @param string $class The extension class name
      *
      * @return bool Whether the extension is registered or not
      */
-    public function hasExtension($name)
+    public function hasExtension($class)
     {
-        return isset($this->extensions[$name]);
+        if (isset($this->legacyExtensionNames[$class])) {
+            $class = $this->legacyExtensionNames[$class];
+            @trigger_error(sprintf('Referencing the "%s" extension by its name (defined by getName()) is deprecated since 1.26 and will be removed in Twig 2.0. Use the Fully Qualified Extension Class Name instead.', $class), E_USER_DEPRECATED);
+        }
+
+        return isset($this->extensions[ltrim($class, '\\')]);
     }
 
     /**
-     * Gets an extension by name.
+     * Gets an extension by class name.
      *
-     * @param string $name The extension name
+     * @param string $class The extension class name
      *
      * @return Twig_ExtensionInterface A Twig_ExtensionInterface instance
      */
-    public function getExtension($name)
+    public function getExtension($class)
     {
-        if (!isset($this->extensions[$name])) {
-            throw new Twig_Error_Runtime(sprintf('The "%s" extension is not enabled.', $name));
+        if (isset($this->legacyExtensionNames[$class])) {
+            $class = $this->legacyExtensionNames[$class];
+            @trigger_error(sprintf('Referencing the "%s" extension by its name (defined by getName()) is deprecated since 1.26 and will be removed in Twig 2.0. Use the Fully Qualified Extension Class Name instead.', $class), E_USER_DEPRECATED);
         }
 
-        return $this->extensions[$name];
+        $class = ltrim($class, '\\');
+
+        if (!isset($this->extensions[$class])) {
+            throw new Twig_Error_Runtime(sprintf('The "%s" extension is not enabled.', $class));
+        }
+
+        return $this->extensions[$class];
     }
 
     /**
@@ -803,19 +816,25 @@ class Twig_Environment
      */
     public function addExtension(Twig_ExtensionInterface $extension)
     {
-        $name = $extension->getName();
+        $class = get_class($extension);
 
         if ($this->extensionInitialized) {
-            throw new LogicException(sprintf('Unable to register extension "%s" as extensions have already been initialized.', $name));
+            throw new LogicException(sprintf('Unable to register extension "%s" as extensions have already been initialized.', $class));
         }
 
-        if (isset($this->extensions[$name])) {
-            @trigger_error(sprintf('The possibility to register the same extension twice ("%s") is deprecated since version 1.23 and will be removed in Twig 2.0. Use proper PHP inheritance instead.', $name), E_USER_DEPRECATED);
+        $m = new ReflectionMethod($extension, 'getName');
+        $legacyName = 'Twig_Extension' !== $m->getDeclaringClass()->getName() ? $extension->getName() : null;
+
+        if (isset($this->extensions[$class]) || (null !== $legacyName && isset($this->legacyExtensionNames[$legacyName]))) {
+            unset($this->extensions[$this->legacyExtensionNames[$legacyName]], $this->legacyExtensionNames[$legacyName]);
+            @trigger_error(sprintf('The possibility to register the same extension twice ("%s") is deprecated since version 1.23 and will be removed in Twig 2.0. Use proper PHP inheritance instead.', $class), E_USER_DEPRECATED);
         }
 
         $this->lastModifiedExtension = 0;
-
-        $this->extensions[$name] = $extension;
+        if ($legacyName !== $class) {
+            $this->legacyExtensionNames[$legacyName] = $class;
+        }
+        $this->extensions[$class] = $extension;
     }
 
     /**
@@ -831,11 +850,16 @@ class Twig_Environment
     {
         @trigger_error(sprintf('The %s method is deprecated since version 1.12 and will be removed in Twig 2.0.', __METHOD__), E_USER_DEPRECATED);
 
+        if (isset($this->legacyExtensionNames[$name])) {
+            $name = $this->legacyExtensionNames[$name];
+            @trigger_error(sprintf('Referencing the "%s" extension by its name (defined by getName()) is deprecated since 1.26 and will be removed in Twig 2.0. Use the Fully Qualified Extension Class Name instead.', $name), E_USER_DEPRECATED);
+        }
+
         if ($this->extensionInitialized) {
             throw new LogicException(sprintf('Unable to remove extension "%s" as extensions have already been initialized.', $name));
         }
 
-        unset($this->extensions[$name]);
+        unset($this->extensions[ltrim($name, '\\')]);
     }
 
     /**
@@ -853,7 +877,7 @@ class Twig_Environment
     /**
      * Returns all registered extensions.
      *
-     * @return array An array of extensions
+     * @return Twig_ExtensionInterface[] An array of extensions (keys are for internal usage only and should not be relied on)
      */
     public function getExtensions()
     {
--- a/lib/Twig/Extension.php
+++ b/lib/Twig/Extension.php
@@ -76,4 +76,14 @@ abstract class Twig_Extension implements Twig_ExtensionInterface
     {
         return array();
     }
+
+    /**
+     * {@inheritdoc}
+     *
+     * @deprecated since 1.26 (to be removed in 2.0), not used anymore internally
+     */
+    public function getName()
+    {
+        return get_class($this);
+    }
 }
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,6 @@
-* 1.25.1 (2016-XX-XX)
+* 1.26.0 (2016-XX-XX)
 
- * n/a
+ * deprecated Twig_ExtensionInterface::getName()
 
 * 1.25.0 (2016-09-21)
 
--- a/composer.json
+++ b/composer.json
@@ -40,7 +40,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "1.25-dev"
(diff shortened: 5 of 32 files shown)


lib/Twig/Extension.php
```diff
@@ -76,4 +76,12 @@ public function getGlobals()
     {
         return array();
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getName()
+    {
+        return '';
```
xabbuh: This should probably return the FQCN to not break existing third party code that relies on the extension name.

lib/Twig/Environment.php
```diff
@@ -803,18 +814,28 @@ public function getExtension($name)
...
     public function addExtension(Twig_ExtensionInterface $extension)
     {
-        $name = $extension->getName();
+        $name = get_class($extension);
+
+        $m = new ReflectionMethod($extension, 'getName');
+        $legacyName = null;
+        // we don't trigger the deprecation notice for the internal extensions
+        // as people might have extended them (even if this is deprecated and not recommended anymore)
+        if ('Twig_Extension' !== $m->getDeclaringClass()->getName() && 'Twig_Extension_' !== substr($name, 0, 15) && 'Mock_Twig_Extension' !== substr($name, 0, 19)) {
```
HeahDude: Some user extensions may be prefixed with `Twig_Extension` as well, at least we cannot be sure of it, using the full extension names of the core would be safer, wouldn't it?

lib/Twig/Extension.php
HeahDude: @xabbuh Is that needed in the abstract class?

lib/Twig/Environment.php
```diff
@@ -789,6 +795,11 @@ public function hasExtension($name)
      */
     public function getExtension($name)
     {
+        if (isset($this->legacyExtensionNames[$name])) {
+            $name = $this->legacyExtensionNames[$name];
+            @trigger_error(sprintf('Referencing the "%s" extension by its name (defined by getName()) is deprecated since 1.26 and will be removed in Twig 2.0. Use the Fully Qualified Extension Class Name instead.', $name), E_USER_DEPRECATED);
+        }
+
```
ro0NL: Should normalize name here.. `getExtension('\Something');` is equal to `getExtension('Something')`.

Perhaps rename `$name` to `$className` to clarify.

lib/Twig/Environment.php
stof: @HeahDude the core references extensions using their class names. But it must still provide the existing names in extensions for BC reasons. And so we want to avoid deprecations for this case

lib/Twig/Environment.php
```diff
@@ -803,18 +814,28 @@ public function getExtension($name)
...
     {
-        $name = $extension->getName();
+        $name = get_class($extension);
+
+        $m = new ReflectionMethod($extension, 'getName');
+        $legacyName = null;
+        // we don't trigger the deprecation notice for the internal extensions
+        // as people might have extended them (even if this is deprecated and not recommended anymore)
+        if ('Twig_Extension' !== $m->getDeclaringClass()->getName() && 'Twig_Extension_' !== substr($name, 0, 15) && 'Mock_Twig_Extension' !== substr($name, 0, 19)) {
+            @trigger_error(sprintf('Defining the getName() method in the "%s" extension is deprecated since 1.26 and will be removed in 2.0. It\'s not used anymore internally.', $name), E_USER_DEPRECATED);
```
stof: should we really trigger a deprecation when they define the method ? Defining an unused method is harmless, and allows to keep compat with older versions

lib/Twig/Environment.php
```diff
@@ -803,18 +814,28 @@ public function getExtension($name)
...
+            $legacyName = $extension->getName();
+        }
 
         if ($this->extensionInitialized) {
             throw new LogicException(sprintf('Unable to register extension "%s" as extensions have already been initialized.', $name));
         }
 
-        if (isset($this->extensions[$name])) {
-            @trigger_error(sprintf('The possibility to register the same extension twice ("%s") is deprecated since version 1.23 and will be removed in Twig 2.0. Use proper PHP inheritance instead.', $name), E_USER_DEPRECATED);
+        if (null !== $legacyName && isset($this->legacyExtensionNames[$legacyName])) {
```
stof: this change does not prevent registering 2 extensions with the same name anymore, so it reverts an existing deprecation. Is it expected ?

stof: And for a reused legacy name, shouldn't this replace the existing extension ? This will not be the case currently, as it does not unregister it (it only unregisters the legacy alias).

doc/functions/date.rst
```diff
@@ -43,6 +43,9 @@ If no argument is passed, the function returns the current date:
         $twig = new Twig_Environment($loader);
         $twig->getExtension('core')->setTimezone('Europe/Paris');
 
+        // as of Twig 1.26
```
stof: should be moved before the legacy way IMO (and to be consistent with the symfony documentation standards, the deprecated way should be removed entirely)

lib/Twig/Extension.php
xabbuh: Imo yes, otherwise might lead to unexpected behaviour as you don't have to implement the method anymore, but third-party code might still deal with the returned value here (having several extensions returning empty strings might then mess things up).

lib/Twig/Environment.php
HeahDude: Yes but the check is only against the prefix, I was suggesting to test the all name, before triggering a depreciation notice or it won't be for user land prefixed extensions.

fabpot: @HeahDude If you are using MY namespace, you are on your own :) Nobody should be using Twig_* for anything.

HeahDude: ok :)

lib/Twig/Extension.php
fabpot: done

doc/functions/date.rst
fabpot: done

lib/Twig/Environment.php

lib/Twig/Environment.php
fabpot: fair enough, removed

lib/Twig/Environment.php
fabpot: fixed

fabpot: fixed

lib/Twig/Environment.php
ro0NL: `$class` :smile:

fabpot: indeed, fixed

lib/Twig/Environment.php
```diff
@@ -803,19 +816,30 @@ public function getExtension($name)
...
+        // we don't trigger the deprecation notice for the internal extensions
+        // as people might have extended them (even if this is deprecated and not recommended anymore)
+        if ('Twig_Extension' !== $m->getDeclaringClass()->getName() && 'Twig_Extension_' !== substr($class, 0, 15) && 'Mock_Twig_Extension' !== substr($class, 0, 19)) {
+            $legacyName = $extension->getName();
         }
 
-        $this->lastModifiedExtension = 0;
+        if (isset($this->extensions[$class]) || (null !== $legacyName && isset($this->legacyExtensionNames[$legacyName]))) {
+            unset($this->extensions[$this->legacyExtensionNames[$legacyName]], $this->legacyExtensionNames[$legacyName]);
+            @trigger_error(sprintf($legacyName.' The possibility to register the same extension twice ("%s") is deprecated since version 1.23 and will be removed in Twig 2.0. Use proper PHP inheritance instead.', $class), E_USER_DEPRECATED);
```
HeahDude: Is that intended to have `$legacyName` at the beginning of the exception message?

fabpot: very good catch, debug left over, removed now.

lib/Twig/Environment.php
```diff
@@ -803,19 +816,30 @@ public function getExtension($name)
...
+            throw new LogicException(sprintf('Unable to register extension "%s" as extensions have already been initialized.', $class));
         }
 
-        if (isset($this->extensions[$name])) {
-            @trigger_error(sprintf('The possibility to register the same extension twice ("%s") is deprecated since version 1.23 and will be removed in Twig 2.0. Use proper PHP inheritance instead.', $name), E_USER_DEPRECATED);
+        $m = new ReflectionMethod($extension, 'getName');
+        $legacyName = null;
+        // we don't trigger the deprecation notice for the internal extensions
+        // as people might have extended them (even if this is deprecated and not recommended anymore)
+        if ('Twig_Extension' !== $m->getDeclaringClass()->getName() && 'Twig_Extension_' !== substr($class, 0, 15) && 'Mock_Twig_Extension' !== substr($class, 0, 19)) {
```
xabbuh: If I do not miss anything, we can always call `getName()` here as we do no longer trigger a deprecation (the further calls to `getName()` down below could then simply be replaced with `$legacyName` which should always be the FQCN for extensions following the new way.

lib/Twig/Environment.php
```diff
@@ -853,7 +883,7 @@ public function setExtensions(array $extensions)
     /**
      * Returns all registered extensions.
      *
-     * @return array An array of extensions
+     * @return array An array of extensions (keys are for internal usage only and should not be relied on)
```
xabbuh: To make the intent more clear `array` could be replaced with `Twig_ExtensionInterface[]`.

lib/Twig/Environment.php
```diff
@@ -771,29 +772,41 @@ public function initRuntime()
...
+     * @param string $class The extension class name
      *
      * @return bool Whether the extension is registered or not
      */
-    public function hasExtension($name)
+    public function hasExtension($class)
     {
-        return isset($this->extensions[$name]);
+        $class = ltrim($class, '\\');
+        if (isset($this->legacyExtensionNames[$class])) {
```
stof: this should be done before the ltrim IMO, as legacy names are not class names

lib/Twig/Environment.php
```diff
@@ -771,29 +772,41 @@ public function initRuntime()
...
      *
      * @return Twig_ExtensionInterface A Twig_ExtensionInterface instance
      */
-    public function getExtension($name)
+    public function getExtension($class)
     {
-        if (!isset($this->extensions[$name])) {
-            throw new Twig_Error_Runtime(sprintf('The "%s" extension is not enabled.', $name));
+        $class = ltrim($class, '\\');
+        if (isset($this->legacyExtensionNames[$class])) {
```
stof: same here about the ltrim

lib/Twig/Environment.php
stof: We should keep the check for ``'Twig_Extension' !== $m->getDeclaringClass()->getName()`` IMO, as it means that the extension does not have a legacy name in this case. But other checks are indeed useless

lib/Twig/Environment.php
```diff
@@ -803,19 +816,30 @@ public function getExtension($name)
...
 
-        $this->lastModifiedExtension = 0;
+        if (isset($this->extensions[$class]) || (null !== $legacyName && isset($this->legacyExtensionNames[$legacyName]))) {
+            unset($this->extensions[$this->legacyExtensionNames[$legacyName]], $this->legacyExtensionNames[$legacyName]);
+            @trigger_error(sprintf('The possibility to register the same extension twice ("%s") is deprecated since version 1.23 and will be removed in Twig 2.0. Use proper PHP inheritance instead.', $class), E_USER_DEPRECATED);
+        }
 
-        $this->extensions[$name] = $extension;
+        $this->lastModifiedExtension = 0;
+        if ($extension->getName() !== $class) {
```
stof: you could reuse ``$legacyName``

lib/Twig/Environment.php
```diff
@@ -831,6 +855,12 @@ public function removeExtension($name)
     {
         @trigger_error(sprintf('The %s method is deprecated since version 1.12 and will be removed in Twig 2.0.', __METHOD__), E_USER_DEPRECATED);
 
+        $name = ltrim($name, '\\');
```
stof: should be after the legacy name check

lib/Twig/Extension.php
```diff
@@ -76,4 +76,12 @@ public function getGlobals()
     {
         return array();
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getName()
+    {
```
stof: should we mark this method as deprecated, to warn people calling it ?

lib/Twig/Environment.php

lib/Twig/Environment.php
fabpot: indeed, done

lib/Twig/Environment.php

lib/Twig/Environment.php

lib/Twig/Environment.php

lib/Twig/Environment.php

lib/Twig/Extension.php
fabpot: It is marked as deprecated in the interface. Adding a `@deprecated` annotation here would generate deprecation notices when running the tests, that's why I've not added it here.

fabpot:
As the new Github reviews are a mess, just a comment to say that all comments have taken into account. This one is ready IMHO.

lib/Twig/Extension.php
stof: Well, is Twig really calling this method for a class *not* overriding it ? I don't think so. and classes overriding it would *not* trigger a deprecation there

ro0NL:
:tada:

barden42:
Hello,

I run composer update this morning and now I'm getting those errors :

- Referencing the "Twig_Extension_Core" extension by its name (defined by getName()) is deprecated since 1.26 and will be removed in Twig 2.0. Use the Fully Qualified Extension Class Name instead. 
- Referencing the "Symfony\Bridge\Twig\Extension\FormExtension" extension by its name (defined by getName()) is deprecated since 1.26 and will be removed in Twig 2.0. Use the Fully Qualified Extension Class Name instead.

What is impacted by your changes ? Because I don't find what is wrong in my code.

stof:
@barden42 these deprecation warnings are coming from code in Symfony itself. Symfony has already been updated to use the new API, but these changes are not yet part of a release. So you have 2 solutions for now:

- ignore these warnings for a few days/weeks until the next Symfony release (which will make them disappear)
- update Symfony to the dev version (which will get you the update already)

barden42:
Thanks @stof 
My problem is that this error make my builds fail, so I'll update symfony to dev version. What is this version ? 2.8.*@dev ?

xabbuh:
@barden42 If you are currently on Symfony 2.8.x, it is indeed `2.8.*@dev`. Another option is to restrict Twig to `<=1.25` if you prefer to not install not released versions.

xabbuh:
@barden42 By the way, Symfony 2.8.12 has been released in the meantime.

danrot:
It seems as if it is not possible anymore to register the same class as twig extension multiple times... Is there any workaround for that?

We thought about adding a decorator for twig extensions which handles certain functionality, and register them using this decorator twig extension. But since the class name is used now this doesn't seem to be possible anymore. Before it probably was because the `getName` call could be delegated to the inner twig extension, which ended up using different names.

PR #2152, opened 2016-09-28
normalizePath: Do not mangle phar URLs.

apapsch:
Removing consecutive slashes breaks phar URLs. Example:
phar:///foo/bar is turned into phar:/foo/bar.

This change preserves the three slashes if path starts with phar://.


fabpot:
Can you add some unit tests?

fabpot:
Thank you @apapsch.

betzholz:
THX works

PR #2154, opened 2016-09-29
Repeat filter

elpiafo:
I didn't find an easy way to repeat a pattern in Twig, like [str_repeat](http://php.net/str-repeat) PHP function.

I created this basic filter in Twig.

### DIFF
--- a/lib/Twig/Extension/Core.php
+++ b/lib/Twig/Extension/Core.php
@@ -160,7 +160,7 @@ class Twig_Extension_Core extends Twig_Extension
             new Twig_Filter('striptags', 'strip_tags'),
             new Twig_Filter('trim', 'trim'),
             new Twig_Filter('nl2br', 'nl2br', array('pre_escape' => 'html', 'is_safe' => array('html'))),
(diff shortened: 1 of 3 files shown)


lib/Twig/Extension/Core.php
```diff
@@ -160,6 +160,7 @@ public function getFilters()
             new Twig_Filter('striptags', 'strip_tags'),
             new Twig_Filter('trim', 'trim'),
             new Twig_Filter('nl2br', 'nl2br', array('pre_escape' => 'html', 'is_safe' => array('html'))),
+            new Twig_Filter('repeat', 'str_repeat', array('pre_escape' => 'html', 'is_safe' => array('html'))),
```
stof: -1 for pre-escaping and being safe. You should remove both, letting the escaping being done after it.

stof:
I'm not sure this filter belongs to the core. I don't think it solves a common need (and adding it in a project needing it is dead easy)

fabpot:
Thanks for your contribution. I'm going to close this PR without merging as it is a very simple one that does not need to be in the core (not broad enough). Thanks.

PR #2158, opened 2016-10-03
Use cryptographically secure functions, if possible

oittaa:
PHP 7 and [random_compat](https://github.com/paragonie/random_compat) library provide proper `random_bytes` and `random_int` functions. Use those, where possible, by adding two wrapper functions: `twig_random_bytes` and `twig_random_int`

For example:
`hash('sha256', uniqid(mt_rand(), true), false)` -> `bin2hex(twig_random_bytes(32))`
`mt_rand` -> `twig_random_int`

### DIFF
--- a/lib/Twig/Compiler.php
+++ b/lib/Twig/Compiler.php
@@ -266,6 +266,6 @@ class Twig_Compiler
 
     public function getVarName()
     {
-        return sprintf('__internal_%s', hash('sha256', uniqid(mt_rand(), true), false));
+        return sprintf('__internal_%s', bin2hex(twig_random_bytes(32)));
     }
 }
--- a/lib/Twig/Environment.php
+++ b/lib/Twig/Environment.php
@@ -346,7 +346,7 @@ class Twig_Environment
      */
     public function createTemplate($template)
     {
-        $name = sprintf('__string_template__%s', hash('sha256', uniqid(mt_rand(), true), false));
+        $name = sprintf('__string_template__%s', bin2hex(twig_random_bytes(32)));
 
         $loader = new Twig_Loader_Chain(array(
             new Twig_Loader_Array(array($name => $template)),
--- a/lib/Twig/Extension/Core.php
+++ b/lib/Twig/Extension/Core.php
@@ -321,6 +321,48 @@ function twig_cycle($values, $position)
 }
 
 /**
+ * Wrapper for random bytes functions
+ *
+ * @param int               $length The length of the random string that should be returned in bytes.
+ *
+ * @return string Returns a string containing the requested number of random bytes
+ */
+function twig_random_bytes($length) {
+    if (function_exists('random_bytes')) {
+        return random_bytes($length);
+    }
+
+    // Warning: Might not be cryptographically secure!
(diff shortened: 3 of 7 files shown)


fabpot:
I don't understand why we would need "cryptographically secure functions" when generating a cache class. Can you elaborate?

oittaa:
Is there some reason to not use the best available functions?

fabpot:
Please, define "best" in this context. I don't see the value, I don't what you are trying to fix.

oittaa:
You can read from the official PHP documentation that `random_int` gives better values, which are more suitable in many situations.

fabpot:
But here, we don't need better values, we don't need cryptographic secure values.

oittaa:
So you want to provide [random](https://github.com/twigphp/Twig/blob/master/doc/functions/random.rst) function, but insist on using `mt_rand`? 

OK, then there's nothing I can do.

xabbuh:
@oittaa All changes except for those done for the `twig_random()` method are not related to the  exposed `random()` function. I think you should at least revert all those changes.

nicolas-grekas:
Making the functions *conditionally* secure means they are not secure. Since these were never specified as secure, they should not be changed.
If you need secure random numbers, you should create new functions that have this in their contract.
For twig generated files, we just need something unique name; there is no *better* in taking a secure random source to do so.
:-1:

stof:
I agree with @nicolas-grekas. Changing internal usage to use a secure random source brings no benefit at all. And changing the ``random()`` Twig function to be secure **sometimes** (with no way to know whether it was secure or no) is useless as well.

Thus, your codebase is very fragile: all classes were you used ``twig_random_bytes`` in their internal implementation are working only because the Twig_Extension_Core class had been autoloaded before their usage...

sstok:
There are two concerns with this PR:

The name must be only unique within the context it's used in, there is no need to use a heavy crypto system here, in fact this could an anti-pattern as you are using something that is cryptographically secure, while that's not a requirement and thus unnecessarily uses the random pool.

`// Warning: Not cryptographically secure!`
Do I need to explain this? 🙂 

https://github.com/paragonie/random_compat#usability-notice :
> If PHP cannot safely generate random data, this library will throw an Exception. It will never fall back to insecure random data. If this keeps happening, upgrade to a newer version of PHP immediately.

This is a classic example of not understanding crypto security (no offense), when it comes to cryptography you must ensure: the system is secure and error when it's not. You can't fallback or try to repair something by making it less secure, that's just wrong! This is not about usability, it's about security and when that breaks "bad things will happen ™".

But, please do learn more about cryptography (before you try to implement your own):

http://www.cryptofails.com/post/75204435608/write-crypto-code-dont-publish-it
https://paragonie.com/blog (lots of useful articles)
https://www.owasp.org/index.php/Guide_to_Cryptography
https://twitter.com/ircmaxell/status/653950793830297603

I'm just being harsh here as bad security advise (for any reason) is dangerous for newcomers 👍

oittaa:
> unnecessarily uses the random pool.

You're the one who should learn more about crypto before posting.

https://blog.cr.yp.to/20140205-entropy.html

PR #2159, opened 2016-10-04
Typo fix

dragoonis:


dragoonis:
Okay then. Feel free to close.

stof:
I cannot close it. I'm not a collaborator on this repo

fabpot:
Indeed, NUL is the abbreviation for the null character (https://en.wikipedia.org/wiki/Null_character). And that helps make the difference with `null`, the PHP value.

dragoonis:
Makes sense.

PR #2161, opened 2016-10-05
fixed regression on static calls for functions/filters/tests

### LINKED ISSUE #2160: Runtime extension PR #1913 & magic methods
I'm currently experiencing what i believe is a breaking change after upgrading from v1.25.0 to v1.26.0.

Exception that is shown in the application is:
```
An exception has been thrown during the compilation of a template ("Method Illuminate\Auth\AuthManager::user() does not exist") in "layouts/user.twig".
```

I've traced this to the PR #1913, if i replace the `lib/Twig/Node/Expression/Call.php` file with the one from v1.25.0 everything works fine.

At [line 263](https://github.com/twigphp/Twig/pull/1913/files#diff-2089d573c57338c74075cb6b719c3b94R262) a new ReflectionMethod instance is created, but the method doesn't exist (`user` in the `AuthManager` class in this case). This is true, but the method should be proxied [trough a magic `__call` method](https://github.com/illuminate/auth/blob/master/AuthManager.php#L294) to the `guard` where the `user` method resides. This behaviour worked fine in v1.25.0 as it didn't went trough the `reflectCallable` method.

For reference, the `rcrowe/twigbridge` package which includes the extension: https://github.com/rcrowe/TwigBridge/blob/master/src/Extension/Laravel/Auth.php

Have analysed this to the best of my knowledge, but wasn't able to implement a (good) fix.

If you need any more info, please let me know! Thanks.

fabpot:
fixes #2160

What about deprecate this possibility? Using `__call()`/`__staticCall()` means that we don't support named arguments for instance.

### DIFF
--- a/lib/Twig/Node/Expression/Call.php
+++ b/lib/Twig/Node/Expression/Call.php
@@ -219,6 +219,9 @@ abstract class Twig_Node_Expression_Call extends Twig_Node_Expression
     private function getCallableParameters($callable, $isVariadic)
     {
         list($r, $_) = $this->reflectCallable($callable);
+        if (null === $r) {
+            return array();
+        }
 
         $parameters = $r->getParameters();
         if ($this->hasNode('node')) {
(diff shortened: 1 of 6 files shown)


lib/Twig/Node/Expression/Call.php
```diff
@@ -259,14 +259,24 @@ private function reflectCallable($callable)
         }
 
         if (is_array($callable)) {
+            if (!method_exists($callable[0], $callable[1])) {
+                // __call()
+                return;
```
stof: returning null will break the compilation. ``compileCallable`` and ``getCallableParameters`` expect an array with 2 elements

test/Twig/Tests/IntegrationTest.php
```diff
@@ -230,4 +233,22 @@ public function is_multi_word($value)
     {
         return false !== strpos($value, ' ');
     }
+
+    public function __call($method, $arguments)
+    {
+        if ('magicCall' !== $method) {
+            throw new LogicException('Unexpected call to __call');
```
stof: should be ``BadMethodCallException`` here to follow the conventions used elsewhere

fabpot: ok

PR #2163, opened 2016-10-05
default implementation of getDebugInfo() for BC

xabbuh:
eZ updated the code on their side in the meantime (see ezsystems/LegacyBridge#63), but there may be other projects doing similar things (fixes https://github.com/twigphp/Twig/pull/2129#discussion_r80250969).


fabpot:
Thank you @xabbuh.

PR #2164, opened 2016-10-05
removed template source code for generated template classes when debug is disabled

fabpot:


mvrhov:
Thanks.

lib/Twig/Lexer.php
```diff
@@ -136,7 +136,7 @@ public function tokenize($code, $filename = null)
             mb_internal_encoding($mbEncoding);
         }
 
-        return new Twig_TokenStream($this->tokens, $this->filename, $code);
+        return new Twig_TokenStream($this->tokens, $this->filename, $this->env->isDebug() ? $code : '');
```
stof: should it really be removed from the TokenStream itself ? IMO, this check belongs to the compilation step instead

LionsAd: I agree, it breaks my code that worked fine in 1.25.

nicolas-grekas:
See #2169

Issue #2167, opened 2016-10-06
Move isDebug() check to compile time instead of removing the code from the Token_Stream.


LionsAd:
In 1.25.0 / 1.26.0 I could do:

```php
$source = $this->parser->getStream()->getSource();
```

In 1.26.1 this returns an empty string.

I can see that we don't want to pollute templates with the source code, but could we make the source available in the Twig_TokenStream.

Our use case is that in Drupal 8 we want to allow 'drilling' down into templates for which we need the source code at the parser level.

For now I can pin the version to 1.26.0, but it would be really good if the added check was moved to compile time.

stof:
what do you mean by "drilling down" ?

LionsAd:
We do have preprocessors for themeable objects, for which we historically use so called render arrays.

e.g. you might have one that looks like:

```php

$build = [
  '#theme' => 'block',
  '#block' => $block,
];
```

which in the end after the preprocess and rendering steps, will load a template called block.tpl.php.

We have that pretty granular, so we have the same for e.g. 'field'.

Now many themers want to use a template within a parent template:

e.g.

```twig
{{ body }|

<span class="sidebar">
{% inline_template field_image %}
Label: {{ label }}
{% endinline_template %}
</span>
```

While it would be possible to implement that inline template purely with twig blocks (using __internal), that would make it impossible to use features like {% extend %} or other things.

So the idea is to just extract the source out of it and instead use:

```twig
Label: {{ label }}
```

as the inline template (hashed by it's content), which means that the inner contents of the inline template behave as if they were in its own file and included. (e.g. like loadTemplate) with the difference that all the preprocessors have run and post-processors run afterwards, which is important for admin sensitive markup like contextual links.

e.g. my template then looks like:

```php
        echo $this->env->getExtension('Twig_Extension_Sandbox')->ensureToStringAllowed($this->env->getExtension('Drupal\Core\Template\TwigExtension')->escapeFilter($this->env, $this->env->getExtension('Drupal\Core\Template\TwigExtension')->renderInlineTemplate($this->env, $context, (string) $this->renderBlock("inline_template___internal_d57134179531fdfd2158ddcf7e945e3c806a59a11ba7d033752f94e4692d9085", $context, $blocks), (isset($context["field_image"]) ? $context["field_image"] : null)), "html", null, true));

// ...

[...]

PR #2168, opened 2016-10-06
Move check for isDebug() to determine to show source to Parser.

LionsAd:
This might need a new test to ensure isDebug() disables showing source in getSource() method.

Closes: #2167

### DIFF
--- a/lib/Twig/Parser.php
+++ b/lib/Twig/Parser.php
@@ -114,7 +114,7 @@ class Twig_Parser implements Twig_ParserInterface
             throw $e;
         }
 
-        $node = new Twig_Node_Module(new Twig_Node_Body(array($body)), $this->parent, new Twig_Node($this->blocks), new Twig_Node($this->macros), new Twig_Node($this->traits), $this->embeddedTemplates, $this->getFilename(), $stream->getSource());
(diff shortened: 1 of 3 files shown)


test/Twig/Tests/LexerTest.php
```diff
@@ -16,7 +16,7 @@ public function testStreamSource()
         $lexer = new Twig_Lexer($env);
 
         $env->disableDebug();
-        $this->assertSame('', $lexer->tokenize('foo')->getSource());
+        $this->assertSame('foo', $lexer->tokenize('foo')->getSource());
```
stof: testing the difference between the debug and non-debug mode here is useless now.

this needs to be tested in the place where it makes a difference now.

lib/Twig/Parser.php
```diff
@@ -114,7 +114,7 @@ public function parse(Twig_TokenStream $stream, $test = null, $dropNeedle = fals
             throw $e;
         }
 
-        $node = new Twig_Node_Module(new Twig_Node_Body(array($body)), $this->parent, new Twig_Node($this->blocks), new Twig_Node($this->macros), new Twig_Node($this->traits), $this->embeddedTemplates, $this->getFilename(), $stream->getSource());
+        $node = new Twig_Node_Module(new Twig_Node_Body(array($body)), $this->parent, new Twig_Node($this->blocks), new Twig_Node($this->macros), new Twig_Node($this->traits), $this->embeddedTemplates, $this->getFilename(), $this->env->isDebug() ? $stream->getSource() : '');
```
stof: I would not do the check here. I would always pass the source to the node, and check for the debug mode when compiling the node

LionsAd:
Closing in favor of https://github.com/twigphp/Twig/pull/2169

PR #2169, opened 2016-10-06
Move source removal to Module instead of Lexer

nicolas-grekas:
See comments here:
https://github.com/twigphp/Twig/pull/2164/files#r82133239

ping @stof @LionsAd

(side note: should getDebugInfo be conditionally populated also?)

### DIFF
--- a/lib/Twig/Lexer.php
+++ b/lib/Twig/Lexer.php
@@ -136,7 +136,7 @@ class Twig_Lexer implements Twig_LexerInterface
             mb_internal_encoding($mbEncoding);
         }
 
-        return new Twig_TokenStream($this->tokens, $this->filename, $this->env->isDebug() ? $code : '');
+        return new Twig_TokenStream($this->tokens, $this->filename, $code);
     }
 
     protected function lexData()
--- a/lib/Twig/Node/Module.php
+++ b/lib/Twig/Node/Module.php
@@ -94,7 +94,9 @@ class Twig_Node_Module extends Twig_Node
 
         $this->compileDebugInfo($compiler);
 
-        $this->compileGetSource($compiler);
+        if ($compiler->getEnvironment()->isDebug()) {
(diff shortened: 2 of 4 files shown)


stof:
:+1: 

This means that ``getSource`` should not be changed to abstract in master anymore though

LionsAd:
👍 

Looks great to me, I think it would make sense to make getDebugInfo() conditional on isDebug(), too. (as the name says)

And yes both methods could no longer be abstract in 2.x/master.

lib/Twig/Node/Module.php
```diff
@@ -94,7 +94,9 @@ protected function compileTemplate(Twig_Compiler $compiler)
 
         $this->compileDebugInfo($compiler);
 
-        $this->compileGetSource($compiler);
+        if ($compiler->getEnvironment()->isDebug()) {
+            $this->compileGetSource($compiler);
+        }
```
fabpot: I would always compile the `getSource()` method but make it return an empty string if debug is disabled.

lib/Twig/Lexer.php
```diff
@@ -136,7 +136,7 @@ public function tokenize($code, $filename = null)
             mb_internal_encoding($mbEncoding);
         }
 
-        return new Twig_TokenStream($this->tokens, $this->filename, $this->env->isDebug() ? $code : '');
+        return new Twig_TokenStream($this->tokens, $this->filename, $code);
```
fabpot: I would revert this totally and remove the `$code` from `Twig_TokenStream` (this is purely internal and nobody should rely on this).

fabpot:
`getDebugInfo()` must always be populated for two reasons: first, it does not add that much code, then, it is used to give better exception messages (which is also useful in production).

nicolas-grekas:
Closing in favor of #2173

PR #2170, opened 2016-10-06
Added source file in debug info when available

nicolas-grekas:
We need the source file to create IDE links from `$template` objects found in e.g. exceptions'' trace/context.

### DIFF
--- a/lib/Twig/Node/Module.php
+++ b/lib/Twig/Node/Module.php
@@ -384,10 +384,18 @@ class Twig_Node_Module extends Twig_Node
 
     protected function compileDebugInfo(Twig_Compiler $compiler)
     {
+        $debugInfo = array_reverse($compiler->getDebugInfo(), true);
+        $loader = $compiler->getEnvironment()->getLoader();
+        $name = $this->getAttribute('filename');
+
+        if ($loader instanceof Twig_Loader_Filesystem && file_exists($file = $loader->getCacheKey($name))) {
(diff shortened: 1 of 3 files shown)


lib/Twig/Node/Module.php
```diff
@@ -384,10 +384,22 @@ protected function compileIsTraitable(Twig_Compiler $compiler)
 
     protected function compileDebugInfo(Twig_Compiler $compiler)
     {
+        $debugInfo = array_reverse($compiler->getDebugInfo(), true);
+        $env = $compiler->getEnvironment();
+
+        switch (false) {
+            case $env->isDebug():
+            case ($loader = $env->getLoader()) instanceof Twig_Loader_Filesystem:
+            case file_exists($file = $loader->getCacheKey($this->getAttribute('filename'))):
```
stof: This code defining a variable in a case statement to use it in the next one (but only when the condition is false) is totally unreadable.

nicolas-grekas: at least the PHP engine understands me ;)

nicolas-grekas: updated to a more boring CS :)

nicolas-grekas:
I can't see any better way, except maybe [undeprecating `getEnvironment()`](https://github.com/twigphp/Twig/pull/2106).

nicolas-grekas:
Closed in favor of #2182

PR #2176, opened 2016-10-11
Improved Sandbox error handling

kitbs:
On the model of the SecurityNotAllowedFilterError, SecurityNotAllowedFunctionError and SecurityNotAllowedTagError, I added SecurityNotAllowedMethodError and SecurityNotAllowedPropertyError to allow the user to retrieve both the $className and $methodName/$propertyName from the exception.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,6 @@
 * 1.26.2 (2016-XX-XX)
 
- * n/a
+ * improved debugging with Twig_Sandbox_SecurityError exceptions for disallowed methods and properties
 
 * 1.26.1 (2016-10-05)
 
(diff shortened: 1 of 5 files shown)


stof:
Tests must be updated to ensure that the right exception classes are used for these cases

fabpot:
Looks good to me. Can you add some tests? Also, can you add an entry in the CHANGELOG file? Thanks.

kitbs:
Absolutely. Will work on that now. Thanks @fabpot and @stof

kitbs:
I've enhanced the existing tests in Twig_Tests_Extension_SandboxTest to assert the new exception classes and their methods (and the older exception classes, which weren't previously covered). I also split the large test out into several more discrete tests.

Hope this is all okay.

kitbs:
Sorry, removed the PHP 5.5 `Class::class` syntax in favour of plain strings.

fabpot:
LGTM

stof:
:+1:

fabpot:
Thank you @kitbs.

PR #2179, opened 2016-10-14
undeprecate interface for a smooth upgrade path

xabbuh:
The interface should then become deprecated in Twig 2.1 (see https://github.com/symfony/symfony/issues/18427#issuecomment-253293223).

### DIFF
--- a/lib/Twig/ExistsLoaderInterface.php
+++ b/lib/Twig/ExistsLoaderInterface.php
@@ -11,8 +11,6 @@
 
 /**
  * Empty interface for Twig 1.x compatibility.
- *
- * @deprecated to be removed in 3.0
  */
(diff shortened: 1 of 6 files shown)


fabpot:
What about the other deprecations in 2.0? Shouldn't we remove them all?

xabbuh:
@fabpot makes sense, code is updated

fabpot:
I would not add TODOs. The best I think would be to remove everything and submit a PR re-adding everything back just after this one is merged. So that we can easily keep track of everything without polluting 2.0. WDYT?

xabbuh:
Fine for me too. I simply removed these things so that we could just revert this one commit later for 2.1.

fabpot:
Thank you @xabbuh.

PR #2181, opened 2016-10-14
added deprecations for 2.x

fabpot:
To be merged when we feel like there is enough changes for 3.0.

### DIFF
--- a/doc/deprecated.rst
+++ b/doc/deprecated.rst
@@ -60,3 +60,20 @@ Environment
 * As of Twig 2.7, the ``Twig\Environment::getBaseTemplateClass()`` and
   ``Twig\Environment::setBaseTemplateClass()`` methods are deprecated and will
   be removed in Twig 3.0.
+
+Interfaces
+----------
+
+* As of Twig 2.7, the empty ``Twig\Loader\ExistsLoaderInterface`` interface is
+  deprecated and will be removed in Twig 3.0.
+
+* As of Twig 2.7, the ``Twig\Extension\InitRuntimeInterface`` interface is
+  deprecated and will be removed in Twig 3.0.
+
+Miscellaneous
+-------------
+
+* As of Twig 2.7, the ``Twig_SimpleFilter``, ``Twig_SimpleFunction``, and
+  ``Twig_SimpleTest`` empty classes are deprecated and will be removed in Twig
+  3.0. Use ``Twig\TwigFilter``, ``Twig\TwigFunction``, and ``Twig\TwigTest``
+  respectively.
--- a/lib/Twig/SimpleFilter.php
+++ b/lib/Twig/SimpleFilter.php
@@ -16,7 +16,10 @@ use Twig\TwigFilter;
  */
 class_exists(TwigFilter::class);
 
+@trigger_error(sprintf('Using the "Twig_SimpleFilter" class is deprecated since Twig version 2.7, use "Twig\TwigFilter" instead.'), E_USER_DEPRECATED);
+
 if (false) {
+    /** @deprecated since Twig 2.7, use "Twig\TwigFilter" instead */
     final class Twig_SimpleFilter extends TwigFilter
     {
     }
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,12 +1,14 @@
 * 2.7.0 (2019-XX-XX)
 
+ * deprecated Twig\Extension\InitRuntimeInterface
+ * deprecated Twig\Loader\ExistsLoaderInterface
  * deprecated PSR-0 classes in favor of namespaced ones
  * made namespace classes the default classes (PSR-0 ones are aliases now)
  * added Twig\Loader\ChainLoader::getLoaders()
  * removed duplicated directory separator in FilesystemLoader
  * deprecated the "base_template_class" option on Twig\Environment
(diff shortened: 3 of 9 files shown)


lib/Twig/Parser.php
```diff
@@ -259,8 +259,14 @@ public function setMacro($name, Twig_Node_Macro $node)
     {
         $this->macros[$name] = $node;
     }
+
+    /**
+     * @deprecated since 2.0. Will be removed in 3.0. There is no reserved macro names anymore
```
xabbuh: The version number must be bumped to 2.1.

lib/Twig/Parser.php
```diff
@@ -259,8 +259,14 @@ public function setMacro($name, Twig_Node_Macro $node)
...
+     */
     public function isReservedMacroName($name)
     {
+        @trigger_error('The '.__METHOD__.' method is deprecated since version 2.0 and will be removed in 3.0.', E_USER_DEPRECATED);
```
xabbuh: Same here

fabpot:
This one has been rebased and some more deprecations added.

lib/Twig/ExtensionInterface.php
```diff
@@ -33,21 +33,21 @@ public function getNodeVisitors();
...
-     * @return Twig_SimpleFilter[]
+     * @return Twig_Filter[]
      */
     public function getFilters();
 
     /**
      * Returns a list of tests to add to the existing list.
      *
-     * @return Twig_SimpleTest[]
+     * @return Twig_Test[]
```
stof: this should have been part of 2.0 actually

fabpot: yes, I will make another PR for things that can be done on 2.0

fabpot: see #2332

stof:
The issue with these deprecation triggers in 2.1 is that they make it impossible to support 1.x and 2.1 at the same time without deprecation warnings (as it requires using the 1.x classes).

And as Twig 2 requires PHP 7, Symfony will have to keep support for Twig 1.x until Symfony 4.0. Twig 2.1 will be released much earlier than that.

As these old classes are just empty child classes, there is no maintenance overhead in having them. So I suggest delaying the deprecation warning until the work on Twig 3.x really start. If Twig 3 is due in years from now, we can buy us time before requiring all packages providing Twig extensions to do a switch between both classes to choose the one to use without deprecation (switch which is not easy to do as there is no direct class_exists check covering it as the class name is also used in Twig 1.x for a different purpose).

fabpot:
@stof I agree. Having a ready-to-be-merged PR is a good idea anyway as with time, we forget about those things. So, I will keep that one open until we feel the need to merge it.

stof:
Then if it is not merged yet, OK.
Can you add a TODO in the description about updating the version numbers in the diff when merging it for real ?

doc/deprecated.rst
```diff
@@ -5,4 +5,25 @@ This document lists deprecated features in Twig 2.x. Deprecated features are
...
+* As of Twig 2.1, the ``Twig_Parser::isReservedMacroName()`` function is
+  deprecated and will be removed in Twig 3.0. It always returns ``false``
+  anyway as Twig 2 does not have any reserved macro names.
+
+Miscellaneous
+-------------
+
+* As of Twig 2.1, the ``Twig_SimpleFilter``, ``Twig_SimpleFunction``, and
+  ``Twig_SimpleTest`` empty classes are deprecated and will be removed in Twig
+  3.0. Use ``Twig_Filter``, ``Twig_Function``, and ``Twig_Test`` respectively.
```
stof: should use the namespaced names instead (which are compatible with Twig 1.x)

fabpot: Done

doc/deprecated.rst
```diff
@@ -45,3 +45,21 @@ Parser
...
   ``false`` anyway as Twig 2 does not have any reserved macro names.
+
+Interfaces
+----------
+
+* As of Twig 2.1, the empty ``Twig_ExistsLoaderInterface`` /
+  ``Twig\Loader\ExistsLoaderInterface`` interface is deprecated and will be
+  removed in Twig 3.0.
+
+* As of Twig 2.1, the ``Twig_Extension_InitRuntimeInterface`` /
```
stof: the version number should be updated, as it is not 2.1 anymore.

lib/Twig/SimpleFilter.php
```diff
@@ -11,6 +11,8 @@
 
 /**
  * For Twig 1.x compatibility.
+ *
+ * @deprecated to be removed in 3.0
```
stof: this is the wrong place for it. It should be added on the class definition below.

lib/Twig/ExtensionSet.php
```diff
@@ -452,16 +454,28 @@ private function initExtension(Twig_ExtensionInterface $extension)
     {
         // filters
         foreach ($extension->getFilters() as $filter) {
+            if ($filter instanceof Twig_SimpleFilter) {
```
stof: Does this actually work, given that `Twig_SimpleFilter` is implemented as a class alias in 2.x ?

fabpot:
New version, updated based on the latest changes.

PR #2182, opened 2016-10-15
added Twig_Source to hold information about the original template

fabpot:
Alternative to #2170

### DIFF
--- a/lib/Twig/Loader/Chain.php
+++ b/lib/Twig/Loader/Chain.php
@@ -14,7 +14,7 @@
  *
  * @author Fabien Potencier <fabien@symfony.com>
  */
-class Twig_Loader_Chain implements Twig_LoaderInterface, Twig_ExistsLoaderInterface
+class Twig_Loader_Chain implements Twig_LoaderInterface, Twig_ExistsLoaderInterface, Twig_SourceContextLoaderInterface
 {
     private $hasSourceCache = array();
     protected $loaders = array();
@@ -61,6 +61,35 @@ class Twig_Loader_Chain implements Twig_LoaderInterface, Twig_ExistsLoaderInterf
         }
 
         throw new Twig_Error_Loader(sprintf('Template "%s" is not defined%s.', $name, $exceptions ? ' ('.implode(', ', $exceptions).')' : ''));
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getSourceContext($name)
+    {
+        $exceptions = array();
+        foreach ($this->loaders as $loader) {
+            if (!$loader instanceof Twig_SourceContextLoaderInterface) {
+                continue;
+            }
+
+            if ($loader instanceof Twig_ExistsLoaderInterface && !$loader->exists($name)) {
+                continue;
+            }
+
+            try {
+                return $loader->getSourceContext($name);
+            } catch (Twig_Error_Loader $e) {
+                $exceptions[] = $e->getMessage();
+            }
+        }
+
+        if ($exceptions) {
+            throw new Twig_Error_Loader(sprintf('Template "%s" is not defined%s.', $name, $exceptions ? ' ('.implode(', ', $exceptions).')' : ''));
+        }
+
+        return new Twig_Source($this->getSource($name), $name);
     }
 
     /**
--- a/lib/Twig/Loader/Filesystem.php
+++ b/lib/Twig/Loader/Filesystem.php
@@ -14,7 +14,7 @@
  *
  * @author Fabien Potencier <fabien@symfony.com>
  */
-class Twig_Loader_Filesystem implements Twig_LoaderInterface, Twig_ExistsLoaderInterface
+class Twig_Loader_Filesystem implements Twig_LoaderInterface, Twig_ExistsLoaderInterface, Twig_SourceContextLoaderInterface
 {
     /** Identifier of the main namespace. */
     const MAIN_NAMESPACE = '__main__';
@@ -129,6 +129,16 @@ class Twig_Loader_Filesystem implements Twig_LoaderInterface, Twig_ExistsLoaderI
     public function getSource($name)
     {
         return file_get_contents($this->findTemplate($name));
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getSourceContext($name)
+    {
+        $path = $this->findTemplate($name);
+
+        return new Twig_Source(file_get_contents($path), $name, $path);
     }
 
     /**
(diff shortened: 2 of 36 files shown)


lib/Twig/Node/Module.php
```diff
@@ -40,8 +47,11 @@ public function __construct(Twig_NodeInterface $body, Twig_Node_Expression $pare
 
         // embedded templates are set as attributes so that they are only visited once by the visitors
         parent::__construct($nodes, array(
-            'source' => $source,
-            'filename' => $filename,
+            'source' => $source->getCode(),
+            // to be renamed in 2.0 to name
```
nicolas-grekas: then let's add name right now?

lib/Twig/Node/Module.php
```diff
@@ -40,8 +47,11 @@ public function __construct(Twig_NodeInterface $body, Twig_Node_Expression $pare
...
+            'filename' => $source->getName(),
+            // to be renamed in 2.0 to filename
```
nicolas-grekas: but that would be a hard break, no continuous upgrade path

lib/Twig/Loader/Filesystem.php
```diff
@@ -14,7 +14,7 @@
  *
  * @author Fabien Potencier <fabien@symfony.com>
  */
-class Twig_Loader_Filesystem implements Twig_LoaderInterface, Twig_ExistsLoaderInterface
+class Twig_Loader_Filesystem implements Twig_LoaderInterface, Twig_ExistsLoaderInterface, Twig_SourceContextLoaderInterface
```
stof: the chain loader should implement Twig_SourceContextLoaderInterface too, otherwise we loose the capabilities when using multiple loaders

lib/Twig/Template.php
```diff
@@ -65,6 +65,16 @@ public function getSource()
     }
 
     /**
+     * Returns information about the original template source code.
+     *
+     * @return Twig_Source
+     */
+    public function getSourceContext()
+    {
+        return new Twig_Source('');
```
stof: ``$this->getTemplateName()`` should be passed as second argument here

lib/Twig/Source.php
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * @author Fabien Potencier <fabien@symfony.com>
+ */
+class Twig_Source
+{
+    private $code;
+    private $name;
+    private $filename;
+
+    public function __construct($code, $name = null, $filename = null)
```
stof: please add phpdoc to document the variable types

lib/Twig/SourceContextLoaderInterface.php
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * (c) 2016 Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+interface Twig_SourceContextLoaderInterface
+{
+    /**
```
stof: missing phpdoc for the argument

lib/Twig/TokenStream.php
```diff
@@ -32,9 +32,19 @@ class Twig_TokenStream
      */
     public function __construct(array $tokens, $filename = null, $source = null)
     {
+        if (!$filename instanceof Twig_Source) {
+            if (null !== $filename || null !== $source) {
+                @trigger_error(sprintf('Passing a string as the $filename argument of %s() is deprecated since version 1.27. Pass a Twig_Source instance instead.', __METHOD__), E_USER_DEPRECATED);
+            }
+            $this->source = new Twig_Source($filename, $source);
```
stof: this is wrong. the first argument of ``Twig_Source`` is the code

lib/Twig/TokenParser/AutoEscape.php
```diff
@@ -39,7 +39,7 @@ public function parse(Twig_Token $token)
         } else {
             $expr = $this->parser->getExpressionParser()->parseExpression();
             if (!$expr instanceof Twig_Node_Expression_Constant) {
-                throw new Twig_Error_Syntax('An escaping strategy must be a string or a bool.', $stream->getCurrent()->getLine(), $stream->getFilename());
+                throw new Twig_Error_Syntax('An escaping strategy must be a string or a bool.', $stream->getCurrent()->getLine(), $this->parser->getFilename());
```
stof: why not using ``$stream->getSourceContext()->getName()`` directly to avoid 1 level in the stack each time ? the parser will forward the call to the stream anyway

lib/Twig/TokenStream.php
stof: Please also add legacy tests covering the deprecated APIs. They would catch such mistakes (and forbid breaking BC by mistake)

fabpot: fixed, tests coming

lib/Twig/SourceContextLoaderInterface.php
fabpot: fixed

lib/Twig/Source.php
fabpot: fixed

lib/Twig/Loader/Filesystem.php
fabpot: done

fabpot:
Rebased on top of #2183, reviewing the second commit only gives you less noise. Renamed `getFilename()` to get `getPath()`. I think it's much better that way.

fabpot:
Tests added, not WIP anymore

lib/Twig/Loader/Chain.php
```diff
@@ -66,6 +66,35 @@ public function getSource($name)
     /**
      * {@inheritdoc}
      */
+    public function getSourceContext($name)
+    {
+        $exceptions = array();
+        foreach ($this->loaders as $loader) {
+            if (!$loader instanceof Twig_SourceContextLoaderInterface) {
+                continue;
```
stof: this is broken in case this loader was the one loading the template, but a next loader could load a template with the same name

fabpot: fixed in #2192

marc1706:
Causes a BC break when upgrading from 1.26 and prior to twig 1.27 as described here:
https://github.com/twigphp/Twig/commit/6c76583e23537b2503ccd555d053c09a8250b8ac#commitcomment-19686237

xabbuh:
@stof But that doesn't mean that someone could have implemented the interface before which will no indeed break their code when updating to Twig 1.27.

marc1706:
My reply to @stof 's comment:
https://github.com/twigphp/Twig/commit/6c76583e23537b2503ccd555d053c09a8250b8ac#commitcomment-19693147

Should probably continue the discussion here as that's more accessible to other users than having the discussion in the commit.

PR #2183, opened 2016-10-17
deprecated Parser::getFilename()

fabpot:
The name is misleading as it returns the logical template name and not the proper filesystem path. In #2182, this is then replaced with a proper object. This PR simplifies the code review for #2182 by avoiding the pollution of all the changes in the token parser classes.

### DIFF
--- a/lib/Twig/Parser.php
+++ b/lib/Twig/Parser.php
@@ -52,8 +52,13 @@ class Twig_Parser implements Twig_ParserInterface
         return sprintf('__internal_%s', hash('sha256', uniqid(mt_rand(), true), false));
     }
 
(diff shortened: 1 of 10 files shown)


lib/Twig/Parser.php
```diff
@@ -52,8 +52,13 @@ public function getVarName()
         return sprintf('__internal_%s', hash('sha256', uniqid(mt_rand(), true), false));
     }
 
+    /**
+     * @deprecated since 1.20 (to be removed in 2.0). Use $parser->getStream()->getFilename() instead.
```
xabbuh: since 1.27

PR #2192, opened 2016-10-19
fixed Twig_Loader_Chain logic

fabpot:


stof:
@fabpot should we force loaders to implement ``getSourceContext`` in 2.x, as we did already for ``exists`` ?

fabpot:
I thought about that, but I'm not sure if it is worth it. I would prefer to deprecate `getSource()` and only keep `getSourceContext()` in 2.0 as both are quite redundant. I'm not sure if `getSource()` usage is widespread though. WDYT?

stof:
Well, we need to think about the migration path for custom loaders.

Once we have ``getSourceContext`` in all loaders, this is indeed quite redundant to keep ``getSource`` (as it ends up being ``getSourceContext()->getCode()``).

Btw, if we want to go this way, I think the template name should not be optional (and should not allow null) in ``Twig_Source``

fabpot:
The template name is optional as sometimes (mostly in tests), you don't need a template name.

stof:
@fabpot a template cannot be compiled without a name. So making it optional would allow loaders to return broken object if they don't provide the name in the object they build.
AFAIK, the only place not *needing* a name are unit tests of the parser level, at the condition that they never need ``$parser->getFilename()`` and related APIs to access the template name (and most parsers need it, even though tests don't fail because there is no scalar typehint forbidding to pass ``null`` in the ``string`` argument)

fabpot:
see #2197

PR #2194, opened 2016-10-20
deprecated not supported behavior

fabpot:


stof:
AFAIK, this break BC for loaders extending ``findTemplate`` without adding support for ``false`` as second argument, which is why it was done this way in the first place.

So this should be done only in 2.x (but we may trigger a deprecation warning if we enter the catch block)

fabpot:
Updated this PR with the deprecation instead

lib/Twig/Loader/Filesystem.php
```diff
@@ -163,6 +163,8 @@ public function exists($name)
         try {
             return false !== $this->findTemplate($name, false);
         } catch (Twig_Error_Loader $exception) {
+            @trigger_error(sprintf('In %s::findTemplate(), you must accept a second argument that when set to "false" returns "false" instead of throwing an exception. Not supporting this argument is deprecated since version 1.27.', __CLASS__), E_USER_DEPRECATED);
```
stof: you should use ``get_class($this)`` instead of ``__CLASS__`` (or detect the declaring class of ``findTemplate`` to be even more precise).
Otherwise, the message says that the issue is in Twig_Loader_Filesystem while it is actually in a child class

PR #2195, opened 2016-10-20
fixed the filesystem loader with relative paths

### LINKED ISSUE #2145: BC break regarding relative paths in 1.25 due to realpath changes (#2127)
The 1.25 version does not allow relative loading paths ($this->paths) in the filesystem loader anymore.

The new `normalizePath()` method is more strict than `realpath()` as it does not allow relative paths (any extra `../` will be ignored). It means that if you have a FilesystemLoader configured with a relative directory in the paths list, the result of `findTemplate()` will be different than in the 1.24 version.

The result will also be wrong because the `is_file()` check (https://github.com/twigphp/Twig/blob/1.x/lib/Twig/Loader/Filesystem.php#L200) will succeed but the returned path won't lead to an existing file, as a result a warning will be thrown by `getSource()`.

Reproducer (expecting true with both 1.24 and 1.25): 
```
<?php

include 'vendor/autoload.php';

$parts = explode('/', __FILE__);
$filename = array_pop($parts);
$dirname = array_pop($parts);

$loader = new Twig_Loader_Filesystem('../'.$dirname);
$r = new ReflectionMethod($loader, 'findTemplate');
$r->setAccessible(true);
var_dump(__FILE__ === $r->invoke($loader, $filename));
```

fabpot:
closes #2145 

TODO

 - [x] document the new `$rootPath` option
 - [x] add test with a `$rootPath` different from `getcwd()`
 - [x] test the cache key does not vary when `$rootPath` changes

### DIFF
--- a/lib/Twig/Loader/Filesystem.php
+++ b/lib/Twig/Loader/Filesystem.php
@@ -23,13 +23,21 @@ class Twig_Loader_Filesystem implements Twig_LoaderInterface, Twig_ExistsLoaderI
     protected $cache = array();
     protected $errorCache = array();
 
+    private $rootPath;
+
     /**
      * Constructor.
      *
-     * @param string|array $paths A path or an array of paths where to look for templates
+     * @param string|array $paths    A path or an array of paths where to look for templates
+     * @param string|null  $rootPath The root path common to all relative paths (null for getcwd())
      */
-    public function __construct($paths = array())
+    public function __construct($paths = array(), $rootPath = null)
     {
+        $this->rootPath = (null === $rootPath ? getcwd() : $rootPath).DIRECTORY_SEPARATOR;
+        if (false !== $realPath = realpath($rootPath)) {
+            $this->rootPath = $realPath.DIRECTORY_SEPARATOR;
+        }
+
         if ($paths) {
             $this->setPaths($paths);
         }
@@ -81,7 +89,7 @@ class Twig_Loader_Filesystem implements Twig_LoaderInterface, Twig_ExistsLoaderI
      * Adds a path where templates are stored.
      *
      * @param string $path      A path where to look for templates
-     * @param string $namespace A path name
+     * @param string $namespace A path namespace
      *
      * @throws Twig_Error_Loader
      */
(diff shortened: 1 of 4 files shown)


lib/Twig/Loader/Filesystem.php
```diff
@@ -23,13 +23,17 @@ class Twig_Loader_Filesystem implements Twig_LoaderInterface, Twig_ExistsLoaderI
...
      * Constructor.
      *
-     * @param string|array $paths A path or an array of paths where to look for templates
+     * @param string|array $paths    A path or an array of paths where to look for templates
+     * @param string       $rootPath The root path common to all relative paths
      */
-    public function __construct($paths = array())
+    public function __construct($paths = array(), $rootPath = null)
     {
+        $this->rootPath = realpath(null !== $rootPath ? $rootPath : getcwd()).DIRECTORY_SEPARATOR;
```
stof: this is the place being broken if ``$rootPath`` is inside a phar, as ``$this->rootPath`` will be ``false``

fabpot: I would say that we don't want to support this use case (if everything is in a phar, then using an absolute path is not really an issue anymore). So, if `realpath` returns false here, I would prefer to throw an exception.

stof: well, if someone uses relative path for the loader paths and passes ``__DIR__`` as root path (instead of making all paths absolute in the loader paths), your suggestion would break the project once it gets bundled as phar.
This seems worse than forbidding passing relative paths to the loader altogether.

fabpot: fixed

fabpot:
Ready.

lib/Twig/Loader/Filesystem.php
```diff
@@ -23,13 +23,21 @@ class Twig_Loader_Filesystem implements Twig_LoaderInterface, Twig_ExistsLoaderI
...
     protected $errorCache = array();
 
+    private $rootPath;
+
     /**
      * Constructor.
      *
-     * @param string|array $paths A path or an array of paths where to look for templates
+     * @param string|array $paths    A path or an array of paths where to look for templates
+     * @param string       $rootPath The root path common to all relative paths
```
Tobion: string|null

fabpot: fixed

PR #2197, opened 2016-10-21
changed default arg values of Twig_Source constructor to empty strings

fabpot:


stof:
the name is still optional, which breaks compilation of templates, as Twig now always expects the name to be populated. It reads it from the Twig_Source instead of storing it in the stream itself: https://github.com/twigphp/Twig/blob/9c502248378e9422759fab05464db774966c22c7/lib/Twig/Node/Module.php#L52 for instance (but many other places too)

btw, Twig_Environment should probably even have a safeguard ensuring that the Twig_Source given by the loader has the same name than the one it asked for.

fabpot:
I don't understand your concern. An empty name is a valid name. I don't see why we would require a non-empty name.

fabpot:
Having a template name helps a lot (debugging, ...), but it's not a hard requirement.

fabpot:
nervermind

PR #2199, opened 2016-10-22
make getSourceContext a requirement for any loader

fabpot:
I'm going to follow the same protocol as for `Twig_ExistsLoaderInterface`.

### DIFF
--- a/lib/Twig/SourceContextLoaderInterface.php
+++ b/lib/Twig/SourceContextLoaderInterface.php
@@ -9,6 +9,13 @@
  * file that was distributed with this source code.
  */
 
+/**
+ * Adds a getSourceContext() method for loaders.
+ *
+ * @author Fabien Potencier <fabien@symfony.com>
+ *
+ * @deprecated since 1.27 (to be removed in 3.0)
+ */
 interface Twig_SourceContextLoaderInterface
 {
     /**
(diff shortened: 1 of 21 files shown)


lib/Twig/SourceContextLoaderInterface.php
```diff
@@ -9,6 +9,13 @@
  * file that was distributed with this source code.
  */
 
+/**
+ * Adds a getSourceContext() method for loaders.
+ *
+ * @author Fabien Potencier <fabien@symfony.com>
+ *
+ * @deprecated since 1.27 (to be removed in 3.0)
```
xabbuh: This deprecation prevents a clean upgrade path, doesn't it? You now need to implement a deprecated interface (`Twig_SourceContextLoaderInterface`) to get rid of a deprecated method (`Twig_LoaderInterface::getSource()`).

fabpot: Not really. Notice that there is no deprecation notice triggered when you are using this interface. That's the same as the `Twig_ExistsLoaderInterface`.

xabbuh: But the `DebugClassLoader` will trigger it, won't it (in Symfony we also do not trigger deprecations manually in interfaces)? Most IDEs will display this as a misuse too.

PR #2204, opened 2016-10-24
Update if.rst

tzfrs:
The explanation sentence for the code snippet tells the user how we can evaluate an expression to true, however the snippet shows how to evaluate to false.

### DIFF
--- a/doc/tags/if.rst
+++ b/doc/tags/if.rst
@@ -8,7 +8,7 @@ In the simplest form you can use it to test if an expression evaluates to
 
 .. code-block:: jinja
 
-    {% if online == false %}
+    {% if offline == true %}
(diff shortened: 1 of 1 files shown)


tzfrs:
Other way would be to change the sentence from `evaluates to true` to `evaluates to false`

fabpot:
I don't see any issue with the current sentence: `test if an expression evaluates to true` means that the result of the expression evaluates to `true`, not that we compare to `true` as a bool. And the result of the expression depends on the value of the `online` or `offline` variable. So, I prefer to keep the docs as is to avoid the confusion between the end result of the expression and the values used in the expression itself.

tzfrs:
Ah, got it. Than I just misinterpreted that sentence. I didn't understand that the docs mean that the expression evaluates to true, but the variable.

PR #2206, opened 2016-10-24
Move tests handling in Twig_ParserExpression instead of Twig_Extension_Core

fabpot:
In my quest to better organize the layers in Twig, I've always find the way we deal with test a bit awkward. This PR moves the code from the Core extension to the Parser, where it belongs, even if if makes hardcoding 'is' and 'is not' twice, but I think SOC is more important here. Especially as I want to remove the `getEnvironment()` from the Parser class at some point. Having everything centralized helps a lot.

### DIFF
--- a/lib/Twig/ExpressionParser.php
+++ b/lib/Twig/ExpressionParser.php
@@ -19,6 +19,8 @@
  * @see http://en.wikipedia.org/wiki/Operator-precedence_parser
  *
  * @author Fabien Potencier <fabien@symfony.com>
+ *
+ * @internal
  */
 class Twig_ExpressionParser
 {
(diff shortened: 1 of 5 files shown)


lib/Twig/ExpressionParser.php
```diff
@@ -44,7 +58,11 @@ public function parseExpression($precedence = 0)
             $op = $this->binaryOperators[$token->getValue()];
             $this->parser->getStream()->next();
 
-            if (isset($op['callable'])) {
+            if ('is not' === $token->getValue()) {
+                $expr = $this->parseNotTestExpression($expr);
+            } elseif ('is' === $token->getValue()) {
+                $expr = $this->parseTestExpression($expr);
+            } elseif (isset($op['callable'])) {
```
fabpot: no, I don't think so, that's the only way to support advanced use cases, even if I doubt people are doing this.

fabpot: Indeed, it was introduced for `is` and `is not` support. Probably not worth keeping this around. I will deprecated that in 1.28.

PR #2212, opened 2016-10-30
Removing @internal tag from Twig_TokenStream::getSourceContext

### LINKED ISSUE #2210: Removing @internal tag
Some [deprecation notices](https://github.com/twigphp/Twig/blob/1.x/lib/Twig/TokenStream.php#L169) suggest to use [function](https://github.com/twigphp/Twig/blob/1.x/lib/Twig/TokenStream.php#L197) which has tag [@internal](https://phpdoc.org/docs/latest/references/phpdoc/tags/internal.html) in [Twig 2](https://github.com/twigphp/Twig/blob/2.x/lib/Twig/TokenStream.php#L151) too, so IDE (e.g. PHPStorm) shows it as strikethrough.

Can we get rid of it?

kubawerlos:
To resolve #2210

### DIFF
--- a/lib/Twig/TokenStream.php
+++ b/lib/Twig/TokenStream.php
@@ -147,8 +147,6 @@ class Twig_TokenStream
      * Gets the source associated with this stream.
      *
      * @return Twig_Source
-     *
-     * @internal
      */
     public function getSourceContext()
     {


fabpot:
see my comment on the issue, but this is definitely not something people should use.

PR #2214, opened 2016-11-01
Allow construction without constructor arguments

gnat42:
Since we can dynamically add templates and its possible to not need to construct the loader without any initial templates. I've added the default argument to the constructor for simpler instantiation IE new Twig_Loader_Array();. This helps us on a project where we've extended the loader, and have no initial templates but need to add constructor parameters for no real reason.


fabpot:
Thank you @gnat42.

PR #2216, opened 2016-11-03
Load templates from cache, even if they have just been compiled

mpdude:
Previously, when the cache was empty, the compiled template would be written to it, but `eval()`d from `$content`. In that case, it is not possible to step through (read: debug) the compiled template because at least xDebug does not have a clue where the code comes from. 

With this change, PHP/xDebug can tell even on the first run (with an empty cache) where the code was loaded from.

### DIFF
--- a/lib/Twig/Environment.php
+++ b/lib/Twig/Environment.php
@@ -420,7 +420,11 @@ class Twig_Environment
                 }
 
                 if (!class_exists($cls, false)) {
-                    // catch that either $this->bcWriteCacheFile was used or $this->cache is implemented as a no-op
+                    /* Last line of defense if either $this->bcWriteCacheFile was used,
+                     * $this->cache is implemented as a no-op or we have a race condition
+                     * where the cache was cleared between the above calls to write to and load from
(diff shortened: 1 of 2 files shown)


stof:
this is how it was done in the past, and this was subject to race conditions under load

mpdude:
For the record, the relevant discussion is #1836

mpdude:
I think this does not re-introduce the problem outlined in #1836 becuase we still have the `eval` block at the end under any condition.

@znerol Would be great to get your feedback as you opened the race condition issue.

fabpot:
LGTM.

I know firsthand that debugging issues when the template was evaluated is a nightmare, so if we can avoid that most of the time, that's a big win.

mpdude:
Changes made to test cases are still somewhat ugly and do not make test intentions clearer. Any ideas how this could be improved?

fabpot:
I think the intention is pretty clear in the code, so that's probably enough.

fabpot:
Thank you @mpdude.

PR #2222, opened 2016-11-07
add machine-readable version constants

xabbuh:
This will make it easier to implement version depending features (see https://github.com/symfony/symfony/pull/20440#discussion_r86840491 for an example).

### DIFF
--- a/lib/Twig/Environment.php
+++ b/lib/Twig/Environment.php
@@ -17,6 +17,11 @@
 class Twig_Environment
 {
     const VERSION = '1.27.1-DEV';
+    const VERSION_ID = 12701;
+    const MAJOR_VERSION = 1;
+    const MINOR_VERSION = 27;
+    const RELEASE_VERSION = 1;
+    const EXTRA_VERSION = 'DEV';
 
     protected $charset;
     protected $loader;


fabpot:
👍

fabpot:
`version_compare` would work well as well, no?

lib/Twig/Environment.php
```diff
@@ -17,6 +17,11 @@
 class Twig_Environment
 {
     const VERSION = '1.27.1-DEV';
+    const VERSION_ID = 12701;
+    const MAJOR_VERSION = 1;
+    const MINOR_VERSION = 27;
+    const RELEASE_VERSION = 1;
+    const EXTRA_VERSION = 'DEV';
```
SpacePossum: what would value will this have in releases?

xabbuh:
@fabpot Should work too. But checking for integer values is faster and allows for greater flexibility.

lib/Twig/Environment.php
xabbuh: In stable releases, this will be the empty string (see https://github.com/symfony/symfony/blob/v2.7.20/src/Symfony/Component/HttpKernel/Kernel.php#L66).

fabpot:
I know and anyway, I like that, same as we have in Symfony.

@SpacePossum For releases, EXTRA would be empty (exact same thing as for Symfony).

SpacePossum:
nice, :+1:  :ship:

fabpot:
Thank you @xabbuh.

PR #2227, opened 2016-11-10
added a note about the default escaping strategy

### LINKED ISSUE #2061: Document default autoescape value?
https://github.com/twigphp/Twig/blob/master/doc/tags/autoescape.rst

I think it would help if people knew what the default escaping strategy was without declaring an autoescape block. (I'm assuming "html".)

fabpot:
fixes #2061

### DIFF
--- a/doc/templates.rst
+++ b/doc/templates.rst
@@ -424,10 +424,8 @@ everything by default.
 
 Twig supports both, automatic escaping is enabled by default.
 
-.. note::
-
-    Automatic escaping is only supported if the *escaper* extension has been
-    enabled (which is the default).
+The automatic escaping strategy can be configured via the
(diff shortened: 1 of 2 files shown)


doc/templates.rst
```diff
@@ -424,10 +424,8 @@ everything by default.
 
 Twig supports both, automatic escaping is enabled by default.
 
-.. note::
-
-    Automatic escaping is only supported if the *escaper* extension has been
-    enabled (which is the default).
+The automatic escaping strategy can be configured via the
+:ref:`autoescape<environment_options>` option and defaults to `html`.
```
javiereguiluz: Double backticks for `html` :)

fabpot: good catch, thanks

PR #2228, opened 2016-11-10
Enhance perf of Template::getAttribute()

nicolas-grekas:
Note that this is unproved statement for now :)
If someone has a benchmark and want to try, please do (and share results).

### DIFF
--- a/lib/Twig/Template.php
+++ b/lib/Twig/Template.php
@@ -10,6 +10,10 @@
  * file that was distributed with this source code.
  */
 
+if (PHP_VERSION_ID >= 50600) {
+    require_once __DIR__.'/twig_call_method.php';
+}
+
 /**
  * Default base class for compiled templates.
  *
@@ -17,6 +21,9 @@
  */
 abstract class Twig_Template implements Twig_TemplateInterface
 {
+    /**
+     * @internal
+     */
     protected static $cache = array();
 
     protected $parent;
@@ -467,6 +474,8 @@ abstract class Twig_Template implements Twig_TemplateInterface
      * @return mixed The attribute value, or a Boolean when $isDefinedTest is true, or null when the attribute is not set and $ignoreStrictCheck is true
      *
      * @throws Twig_Error_Runtime if the attribute does not exist and Twig is running in strict mode and $isDefinedTest is false
+     *
+     * @internal
      */
     protected function getAttribute($object, $item, array $arguments = array(), $type = self::ANY_CALL, $isDefinedTest = false, $ignoreStrictCheck = false)
     {
@@ -474,7 +483,7 @@ abstract class Twig_Template implements Twig_TemplateInterface
         if (self::METHOD_CALL !== $type) {
             $arrayItem = is_bool($item) || is_float($item) ? (int) $item : $item;
 
-            if ((is_array($object) && array_key_exists($arrayItem, $object))
+            if ((is_array($object) && (isset($object[$arrayItem]) || array_key_exists($arrayItem, $object)))
                 || ($object instanceof ArrayAccess && isset($object[$arrayItem]))
             ) {
                 if ($isDefinedTest) {
@@ -555,37 +564,54 @@ abstract class Twig_Template implements Twig_TemplateInterface
         $class = get_class($object);
 
         // object method
-        if (!isset(self::$cache[$class]['methods'])) {
+        if (!isset(self::$cache[$class])) {
             // get_class_methods returns all methods accessible in the scope, but we only want public ones to be accessible in templates
             if ($object instanceof self) {
                 $ref = new ReflectionClass($class);
                 $methods = array();
 
                 foreach ($ref->getMethods(ReflectionMethod::IS_PUBLIC) as $refMethod) {
-                    $methodName = strtolower($refMethod->name);
-
                     // Accessing the environment from templates is forbidden to prevent untrusted changes to the environment
-                    if ('getenvironment' !== $methodName) {
-                        $methods[$methodName] = true;
+                    if ('getenvironment' !== strtolower($refMethod->name)) {
+                        $methods[] = $refMethod->name;
                     }
                 }
-
-                self::$cache[$class]['methods'] = $methods;
             } else {
-                self::$cache[$class]['methods'] = array_change_key_case(array_flip(get_class_methods($object)));
+                $methods = get_class_methods($object);
             }
+            $cache = array();
+
+            foreach ($methods as $method) {
+                $cache[$method] = $method;
+                $cache[$lcName = strtolower($method)] = $method;
+
+                if ('g' === $lcName[0] && 0 === strpos($lcName, 'get')) {
+                    $name = substr($method, 3);
+                    $lcName = substr($lcName, 3);
+                } elseif ('i' === $lcName[0] && 0 === strpos($lcName, 'is')) {
+                    $name = substr($method, 2);
+                    $lcName = substr($lcName, 2);
+                } else {
+                    continue;
+                }
+
+                if (!isset($cache[$name])) {
+                    $cache[$name] = $method;
+                }
+                if (!isset($cache[$lcName])) {
+                    $cache[$lcName] = $method;
+                }
+            }
+            self::$cache[$class] = $cache;
         }
 
         $call = false;
-        $lcItem = strtolower($item);
-        if (isset(self::$cache[$class]['methods'][$lcItem])) {
-            $method = (string) $item;
-        } elseif (isset(self::$cache[$class]['methods']['get'.$lcItem])) {
-            $method = 'get'.$item;
-        } elseif (isset(self::$cache[$class]['methods']['is'.$lcItem])) {
-            $method = 'is'.$item;
-        } elseif (isset(self::$cache[$class]['methods']['__call'])) {
-            $method = (string) $item;
+        if (isset(self::$cache[$class][$item])) {
+            $method = self::$cache[$class][$item];
+        } elseif (isset(self::$cache[$class][$lcItem = strtolower($item)])) {
+            $method = self::$cache[$class][$lcItem];
+        } elseif (isset(self::$cache[$class]['__call'])) {
+            $method = $item;
             $call = true;
         } else {
             if ($isDefinedTest) {
@@ -596,7 +622,7 @@ abstract class Twig_Template implements Twig_TemplateInterface
                 return;
             }
 
-            throw new Twig_Error_Runtime(sprintf('Neither the property "%1$s" nor one of the methods "%1$s()", "get%1$s()"/"is%1$s()" or "__call()" exist and have public access in class "%2$s".', $item, get_class($object)), -1, $this->getTemplateName());
+            throw new Twig_Error_Runtime(sprintf('Neither the property "%1$s" nor one of the methods "%1$s()", "get%1$s()"/"is%1$s()" or "__call()" exist and have public access in class "%2$s".', $item, $class), -1, $this->getTemplateName());
         }
 
         if ($isDefinedTest) {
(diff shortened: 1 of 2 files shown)


lib/Twig/Template.php
```diff
@@ -607,15 +618,33 @@ protected function getAttribute($object, $item, array $arguments = array(), $typ
...
+            } catch (BadMethodCallException $e) {
+                if ($ignoreStrictCheck || !$this->env->isStrictVariables()) {
+                    return;
+                }
+                throw $e;
             }
-            throw $e;
+        } elseif ($a = $arguments) {
+            switch (false) {
+                case isset($a[1]): $ret = $object->$method($a[0]); break;
```
stof: what if the argument is provided as ``null`` ? This will break things. You should switch on ``count($arguments)`` instead

lib/Twig/Template.php
```diff
@@ -607,15 +618,33 @@ protected function getAttribute($object, $item, array $arguments = array(), $typ
...
+                case isset($a[2]): $ret = $object->$method($a[0], $a[1]); break;
```
stof: the old code allows to have associative arrays (even though keys are ignored), because ``call_user_func_array`` ignores keys (except in HHVM 3.14 but this looks like a bug as older HHVM version behave like PHP): https://3v4l.org/M3h42
So this cannot be accepted as it breaks BC (or you need to do ``$a = array_values($arguments)``, but this might actually negate the micro-optimization you are adding)

lib/Twig/Template.php
nicolas-grekas: fixed

lib/Twig/Template.php
```diff
@@ -607,15 +618,33 @@ protected function getAttribute($object, $item, array $arguments = array(), $typ
...
+            try {
+                $ret = $object->__call($method, $arguments);
+            } catch (BadMethodCallException $e) {
+                if ($ignoreStrictCheck || !$this->env->isStrictVariables()) {
+                    return;
+                }
+                throw $e;
             }
-            throw $e;
+        } elseif ($a = $arguments) {
```
stof: the case of 0 arguments could be handled in the switch, right ?

lib/Twig/Template.php
```diff
@@ -607,15 +618,33 @@ protected function getAttribute($object, $item, array $arguments = array(), $typ
...
+            switch (count($a)) {
+                case 1: $ret = $object->$method($a[0]); break;
```
stof: my comment in https://github.com/twigphp/Twig/pull/2228#discussion_r87375362 about the BC break is still valid

nicolas-grekas: BC break removed as far as public interfaces are concerned: there is now no way to call this method with an associative array by writing twig code.
Since the method is protected, there is a theoretical possibility for a class extending `Twig_Template` and calling it with an associative array. But not a Twig generated one.
But seriously, such a narrow case should not be taken as a blocker here IMHO.

lib/Twig/Template.php
```diff
@@ -555,37 +555,48 @@ protected function getAttribute($object, $item, array $arguments = array(), $typ
         $class = get_class($object);
 
         // object method
-        if (!isset(self::$cache[$class]['methods'])) {
+        if (null === $methodsCache = &self::$cache[$class]['methods']) {
```
nicolas-grekas: optim 1: less indirection (single local var *vs* static prop + key 1 + key 2)

lib/Twig/Template.php
```diff
@@ -555,37 +555,48 @@ protected function getAttribute($object, $item, array $arguments = array(), $typ
...
+                }
+
+                if (!isset($methodsCache[$name])) {
+                    $methodsCache[$name] = $method;
+                }
             }
         }
 
         $call = false;
-        $lcItem = strtolower($item);
```
nicolas-grekas: optim 2: no call to strtolower when not required

lib/Twig/ExpressionParser.php
```diff
@@ -350,8 +350,18 @@ public function getFunctionNode($name, $line)
...
                 }
+                if (count($args) === 2) {
+                    $arguments = null;
+                } elseif ($args->getNode(2) instanceof Twig_Node_Expression_Array) {
+                    $arguments = new Twig_Node_Expression_Array(array(), $line);
+                    foreach ($args->getNode(2)->getKeyValuePairs() as $pair) {
+                        $arguments->addElement($pair['value']);
+                    }
+                } else {
+                    $arguments = $args->getNode(2);
```
stof: this can be a dynamic expression, which means it is still possible to evaluate this node to an associative array

lib/Twig/ExpressionParser.php
```diff
@@ -350,8 +350,18 @@ public function getFunctionNode($name, $line)
                 if (count($args) < 2) {
                     throw new Twig_Error_Syntax('The "attribute" function takes at least two arguments (the variable and the attributes).', $line, $this->parser->getStream()->getSourceContext()->getName());
                 }
+                if (count($args) === 2) {
+                    $arguments = null;
+                } elseif ($args->getNode(2) instanceof Twig_Node_Expression_Array) {
+                    $arguments = new Twig_Node_Expression_Array(array(), $line);
```
stof: this should be the line of ``$args->getNode(2)`` instead

lib/Twig/Template.php
```diff
@@ -555,37 +555,48 @@ protected function getAttribute($object, $item, array $arguments = array(), $typ
...
-        if (isset(self::$cache[$class]['methods'][$lcItem])) {
-            $method = (string) $item;
-        } elseif (isset(self::$cache[$class]['methods']['get'.$lcItem])) {
-            $method = 'get'.$item;
```
nicolas-grekas: optim 3: zero string copy vs concat or cast to string (which is free for strings on PHP7, but copies on PHP5)

lib/Twig/Template.php
```diff
@@ -555,37 +555,48 @@ protected function getAttribute($object, $item, array $arguments = array(), $typ
...
-        } elseif (isset(self::$cache[$class]['methods']['is'.$lcItem])) {
-            $method = 'is'.$item;
-        } elseif (isset(self::$cache[$class]['methods']['__call'])) {
-            $method = (string) $item;
+        if (isset($methodsCache[$item])) {
+            $method = $methodsCache[$item];
+        } elseif (isset($methodsCache[$lcItem = strtolower($item)])) {
+            $method = $methodsCache[$lcItem];
+        } elseif (isset($methodsCache['__call'])) {
```
nicolas-grekas: optim 4: less lookups for the worst case

lib/Twig/Template.php
```diff
@@ -607,15 +618,33 @@ protected function getAttribute($object, $item, array $arguments = array(), $typ
...
+            switch (count($a)) {
+                case 1: $ret = $object->$method($a[0]); break;
+                case 2: $ret = $object->$method($a[0], $a[1]); break;
+                case 3: $ret = $object->$method($a[0], $a[1], $a[2]); break;
+                case 4: $ret = $object->$method($a[0], $a[1], $a[2], $a[3]); break;
+                case 5: $ret = $object->$method($a[0], $a[1], $a[2], $a[3], $a[4]); break;
+                case 6: $ret = $object->$method($a[0], $a[1], $a[2], $a[3], $a[4], $a[5]); break;
+                case 7: $ret = $object->$method($a[0], $a[1], $a[2], $a[3], $a[4], $a[5], $a[6]); break;
+                case 8: $ret = $object->$method($a[0], $a[1], $a[2], $a[3], $a[4], $a[5], $a[6], $a[7]); break;
+                case 9: $ret = $object->$method($a[0], $a[1], $a[2], $a[3], $a[4], $a[5], $a[6], $a[7], $a[8]); break;
```
nicolas-grekas: optim 5: no call_user_func_array indirection in most cases

lib/Twig/Template.php
nicolas-grekas: yes, yet not calling count when not required is another optim *vs* handling in the switch

lib/Twig/Environment.php
```diff
@@ -104,6 +104,9 @@ public function __construct(Twig_LoaderInterface $loader = null, $options = arra
         } else {
             @trigger_error('Not passing a Twig_LoaderInterface as the first constructor argument of Twig_Environment is deprecated since version 1.21.', E_USER_DEPRECATED);
         }
+        if (PHP_VERSION_ID >= 50600) {
+            require_once __DIR__.DIRECTORY_SEPARATOR.'twig_call_method.php';
```
stof: shouldn't this be done at the top of the Twig_Template file instead, as this is where the function is called ?

lib/Twig/twig_call_method.php
```diff
@@ -0,0 +1,6 @@
+<?php
+
+function twig_call_method($object, $method, $arguments)
```
stof: please tag it as ``@internal``

lib/Twig/Template.php
```diff
@@ -607,15 +618,23 @@ protected function getAttribute($object, $item, array $arguments = array(), $typ
...
+                if ($ignoreStrictCheck || !$this->env->isStrictVariables()) {
+                    return;
+                }
+                throw $e;
             }
-            throw $e;
+        } elseif (!$arguments) {
+            $ret = $object->$method();
+        } elseif (PHP_VERSION_ID >= 50600) {
+            $ret = twig_call_method($object, $method, $arguments);
```
nicolas-grekas: optim 5 bis: use argument unpacking in a wrapper.
As surprising as this seems, its faster to call this wrapper that using `call_user_func_array` directly. Should be related to the extra array allocation that it requires.
A micro-benchm shows this on my PHP 7.1:
- x1.0: call_user_func_array(reference)
- x1.8: switch (but unsafe as spotted by @stof 
- x2.3: twig_call_method
- x3.5: unpacking (but Twig's min version is PHP 5.2, not 5.6)

stof:
As you change the structure of the cache, you need to update the C extension too, as it also populates and reads the cache property. Otherwise, it would break things when mixing usage of ``twig_get_attribute`` and ``Twig_Template::getAttribute()`` in the same process

lib/Twig/Environment.php
nicolas-grekas: good idea, done

lib/Twig/twig_call_method.php
nicolas-grekas: done

nicolas-grekas:
ping @jpauli, previous comment is for you :)

lib/Twig/Template.php
xavismeh: I don't understand at this point why you're not using unpacking since PHP version, at this point **is** >= 5.6.
However, for 5.[2-5], can't you relay on `twig_call_method`?
Not sure if I'm missing something on this though

fabpot: Because it would generate a syntax error on older version

xavismeh: What about adding it into polyfill libs?

nicolas-grekas:
@stof can you explain how this situation could happen btw?

stof:
something doing calls to ``Twig_Template::getAttribute`` explicitly in some code, while compiled templates are using the extension

fabpot:
That's not supported. `getAttribute()` is an implementation detail, and people should rely on the compilation code to call attribute in templates, so I don't see any valid use case.

fabpot:
Thank you @nicolas-grekas.

lib/Twig/Template.php
```diff
@@ -555,37 +564,54 @@ protected function getAttribute($object, $item, array $arguments = array(), $typ
...
+            $cache = array();
+
+            foreach ($methods as $method) {
+                $cache[$method] = $method;
+                $cache[$lcName = strtolower($method)] = $method;
+
+                if ('g' === $lcName[0] && 0 === strpos($lcName, 'get')) {
+                    $name = substr($method, 3);
+                    $lcName = substr($lcName, 3);
+                } elseif ('i' === $lcName[0] && 0 === strpos($lcName, 'is')) {
```
staabm: You could save one strpos call by checking lcName[1] instead

nicolas-grekas: Sure, after checking offset 1 exists!

oleg-andreyev:
Testing my project, well and first results:

image

compering v1.26.1 vs bdbfb15

Issue #1979, opened 2016-01-29
Null-coalescing operator (??): The "defined" test only works with simple variables


mnpenner:
The null-coalescing operator (??) should work with "non-simple" variables (e.g. functions or filters that return `null`). 

As an example,

``` twig
{{ "A" ?? log('XXX') ?? "B" }}
```

This throws an exception, where `log()` is a function that returns `null`. I don't believe it should throw.

~~I don't have PHP7 to verify with, but I'm pretty sure that PHP7 wouldn't throw an exception.~~

``` php
Psy Shell v0.6.1 (PHP 7.0.2 — cli) by Justin Hileman
>>> function a() { return null; }
=> null
>>> a() ?? 'b'
=> "b"
```

gadelat:
Here is the actual error to make this report more useful

> Uncaught Twig_Error_Syntax: The "defined" test only works with simple variables in "index.twig" at line 12. in {redacted}/twig/lib/Twig/Node/Expression/Test/Defined.php:37

From the message it's obvious this is not a bug and this issue is a feature request

mnpenner:
Never claimed it was a bug; just said it's inconsistent with the `??` operator in PHP 7. It would make sense to have it behave the same.

So yes, please consider this a feature request.

chiqui3d:
Hello @fabpot, for me not work with:

`{{observa[1] is defined ?? observa[1]}}`

error:

`The "defined" test only works with simple variables.
`

xabbuh:
I think you may want to use something like `{{ observa[1] ?? '' }}`.

chiqui3d:
thanks but not work.

I finally put it that way:

`{{observa[1] is defined ? observa[1] : "" }}`

apfelbox:
@chiqui3d quick heads up:

You don't need the empty state, this works as well: `{{ observa[1] is defined ? observa[1] }}` (as well as `{{ observa[1] | default("") }}`).

chiqui3d:
wonderful, thanks very much

PR #2229, opened 2016-11-10
Bump minimum version to PHP 5.6

fabpot:
Twig 2.0 will be released really soon now (hopefully by the end of the year). So, it's time again to reconsider the minimum PHP version we want to support.

I would like to bump it to PHP 5.6 for the following reasons:

* PHP 5.5 is not supported anymore (not even security fixes now);

* the latest Ubuntu LTS is already using PHP 7!

* #2228 allows to improve performance significantly and having this without the `if`statement makes the optimization more interesting.

What do you think?


MaxVandervelde:
All for it, but is there a plan to support Twig 1.x for any amount of time, for people who may not be able to upgrade quickly?

mickburkesnr:
You should bump it up to 5.6 to keep it inline with the rest of the software Twig is used in. I don't see a reason to keep it so low really.

fabpot:
@MaxVandervelde Great question and the answer is YES. I don't plan to stop maintaining 1.x. So, that will still be an option for the foreseeable future.

Crell:
If 1.x is going to live on for the 5.5-dependent projects for the foreseeable future, I'd say go for it.  5.6, maybe even 7. :smile: (By the time the 5.5-dependent projects drop 5.5, it will be time for them to use 7.something.)

SpacePossum:
like to see 7 as min. as it would be even faster : )
but dropping 5.5 is always a good start :+1:

Seldaek:
I'd vote for 7, but at least 5.6 sounds reasonable at this point. If you have a major cycle of 3-4 years you have to be fairly aggressive when bumping the requirements IMO.

xabbuh:
I would be in favour of PHP 7 as well. However, this means that Twig 1.x would have best be supported at until November 2020 when Symfony 3.4 will reach EOM.

.travis.yml
```diff
@@ -8,7 +8,6 @@ cache:
         - $HOME/.composer/cache/files
 
 php:
-    - 5.5
     - 5.6
     - 7.0
```
alcaeus: You could also add in 7.1 since it's going to be released soon.

composer.json
```diff
@@ -27,7 +27,7 @@
         "forum": "https://groups.google.com/forum/#!forum/twig-users"
     },
     "require": {
-        "php": ">=5.5",
+        "php": ">=5.6",
```
alcaeus: I wouldn't trust PHP to be backwards compatible in the future, maybe use `5.6 || ^7.0`? You can always open it up for PHP 8 if its backwards compatible, but I wouldn't blindly trust it.

SpacePossum: `^5.6|^7.0` since PHP6 is not supported ^_^

MiguelAlcaino:
I agree with the 5.6 version. But there are a lot of ubuntu 14.04 around there with default php5.5. Anyway I think is good to apply some pressure sometimes.

mburtscher:
I'd put it to `^7.0`. New major version will be used primarily in new projects where PHP 7 shouldn't be a problem.

prisis:
Im too for php7

alcaeus:
In my opinion, moving to PHP 7+ would only make sense if you're going to add scalar type hints and possibly enable strict types. If that's not happening for Twig 2.0 I don't really see a reason to drop support for PHP 5.6 at this time. I don't see a problem dropping 5.6 in a minor version release since that's not a BC break.

jepster:
Since it is a major release, it should go with php's major release also: php 7.0.

iltar:
If 1.x will still be supported for a while, why not go 7.0? Would be greatly beneficial for development.

MRJBGO:
I would go for 7.0 as min version as 1.x will still be supported for the moment

MJBGO:
I would go for PHP 7.0 as Twig 1.x will still be supported for the moment and it is a major release

fabpot:
Dropping 5.5 makes sense also because we have an immediate benefit in terms of performance (not that 5.6 is faster, but just because we can use a new 5.6 feature that makes our code faster). Moving to PHP 7 does not help in any way (as we don't have any PHP 7 specific optimizations in the code). So, it's not really worth it.

antonydandrea:
May I ask what changes we can expect on twig 2?

fabpot:
One benefit of moving to PHP 7 is the ability to fix this issue: #1979 Worth it?

ruudk:
I vote for PHP 7

acrobat:
I would go for php 7

jdeniau:
PHP 5.6 will continue to live until 31 Dec 2018 so keeping it is a good think if it does not change anything. 

As you mentionned, PHP7 may solve some issue, but you still can drop PHP5 support on Twig 2.1. 
If someone try to use Twig ^2.0 with PHP5, composer will just not install Twig 2.1 and keep 2.0 as it does in every conflict package resolution.

antonydandrea:
It depends how long you anticipate maintaining Twig 1.X for. You could make 2.0 only from 7 (which from a version number point of view would be nice), but you will have PHP 5.6 until at least the end of 2018 staying on 1.X

andrerom:
+1 on 5.6, it's still supported until end of 2018, bith Debian and RHEL are still on it. Also several extensions are still missing or somewhat unstable on 7.

Possible to consider 7 for 2.1?

MAXakaWIZARD:
+1 for PHP7

jepster:
@fabpot: may PHP 7 is not that benefitial for Twig. But may it would bump the usage of PHP 7. What do you think?

dominikzogg:
I vote for #php7

freekmurze:
+1 for PHP 7

Being able to use the null-coalescing operator will make the code more clear, which is a nice benefit.

Upping the required PHP version will nudge people on PHP 5 that want to use Twig towards upgrading their PHP version. People that cannot upgrade to PHP 7 could still use Twig 1.0.

kpodemski:
1.x for legacy PHP, v2 for a good stuff, i mean PHP 7!

hollodotme:
+1 for php7

gnat42:
Bumping it is fine. I'm not sure between 5.6 and 7.0 as min version. If 7 doesn't really get you any optimization then 5.6 seems as good as any depending on expected lifecycle of twig 2

iltar:
I agree with the replies that mention forcing an upgrade. I know that there are companies that will only update software if they really have to and don't mind running legacy "because it still works". This halts progression

joelpittet:
We did some preliminary work to hopefully get 2.0 into Drupal 8 but a bump to 5.6 may delay that effort because core is at 5.5. I can't say if that will change or not in the 8.x life cycle but my guess would be it wouldn't. Sorry for being the only person on the negative side of this issue. Tricky to align dependencies and not hold them back at the same time:(

javiereguiluz:
If Twig follows the same philosophy as Symfony (and it should) the minimum version bump will barely change anything in this project. The existing code won't be refactored to use the amazing 5.6/7.0 features (to avoid merging conflicts between Twig 1.x and 2.x).

So, in theory Twig will use 5.6/7.0, but in practice 99% of code will still be 5.2/5.3 and only the new features introduced from now on will be 5.6/7.0. Although this may look disappointing for developers, I agree that it's the wise thing to do to ease the maintenance of the project.

xabbuh:
@joelpittet Well, but you could still support Twig 1.x and 2.x at the same time in Drupal and let the user choose which version they want to pick. Developers running PHP > 5.5 can then choose Twig 2.x. That's how we do it in Symfony and it works quite well. If you need any help, feel free to ping us. :)

tgabi333:
7, not a question for me

fabpot:
@javiereguiluz Not sure I understand what you're trying to say here. We are using all the amazing features of PHP 5.6 and even 7.0 in Twig 1.x AND 2.0. As you can see in the codebase, depending on the version you are using, we optimize the generated code. So, there is nothing disappointing here.

Now, if you are talking about namespaces, which is the only "feature" I can think of that we are not using in Twig, well, I couldn't care less. It does not bring any value to the project, I does not make it better, cleaner, or anything. Twig is primarily a templating system, and the focus is on the Twig language, not PHP. 80% of Twig core in internal and should never be "used" by Twig users.

Or do you have something else in mind?

fabpot:
@joelpittet Twig 1.x and 2.0 have the exact same features, so Drupal can still be compatible with both versions. Bumping the minimum version of Twig only means that Composer will install Twig 2.0  only for people on PHP 5.6 and up.

The only downside I can think of for Drupal is that you cannot drop support for Twig 1.x, but even if you could, it would not change anything in the way you are using Twig internally. And that's the strategy we have adopted in Symfony, we won't drop Twig 1.x anytime soon as there is no need to.

surgun9:
5.6/7.0

dominikzogg:
@fabpot whats about scalar type hints, using strict, to get a event more stable codebase?

javiereguiluz:
@fabpot I wasn't thinking about the namespaces. Let me explain it with a simple example using the `write()` method of `Twig_Compiler`.

### Scenario 1

Twig code uses PHP 5.2 and Composer requires PHP 5.2

| Code | Composer
| ----- | -----------
| before_code | before_composer

### Scenario 2

Composer requires PHP 7.0 but Twig code is unchanged (it still "uses" PHP 5.2)

| Code | Composer
| ----- | -----------
| before_code | after_composer

### Scenario 3

Composer requires PHP 7.0 and Twig code is carefully refactored to leverage the amazing PHP 7.0 features

| Code | Composer
| ----- | -----------
| after_code | after_composer

---

Some people may think that we're going for Scenario 3 ... but we're talking about Scenario 2 (the same we did for Symfony). So, most of the Twig code will "never" leverage PHP 7.0 features. Personally I agree with this, because otherwise the project maintenance will be a nightmare.

joelpittet:
Ah thanks @fabpot don't think about supporting both. There was a bit of reworking of some things we did but iirc they could be done in a way that would satisfy both.

fabpot:
@joelpittet If supporting both Twig 1.x (latest version) and 2.0 is not possible, open an issue as this should be considered as a bug.

Thijzer:
Consider this questions. What benefits do you get from enforcing 7.0. You could refactor 2.* suited for 7.0 without breaking anything. Not moving to 7.0 would also shorten the lifespan of the 2. * release.  Also consider the pace php releases are growing into. I just don't see a negative issue not enforcing 7.0.

xabbuh:
@javiereguiluz I do not see any issues with that. It does not really matter how the compiler is written at all. What's really import is what is the output of the compiler. And that can actually be depending on the PHP version. Though I currently do not see anything in compiled code that could be improved for PHP 7.

joelpittet:
@fabpot one other thing if we were supporting both Twig 1.x and 2.x through composer, for smaller shops if they build on their development machine with >= PHP 5.6 and it pulls down Twig 2.x and deploy on production with PHP 5.5, there could be quite a bit of WTF noise (though I'm not sure what features of 5.6 only would be used). Ideally they would have a build script, and a staging environment, but I don't count a good chunk of people being that lucky.

Regarding the issue we are facing, it's really currently on the Drupal side of things, not a bug in Twig for the upgrade to 2.x. Here's the issue for reference: https://www.drupal.org/node/2568181

Seldaek:
@joelpittet If people use a different version locally than in prod, they really should define a platform config in composer.json, see https://getcomposer.org/doc/06-config.md#platform

angelk:
+1 for php7

joelpittet:
@Seldaek oh that's pretty sweet, good to know:)

SpacePossum:
when Twig 2 is set to 7.1 there would be gain in some core filter and functions from `Iterable`, losing the `is_array or Traversable` checks and some `iterator_to_array` calls

fabpot:
@SpacePossum I hadn't had look but can't we make have a special path for when people are using PHP 7.1, do we really need to wait for the minimum version to be 7.1 before benefiting from this?

stof:
Well, replace the existing checks with a ``is_iterable`` check does not bring lots of benefits IMO, as it is only inside the implementation. Doing a conditional here does not bring much benefit IMO

Issue #1882, opened 2015-10-20
Add the new function to obtain the names of all blocks


hason:
Suggested names:
- `blocks`
- `block_names`

Use case:

``` twig
{# grid.html.twig #}
<div class="row">
    {% set columns = columns|split(',') %}
    {% for i, block in blocks() if block starts with 'col' %}
        <div class="col-md-{{ cycle(columns, i) }}">{{ block(block) }}</div>
    {% endfor %}
</div>

{# index.html.twig #}
{% embed "grid.html.twig" with { columns: "2,4,4,2" } %}
    {% block col1 %}foo{% endblock %}
    {% block col2 %}foo{% endblock %}
    {% block col3 %}foo{% endblock %}    
    {% block col4 %}foo{% endblock %}
{% endembed %}
```

Issue #1873, opened 2015-10-13
Provide api for rendering blocks separately


Koc:
Does it possible change something here for 2.0?

ref https://github.com/twigphp/Twig/issues/676

stof:
Well, adding a new API does not break BC, so this could even go in the 1.x branch, if we decide on a design to do it

alexweissman:
Just for context, the question is whether `renderBlock` will become an API method (as opposed to being labeled "internal").  The problem right now is:

- `renderBlock` is really useful for rendering email templates (i.e., one block for the subject, and another block for the body).
- Unfortunately, `renderBlock` does not seem to run in the global context, so any global Twig variables do not get passed in.
- Also, something about `renderBlock` and buffering.

nakashu:
A use case I wanted to solve many times with block rendering is ajax responses.

```twig
{% extends 'layout.html.twig' %}

{% block breadcrumb %}... breadcrumb ... {% endblock %}
{% block content %}
.. content that should be changed after ajax request ..
.. with more complex logic, like form errors rendering etc .. 
{% endblock %}
```

Here you could just render the content block, based on if the request is ajax else render the whole page with extends ... Although it could be done by using %include 'content.html.twig'%, at least for me this seems like a nicer solution.

```php
return $request->isXmlHttpRequest()
    ? $this->renderBlock('page.html.twig')
    : $this->render('page.html.twig');
// or handle it in listener
```

If anoyone got a nice solutions for returning partially rendered twig tempales for ajax, let me know, I'll be gratefull.

stof:
this can also be done using conditional inheritance (this is what the Symfony WebProfiler is doing):

```twig
{# ajax_layout.html.twig #}
{% block content '' %}
```


```twig
{% extends is_ajax ? 'ajax_layout.html.twig' : 'layout.html.twig' %}

{% block breadcrumb %}... breadcrumb ... {% endblock %}
{% block content %}
.. content that should be changed after ajax request ..
.. with more complex logic, like form errors rendering etc .. 
{% endblock %}
```

nakashu:
that looks really nice thanks @stof.

nakashu:
tested out @stof 's suggestion and with just tiny change to make it more dynamic, you can render any block from a larger template, thus also covering the email use case:

```twig
{# block.html.twig #}
{% if _block is defined %}
    {{ block(_block) }} {# render only block with specified name in the variable #}
{% endif %}
```

```twig
[...]

PR #2236, opened 2016-11-10
Expose a way to access template data and methods in a portable way

fabpot:
`Twig_Template` is an internal class, and most of its methods are marked as being `@internal` as they are not "safe" to use by end users.

But some of its features are interesting like `renderBlock()` which is very useful when rendering emails for instance (see #1873).

This PR tries to address both issues by marking the whole `Twig_Template` class as being internal  and by exposing a new way to safely interact with a template besides the obvious `render()`/`display()` methods via `$twig->load("...")`.

If also add some convenient methods like `hasBlock()` or `getBlocks()` (see ~#1831~ and #1882):

```php
$template = $twig->load('index');

echo $template->render($context);
$template->display($context);

if ($template->hasBlock('name') {
   $template->displayBlock('name', $context);
}

foreach ($template->getBlocks() as $block) {
    echo $template->render($block, $context);
}
```

### DIFF
--- a/lib/Twig/Template.php
+++ b/lib/Twig/Template.php
@@ -17,7 +17,13 @@ if (PHP_VERSION_ID >= 50600) {
 /**
  * Default base class for compiled templates.
  *
+ * This class is an implementation detail of how template compilation currently
+ * works, which might change. It should never be used directly. Use $twig->load()
+ * instead, which returns an instance of Twig_TemplateWrapper.
+ *
  * @author Fabien Potencier <fabien@symfony.com>
+ *
+ * @internal
  */
 abstract class Twig_Template implements Twig_TemplateInterface
 {
@@ -307,18 +313,34 @@ abstract class Twig_Template implements Twig_TemplateInterface
     }
 
     /**
-     * Returns all block names.
+     * Returns all block names in the current context of the template.
      *
-     * This method is for internal use only and should never be called
-     * directly.
+     * This method checks blocks defined in the current template
+     * or defined in "used" traits or defined in parent templates.
+     *
+     * @param string $name    The block name
+     * @param array  $context The context
+     * @param array  $blocks  The current set of blocks
      *
      * @return array An array of block names
      *
      * @internal
      */
-    public function getBlockNames()
+    public function getBlockNames(array $context = null, array $blocks = array())
     {
-        return array_keys($this->blocks);
+        if (null === $context) {
+            @trigger_error('The '.__METHOD__.' method is internal and should never be called; calling it directly is deprecated since version 1.28 and won\'t be possible anymore in 2.0.', E_USER_DEPRECATED);
+
+            return array_keys($this->blocks);
+        }
+
+        $names = array_merge(array_keys($blocks), array_keys($this->blocks));
+
+        if (false !== $parent = $this->getParent($context)) {
+            $names = array_merge($names, $parent->getBlockNames($context));
+        }
+
+        return array_unique($names);
     }
 
     protected function loadTemplate($template, $templateName = null, $line = null, $index = null)
@@ -329,6 +351,10 @@ abstract class Twig_Template implements Twig_TemplateInterface
             }
 
             if ($template instanceof self) {
+                return $template;
+            }
+
+            if ($template instanceof Twig_TemplateWrapper) {
                 return $template;
             }
 
@@ -654,9 +680,10 @@ abstract class Twig_Template implements Twig_TemplateInterface
             throw $e;
         }
 
-        // useful when calling a template method from a template
-        // this is not supported but unfortunately heavily used in the Symfony profiler
+        // @deprecated in 1.28
         if ($object instanceof Twig_TemplateInterface) {
+            @trigger_error('Using the dot notation on an instance of '.__CLASS.' is deprecated since version 1.28 and won\'t be supported anymore in 2.0.', E_USER_DEPRECATED);
+
             return $ret === '' ? '' : new Twig_Markup($ret, $this->env->getCharset());
         }
 
--- a/lib/Twig/TemplateWrapper.php
+++ b/lib/Twig/TemplateWrapper.php
@@ -1,0 +1,134 @@
+<?php
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) 2016 Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+/**
+ * Exposes a template to userland.
+ *
+ * @author Fabien Potencier <fabien@symfony.com>
+ */
+final class Twig_TemplateWrapper
+{
+    private $env;
+    private $template;
+
+    /**
+     * This method is for internal use only and should never be called
+     * directly (use Twig_Environment::load() instead).
+     *
+     * @internal
+     */
+    public function __construct(Twig_Environment $env, Twig_Template $template)
+    {
+        $this->env = $env;
+        $this->template = $template;
+    }
+
+    /**
+     * Renders the template.
+     *
+     * @param array $context An array of parameters to pass to the template
+     *
+     * @return string The rendered template
+     */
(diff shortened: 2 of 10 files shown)


fabpot:
I've made the context a constructor argument instead of an argument of each method for two main reasons:

 * we might be able to optimize the context in a future version;
 * being able to call blocks with different context seems messy.

Not sure if that's a good idea or not.

fabpot:
Comparison of how to do it now (but does not really work well), and the new safe way:

```php
// old unsafe way
$template = $twig->loadTemplate('mail.twig');
$parameters  = array(
    'customer' => $customer
);

$subject  = $template->renderBlock('subject',   $parameters);
$bodyHtml = $template->renderBlock('body_html', $parameters);
$bodyText = $template->renderBlock('body_text', $parameters);

// new safe way
$template = $twig->load('mail.twig', array(
    'customer' => $customer
));

$subject  = $template->renderBlock('subject');
$bodyHtml = $template->renderBlock('body_html');
$bodyText = $template->renderBlock('body_text');
```

fabpot:
see symfony/symfony#20484 for a usage example (I've also replaced renderblock by displayblock there but that's not related to the changes here, just avoiding an unneeded indirection).

fabpot:
The second commit add a `template_block()` function that would allow to remove the hacks we have in the Symfony web profiler templates:

```jinja2
{{ template_block('template', 'block', context) }}
```

fabpot:
If we expose a `template_block()` function, do we need a `template_has_block()`? Not sure, especially has trying to render a non-existing block just returns an empty string. I don't want to pollute too much and we don't have `has_block()` either (but asked in #1831).

lib/Twig/Node/Expression/TemplateBlockReference.php
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        parent::__construct($nodes, array('output' => false), $lineno);
+    }
+
+    public function compile(Twig_Compiler $compiler)
+    {
+        if ($this->getAttribute('output')) {
+            $compiler
+                ->addDebugInfo($this)
```
stof: shouldn't this always be added ?

lib/Twig/Node/Expression/TemplateBlockReference.php
```diff
@@ -0,0 +1,54 @@
...
+                ->subcompile($this->getNode('name'))
+                ->raw(', ')
+                ->subcompile($this->getNode('context'))
+                ->raw(");\n")
+            ;
+        } else {
+            $compiler
+                ->raw('$this->env->load(')
+                ->subcompile($this->getNode('template'))
+                ->raw(')->displayBlock(')
```
stof: should it really use display here ? I think it should be ``renderBlock``

lib/Twig/Template.php
```diff
@@ -650,8 +656,7 @@ protected function getAttribute($object, $item, array $arguments = array(), $typ
             throw $e;
         }
 
-        // useful when calling a template method from a template
-        // this is not supported but unfortunately heavily used in the Symfony profiler
+        // @deprecated in 1.28
```
stof: should it throw a deprecation warning ?

Koc:
+1 for public api
+0 for internal implementation

fabpot:
@Koc What do you mean by "internal implementation" vs "public API"?

Koc:
I mean that is useful for end twig user and solves my issue but I have nothing to say about implementation.

lib/Twig/Node/Expression/TemplateBlockReference.php
fabpot: fixed

lib/Twig/Node/Expression/TemplateBlockReference.php
fabpot: we don't add this for inlined code (except for context accesses)

lib/Twig/Node/Expression/TemplateBlockReference.php
```diff
@@ -0,0 +1,54 @@
...
+
+        parent::__construct($nodes, array('output' => false), $lineno);
+    }
+
+    public function compile(Twig_Compiler $compiler)
+    {
+        if ($this->getAttribute('output')) {
+            $compiler
+                ->addDebugInfo($this)
+                ->write('$this->env->load(')
```
stof: Technically, the compiled code could avoid using the Twig_TemplateWrapper layer, as you don't need the output buffer cleanup logic here (it will be cleaned by the rendering of the template using the function already when the exception bubbles to it). This will avoid this level of call stack (and of try/catch). And the Twig code can relies on the Twig internal API.

Btw, this is why rendering blocks directly in the form theming was safe regarding the output buffer: it was always called inside a Twig call cleaning the buffer (unless someone was doing a crazy thing to call the TwigFormRenderer from a different templating engine than Twig). The email rendering use case was the unsafe one, but this one will not rely on ``template_block``

lib/Twig/Template.php
fabpot: done

fabpot:
I've extracted the second commit to #2245 and removed comments about it here.

lib/Twig/TemplateWrapper.php
```diff
@@ -0,0 +1,129 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+/**
+ * Exposes a template to userland.
+ *
+ * @author Fabien Potencier <fabien@symfony.com>
+ */
+final class Twig_TemplateWrapper
```
stof: Should we really call it a wrapper ? The fact that it is a wrapper around a Twig_Template object is also an implementation detail. I'm wondering whether we can find a better name here (I have no good idea though)

lib/Twig/TemplateWrapper.php
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    /**
+     * Returns defined block names in the template.
+     *
+     * @return string[] An array of defined template block names
+     */
+    public function getBlocks()
+    {
+        // FIXME
```
stof: should be done, or should be removed

lib/Twig/TemplateWrapper.php
```diff
@@ -0,0 +1,129 @@
...
+                ob_end_clean();
+            }
+
+            throw $e;
+        }
+
+        return ob_get_clean();
+    }
+
+    public function getSourceContext()
```
stof: please add the phpdoc for the return type, as this is the public API

lib/Twig/TemplateWrapper.php
fabpot: I would have loved to use `Twig_Template`, but that's not possible. So, I used a descriptive name instead. But I agree that the name is not ideal.

fabpot:
This one is finished now (docs and tests added). If nobody has a better name than ``Twig_TemplateWrapper``, I will merge this one as is.

lib/Twig/TemplateWrapper.php
Koc: Is it possible to use `Twig_Template` for 2.0?

stof: Changing the meaning of Twig_Template between 1.x and 2.x means it is very hard to write forward-compatible code. So -1 for this.

SpacePossum:
awesome, thanks!

lib/Twig/Template.php
```diff
@@ -654,9 +680,10 @@ protected function getAttribute($object, $item, array $arguments = array(), $typ
...
         if ($object instanceof Twig_TemplateInterface) {
+            @trigger_error('Using the dot notation on an instance of '.__CLASS.' is deprecated since version 1.28 and won\'t be supported anymore in 2.0.', E_USER_DEPRECATED);
```
gharlan: typo: `__CLASS` -> `__CLASS__`

SpacePossum: https://github.com/twigphp/Twig/pull/2263

PR #2238, opened 2016-11-11
added support for PHP 7 null coalescing operator

### LINKED ISSUE #1979: Null-coalescing operator (??): The "defined" test only works with simple variables

fabpot:


stof:
Does it work fine with the usage of ``twig_attribute(a, b) ?? 'other'`` ?

fabpot:
What do you mean by "it works"? Yes it still works, but does not use the PHP 7 operator as a is not necessarily an array.

lib/Twig/Node/Expression/NullCoalesce.php
```diff
@@ -20,4 +20,20 @@ public function __construct(Twig_NodeInterface $left, Twig_NodeInterface $right,
 
         parent::__construct($test, $left, $right, $lineno);
     }
+
+    public function compile(Twig_Compiler $compiler)
+    {
+        if (PHP_VERSION_ID >= 70000 && $this->getNode('expr2') instanceof Twig_Node_Expression_Name) {
```
fabpot: This is just one case where we can "optimize" the ternary operator. PHP 7 also supports more complex expressions that can return `null`. The rule here seems to be that you cannot have more than one level of undefinedness.

So, for instance, if `log` is defined, `log("foo") ?? "NOPE"` works, but `log($a["foo"]) ?? "NOPE"` does not if `$a["foo"]` is not defined. If we want to convert the code to always use `??` when possible, this should probably be implemented in the Optimizer node visitor, not here. But the question is: is it worth it in terms of performance? It cleans up the generated code quite a bit though.

PR #2239, opened 2016-11-11
Add "is defined" support for block()

### LINKED ISSUE #1821: Add a new test to check if a block exists

fabpot:
replaces #1831, fixes #1821

I think reusing the semantic of the `defined` test is more idiomatic and easily discoverable.

/cc @hason


lib/Twig/Node/Expression/BlockReference.php
```diff
@@ -19,28 +19,36 @@ class Twig_Node_Expression_BlockReference extends Twig_Node_Expression
...
+        if ($this->getAttribute('is_defined_test')) {
             $compiler
-                ->raw('$this->renderBlock(')
+                ->raw('$this->blockExists(')
                 ->subcompile($this->getNode('name'))
                 ->raw(', $context, $blocks)')
             ;
+        } else {
+            if ($this->getAttribute('as_string')) {
+                $compiler->raw('(string) ');
```
stof: this casting is useless when being in output mode, right ?

fabpot:
Thank you @hason.

PR #2241, opened 2016-11-11
added some more tests

fabpot:


fabpot:
Some of theses tests are already part of the testing of attributes, but I prefer to have them in defined as well.

PR #2242, opened 2016-11-11
added 'is defined' support for constant

fabpot:


lib/Twig/Node/Expression/Function.php
```diff
@@ -27,7 +27,12 @@ public function compile(Twig_Compiler $compiler)
         $this->setAttribute('needs_context', $function->needsContext());
         $this->setAttribute('arguments', $function->getArguments());
         if ($function instanceof Twig_FunctionCallableInterface || $function instanceof Twig_SimpleFunction) {
-            $this->setAttribute('callable', $function->getCallable());
+            $callable = $function->getCallable();
+            if ('constant' === $name && $this->getAttribute('is_defined_test')) {
```
stof: wouldn't it be better to use a special node for the ``constant`` function instead of hardcoding this in the generic node ?

PR #2244, opened 2016-11-12
New method to get up-to-date template timestamp loaded

micaelmalta:
Since header last-modified is important to Google, we have to know when template rendered has last changed

In controller, you can get up-to-date template timestamp loader after render method

Here an example:

`$content           = $this->renderView($templateName);
$templateFreshness = $this->container->get('twig')->getTemplateLoaderFreshness(); 

if($templateFreshness && $templateFreshness > $articleDateModified) {
    $articleDateModified = $templateFreshness;
}

$lastModified = new \DateTime();
$lastModified->setTimestamp($articleDateModified);
    
$response = new Response();
$response->setLastModified($lastModified);
$response->setEtag($etag);

if ($response->isNotModified($request)) {
    return $response;
}

$response->setContent($content);

return $response;`

### DIFF
--- a/lib/Twig/Environment.php
+++ b/lib/Twig/Environment.php
@@ -501,6 +501,7 @@ class Twig_Environment
     public function getTemplateLoaderFreshness()
     {
         $cache = $this->getCache(false);
+
         return $this->getLoader()->getTemplateFreshness($cache);
     }
 
--- a/lib/Twig/Environment.php
+++ b/lib/Twig/Environment.php
@@ -494,6 +494,17 @@ class Twig_Environment
     }
 
     /**
+     * Returns up-to-date timestamp of all template loaded.
+     *
+     * @return string
+     */
+    public function getTemplateLoaderFreshness()
+    {
+        $cache = $this->getCache(false);
+        return $this->getLoader()->getTemplateFreshness($cache);
+    }
+
+    /**
      * Tries to load a template consecutively from an array.
      *
      * Similar to loadTemplate() but it also accepts Twig_TemplateInterface instances and an array
--- a/lib/Twig/Loader/Chain.php
+++ b/lib/Twig/Loader/Chain.php
@@ -166,4 +166,12 @@ class Twig_Loader_Chain implements Twig_LoaderInterface, Twig_ExistsLoaderInterf
 
         throw new Twig_Error_Loader(sprintf('Template "%s" is not defined%s.', $name, $exceptions ? ' ('.implode(', ', $exceptions).')' : ''));
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getTemplateFreshness(Twig_CacheInterface $cache)
+    {
+        return time();
+    }
 }
--- a/lib/Twig/Loader/Filesystem.php
+++ b/lib/Twig/Loader/Filesystem.php
@@ -302,4 +302,35 @@ class Twig_Loader_Filesystem implements Twig_LoaderInterface, Twig_ExistsLoaderI
             || null !== parse_url($file, PHP_URL_SCHEME)
         ;
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getTemplateFreshness(Twig_CacheInterface $cache)
+    {
+        $cacheKey = $cache->generateKey(
+            md5(serialize($this->cache)),
+            get_class()
+        );
+
+        ob_start();
+        $cache->load($cacheKey);
+        $lastDate = ob_get_contents();
+        ob_end_clean();
+
+        if(!$lastDate){
+            $lastModified = 0;
+            foreach($this->cache as $name=>$v){
+                $dateModified = filemtime($this->findTemplate($name));
+                if($dateModified > $lastModified) {
+                    $lastModified = $dateModified;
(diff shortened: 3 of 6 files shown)


lib/Twig/Loader/Filesystem.php
```diff
@@ -302,4 +302,34 @@ private function isAbsolutePath($file)
...
+
+        ob_start();
+        $cache->load($cacheKey);
+        $lastDate = ob_get_contents();
+        ob_end_clean();
+
+        if (!$lastDate) {
+            $lastModified = 0;
+            foreach ($this->cache as $name => $v) {
+                $dateModified = filemtime($this->findTemplate($name));
```
stof: Doing such calls in prod is not a good idea.

And in autoreload mode, your freshness cache will not be refreshed when refreshing the template, which does not make any sense.

lib/Twig/Loader/Filesystem.php
```diff
@@ -302,4 +302,34 @@ private function isAbsolutePath($file)
...
         ;
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getTemplateFreshness(Twig_CacheInterface $cache)
+    {
+        $cacheKey = $cache->generateKey(
+            md5(serialize($this->cache)),
```
stof: this cache key will depend of which templates have already been loaded in this process, and of the order in which they have been loaded. This looks like a bad idea to me.

lib/Twig/Loader/Chain.php
```diff
@@ -166,4 +166,12 @@ public function isFresh($name, $time)
 
         throw new Twig_Error_Loader(sprintf('Template "%s" is not defined%s.', $name, $exceptions ? ' ('.implode(', ', $exceptions).')' : ''));
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getTemplateFreshness(Twig_CacheInterface $cache)
+    {
+        return time();
```
stof: this means that as soon as several loaders are used, any logic implemented in the Filesystem loader gets ignored and you will get ``time()`` each time. I fail to see how this API can be useful

lib/Twig/Environment.php
```diff
@@ -494,6 +494,18 @@ public function isTemplateFresh($name, $time)
     }
 
     /**
+     * Returns up-to-date timestamp of all template loaded.
+     *
+     * @return string
+     */
+    public function getTemplateLoaderFreshness()
```
stof: Btw, even the name is weird. This does not return a "freshness" (what is this?) but a timestamp.

stof:
Btw, the phpdoc of your loader implies that a loader is expected to track which templates have already been loaded using it. But there is no such requirement for loaders today, and forcing them to become stateful is a bad idea.

lib/Twig/Loader/Filesystem.php
mmicael1: As the order of include template change the rendering of a template, yes the order is important

lib/Twig/Loader/Filesystem.php
```diff
@@ -302,4 +302,34 @@ private function isAbsolutePath($file)
...
+            get_class()
+        );
+
+        ob_start();
+        $cache->load($cacheKey);
```
mmicael1: Twig does not have other cache implementation today

lib/Twig/Loader/Filesystem.php
```diff
@@ -302,4 +302,34 @@ private function isAbsolutePath($file)
...
+        $lastDate = ob_get_contents();
+        ob_end_clean();
+
+        if (!$lastDate) {
+            $lastModified = 0;
+            foreach ($this->cache as $name => $v) {
```
mmicael1: I dont understand. As all the templates loaded can change the html output, it's what we need

lib/Twig/Loader/Filesystem.php
mmicael1: That's why i have cached it.

autoreload in production il also a very bad idea

lib/Twig/LoaderInterface.php
```diff
@@ -52,4 +52,13 @@ public function getCacheKey($name);
...
     public function isFresh($name, $time);
+
+    /**
+     * Returns up-to-date timestamp of all template loaded.
+     *
+     * @param Twig_CacheInterface $cache
+     *
+     * @return string
+     */
+    public function getTemplateFreshness(Twig_CacheInterface $cache);
```
mmicael1: So what do you suggest ?

mmicael1:
So what do you suggest to know if twig template has changed ?

lib/Twig/Loader/Filesystem.php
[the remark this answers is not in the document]
robfrawley: @mmicael1 The important \@stof is conveying is:

> Twig_CacheInterface is meant to cache class definition for compiled templates [and] **reusing it for a totally different caching need is a bad idea**, as the implementation may not be a generic cache.

In other words, define something else for your use-case, don't re-use that cache implementation.

fabpot:
I'm closing this pull request as the implementation cannot work.

PR #2245, opened 2016-11-12
added support for a custom template on the block() function

fabpot:
`block()` allows one to render a block of the current template.

This PR adds the possibility to use `block()` to render a block of another template:

```twig
{{ block('footer', template_name) }}
```

As it uses the same logic as for the regular `block()`, you can also use it in a test:

```twig
{% if block('footer', template_name) is defined %}
    ...
{% endif %}
```

This removes the needs to use the internal `Twig_Template::renderBlock()` method in the Symfony Web Profiler for instance. And combined with #2236, it allows us to really mark the whole `Twig_Template` class as being internal.

I'm aware that #1302 asked for being able to pass a context to `block()` (implemented in #1433), but I prefer not to in favor of adding the `with` tag (see #719 and #1054).

When the `with` will be implemented, the code in the Symfony Web Profiler will become something along the lines of:

```twig
        {% with {
            'collector': profile.getcollector(name),
            'profiler_url': profiler_url,
            'token': profile.token,
            'name': name
          } %}
            {{ block('toolbar', template) }}
        {% endwith %}
```

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
 * 1.28.0 (2016-XX-XX)
 
+ * added support for a custom template on the block() function
  * added "is defined" support for block() and constant()
  * optimized the way attributes are fetched
 
--- a/doc/functions/block.rst
+++ b/doc/functions/block.rst
@@ -1,6 +1,9 @@
 ``block``
 =========
 
+.. versionadded: 1.28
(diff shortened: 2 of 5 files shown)


stof:
👍 for this. Looks better to reuse it ``block`` rather than adding a new ``template_block``

PR #2246, opened 2016-11-13
Add a "with" tag

### LINKED ISSUE #719: Add a "with" tag

fabpot:
Closes #719, alternative implementation of #1054

### DIFF
--- a/lib/Twig/Node/With.php
+++ b/lib/Twig/Node/With.php
@@ -1,0 +1,62 @@
+<?php
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) 2016 Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+/**
+ * Represents a nested "with" scope.
+ *
+ * @author Fabien Potencier <fabien@symfony.com>
+ */
+class Twig_Node_With extends Twig_Node
+{
+    public function __construct(Twig_Node $body, Twig_Node $variables = null, $only = false, $lineno, $tag = null)
+    {
+        $nodes = array('body' => $body);
+        if (null !== $variables) {
+            $nodes['variables'] = $variables;
+        }
+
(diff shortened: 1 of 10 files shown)


lib/Twig/Node/With.php
```diff
@@ -0,0 +1,62 @@
...
+                ->raw(";\n")
+                ->write(sprintf("if (!is_array(\$%s)) {\n", $varsName))
+                ->indent()
+                ->write("throw new Twig_Error_Runtime('Variables passed to the \"with\" tag must be a hash.');\n")
+                ->outdent()
+                ->write("}\n")
+            ;
+
+            if ($this->getAttribute('only')) {
+                $compiler->write("\$context = array('_parent' => \$context);\n");
```
stof: this still gives access to the parent scope here. Shouldn't we store the old scope in a variable in the compiled template instead of storing it inside the new context ? Doing ``{% set _parent = ... %}`` inside the ``{% with %}`` allows to replace the whole scope outside the tag

guilliamxavier:
Maybe related to stof's question, I found out that the `with` tag does not behave the same as the `for` tag when you reassign a variable declared before (even though they both create a nested scope):

```twig
{% set foo = 0 -%}
{{ foo }}
{% for var in [1] %}
    {{ foo }}
    {% set foo = var -%}
    {{ foo }}
{% endfor %}
{{ foo }}
```

outputs:

```
0
    0
    1
1
```

but if I change the `for` into an equivalent `with`:

```diff
-{% for var in [1] %}
+{% with { var: 1 } %}

-{% endfor %}
+{% endwith %}
```

then the output is different at the last line:

```
0
    0
    1
0
```

Is this intentional/desirable/avoidable?

stof:
The with tag behaves the same than the block tag: they create an isolated scope. `for` does not create an isolated scope. It only cleans some scope after it.

I'm not sure the access to the parent scope was intended.

PR #2248, opened 2016-11-13
renamed blockExists to hasBlock

fabpot:
As `hasBlock()` is marked as internal and because it's not even used anywhere in Twig, I've replaced its implementation with the one of `blockExists()`, and I also kept BC with a deprecation notice just in case someone is using it.

### DIFF
--- a/lib/Twig/Template.php
+++ b/lib/Twig/Template.php
@@ -270,29 +270,40 @@ abstract class Twig_Template implements Twig_TemplateInterface
     }
 
     /**
-     * Returns whether a block exists or not.
+     * Returns whether a block exists or not in the current context of the template.
      *
-     * This method is for internal use only and should never be called
-     * directly.
+     * This method checks blocks defined in the current template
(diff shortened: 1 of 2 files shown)


lib/Twig/Template.php
```diff
@@ -270,29 +270,40 @@ public function renderBlock($name, array $context, array $blocks = array(), $use
...
-     * @see blockExists
-     *
      * @internal
      */
-    public function hasBlock($name)
+    public function hasBlock($name, array $context = null, array $blocks = array())
     {
-        return isset($this->blocks[(string) $name]);
+        if (null === $context) {
+            @trigger_error('The '.__METHOD__.' method is internal and should never be called; calling it directly is deprecated since version 1.28 and won\'t be possible anymore in 2.0.', E_USER_DEPRECATED);
```
stof: the deprecation warning is weird, as the method won't be removed entirely, but only this signature.

PR #2249, opened 2016-11-13
Remove optimization as it's not compatible with Symfony cache system

### LINKED ISSUE #2243: Using twig_call_method.php is not compatible with class cache
Using the latest twig master with the Symfony TwigBundle produces this error 
Fatal error: Symfony\Component\ClassLoader\ClassCollectionLoader::main(): Failed opening required '/Users/davidstone/cbsi/gamespot/app/cache/dev/twig_call_method.php' 

This was introduced in https://github.com/twigphp/Twig/commit/bdbfb151115be4eea22e96abd4fd01aa385516aa#diff-8a1ffef1db834d4f5fa55d72a2f43894R14
and happens because the contents of the class file is copied into classes.php  but not the file twig_call_method.php. 

I recommend we remove twig_call_method.php and use only call_user_func_array in the 1.x branch and use $object->$method(...$arguments); in the 2.0 branch since it requires php 5.6.
I would guess the extra function call to twig_call_method() counteracts any performance benefit the new ... language construct provides.

fabpot:
fixes #2243

@nicolas-grekas Instead of fixing Symfony cache system (which will only happen in a new patch release), I propose to revert these changes. It means no optimization for Twig 1.x, but the optimization is the default in Twig 2.0. So, I don't think this is a big deal.

### DIFF
--- a/lib/Twig/Template.php
+++ b/lib/Twig/Template.php
@@ -10,10 +10,6 @@
  * file that was distributed with this source code.
  */
 
-if (PHP_VERSION_ID >= 50600) {
-    require_once __DIR__.'/twig_call_method.php';
-}
-
 /**
  * Default base class for compiled templates.
  *
@@ -635,8 +631,6 @@ abstract class Twig_Template implements Twig_TemplateInterface
         try {
             if (!$arguments) {
                 $ret = $object->$method();
-            } elseif (PHP_VERSION_ID >= 50600) {
-                $ret = twig_call_method($object, $method, $arguments);
             } else {
                 $ret = call_user_func_array(array($object, $method), $arguments);
             }
--- a/lib/Twig/twig_call_method.php
+++ b/lib/Twig/twig_call_method.php
@@ -1,9 +1,0 @@
-<?php
-
-/**
- * @internal
- */
-function twig_call_method($object, $method, $arguments)
-{
-    return $object->$method(...$arguments);
-}


nicolas-grekas:
:+1:

stof:
would be great to still loo at the Symfony patch system though, as it still means there is a bug in it

lib/Twig/Template.php
```diff
@@ -10,10 +10,6 @@
  * file that was distributed with this source code.
  */
 
-if (PHP_VERSION_ID >= 50600) {
-    require_once __DIR__.'/twig_call_method.php';
```
stof: on a side note, Twig 1.x advocates 5.2.7 as min PHP version, which was also broken here

nicolas-grekas:
I don't think there is any bug in Symfony. Inlining works for most classes, not for all of them? => don't use the feat when it doesn't work. Straighforward :)

lib/Twig/Template.php
nicolas-grekas: It wasn't broken: the `if` statement guards from 5.2 failures about `__DIR__`

nicolas-grekas:
BTW, the workaround here would be easy on the 1.x branch: just turn the function into a static method.

stof:
@nicolas-grekas the inlining should be rewriting ``__DIR__``, or forbid inlining such classes. but it should not inline them and break the path silently

lib/Twig/Template.php
stof: hmm, undefined constants are indeed fine in non-executed paths. I forgot that

nicolas-grekas:
See https://github.com/symfony/symfony/pull/20455/files#r87772488

PR #2250, opened 2016-11-13
Add support for PHP 7 null coalescing operator

### LINKED ISSUE #1979: Null-coalescing operator (??): The "defined" test only works with simple variables

fabpot:
This is just one case where we can "optimize" the ternary operator. PHP 7 also supports more complex expressions that can return `null`. The rule here seems to be that you cannot have more than one level of undefinedness.

So, for instance, if `log` is defined, `log("foo") ?? "NOPE"` works, but `log($a["foo"]) ?? "NOPE"` does not if `$a["foo"]` is not defined. If we want to convert the code to always use `??` when possible, this should probably be implemented in the Optimizer node visitor, not here. But the question is: is it worth it in terms of performance? It cleans up the generated code quite a bit though.

closes #1979

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
 * 1.28.0 (2016-XX-XX)
 
+ * added support for the PHP 7 null coalescing operator for the ?? Twig implementation
  * exposed a way to access template data and methods in a portable way
  * changed context access to use the PHP 7 null coalescing operator when available
  * added the "with" tag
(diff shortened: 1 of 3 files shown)


stof:
👍

PR #2253, opened 2016-11-16
Use ** operator over function pow()

SpacePossum:
With Twig PHP dep. raised to 5.6 the arithmetic operator [`**`](http://php.net/manual/en/language.operators.arithmetic.php) can be used.

Besides slightly faster it is also harder to override an operator than it is the `pow()` function.


lib/Twig/Node/Expression/Binary/Power.php
```diff
@@ -13,11 +13,9 @@ class Twig_Node_Expression_Binary_Power extends Twig_Node_Expression_Binary
     public function compile(Twig_Compiler $compiler)
     {
         $compiler
-            ->raw('pow(')
             ->subcompile($this->getNode('left'))
-            ->raw(', ')
+            ->raw(' ** ')
             ->subcompile($this->getNode('right'))
```
stof: You can actually remove the ``compile`` method entirely, as the parent implementation will already do things properly.

SpacePossum: awesome :)  :+1:

stof:
The tests should be added in 1.x (btw, are they really missing ?)

SpacePossum:
test for this is in https://github.com/twigphp/Twig/pull/2255

fabpot:
master is Twig 3.x, so this should be switched to 2.x.

fabpot:
Thank you @SpacePossum.

PR #2254, opened 2016-11-16
Clean ups

SpacePossum:


lib/Twig/TemplateWrapper.php
```diff
@@ -48,7 +48,7 @@ public function render($context = array())
      *
      * @param array $context An array of parameters to pass to the template
      */
-    public function display($context = array())
+    public function display(array $context = array())
```
stof: Such changes must be done in 1.x, not in master

SpacePossum: Wouldn't be BC break to change the signature of a public method?

stof: Twig_TemplateWrapper has not been released yet

SpacePossum: ah check, will update :+1:

lib/Twig/Extension/Core.php
```diff
@@ -408,7 +408,7 @@ function twig_date_converter(Twig_Environment $env, $date = null, $timezone = nu
         return false !== $timezone ? $date->setTimezone($timezone) : $date;
     }
 
-    if ($date instanceof DateTime || $date instanceof DateTimeInterface) {
```
SpacePossum: note to self, cannot be backported to 1.x line, `DateTimeInterface` introduced in 5.5

stof:
Anything which can be applied in 1.x should be applied there, to simplify maintenance (as we merge 1.x into master regularly)

lib/Twig/Loader/Filesystem.php
```diff
@@ -74,7 +74,7 @@ public function getNamespaces()
     public function setPaths($paths, $namespace = self::MAIN_NAMESPACE)
     {
         if (!is_array($paths)) {
-            $paths = array($paths);
+            $paths = (array) $paths;
```
stof: this does not always have the same behavior

lib/Twig/Loader/Filesystem.php
```diff
@@ -285,7 +285,7 @@ private function isAbsolutePath($file)
     {
         return strspn($file, '/\\', 0, 1)
             || (strlen($file) > 3 && ctype_alpha($file[0])
-                && substr($file, 1, 1) === ':'
+                && ':' === $file[0]
```
stof: this is wrong. ``substr($file, 1, 1)`` is ``file[1]``, not ``file[0]``

lib/Twig/Loader/Filesystem.php
SpacePossum: true, but here it does, `$paths` must a string at this point (as hinted through the PHPDoc), passing something else will cause problems when later on `addPath` is called

lib/Twig/Loader/Filesystem.php
SpacePossum: :+1:  (proves missing test for this path btw.)

SpacePossum: is something like appveyor for Windows test left out of the repo by design or because of lack of time to set it up?

fabpot:
master is Twig 3.x, so this should be switched to 2.x.

lib/Twig/Environment.php
```diff
@@ -80,7 +80,7 @@ class Twig_Environment
      * @param Twig_LoaderInterface $loader
      * @param array                $options An array of options
      */
-    public function __construct(Twig_LoaderInterface $loader, $options = array())
+    public function __construct(Twig_LoaderInterface $loader, array $options = array())
```
fabpot: Must be reverted like other `array` type hints.

lib/Twig/Loader/Filesystem.php
fabpot: to be reverted as well.

lib/Twig/NodeVisitor/Optimizer.php
```diff
@@ -89,13 +89,14 @@ private function optimizePrintNode(Twig_Node $node, Twig_Environment $env)
...
-            $node->getNode('expr') instanceof Twig_Node_Expression_BlockReference ||
-            $node->getNode('expr') instanceof Twig_Node_Expression_Parent
+            $exprNode instanceof Twig_Node_Expression_BlockReference ||
+            $exprNode instanceof Twig_Node_Expression_Parent
         ) {
-            $node->getNode('expr')->setAttribute('output', true);
+            $exprNode->setAttribute('output', true);
 
-            return $node->getNode('expr');
+            return $exprNode;
```
fabpot: This change can probably be made in 1.x as well, right?

SpacePossum: indeed, move to other PR :+1:

stof:
> master is Twig 3.x

you are starting the dev of 3.x before even releasing 2.0 ?
And is it really worth starting to maintain a third major version already ?

fabpot:
Thank you @SpacePossum.

PR #2256, opened 2016-11-16
1 x cleanups

SpacePossum:


lib/Twig/NodeVisitor/Escaper.php
```diff
@@ -137,7 +137,7 @@ protected function needEscaping(Twig_Environment $env)
             return $this->statusStack[count($this->statusStack) - 1];
         }
 
-        return $this->defaultStrategy ? $this->defaultStrategy : false;
```
stof: cannot be done in 1.x due to the min supported PHP version

lib/Twig/NodeVisitor/Escaper.php
```diff
@@ -137,7 +137,7 @@ protected function needEscaping(Twig_Environment $env)
...
+        return $this->defaultStrategy ? : false;
```
stof: btw, when writing a short ternary, there is no space in it in PSR-2

lib/Twig/Extension/Core.php
```diff
@@ -652,11 +652,8 @@ function twig_array_merge($arr1, $arr2)
 function twig_slice(Twig_Environment $env, $item, $start, $length = null, $preserveKeys = false)
 {
     if ($item instanceof Traversable) {
-        if ($item instanceof IteratorAggregate) {
+        if ($start >= 0 && $length >= 0 && $item instanceof IteratorAggregate) {
             $item = $item->getIterator();
-        }
-
-        if ($start >= 0 && $length >= 0 && $item instanceof Iterator) {
```
SpacePossum: I don't think so, testing for `IteratorAggregate` and so change item to an `Iterator` makes the next test `instanceof Iterator` useless, so these can be combined

SpacePossum: oh wait, `getIterator` doesn't always return an `Iterator`... joy.. thanks, will revert!

stof: and you can also pass an Iterator directly, which would not be an IteratorAggregate. This would be the main thing you were breaking

lib/Twig/Extension/Core.php
```diff
@@ -991,7 +990,7 @@ function twig_escape_filter(Twig_Environment $env, $string, $strategy = 'html',
                 $string = twig_convert_encoding($string, 'UTF-8', $charset);
             }
 
-            if (0 == strlen($string) ? false : (1 == preg_match('/^./su', $string) ? false : true)) {
+            if ('' === $string ? false : (1 === preg_match('/^./su', $string) ? false : true)) {
```
iltar: isn't `1 !== preg_match(...)` the same as the inverse ternary? This case is used below more often

SpacePossum: like this https://github.com/twigphp/Twig/pull/2256/commits/15a7677aa271be4540860e48877c4f79e153c2ba ?

lib/Twig/TemplateWrapper.php
```diff
@@ -38,7 +38,7 @@ public function __construct(Twig_Environment $env, Twig_Template $template)
      *
      * @return string The rendered template
      */
-    public function render($context = array())
+    public function render(array $context = array())
```
fabpot: All array type hints must be removed. I don't want them.

lib/Twig/Extension/Core.php
```diff
@@ -1116,7 +1116,7 @@ function _twig_escape_css_callback($matches)
     // \xHH
     if (!isset($char[1])) {
         $hex = ltrim(strtoupper(bin2hex($char)), '0');
-        if (0 === strlen($hex)) {
+        if ('' === $hex) {
```
fabpot: What makes you think that this is better. I'm sure that the current version is actually faster. Have you some numbers to back this change? In any case, I think the different is really small and that it's just not worth it.

SpacePossum: >Have you some numbers to back this change?

no, the rationale behind is that it would save a function call (but I didn't check the implementation of `strlen` in a long time)

> In any case, I think the different is really small and that it's just not worth it.

agreed, reverted :)

lib/Twig/TemplateWrapper.php
```diff
@@ -99,6 +99,8 @@ public function renderBlock($name, $context = array())
      *
      * @param string $name    The block name to render
      * @param array  $context An array of parameters to pass to the template
+     *
+     * @return string
```
fabpot: not sure it adds value, but if you add it here, it should be added to `display()` as well (in the same file).

lib/Twig/TemplateWrapper.php
```diff
@@ -47,6 +47,8 @@ public function render($context = array())
      * Displays the template.
      *
      * @param array $context An array of parameters to pass to the template
+     *
+     * @return string The rendered template
```
fabpot: Actually, I read too fast. `display` does not return anything. Instead, the `return` should be removed in the method code.

lib/Twig/TemplateWrapper.php
fabpot: Same here, the `displayBlock` method does not return anything. Here the code is just wrong and should be fixed instead.

lib/Twig/TemplateWrapper.php
```diff
@@ -102,26 +102,7 @@ public function renderBlock($name, $context = array())
...
-        } catch (Throwable $e) {
-            while (ob_get_level() > $level) {
-                ob_end_clean();
-            }
-
-            throw $e;
-        }
-
-        return ob_get_clean();
+        $this->template->displayBlock($name, $this->env->mergeGlobals($context));
```
fabpot: I've fixed this in #2257 as renderBlock() code is wrong.

SpacePossum: you want my fix of `display()` in there as well, than I can revert that change here as well and have no more changes on the file in this PR

fabpot: Can you revert this as it will help when merging, thanks.

SpacePossum: done

fabpot:
Thank you @SpacePossum.

PR #2262, opened 2016-11-18
Call clearstatcache() before calling is_dir().

hanabokuro:
The result of is_dir() is cached.
So call clearstatcache() before calling 2nd is_dir().

### DIFF
--- a/lib/Twig/Cache/Filesystem.php
+++ b/lib/Twig/Cache/Filesystem.php
@@ -58,8 +58,11 @@ class Twig_Cache_Filesystem implements Twig_CacheInterface
     {
         $dir = dirname($key);
         if (!is_dir($dir)) {
-            if (false === @mkdir($dir, 0777, true) && !is_dir($dir)) {
-                throw new RuntimeException(sprintf('Unable to create the cache directory (%s).', $dir));
+            if (false === @mkdir($dir, 0777, true)) {
+                clearstatcache(false, $dir);
+                if (!is_dir($dir)) {
+                    throw new RuntimeException(sprintf('Unable to create the cache directory (%s).', $dir));
+                }
             }
         } elseif (!is_writable($dir)) {
             throw new RuntimeException(sprintf('Unable to write in the cache directory (%s).', $dir));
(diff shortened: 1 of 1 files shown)


lib/Twig/Cache/Filesystem.php
```diff
@@ -58,8 +58,11 @@ public function write($key, $content)
     {
         $dir = dirname($key);
         if (!is_dir($dir)) {
-            if (false === @mkdir($dir, 0777, true) && !is_dir($dir)) {
-                throw new RuntimeException(sprintf('Unable to create the cache directory (%s).', $dir));
+            if (false === @mkdir($dir, 0777, true)) {
+                clearstatcache(false, $dir);
+                if (!is_dir($dir)) {
+                    throw new RuntimeException(sprintf('Unable to create the cache directory (%s).', $dir));
+                }
```
SpacePossum: If this is done to deal with a race condition where the directory is made by another process  than the directory might not be writable at this point, which is not checked as is done on line 67.

The call to [clearstatcache](http://php.net/manual/en/function.clearstatcache.php) will not do what you expected, as the second param (`$dir`) is only used if the first param is `true`.

hanabokuro: mkdir()  will create directory & set directory permission atomically.
So no need to care race condition problem with is_writable() ?

fabpot:
Thank you @hanabokuro.

Tobion:
This contradicts #1843 again.

@hanabokuro in which case did you experience this problem?

fabpot:
@hanabokuro Can you tell use when you need this?

hanabokuro:
It happens when two processes try to create a cache at the same time when the cache does not yet exist just after deployment.

* Process a calls Twig_Cache_Filesystem::write() and call is_dir() and got false.
* Process b calls Twig_Cache_Filesystem::write() and call is_dir() and got false.
* Process a call mkdir() and got success.
* Process a call mkdir() and got false. because Process a made that dir.
   And throw RuntimeException(sprintf('Unable to create the cache directory (%s).', $dir));

Tobion:
The second `is_dir` check should not return false then and be cached. Can you try to reproduce a negative is_dir check which is cached?

hanabokuro:
realpath_cache is local cache, not system wide.
So process a has cache, but Process b doesnt have.

Tobion:
That does not explain anything. A negative is_dir check is NOT cached. So `clearstatcache` makes no difference.

fabpot:
Closing.

PR #2265, opened 2016-11-18
Optimize usage of Traversable/Iterator

SpacePossum:


lib/Twig/Extension/Core.php
```diff
@@ -821,10 +821,32 @@ function _twig_default_filter($value, $default = '')
 function twig_get_array_keys_filter($array)
 {
     if ($array instanceof Traversable) {
-        return array_keys(iterator_to_array($array));
+        if ($array instanceof IteratorAggregate) {
+            $array = $array->getIterator();
+        } // else $array must be an instance of IteratorAggregate here, we could call getIterator on that one etc. until we get an iterator
+
+        if ($array instanceof Iterator) {
```
SpacePossum: the main advantage here is not accessing the objects, so classes implementing Iterator that use lazy loading of the objects are not triggered to do so on key request only

lib/Twig/Extension/Core.php
```diff
@@ -821,10 +821,32 @@ function _twig_default_filter($value, $default = '')
...
+            while($array->valid()) {
+                $keys[] = $array->key();
+                $array->next();
+            }
+
+            return $keys;
+        }
+
+        $keys = [];
+        foreach ($array as $key => $item) {
```
SpacePossum: advantage here is the removing of the whole iterator to array copy

lib/Twig/Extension/Core.php
```diff
@@ -901,9 +923,25 @@ function twig_in_filter($value, $compare)
     } elseif (is_string($compare) && (is_string($value) || is_int($value) || is_float($value))) {
         return '' === $value || false !== strpos($compare, (string) $value);
     } elseif ($compare instanceof Traversable) {
-        return in_array($value, iterator_to_array($compare, false), is_object($value) || is_resource($value));
+        if (is_object($value) || is_resource($value)) {
```
SpacePossum: Advantage here is not copying the data from the Traversable object to an array and then start searching for the item.

lib/Twig/Extension/Core.php
```diff
@@ -821,10 +821,32 @@ function _twig_default_filter($value, $default = '')
...
+        $keys = [];
+        foreach ($array as $key => $item) {
+            $keys[] = $key;
+        }
+
+        return $keys;
     }
 
     if (!is_array($array)) {
+        @trigger_error('Using "keys" must be done on an array or Traversable, deprecated since version 1.29 and will be removed in Twig 2.0', E_USER_DEPRECATED);
```
stof: this is not consistent with the way ``{% for %}`` works in Twig though

lib/Twig/Extension/Core.php
```diff
@@ -821,10 +821,32 @@ function _twig_default_filter($value, $default = '')
...
+            $array->rewind();
+            while($array->valid()) {
+                $keys[] = $array->key();
+                $array->next();
+            }
+
+            return $keys;
+        }
+
+        $keys = [];
```
stof: short array syntax cannot be used in Twig 1.x due to supported PHP versions (and should not be used in 2.x either to keep things maintainable when merging 1.x into 2.x)

lib/Twig/Extension/Core.php
```diff
@@ -901,9 +923,25 @@ function twig_in_filter($value, $compare)
...
+                if ($item == $value) {
+                    return true;
+                }
+            }
+        }
+
+        return false;
     }
 
+    @trigger_error('Using "in" must be done on an array, string or Traversable, deprecated since version 1.29 and will be removed in Twig 2.0', E_USER_DEPRECATED);
```
stof: the message is not easy to understand IMO. It is not clear about what is deprecated

lib/Twig/Extension/Core.php
```diff
@@ -821,10 +821,32 @@ function _twig_default_filter($value, $default = '')
 function twig_get_array_keys_filter($array)
 {
     if ($array instanceof Traversable) {
-        return array_keys(iterator_to_array($array));
+        if ($array instanceof IteratorAggregate) {
+            $array = $array->getIterator();
+        } // else $array must be an instance of IteratorAggregate here, we could call getIterator on that one etc. until we get an iterator
```
stof: this sentence is weird. And the part ``else $array must be an instance of IteratorAggregate here`` looks wrong to me

lib/Twig/Extension/Core.php
SpacePossum: So something like:
`$twigParams['c'] = function(){ return 1;};`

`{% for a,b in c %}`

is legit?

stof: yes (but useless as it does always an empty loop). Look at ``twig_ensure_traversable`` (which would be better named ``twig_ensure_iterable`` now that PHP has a proper name for this concept, but this is not a big issue as it is internal anyway)

lib/Twig/Extension/Core.php
SpacePossum: Removed.

What I was thinking of when making the note is that when we know the object being an instance of `Traversable` we can always get the `Iterator`. Now we only try it one time with `getIterator()`.
We could do something like:
```php
if ($array instanceof Traversable) { // so either implements IteratorAggregate or Iterator.
    if ($array instanceof IteratorAggregate) {
        do {
            // returns either a IteratorAggregate or Iterator.
            $array = $array->getIterator(); 
        } while($array instanceof IteratorAggregate);
    }

    // $array is always an instance of Iterator here
}
```

lib/Twig/Extension/Core.php
```diff
@@ -821,11 +821,27 @@ function _twig_default_filter($value, $default = '')
 function twig_get_array_keys_filter($array)
 {
     if ($array instanceof Traversable) {
-        return array_keys(iterator_to_array($array));
-    }
+        if ($array instanceof IteratorAggregate) {
+            $array = $array->getIterator();
+        }
 
-    if (!is_array($array)) {
```
stof: this case was lost entirely, which is a BC break

SpacePossum: ah, must how thrown it away when I removed the deprecation warning, fixed, thanks

lib/Twig/Extension/Core.php
derrabus: Or shorter:

```php
while ($array instanceof IteratorAggregate) {
    $array = $array->getIterator();
}
```

SpacePossum: :+1: nice, thanks :)
I might make a PR out of this after this one, trying to keep the scope down (which I always find hard ;) )

SpacePossum:
`merge` and `join` can be optimized as well, however I don't have the time (soon) to do this.
The current fixes are the ones with the most gain.

lib/Twig/Extension/Core.php
```diff
@@ -821,7 +821,27 @@ function _twig_default_filter($value, $default = '')
...
+        if ($array instanceof Iterator) {
+            $keys = array();
+            $array->rewind();
+            while ($array->valid()) {
+                $keys[] = $array->key();
+                $array->next();
+            }
+
+            return $keys;
+        }
```
nikic: Why this code instead of just the foreach below? At least for any built-in iterator, this is going to be slower.

stof: This was requested for iterators doing some lazy-building of the value in ``current()`` to avoid doing the heavy work when only keys are needed

SpacePossum: It depends on the implementation of the Iterator. The case here is getting the keys only, we are not interested in the items.
For example consider something like this:
```php
<?php

final class Repository implements Iterator {

    private $dataBaseIds;
    private $position = 0;

    public function __construct(array $dataBaseIds) {

        $this->dataBaseIds = $dataBaseIds;
    }

    public function rewind() {
        $this->position = 0;
    }

    public function current() {
        return $this->loadItemFromDatabase($this->dataBaseIds[$this->position]);
    }

    public function key() {
        return $this->dataBaseIds[$this->position];
    }

    public function next() {
        ++$this->position;
    }

    public function valid() {
        return $this->position < count($this->dataBaseIds);
    }

    private function loadItemFromDatabase($id)
    {
        echo "\nloading ".$id;
        sleep(1);
        return 'item:'.$id;
    }
}

// Current Twig implementation
function getKeysCurrentTwigImplementation(Iterator $r)
{
    $keys = [];
    foreach ($r as $key => $item) {
        $keys[] = $key;
    }

    return $keys;
}

// PR'ed version
function getKeysPR(Iterator $r)
{
    $keys = [];

    $r->rewind();
    while ($r->valid()) {
        $keys[] = $r->key();
        $r->next();
    }

    return $keys;
}


$r = new Repository([8,9,10]);
$keys = getKeysCurrentTwigImplementation($r);
$keys = getKeysPR($r);

var_dump($keys);
```

The trick is that the `foreach` `key` `value` way always calls the `current` on the iterator, even when it is not used.
Ideal would be that PHP would allow for something like;
```php
foreach($iterator => $key;) {
}
```
so we would get the key, but not the value (another benefit would be that no 'unused variable' has to be made, keeping SCA's happy).

lib/Twig/Extension/Core.php
```diff
@@ -901,7 +921,21 @@ function twig_in_filter($value, $compare)
     } elseif (is_string($compare) && (is_string($value) || is_int($value) || is_float($value))) {
         return '' === $value || false !== strpos($compare, (string) $value);
     } elseif ($compare instanceof Traversable) {
-        return in_array($value, iterator_to_array($compare, false), is_object($value) || is_resource($value));
+        if (is_object($value) || is_resource($value)) {
+            foreach ($compare as $key => $item) {
```
SpacePossum: :+1: removed

lib/Twig/Extension/Core.php
nikic: If you've hit this problem with real-world iterators, this seems reasonable. I just wouldn't have expected that iterators that have both useful keys and a key/value asymmetry occur sufficiently commonly in practice to merit this special-casing.

There is an RFC (https://wiki.php.net/rfc/foreach_void) for a `foreach ($it as $key => void)` syntax. IIRC this RFC never went to discussion, due to bad reception OTR, but maybe this is a good example to reconsider it.

SpacePossum: I think iterators like that are becoming more and more popular because of the `Iterable` addition to PHP 7 (which is awesome btw. :) ). Iterators like this are very powerful and reduce a lot of boiler template code.

I'm not sure what OTR mean, but I think the RFC is nice.

Sorry @jbafford for pinging you here, I couldn't open an issue over @ https://github.com/jbafford/php-src/tree/foreachvoid to get in contact with you. Are you still working on the RFC or willing to pick it up again?

SpacePossum:
@fabpot I know you are not a (big) fan of micro optimizations, code style tweaks and/or bloating the code base for a small gain or supporting an edge case.
However I think the changes in this PR will be big performance gain, not only for the advanced iterator case, but also for build in iterators as this PR removes not need `iterator_to_array` calls which reduces the creating, copying and re-iterating arrays.

(btw. do you want me to add the CS changes as proposed by fabbot.io in this PR? I did a run on the whole code base with the current CS config and there a lot changes you might (not) want, maybe better to do all those in one big CS PR?)

fabpot:
@SpacePossum Regarding fabbot's suggestion, I'm not interested in them. Unfortunately, PHP-CS-Fixer 2 broke quite a few things for us. And as you can see, it wants to add `\` while it's not needed as we are not using namespaces. Everything else is details I don't care about... at least in a PR. I need to do it (unfortunately) myself to easily propagate them to all branches.

SpacePossum:
>Regarding fabbot's suggestion, I'm not interested in them. Unfortunately, PHP-CS-Fixer 2 broke quite a few things for us

sorry to hear, we tried to be as careful as we could

> And as you can see, it wants to add \ while it's not needed as we are not using namespaces. 

just a hint, it is done by the rule `php_unit_fqcn_annotation` which is in the default set for Symfony. Maybe it is worth it to create a `.php_cs.dist` file in the Twig repo to tweak it. But than again I don't know the internals of `fabbot.io`.

fabpot:
Ok, I'm going to submit a PR to tweak PHP-CS-Fixer configuration.

Do you know which rule I need to disable to avoid changes like this:

```php
-    public function __construct(Twig_NodeInterface $name, $template = null, $lineno, $tag = null)
+    public function __construct(Twig_NodeInterface $name, $template, $lineno, $tag = null)
```

And the one to avoid changing heredocs?

SpacePossum:
>Do you know which rule I need to disable to avoid changes like this:

that would be:
`no_unreachable_default_argument_value`

>And the one to avoid changing heredocs?

my guess (pretty sure though) is:
`heredoc_to_nowdoc`

fabpot:
@SpacePossum You're right on both of them :) Thanks. Last one is about one-line anonymous functions that are expanded on several lines, I've just had a look, and I can't figure out which rule it might be. #2308 for the config.

SpacePossum:
>Last one is about one-line anonymous functions that are expanded on several lines

@fabpot thats the same as this [one](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2413#issuecomment-267915932) I'm afraid, i.e. it needs a new configuration option on the `braces`-fixer we don't have yet.

lib/Twig/Extension/Core.php
jbafford: @SpacePossum Yes, actually. I rebased it against master about two weeks ago; I just needed to finish building test cases to show that there actually is a performance benefit before raising the issue again. (One of the issues I came across last time was that people seemed to think that `foreach($arr as $key => $unused)` or `foreach($arr as $key => $_)` were perfectly reasonable alternatives, which is to say, they didn't see that ignoring values was all that useful.)

Is there a specific real-world use case of this that I can look at to use as a performance example?

lib/Twig/Extension/Core.php
fabpot: Can you add a test to cover this change, like throwing an exception when trying to get the value?

SpacePossum: I've added the test. I've also added a test to make sure the iterator is not iterated to over all elements if not needed. Other changes are to make sure all code paths are tested.

lib/Twig/Extension/Core.php
SpacePossum: Awesome, thanks for all the work :)
I'm looking for a real-world cases that are opensource.
However such cases are plenty and easy to think of. When iterate data:
- with fix length/size files data (for example binary files)
- with delimiters ('csv', 'json')
- such as database result sets based on identifiers
- such as cached identifiers with the values in some other source (like a database)

Whenever you want filter by key (or when only interested in the key for whatever reason) there is a clear advantage by not fetching the value (for example hydrating an item by doctrine2 is pretty expensive especially if you are not going to need it).

For example:
small gain [here](https://github.com/doctrine/doctrine2/blob/24377156b41a5fca37b26579c28f824f779d41ef/lib/Doctrine/ORM/Query/TreeWalkerChainIterator.php#L64) and even for more gain after some changes in classes like [this](https://github.com/doctrine/doctrine2/blob/cd11723e6310b5973f50e7d9819c58916721d290/lib/Doctrine/ORM/Internal/Hydration/IterableResult.php)

lib/Twig/Extension/Core.php
```diff
@@ -821,7 +821,27 @@ function _twig_default_filter($value, $default = '')
...
+                $array->next();
+            }
+
+            return $keys;
+        }
+
+        $keys = array();
+        foreach ($array as $key => $item) {
+            $keys[] = $key;
+        }
```
fabpot: Out of curiosity, `array_keys($array)` would be worse (performance wise)?

stof: We can still reach this point with Traversable objects which are not Iterator (happens for classes implemented in C)

fabpot:
Thank you @SpacePossum.

SpacePossum:
thanks at all reviewers @stof @nikic @derrabus & @fabpot (also for merging :) )
good luck @jbafford with the RFC on PHP, really hope it makes it!

PR #2269, opened 2016-11-18
Bugfix: When rendering blocks of other templates, don't pass the local blocks array

derrabus:
This PR fixes symfony/symfony#20556.

When referring to a block of another template from within a template that also has a block with the same name, Twig renders the wrong block.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,6 @@
 * 1.28.1 (2016-XX-XX)
 
- * n/a
+ * fixed block() function when used with a template argument
 
 * 1.28.0 (2016-11-17)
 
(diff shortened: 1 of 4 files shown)


stof:
Please add a test covering this

derrabus:
The class seems to be completely untested so far. I am a bit unfamiliar with the way the compilation is tested, but I can look into that.

fabpot:
@derrabus I'm going to take care of adding some tests :)

javiereguiluz:
I've just tested this and it fixes the reported issues in Symfony. Thanks @derrabus

fabpot:
I will push a new commit on your branch (the defined block must be changed as well).

derrabus:
I tried to write a test by looking into other tests for expression classes. I'll leave the branch now as it is. Thanks for taking over, @fabpot. :-)

fabpot:
As I have a flight, I will probably work on this later today. Expect a release in the next day. @derrabus Thanks for investigating the issue.

PR #2271, opened 2016-11-19
Deprecate support for mbstring.func_overload != 0

fabpot:
see https://wiki.php.net/rfc/deprecations_php_7_2

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,6 @@
-* 1.28.3 (2016-XX-XX)
+* 1.29.0 (2016-XX-XX)
 
- * n/a
+ * deprecated support for mbstring.func_overload != 0
 
 * 1.28.2 (2016-11-23)
 
--- a/composer.json
+++ b/composer.json
@@ -40,7 +40,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "1.28-dev"
+            "dev-master": "1.29-dev"
         }
     }
 }
--- a/ext/twig/php_twig.h
+++ b/ext/twig/php_twig.h
@@ -15,7 +15,7 @@
 #ifndef PHP_TWIG_H
 #define PHP_TWIG_H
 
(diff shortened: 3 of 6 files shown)


nicolas-grekas:
:+1: (should throw an exception in 2 then?)

stof:
:+1: (I really hope this deprecation will be accepted in PHP 7.2)

SpacePossum:
:+1:

>should throw an exception in 2 then?

:+1: And this cast can be removed than as well https://github.com/twigphp/Twig/blob/1.x/lib/Twig/Extension/Core.php#L677 (https://github.com/twigphp/Twig/pull/1453)

stof:
@SpacePossum no, this case is about using mbstring when possible to support UTF-8 better. It is not related to func_overload

SpacePossum:
Fabien
>Deprecate support for mbstring.func_overload != 0

@stof 
>is not related to func_overload

I'm confused here and so still think the string cast can still be removed as soon as ` support for mbstring.func_overload != 0` is deprecated. However it is a minor that can be put on a backlog somewhere :)

stof:
@SpacePossum I'm talking about the case you linked in your previous comment. This case has **nothing** to do with supporting func_overload. It is about supporting UTF-8 properly when you have mbstring enabled

SpacePossum:
It has anything to do with `func_overload` :)

As with `func_overload` is enabled the function `mb_substr` returns `false` and not `''` on some cases (on some Debain and Ubuntu ('stable') releases). I'm not ref. polyfil here, but direct function calls on the exension. That is why the cast is added.

Quoting from the PR (https://github.com/twigphp/Twig/pull/1453):
>I've added: mbstring.func_overload = 2

So if you deprecate the support for `mbstring.func_overload` on lexer level (i.e. Twig as a whole), you can ditch the edge case on `mb_substr` as well, as it _only_ occurs when using the extension with `func_overload` enabled.

PR #2272, opened 2016-11-21
Fix unconsistent behavior with "get"  and "is"

wolfx:


stof:
We need to add tests for it.

lib/Twig/Template.php
```diff
@@ -603,6 +603,8 @@ protected function getAttribute($object, $item, array $arguments = array(), $typ
                 }
             } else {
                 $methods = get_class_methods($object);
+                // sort values to have consistent behavior with lines 611 to 631
+                sort($methods);
```
stof: AFAIK, the doc already says that ``get*`` wins over ``is*``. This sorting fixes this behavior because the recent refactoring done by @nicolas-grekas lost the precedence order (making the first found win instead)

stof: The sorting should be outside the ``else``, as it is necessary also for ``Twig_Template`` child classes actually.

fabpot:
Thank you @wolfx.

fabpot:
test added in 8778313

stof:
@fabpot for Twig 2.x, we should use a different fix than ``sort`` though, as ``has*`` should be checked after ``is*`` (otherwise it breaks BC when both exist, and this is not a BC break we want), and the sorting would not solve the issue here, as ``has*`` would be sorted before ``is*`` (using sorting was a nice trick for get vs is though)

PR #2273, opened 2016-11-22
Make Twig_Error report real source path when possible

nicolas-grekas:
To get useful stack traces like this one:
capture du 2016-11-22 12-16-07

### DIFF
--- a/lib/Twig/Error.php
+++ b/lib/Twig/Error.php
@@ -39,6 +39,9 @@ class Twig_Error extends Exception
     protected $rawMessage;
     protected $previous;
 
+    private $sourcePath;
+    private $sourceCode;
+
     /**
      * Constructor.
      *
@@ -51,13 +54,23 @@ class Twig_Error extends Exception
      *
      * By default, automatic guessing is enabled.
      *
-     * @param string    $message  The error message
-     * @param int       $lineno   The template line where the error occurred
-     * @param string    $name     The template logical name where the error occurred
-     * @param Exception $previous The previous exception
+     * @param string                  $message  The error message
+     * @param int                     $lineno   The template line where the error occurred
+     * @param Twig_Source|string|null $source   The source context where the error occurred
+     * @param Exception               $previous The previous exception
      */
-    public function __construct($message, $lineno = -1, $name = null, Exception $previous = null)
+    public function __construct($message, $lineno = -1, $source = null, Exception $previous = null)
     {
+        if (null === $source) {
+            $name = null;
+        } elseif (!$source instanceof Twig_Source) {
+            // for compat with the Twig C ext., passing the template name as string is accepted
+            $name = $source;
+        } else {
+            $name = $source->getName();
+            $this->sourceCode = $source->getCode();
+            $this->sourcePath = $source->getPath();
+        }
         if (PHP_VERSION_ID < 50300) {
             $this->previous = $previous;
             parent::__construct('');
@@ -68,7 +81,7 @@ class Twig_Error extends Exception
         $this->lineno = $lineno;
         $this->filename = $name;
 
-        if (-1 === $lineno || null === $name) {
+        if (-1 === $lineno || null === $name || null === $this->sourcePath) {
             $this->guessTemplateInfo();
         }
 
@@ -92,11 +105,11 @@ class Twig_Error extends Exception
      *
      * @return string The name
      *
-     * @deprecated since 1.27 (to be removed in 2.0). Use getTemplateName() instead.
+     * @deprecated since 1.27 (to be removed in 2.0). Use getSourceContext() instead.
      */
     public function getTemplateFile()
     {
-        @trigger_error(sprintf('The "%s" method is deprecated since version 1.27 and will be removed in 2.0. Use getTemplateName() instead.', __METHOD__), E_USER_DEPRECATED);
+        @trigger_error(sprintf('The "%s" method is deprecated since version 1.27 and will be removed in 2.0. Use getSourceContext() instead.', __METHOD__), E_USER_DEPRECATED);
 
         return $this->filename;
     }
@@ -106,11 +119,11 @@ class Twig_Error extends Exception
      *
      * @param string $name The name
      *
-     * @deprecated since 1.27 (to be removed in 2.0). Use setTemplateName() instead.
+     * @deprecated since 1.27 (to be removed in 2.0). Use setSourceContext() instead.
      */
     public function setTemplateFile($name)
     {
-        @trigger_error(sprintf('The "%s" method is deprecated since version 1.27 and will be removed in 2.0. Use setTemplateName() instead.', __METHOD__), E_USER_DEPRECATED);
+        @trigger_error(sprintf('The "%s" method is deprecated since version 1.27 and will be removed in 2.0. Use setSourceContext() instead.', __METHOD__), E_USER_DEPRECATED);
 
         $this->filename = $name;
 
@@ -121,9 +134,13 @@ class Twig_Error extends Exception
      * Gets the logical name where the error occurred.
      *
      * @return string The name
+     *
+     * @deprecated since 1.29 (to be removed in 2.0). Use getSourceContext() instead.
      */
     public function getTemplateName()
     {
+        @trigger_error(sprintf('The "%s" method is deprecated since version 1.29 and will be removed in 2.0. Use getSourceContext() instead.', __METHOD__), E_USER_DEPRECATED);
+
         return $this->filename;
     }
 
@@ -131,10 +148,15 @@ class Twig_Error extends Exception
      * Sets the logical name where the error occurred.
      *
      * @param string $name The name
+     *
+     * @deprecated since 1.29 (to be removed in 2.0). Use setSourceContext() instead.
      */
     public function setTemplateName($name)
     {
+        @trigger_error(sprintf('The "%s" method is deprecated since version 1.29 and will be removed in 2.0. Use setSourceContext() instead.', __METHOD__), E_USER_DEPRECATED);
+
         $this->filename = $name;
+        $this->sourceCode = $this->sourcePath = null;
 
         $this->updateRepr();
     }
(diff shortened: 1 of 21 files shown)


lib/Twig/Error.php
```diff
@@ -34,10 +34,11 @@
 class Twig_Error extends Exception
 {
     protected $lineno;
-    // to be renamed to name in 2.0
+    /** @deprecated to be renamed to name in 2.0 */
     protected $filename;
     protected $rawMessage;
     protected $previous;
+    protected $path;
```
stof: please make new properties private, to avoid having to deal with BC on them

lib/Twig/Error.php
```diff
@@ -34,10 +34,11 @@
 class Twig_Error extends Exception
 {
     protected $lineno;
-    // to be renamed to name in 2.0
+    /** @deprecated to be renamed to name in 2.0 */
```
stof: this is *not* deprecated in 1.x

lib/Twig/Error.php
```diff
@@ -67,8 +69,9 @@ public function __construct($message, $lineno = -1, $name = null, Exception $pre
 
         $this->lineno = $lineno;
         $this->filename = $name;
+        $this->path = $path;
 
-        if (-1 === $lineno || null === $name) {
+        if (-1 === $lineno || null === $name || null === $path) {
```
stof: -1 for the null check on the path, as it would force guessing for any case where the template does not have a path because of not using a template loader relying on the filesystem. And the template guessing is a performance killer, which is why we did our best to provide the template name at compilation time for runtime errors whenever possible.

lib/Twig/Lexer.php
```diff
@@ -26,8 +26,9 @@ class Twig_Lexer implements Twig_LexerInterface
     protected $states;
     protected $brackets;
     protected $env;
-    // to be renamed to $name in 2.0 (where it is private)
+    /** @deprecated to be renamed to $name in 2.0 (where it is private) */
```
stof: same here. It is **not** deprecated

lib/Twig/Lexer.php
```diff
@@ -26,8 +26,9 @@ class Twig_Lexer implements Twig_LexerInterface
...
     protected $filename;
+    protected $path;
```
stof: private please

test/Twig/Tests/ErrorTest.php
```diff
@@ -54,6 +65,37 @@ public function testTwigExceptionAddsFileAndLineWhenMissingWithInheritanceOnDisk
...
+            $this->assertEquals(3, $e->getTemplateLine());
+            $this->assertEquals('index.html', $e->getTemplateName());
+            $this->assertEquals(3, $e->getLine());
+            $this->assertEquals(strtr(dirname(__FILE__).'/Fixtures/errors/index.html', '/', DIRECTORY_SEPARATOR), $e->getFile());
+        }
+
+        try {
+            $template->render(array('foo' => new Twig_Tests_ErrorTest_Foo()));
+
+            $this->fail();
```
stof: please use separate tests instead of doing several tests in the same PHPUnit test. Ii would make things much easier to maintain (and would provide better reporting on failures, as your second test currently does not run at all if the first one fails)

nicolas-grekas: tests splitted

lib/Twig/Error.php
nicolas-grekas: If you look at the full patch, you'll notice that when a no line is provided when building an exception, no path is provided.
And reciprocally when a line is provided, a path is always provided.
This means this added condition doesn't trgger any new "guess" in the code.
Yet, this should be added for consistency IMO, because having a partially guessed state makes no sense.

nicolas-grekas: also note that the default return value of getPath is the empty string, not null.

lib/Twig/Error.php
```diff
@@ -55,8 +57,9 @@ class Twig_Error extends Exception
      * @param int       $lineno   The template line where the error occurred
      * @param string    $name     The template logical name where the error occurred
      * @param Exception $previous The previous exception
+     * @param string    $path     The template source path if any
      */
-    public function __construct($message, $lineno = -1, $name = null, Exception $previous = null)
+    public function __construct($message, $lineno = -1, $name = null, Exception $previous = null, $path = null)
```
nicolas-grekas: done

lib/Twig/Error.php
```diff
@@ -161,6 +174,28 @@ public function setTemplateLine($lineno)
         $this->updateRepr();
     }
 
+    /**
+     * Gets the path of the twig file where the error occurred.
+     *
+     * @return string The name
+     */
+    public function getTemplatePath()
```
nicolas-grekas: I guess this should ow be turned to getSourceContext

lib/Twig/Error.php
```diff
@@ -51,13 +54,23 @@ class Twig_Error extends Exception
...
+     * @param Twig_Source $source   The source context where the error occurred
+     * @param Exception   $previous The previous exception
      */
-    public function __construct($message, $lineno = -1, $name = null, Exception $previous = null)
+    public function __construct($message, $lineno = -1, $source = null, Exception $previous = null)
     {
+        if (null === $source) {
+            $name = null;
+        } elseif (!$source instanceof Twig_Source) {
+            // for compat with the Twig C ext., passing the template name as string is accepted
```
stof: for BC too.

And should we update the C ext to make it call ``getSourceContext`` instead of ``getTemplateName`` ? I would say yes (should be an easy change btw, even for you or me), and then we can deprecate this case

lib/Twig/Error.php
```diff
@@ -51,13 +54,23 @@ class Twig_Error extends Exception
      *
      * By default, automatic guessing is enabled.
      *
-     * @param string    $message  The error message
-     * @param int       $lineno   The template line where the error occurred
-     * @param string    $name     The template logical name where the error occurred
-     * @param Exception $previous The previous exception
+     * @param string      $message  The error message
+     * @param int         $lineno   The template line where the error occurred
+     * @param Twig_Source $source   The source context where the error occurred
```
stof: ``Twig_Source|null`` (or ``Twig_Source|string|null`` if we don't deprecate the case of passing a string)

lib/Twig/Error.php
```diff
@@ -135,6 +148,7 @@ public function getTemplateName()
     public function setTemplateName($name)
```
stof: should we deprecate this method ?

lib/Twig/Error.php
```diff
@@ -161,6 +175,30 @@ public function setTemplateLine($lineno)
         $this->updateRepr();
     }
 
+    /**
+     * Gets the path of the twig file where the error occurred.
+     *
+     * @return string The name
```
stof: wrong phpdoc

lib/Twig/Error.php
```diff
@@ -161,6 +175,30 @@ public function setTemplateLine($lineno)
...
+     */
+    public function getSourceContext()
+    {
+        return $this->filename ? new Twig_Source($this->sourceCode, $this->filename, $this->sourcePath) : null;
+    }
+
+    /**
+     * Sets the path of the twig file where the error occurred.
+     *
+     * @param string $name The name
```
stof: wrong phpdoc

lib/Twig/Error.php
```diff
@@ -161,6 +175,30 @@ public function setTemplateLine($lineno)
...
+     *
+     * @return string The name
+     */
+    public function getSourceContext()
+    {
+        return $this->filename ? new Twig_Source($this->sourceCode, $this->filename, $this->sourcePath) : null;
+    }
+
+    /**
+     * Sets the path of the twig file where the error occurred.
```
stof: wrong description

lib/Twig/Error.php
nicolas-grekas: There are some place in the code base that only have the template name locally, and rely on the guessing mechanism for filling the other info. Thus, I think we should keep this method asis: it would make little sense for these code paths to create a mock Twig_Source instance just to satisfy the new signature.

lib/Twig/Error.php
stof: We should still change the C ext to avoid triggering guessing for the path when using the C ext

lib/Twig/Error.php
```diff
@@ -161,6 +175,28 @@ public function setTemplateLine($lineno)
         $this->updateRepr();
     }
 
+    /**
+     * Gets the source context of the twig template where the error occurred.
```
fabpot: That's "Twig" everywhere, not "twig"

lib/Twig/Error.php
```diff
@@ -161,6 +175,28 @@ public function setTemplateLine($lineno)
...
+     * @return Twig_Source|null
+     */
+    public function getSourceContext()
+    {
+        return $this->filename ? new Twig_Source($this->sourceCode, $this->filename, $this->sourcePath) : null;
+    }
+
+    /**
+     * Sets the source context of the twig template where the error occurred.
+     */
```
fabpot: But honestly, I would remove those phpdocs :)

lib/Twig/Error/Loader.php
```diff
@@ -24,8 +24,18 @@
  */
 class Twig_Error_Loader extends Twig_Error
 {
-    public function __construct($message, $lineno = -1, $filename = null, Exception $previous = null)
+    private $guess = false;
+
+    public function __construct($message, $lineno = -1, $source = null, Exception $previous = null)
+    {
+        parent::__construct($message, false, null, $previous);
+        $this->guess = true;
```
fabpot: I'm confused by this. We should never guess here, but you set it to `true` and I don't anywhere in the code a way to turn it off. Can you explain?

nicolas-grekas: We should never guess at instantiation time, but later, on demand, is needed (when guess is explicitly called.)

fabpot: When is it called on-demand? As noted in the phpdoc of the class, I don't think there is a case where we can guess for such exceptions.

nicolas-grekas: See e.g. https://github.com/twigphp/Twig/pull/2273/files#diff-8a1ffef1db834d4f5fa55d72a2f43894R129

nicolas-grekas: seel also https://github.com/twigphp/Twig/pull/2273/files#diff-8a1ffef1db834d4f5fa55d72a2f43894L214

lib/Twig/Error.php
nicolas-grekas: it is deprecated now

fabpot:
Thank you @nicolas-grekas.

PR #2276, opened 2016-11-23
Support nested comments

WietseWind:
Support nested comments, by checking the remaining amount(s) of comment-open vs comment-close tags.

### DIFF
--- a/lib/Twig/Lexer.php
+++ b/lib/Twig/Lexer.php
@@ -324,10 +324,8 @@ class Twig_Lexer implements Twig_LexerInterface
 
         $this->moveCursor($moveCursorBody.$match[0][0]);
 
-        if($nestedComments > 0)
-        {
-            for($nc=0;$nc<$nestedComments;$nc++)
-            {
+        if($nestedComments > 0) {
+            for($nc=0;$nc<$nestedComments;$nc++) {
                 $this->lexComment();
             }
         }
--- a/lib/Twig/Lexer.php
+++ b/lib/Twig/Lexer.php
@@ -324,8 +324,8 @@ class Twig_Lexer implements Twig_LexerInterface
 
         $this->moveCursor($moveCursorBody.$match[0][0]);
 
-        if($nestedComments > 0) {
-            for($nc=0;$nc<$nestedComments;$nc++) {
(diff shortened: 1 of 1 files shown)


stof:
this is a BC break though, because ``{#    {#   #}fooo`` would not be valid anymore, and would not display ``fooo`` anymore.

and btw, I don't think other languages allow nested comments either (it would mean that *some* text inside the comment would still be code)

WietseWind:
@stof How do you mean? The example you give would throw an exception either way (`Unclosed comment in "xxxx.twig"`; there's simply a closing-tag missing.

This PR enables you to do the following:

```twig
<p>Let's display this</p>

{# Let's skip this
    {% for i in 1..2 %}
       <li>{{ i }}</li> {# Never mind this comment #}  
    {% endfor %}
#}

<p>And let's display this</p>
```

stof:
It is perfectly valid currently, as the second ``{#`` is **not** a comment opening tag. It is part of the comment text. Here is the proof: https://twigfiddle.com/bqxj15

It only throws this exception in your patched version.
And this is precisely why your feature request is a BC break (there is **no** way to implement it in a BC way.

And btw, PHP does not allow comment nesting either (here is the proof: https://3v4l.org/T6Z2h) and I don't know of any other comment allowing it as it precisely breaks the concept of comments (as some stuff inside the comment becomes code again)

WietseWind:
@stof Thanks for explaining. I think nested comments would make more sense than a second comment-opening tag, but breaking BC would be a problem indeed. I'll close this PR and and it to our own code lib. instead :)

stof:
@WietseWind the concept of nested comments itself is weird, as it means that the content of comments must be parsed to find code in them, defeating the meaning of comments in any language.

And it becomes easy to find stuff breaking it. See this:

```twig
<p>Let's display this</p>

{# Let's skip this
    {% for i in 1..2 %}
       <li>{{ '{#' }}</li>
    {% endfor %}
#}

<p>And let's display this</p>
```

Your code would throw a syntax error by detecting a comment opening tag inside the comment, while it was not a comment opening tag before commenting the code.

WietseWind:
@stof Thanks. You fully convinced me ;)

PR #2277, opened 2016-11-23
Strict blocks

julienfalque:
Not sure if this is a bug or a feature, but using `block('some_undefined_block')` silently fails. This PR introduces a new `strict_blocks` environment option to make Twig throw an exception in such case.

Also, I think this should be the default behavior in 2.x or 3.x by changing the option's default value to `true` or even removing it. WDYT?

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -122,6 +122,12 @@ The following options are available:
   replace them with a ``null`` value. When set to ``true``, Twig throws an
   exception instead (default to ``false``).
 
+* ``strict_blocks`` *boolean*
+
+  If set to ``false``, Twig will silently ignore usage of undefined
+  blocks with the ``block()`` function. When set to ``true``, Twig throws an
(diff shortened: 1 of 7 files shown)


fabpot:
Indeed, that's a "feature". While working on the `block` implementation for Twig 1.28, I also thought that it was kind of weird. But I'm not a big fan of adding more configuration settings. If we think this was a bad decision back then, then let's add a deprecation notice when a block that does not exist is rendered for Twig 1.x, and let throw a proper exception on Twig 2.0.

stof:
👍 for deprecating this instead.

Such configuration is even harmful, as a template written based on the assumption that undefined blocks are fine would break when used in a project using strict blocks. This is the same kind of harm than PHP ini settings changing the behavior of functions, making things very hard (and contrary to PHP, Twig templates have no chance to alter themselves based on this config setting)

PR #2279, opened 2016-11-25
Deprecate usage of undefined blocks

julienfalque:
Alternative to #2277.

### DIFF
--- a/test/Twig/Tests/Fixtures/functions/recursive_block_with_inheritance.test
+++ b/test/Twig/Tests/Fixtures/functions/recursive_block_with_inheritance.test
@@ -12,7 +12,7 @@
 {% block list %}<ul>{{ block('children') }}</ul>{% endblock %}
 {% block children %}{% set currentItem = item %}{% for item in currentItem %}{{ block('item') }}{% endfor %}{% set item = currentItem %}{% endblock %}
 {% block item %}<li>{% if item is not iterable %}{{ block('label') }}{% else %}{{ block('list') }}{% endif %}</li>{% endblock %}
-{% block label %}{{ item }}{{ block('unknown') }}{% endblock %}
+{% block label %}{{ item }}{% endblock %}
 --TEMPLATE(base.twig)--
 {{ block('list') }}
 --DATA--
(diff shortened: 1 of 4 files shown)


test/Twig/Tests/Fixtures/functions/recursive_block_with_inheritance.test
```diff
@@ -12,7 +12,7 @@
 {% block list %}<ul>{{ block('children') }}</ul>{% endblock %}
 {% block children %}{% set currentItem = item %}{% for item in currentItem %}{{ block('item') }}{% endfor %}{% set item = currentItem %}{% endblock %}
 {% block item %}<li>{% if item is not iterable %}{{ block('label') }}{% else %}{{ block('list') }}{% endif %}</li>{% endblock %}
-{% block label %}{{ item }}{{ block('unknown') }}{% endblock %}
```
stof: the case of an unknown block must be kept in a separate test marked as legacy, to ensure we don't break its support in 1.x

test/Twig/Tests/TemplateTest.php
```diff
@@ -234,6 +236,21 @@ public function testGetAttributeWithTemplateAsObjectForDeprecations()
     }
 
     /**
+     * @group legacy
+     * @expectedDeprecation Displaying undefined block "unknown" in template "index.twig" is deprecated since version 1.29 and will throw an exception in 2.0.
+     * @expectedDeprecation Displaying undefined block "unknown" in template "index.twig" is deprecated since version 1.29 and will throw an exception in 2.0.
+     */
+    public function testRenderBlockWithUndefinedBlock()
+    {
+        // to be removed in 2.0
```
stof: this comment must be removed. Tests to be removed in 2.0 are all the legacy tests of 1.x, so there is not need to add this comment

test/Twig/Tests/Fixtures/functions/recursive_block_with_inheritance.test
julienfalque: I added a legacy integration test.

test/Twig/Tests/TemplateTest.php
julienfalque: Comment removed.

fabpot:
Thank you @julienfalque.

PR #2282, opened 2016-12-05
Fix sandbox being left enabled if an exception is thrown while rendering

CarsonF:
What's happening for me:
- An exception is thrown while rendering inside the sandbox
- Render exception response
- WebProfiler toolbar tries to include a file
- Sandbox is still enabled, so a SecurityException is thrown (which hides the real error)

### DIFF
--- a/lib/Twig/Extension/Core.php
+++ b/lib/Twig/Extension/Core.php
@@ -1406,6 +1406,18 @@ function twig_include(Twig_Environment $env, $context, $template, $variables = a
 
             throw $e;
         }
+    } catch (Throwable $e) {
+        if ($isSandboxed && !$alreadySandboxed) {
+            $sandbox->disableSandbox();
+        }
+
+        throw $e;
+    } catch (Exception $e) {
+        if ($isSandboxed && !$alreadySandboxed) {
+            $sandbox->disableSandbox();
+        }
+
+        throw $e;
     }
 
     if ($isSandboxed && !$alreadySandboxed) {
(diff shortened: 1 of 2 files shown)


stof:
could you write a test covering this

lib/Twig/Extension/Core.php
```diff
@@ -1402,6 +1402,12 @@ function twig_include(Twig_Environment $env, $context, $template, $variables = a
 
             throw $e;
         }
+    } catch (\Exception $e) {
```
CarsonF: Does PHP 5 choke on `catch (Throwable $e)`? 

Should that be before or after catching Exception? It probably doesn't matter in this case. I just remember reading somewhere that to make code compatible with PHP 5 `Throwable` should be caught first...

CarsonF:
> could you write a test covering this ?

Since this involves the `include` function and the sandbox I'm not sure where I should add the test?
SandboxTest?

fabpot:
@CarsonF Yes, `SandboxTest` seems to make more sense

CarsonF:
@fabpot I don't think those CS changes should be made in this PR, right?

fabpot:
Indeed, you can ignore them :)

lib/Twig/Extension/Core.php
```diff
@@ -1406,6 +1406,18 @@ function twig_include(Twig_Environment $env, $context, $template, $variables = a
 
             throw $e;
         }
+    } catch (\Throwable $e) {
```
CarsonF: Ah, of course.

test/Twig/Tests/Extension/SandboxTest.php
```diff
@@ -34,6 +34,7 @@ protected function setUp()
             '1_basic' => '{% if obj.foo %}{{ obj.foo|upper }}{% endif %}',
             '1_layout' => '{% block content %}{% endblock %}',
             '1_child' => "{% extends \"1_layout\" %}\n{% block content %}\n{{ \"a\"|json_encode }}\n{% endblock %}",
+            '1_include' => '{{ include("1_basic1", sandboxed=true) }}',
```
CarsonF: Well that would throw `Twig_Error_Loader` and disabling the sandbox is already checked for that exception.
https://github.com/twigphp/twig/blob/1.x/lib/Twig/Extension/Core.php#L1402-L1405

fabpot: nvm

fabpot:
Thank you @CarsonF.

PR #2288, opened 2016-12-08
marked some classes as being final

fabpot:
Not sure about this one. I'd like some reviews before merging (we might make some other ones final and I might have marked some as final which should not). ping @stof @nicolas-grekas

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
 * 1.29.0 (2016-XX-XX)
 
+ * marked some classes as being final (via @final)
  * made Twig_Error report real source path when possible
  * added support for {{ _self }} to provide an upgrade path from 1.x to 2.0 (replaces {{ _self.templateName }})
  * deprecated silent display of undefined blocks
--- a/doc/deprecated.rst
+++ b/doc/deprecated.rst
@@ -209,3 +209,6 @@ Miscellaneous
 
 * As of Twig 1.27, ``Twig_Parser::addHandler()`` and
   ``Twig_Parser::addNodeVisitor()`` are deprecated and will be removed in 2.0.
+
+* As of Twig 1.29, some classes are marked as being final via the `@final`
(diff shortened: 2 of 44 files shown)


lib/Twig/TemplateWrapper.php
```diff
@@ -13,6 +13,8 @@
  * Exposes a template to userland.
  *
  * @author Fabien Potencier <fabien@symfony.com>
+ *
+ * @final
```
stof: useless as it is already final for real

stof:
@fabpot what about Twig_Source, Twig_FileExtensionEscapingStrategy, Twig_NodeTraverser, Twig_TokenStream and profiler-related classes ? Should they also be tagged as ``@final`` ?

lib/Twig/TemplateWrapper.php
fabpot: indeed, fixed

fabpot:
@stof I've updated the code. I've not added `@final` on `Twig_FileExtensionEscapingStrategy` as it might make sense to extend its logic.

stephanvierkant:
What the idea behind this?

I'm getting this error:
> User Deprecated: The "Twig_TokenParser_Include" class is considered final. It may change without further notice as of its next major version. You should not extend it from "Twig_TokenParser_Embed".

I'm using Twig v2.3.2.

stof:
@fabpot looks like we made a mistake when marking this one as `@final` 😄

xabbuh:
see also #2459

fabpot:
see twigphp/Twig@a083db55

ro0NL:
@fabpot any chance to apply this on 1.x as well :) still get the deprecation.

https://github.com/twigphp/Twig/blob/1.x/lib/Twig/TokenParser/Embed.php#L15

stephanvierkant:
A new version tag would be great for 2.x as well :)

PR #2289, opened 2016-12-08
Remove debugging class

stof:
this file was added in the repo in the 2.x branch during an upstream merge of the 1.x branch. It clearly looks like a mistake.


fabpot:
Good catch, thanks @stof.

PR #2290, opened 2016-12-08
Better testing of operators returned by Extension

SpacePossum:
Add better testing and docs description about operators when using an extension.

### DIFF
--- a/lib/Twig/Environment.php
+++ b/lib/Twig/Environment.php
@@ -1512,8 +1512,15 @@ class Twig_Environment
 
         // operators
         if ($operators = $extension->getOperators()) {
+            if (!is_array($operators)) {
+                throw new InvalidArgumentException(sprintf(
+                    '"%s::getOperators()" must return an array with operators, got "%s".',
+                    get_class($extension), is_object($operators) ? get_class($operators) : gettype($operators).(is_resource($operators) ? '' : '#'.$operators)
+                ));
+            }
+
             if (2 !== count($operators)) {
-                throw new InvalidArgumentException(sprintf('"%s::getOperators()" does not return a valid operators array.', get_class($extension)));
+                throw new InvalidArgumentException(sprintf('"%s::getOperators()" must return an array of 2 elements, got %d.', get_class($extension), count($operators)));
             }
 
             $this->unaryOperators = array_merge($this->unaryOperators, $operators[0]);
--- a/test/Twig/Tests/CustomExtensionTest.php
+++ b/test/Twig/Tests/CustomExtensionTest.php
@@ -1,0 +1,86 @@
+<?php
+
+final class CustomExtensionTest extends \PHPUnit_Framework_TestCase
+{
+    /**
+     * @param Twig_ExtensionInterface $extension
+     * @param string                  $expectedExceptionMessage
+     *
+     * @requires PHP 5.3
+     * @dataProvider provideInvalidExtensions
+     */
+    public function testGetInvalidOperators(\Twig_ExtensionInterface $extension, $expectedExceptionMessage)
+    {
+        $this->setExpectedException('InvalidArgumentException', $expectedExceptionMessage);
+
+        $loader = new \Twig_Loader_Array(array('foo' => '{{ foo }}'));
+        $env = new \Twig_Environment($loader);
+        $env->addExtension($extension);
+
+        $method = new \ReflectionMethod($env, 'initExtensions');
+        $method->setAccessible(true);
+        $method->invoke($env);
+    }
+
+    public function provideInvalidExtensions()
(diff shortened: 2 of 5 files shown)


lib/Twig/Environment.php
```diff
@@ -1402,6 +1402,24 @@ public function computeAlternatives($name, $items)
         return Twig_Error_Syntax::computeAlternatives($name, $items);
     }
 
+    public function addUnaryOperator(array $operator)
+    {
+        if ($this->extensionInitialized) {
+            throw new LogicException('Unable to add unary operator as extensions have already been initialized.');
+        }
+
+        $this->unaryOperators = array_merge($this->unaryOperators, $operator);
```
stof: this implementation is not good. It should register them in the Staging extension instead.

lib/Twig/Environment.php
```diff
@@ -1512,12 +1530,22 @@ protected function initExtension(Twig_ExtensionInterface $extension)
...
+                throw new InvalidArgumentException(sprintf(
+                    '"%s::getOperators()" must return an operators array, got "%s".',
+                    get_class($extension), is_object($operators) ? get_class($operators) : gettype($operators).(is_resource($operators) ? '' : '#'.$operators)
+                ));
             }
 
-            $this->unaryOperators = array_merge($this->unaryOperators, $operators[0]);
-            $this->binaryOperators = array_merge($this->binaryOperators, $operators[1]);
+            switch (count($operators)) {
+                case 1:
```
stof: return an array with a single element in it is not supported. You should not change this behavior

lib/Twig/Environment.php
SpacePossum: there no operators in the staging

lib/Twig/Environment.php
SpacePossum: I would like to see it supported, atm it forces a users to add a second empty array if binary operators are not part of the extension.

lib/Twig/Environment.php
stof: yeah, because it was not possible to register them without extension previously. Your PR should precisely be adding them as it adds the feature.

lib/Twig/Environment.php
stof: there is 2 thing here:

- it is not related to registering operators without extension, so it should not be hidden in an unrelated PR (the PR description never talks about this change, and it is not covered by tests either)
- I would be against it, as it makes the API even harder to understand IMO

SpacePossum: check, I'll take this out :+1:

test/Twig/Tests/CustomExtensionTest.php
```diff
@@ -0,0 +1,102 @@
+<?php
+
+/**
+ * @internal
```
stof: we don't tag tests as internal (they are not in ``lib`` at all anyway)

test/Twig/Tests/CustomExtensionTest.php
```diff
@@ -0,0 +1,102 @@
...
+            array(new InvalidOperatorExtension(new \stdClass()), '"InvalidOperatorExtension::getOperators()" must return an operators array, got "stdClass".'),
+            array(new InvalidOperatorExtension(array(1, 2, 3)), '"InvalidOperatorExtension::getOperators()" must return array of 2 elements, got 3.'),
+        );
+    }
+}
+
+/**
+ * @internal
+ */
+abstract class AbstractInvalidOperatorsExtension implements \Twig_ExtensionInterface
```
stof: no need for this abstract class

lib/Twig/Environment.php
```diff
@@ -1512,8 +1512,15 @@ protected function initExtension(Twig_ExtensionInterface $extension)
...
+            if (!is_array($operators)) {
+                throw new InvalidArgumentException(sprintf(
+                    '"%s::getOperators()" must return an operators array, got "%s".',
+                    get_class($extension), is_object($operators) ? get_class($operators) : gettype($operators).(is_resource($operators) ? '' : '#'.$operators)
+                ));
+            }
+
             if (2 !== count($operators)) {
-                throw new InvalidArgumentException(sprintf('"%s::getOperators()" does not return a valid operators array.', get_class($extension)));
+                throw new InvalidArgumentException(sprintf('"%s::getOperators()" must return array of 2 elements, got %d.', get_class($extension), count($operators)));
```
stof: ``must return an array with 2 elements``

SpacePossum:
I'll retarget to 2.x, supporting PHP5.2 with PHPUnit from the stone ages is no fun nor useful

fabpot:
Thank you @SpacePossum.

PR #2294, opened 2016-12-13
Cleanup

nicolas-grekas:
Two things here:

guessTemplateInfo is private in 2.x
this PR will fix tests on 2.x

setTemplateName is deprecated, and the removed code paths are never run

### DIFF
--- a/lib/Twig/Error/Loader.php
+++ b/lib/Twig/Error/Loader.php
@@ -24,18 +24,15 @@
  */
 class Twig_Error_Loader extends Twig_Error
 {
-    private $guess = false;
-
     public function __construct($message, $lineno = -1, $source = null, Exception $previous = null)
     {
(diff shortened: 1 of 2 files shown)


lib/Twig/Error/Loader.php
```diff
@@ -24,18 +24,10 @@
...
-        parent::__construct($message, false, null, $previous);
-        $this->guess = true;
-    }
-
-    protected function guessTemplateInfo()
-    {
-        if ($this->guess) {
-            parent::guessTemplateInfo();
-        }
+        Exception::__construct('', 0, $previous);
```
fabpot: Breaks PHP 5.3 AFAIU if `$previous` is not null

stof:
👍

fabpot:
Thank you @nicolas-grekas.

PR #2298, opened 2016-12-16
Remove $env argument type from ExpressionParser constructor

fancyweb:
Either the argument type should be removed, or the whole "else" block and the deprecation warning is useless.


stof:
indeed

fabpot:
Thank you @fancyweb.

PR #2300, opened 2016-12-16
removed ltrim because $class variable already ltrim'ed

AmsTaFFix:
removed ltrim because $class variable already ltrim'ed


stof:
👍

fabpot:
Good catch, thanks @AmsTaFFix.

PR #2302, opened 2016-12-17
drop PHP7 support

fabpot:


Koc:
Did you mean php 5.6 support?

fabpot:
@Koc indeed :)

fabpot:
I've dropped HHVM as I don't know how to use a version that is compatible with PHP 7.

fabpot:
Also, I know that the PHP extension could be removed, but Santa Claus might have a great present, so I've kept it for now.

CHANGELOG
```diff
@@ -12,7 +12,7 @@
  * removed Twig_Autoloader (use Composer instead)
  * removed `true` as an equivalent to `html` for the auto-escaping strategy
  * removed pre-1.8 autoescape tag syntax
- * dropped support for PHP 5.2, 5.3, 5.4, and 5.5
+ * dropped support for PHP 5.x
```
michaelcullum: Should probably add HHVM to the changelog also?

fabpot: No, I've just removed HHVM from the Travis matrix, but I've not removed support for HHVM in the code. Actually, as HHVM should be compatible with PHP, we could even not care about HHVM at all as if something does not work, that's probably a bug on HHVM side.

lsmith77: yeah but it would be useful to see that and then report to HHVM devs.

stof:
The default Travis environment runs on Precise, and so on a very old HHVM version (not supporting PHP 7 feature), as HHVM dropped the support of Precise a long time ago. See the Symfony .travis.yml file to see the way to make the HHVM build run on Trusty with an uptodate HHVM version

CHANGELOG
robfrawley: @fabpot You want to implement something along the lines of the following in your travis.yml to support HHVM: https://docs.travis-ci.com/user/languages/php#HHVM-versions-on-Trusty

robfrawley: Actually, as stof mentioned, you already implement this [in the Symfony travis.yml](https://github.com/symfony/symfony/blob/master/.travis.yml#L23).

fabpot:
@stof That does not change anything. HHVM still does not advertize 7.0 as a version.

stof:
@fabpot on recent versions, there is a INI setting to switch between PHP 5 and PHP 7 mode: ``hhvm.php7.all = 1``

I just don't know the location of the HHVM config file on Travis.

michaelcullum:
But will that setting allow it to fulfil the composer dependency requirement of PHP 7? My understanding was that just enabled the BC-breaking PHP 7 changes although I might be mistaken.

fabpot:
As you can see, I have set the ini thing, but that does not change the version of HHVM to 7.

michaelcullum:
Alternatively, when using hhvm you could run composer with `--ignore-platform-reqs`

andrerom:
@michaelcullum: that won't help much, people won't be able to require it on hhvm (without doing he same which is not good for end users).

/cc @Seldaek

robfrawley:
Isn't this a composer issue? Why would HHVM magically report it's version to be PHP 7? Composer needs to do the extra work of looking for the INI setting if it truly intends to support HHVM, right?

Seldaek:
It's HHVM's job to define PHP_VERSION as that is how to we know which version of the php language is supported. The main question to me is more what's the point? Is anyone discussing here actually using HHVM or just discussing for the point of discussing?

robfrawley:
@Seldaek The latter.

robfrawley:
@fabpot The INI setting absolutely does change the version to 7 for HHVM: see https://travis-ci.org/src-run/twig/jobs/184963773#L680 @Seldaek Any idea why composer is unhappy? https://travis-ci.org/src-run/twig/jobs/184963773#L694

robfrawley:
This should provide the desired results (once Composer works as expected):
```yaml
matrix:
  include:
    - php: hhvm-stable
      sudo: required
      dist: trusty
      group: edge
      env: TWIG_EXT=no
    - php: hhvm-stable
      sudo: required
      dist: trusty
      group: edge
      env: TWIG_EXT=no
    - php: 7.0
      env: TWIG_EXT=yes
    - php: 7.1
      env: TWIG_EXT=yes
    - php: 7.0
      env: TWIG_EXT=no
    - php: 7.1
      env: TWIG_EXT=no
  fast_finish: true
```

robfrawley:
Attempting to fix issue with composer via composer/composer#5978

.travis.yml
```diff
@@ -7,32 +7,32 @@ cache:
...
+        - php: 7.0
+          env: TWIG_EXT=no
+        - php: 7.1
+          env: TWIG_EXT=no
+    allow_failures:
+        - php: hhvm-stable
+          sudo: required
+          dist: trusty
+          group: edge
+          env: TWIG_EXT=no
```
stof: I think you can simplify this ``allow_failures`` group to ``php: hhvm-stable`` only

.travis.yml
```diff
@@ -7,32 +7,32 @@ cache:
...
-
-env:
-    - TWIG_EXT=no
-    - TWIG_EXT=yes
+matrix:
+    include:
+        - php: hhvm-stable
+          sudo: required
+          dist: trusty
+          group: edge
```
stof: I think ``group: edge`` is not necessary anymore (as they promoted the image to become the stable one a few weeks ago). I suggest running on the stable image rather than the edge one (which may be unstable due to being updated all the time)

lib/Twig/Node/Expression/Name.php
```diff
@@ -61,21 +61,19 @@ public function compile(Twig_Compiler $compiler)
...
                 $compiler
                     ->raw('(isset($context[')
                     ->string($name)
                     ->raw(']) || array_key_exists(')
                     ->string($name)
                     ->raw(', $context) ? $context[')
                     ->string($name)
-                    ->raw('] : $this->notFound(')
+                    ->raw('] : (function () { throw new Twig_Error_Runtime(\'Variable ')
                     ->string($name)
```
stof: this is fragile, as ``->string($name)`` will output quotes, which may close the string started as ``'Variable `` if you switch the code to use single quotes rather than double quotes in the future.

fabpot: But we won't change that, so I don't think this is a real issue.

robfrawley:
Thanks to @Seldaek the latest snapshot of Composer runs successfully using PHP7-mode-enabled HHVM (see [composer/composer#5978@267920323](https://github.com/composer/composer/issues/5978#issuecomment-267920323)). There are some ligament test issues ATM with HHVM, so it should stay as an `allowed_failure` for now regardless of the Composer fix. I'll open a new PR once I've sorted through the HHVM test failures.

PR #2305, opened 2016-12-18
Fix Travis Build Matrix to Support 7.x and HHVM Only

robfrawley:
As a follow-up to #2302 (with HHVM set as an allowed failure until Composer sorts out its inability to parse the `7.0.99-hhvm` version string, ticket for this here: composer/composer#5978).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,36 +3,37 @@ language: php
 sudo: false
 
 cache:
-    directories:
-        - vendor
-        - $HOME/.composer/cache/files
+  directories:
+    - vendor
+    - $HOME/.composer/cache/files
 
-php:
-    - 5.6
-    - 7.0
-    - 7.1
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -3,36 +3,37 @@ language: php
 sudo: false
 
 cache:
-    directories:
-        - vendor
-        - $HOME/.composer/cache/files
+  directories:
+    - vendor
+    - $HOME/.composer/cache/files
```
fabpot: Can you revert the whitespace changes? I like when the diff is as small as possible :)

fabpot: nevermind

fabpot:
I've cherry-picked your commit on #2302. Thanks.

robfrawley:
:+1:

PR #2310, opened 2016-12-22
PHP7 C extension for Twig

fabpot:
This adds a new C extension for Twig 2+, compatible with PHP 7+.

My first rudimentary benchmarks do not show big improvements in term of performance. So, can YOU test it on your real-world applications running Twig 2 and report the performance impact of using the extension (with opcache of course)? If performance improvements are non-significant, we will drop it for Twig 2.


ext/twig/twig.c
```diff
@@ -0,0 +1,758 @@
...
+
+	zval_ptr_dtor(&constructor_args[1]);
+	zval_ptr_dtor(&constructor_retval);
+
+	Z_ADDREF(ex);
+	zend_throw_exception_object(&ex);
+}
+
+/* {{{ proto mixed twig_get_attribute(Twig_Environment env, Twig_Source source, mixed object, mixed item, array arguments, string type, boolean isDefinedTest, boolean ignoreStrictCheck)
+   A C implementation of TwigTemplate::getAttribute() */
```
stof: should be updated, as the PHP code does not live in ``TwigTemplate::getAttribute()`` anymore

ext/twig/twig.c
```diff
@@ -0,0 +1,758 @@
...
+
+			if (!ret) {
+				ret = &EG(uninitialized_zval);
+			}
+
+			zend_string_release(item_str);
+			RETURN_ZVAL(ret, 1, 0);
+		}
+/*
+		if (Twig_Template::ARRAY_CALL === $type) {
```
stof: this comment is outdated

ext/twig/twig.c
```diff
@@ -0,0 +1,758 @@
...
+	if (result) {
+		ZVAL_COPY(result, &retval);
+	}
+
+	zval_ptr_dtor(&retval);
+
+	return success;
+}
+
+static int twig_has_extension(zval *obj)
```
stof: would be more readable if it was named ``twig_has_sandbox_extension``, as it checks for the sandbox extension, not for any extension (the macro should be renamed too)

lib/Twig/Extension/Core.php
```diff
@@ -1337,3 +1337,195 @@ function twig_array_batch($items, $size, $fill = null)
...
+                $message = sprintf('Impossible to invoke a method ("%s") on a null variable.', $item);
+            } else {
+                $message = sprintf('Impossible to invoke a method ("%s") on a %s variable ("%s").', $item, gettype($object), $object);
+            }
+
+            throw new Twig_Error_Runtime($message, -1, $source);
+        }
+
+        // object property
+        if (Twig_Template::METHOD_CALL !== $type && !$object instanceof Twig_Template) { // Twig_Template does not have public properties, and we don't want to allow access to internal ones
```
stof: shouldn't ``$object instanceof Twig_Template`` be forbidden entirely at the beginning, as it was deprecated in the 1.x branch ? It looks like deprecated APIs were turned into allowed calls instead of forbidden calls when merging 1.x into 2.x.
And if it is indeed changed, it must be reflected in the C extension

lib/Twig/Extension/Core.php
```diff
@@ -1337,3 +1337,195 @@ function twig_array_batch($items, $size, $fill = null)
...
+            if ($object instanceof Twig_Template) {
+                $ref = new ReflectionClass($class);
+                $methods = array();
+                foreach ($ref->getMethods(ReflectionMethod::IS_PUBLIC) as $refMethod) {
+                    $methods[] = $refMethod->name;
+                }
+            } else {
+                $methods = get_class_methods($object);
+            }
+            // sort values to have consistent behavior, so that "get" methods win precedence over "is" methods
```
fabpot: fixed

lib/Twig/Extension/Core.php
```diff
@@ -1337,3 +1337,195 @@ function twig_array_batch($items, $size, $fill = null)
...
+     * @param bool   $isDefinedTest     Whether this is only a defined check
+     * @param bool   $ignoreStrictCheck Whether to ignore the strict attribute check or not
+     *
+     * @return mixed The attribute value, or a Boolean when $isDefinedTest is true, or null when the attribute is not set and $ignoreStrictCheck is true
+     *
+     * @throws Twig_Error_Runtime if the attribute does not exist and Twig is running in strict mode and $isDefinedTest is false
+     *
+     * @internal
+     */
+    function twig_get_attribute(Twig_Environment $env, Twig_Source $source, $object, $item, array $arguments = array(), $type = Twig_Template::ANY_CALL, $isDefinedTest = false, $ignoreStrictCheck = false)
```
fabpot: I've started a branch, but I'm not sure as we won't be able to do it everywhere for BC.

lib/Twig/Extension/Core.php
```diff
@@ -1337,3 +1337,195 @@ function twig_array_batch($items, $size, $fill = null)
...
+                return $object->$item;
+            }
+        }
+
+        $class = get_class($object);
+
+        // object method
+        if (!isset($cache[$class])) {
+            // get_class_methods returns all methods accessible in the scope, but we only want public ones to be accessible in templates
+            if ($object instanceof Twig_Template) {
```
fabpot: removed

lib/Twig/Extension/Core.php
fabpot: fixed

fabpot:
I've extracted everything not related to the C extension to #2316 as the refactoring done there can be merged independently, even if we decide to not have a C extension for Twig 2.

fabpot:
As I don't have any feedback yet on the performance improvement that the C extension brings in, it won't be part of 2.0.

robfrawley:
@fabpot Forgot to post my finding (you may want to give this more time, as many will have been busy with holiday-related activities during their respective holiday celebrations)... but, with that said, I had _no to modest_ improvements on `PHP 7.0.14` with this extension. I picked one of my random projects, and here are the results:

| Template | No Extension | With Extension |
| --- | --- | --- |
| [default/index.html.twig](https://github.com/src-run/srw-client-silverpapillon/blob/master/src/AppBundle/Resources/views/default/index.html.twig) | `40ms` | `40ms` |
| [category/list.html.twig](https://github.com/src-run/srw-client-silverpapillon/blob/master/src/AppBundle/Resources/views/category/list.html.twig) | `19ms` | `18ms` |
| [category/view.html.twig](https://github.com/src-run/srw-client-silverpapillon/blob/master/src/AppBundle/Resources/views/category/view.html.twig) | `34ms` | `34ms` |
| [feed/index.twig.html](https://github.com/src-run/srw-client-silverpapillon/blob/master/src/AppBundle/Resources/views/feed/index.html.twig) | `42ms` | `38ms` |

fabpot:
Thanks @robfrawley for your feedback, this is much appreciated. Your results are in line with mine, no real benefit. I will indeed way more before closing this PR, but I will move forward and release 2.0 in the meantime. The C extension can be part of 2.1.

fabpot:
Last call before I close this issue. Anyone seeing a significant benefit when using the C extension on their projects? It should help on templates where there are many calls using the dot notation like in `x.y`.

anavarre:
FTR we've removed the recommendation entirely for Drupal since we couldn't really determine if there was any real benefit or not - See https://www.drupal.org/node/2568247

fabpot:
Closing as it seems there is no interest in a C extension for Twig.

PR #2311, opened 2016-12-22
Add a simple Twig_RuntimeLoaderInterface implementation

chalasr:
Next to https://github.com/symfony/symfony/pull/21023

This is related to the BC break reported in symfony/symfony#21008 which has been introduced in symfony/symfony#20093 when decoupling extensions from definitions.

What I propose here is to ease the upgrade to symfony 3.2+ by adding a simple `Twig_RuntimeLoaderInterface` implementation here, useful only when using the twig-bridge outside of the symfony fullstack framework (with the Form component for instance).

Upgrading would be as simple as:

```diff
$twig = new Twig_Environment(...);
$rendererEngine = new TwigRendererEngine(array('form_div_layout.html.twig'), $twig);
- $twig->addExtension(new FormExtension(new TwigRenderer($rendererEngine, $csrfTokenManager)));
+ $twig->addExtension(new FormExtension());
+ $twig->addRuntimeLoader(new Twig_RuntimeLoader(array(TwigRenderer::class => new TwigRenderer($rendererEngine, $csrfTokenManager)));
```

Instead of having to write this runtime loader yourself. 
Please see symfony/symfony#21008 for details and a concrete example of how this could help.

### DIFF
--- a/lib/Twig/FactoryRuntimeLoader.php
+++ b/lib/Twig/FactoryRuntimeLoader.php
@@ -1,0 +1,40 @@
+<?php
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+/**
+ * Lazy loads the runtime implementations for a Twig element.
+ *
+ * @author Robin Chalas <robin.chalas@gmail.com>
+ */
+class Twig_FactoryRuntimeLoader implements Twig_RuntimeLoaderInterface
+{
+    private $map;
+
+    /**
+     * @param array $map An array of format [classname => factory callable]
+     */
+    public function __construct(array $map = array())
+    {
+        $this->map = $map;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function load($class)
+    {
+        if (isset($this->map[$class])) {
+            $runtimeFactory = $this->map[$class];
+
+            return $runtimeFactory();
+        }
+    }
+}
--- a/test/Twig/Tests/RuntimeFactoryLoaderTest.php
+++ b/test/Twig/Tests/RuntimeFactoryLoaderTest.php
@@ -1,0 +1,32 @@
+<?php
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+class Twig_Tests_FactoryRuntimeLoaderTest extends PHPUnit_Framework_TestCase
+{
+    public function testLoad()
+    {
+        $loader = new Twig_FactoryRuntimeLoader(array('stdClass' => 'getRuntime'));
+
+        $this->assertInstanceOf('stdClass', $loader->load('stdClass'));
+    }
+
+    public function testLoadReturnsNullForUnmappedRuntime()
+    {
+        $loader = new Twig_FactoryRuntimeLoader();
+
+        $this->assertNull($loader->load('stdClass'));
+    }
+}
+
+function getRuntime()
+{
+    return new stdClass();
+}


stof:
I would not name it ``Twig_RuntimeLoader``, as this gives it a very special name as a runtime loader, which it is a very specific use case instead (and not the one you should use most of the time, as it defeats all performance goals of the system). The name should describe what makes this implementation special.

Btw, to avoid breaking the goal of the loader, a map of ``class => factory callable`` would make more sense, allowing to keep the runtimes lazy-loaded.

It would then end up as this:

```php
$twig->addRuntimeLoader(new Twig_RuntimeLoader(array(
    // even better would be to change the way the engine and the csrf token manager are
    // instantiated to keep them lazy as well, but it depends on how they are used elsewhere
    // in the app
    TwigRenderer::class => function() use($rendererEngine, $csrfTokenManager) {
       return new TwigRenderer($rendererEngine, $csrfTokenManager);
    },
));
```

lib/Twig/RuntimeLoader.php
```diff
@@ -0,0 +1,35 @@
...
+ * (c) Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+/**
+ * @author Robin Chalas <robin.chalas@gmail.com>
+ */
+class Twig_RuntimeLoader implements \Twig_RuntimeLoaderInterface
```
stof: the ``\`` must be removed, as Twig 1.x still supports PHP 5.2

lib/Twig/RuntimeLoader.php
```diff
@@ -0,0 +1,35 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function load($class)
+    {
+        if (isset($this->mapping[$class])) {
+            return $this->mapping[$class];
+        }
+
+        throw new \InvalidArgumentException(sprintf('Class "%s" is not mapped as Twig runtime.', $class));
```
stof: It must return null when it does not support the class according to the interface

lib/Twig/RuntimeLoader.php
```diff
@@ -0,0 +1,42 @@
...
+     */
+    public function load($class)
+    {
+        if (isset($this->map[$class])) {
+            $runtime = $this->map[$class];
+
+            return $runtime();
+        }
+
+        return null;
```
xabbuh: IIRC the `return` statement can then simply be omitted to match the Twig code style rules.

test/Twig/Tests/RuntimeLoaderTest.php
```diff
@@ -0,0 +1,25 @@
...
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+class Twig_Tests_RuntimeLoaderTest extends PHPUnit_Framework_TestCase
```
xabbuh: missing blank line before the class declaration

chalasr:
Fixed line notes and applied @stof suggestion about mapping callables instead of instances.

For the naming of this class, I agree that the current one is unclear and not adapted. Right now I fail at finding something appropriated so if you or @xabbuh have something in mind, please tell me

chalasr:
Changed to Twig_ExtensionRuntimeLoader

chalasr:
Changed to `Twig_FactoryRuntimeLoader`

lib/Twig/FactoryRuntimeLoader.php
```diff
@@ -0,0 +1,40 @@
...
+     * @param array $map An array of format [classname => factory callable]
+     */
+    public function __construct(array $map = array())
+    {
+        $this->map = $map;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
```
fabpot: The phpdoc can be removed.

fabpot:
Thank you @chalasr.

PR #2313, opened 2016-12-22
Update range.rst

DonCallisto:


fabpot:
I made a slightly different change to document this feature in 4034848. Thank you.

PR #2314, opened 2016-12-23
add a comma to clarify the meaning

xabbuh:
Without the comma you could understand that the "if" not only applied
to `-1`, but to `1` as well if you read too quickly.


fabpot:
Thank you @xabbuh.

PR #2315, opened 2016-12-23
modified code to use variadics when possible

fabpot:


lib/Twig/Compiler.php
```diff
@@ -103,9 +103,8 @@ public function raw($string)
      *
      * @return $this
      */
-    public function write()
+    public function write(... $strings)
```
Tobion: there is an extra space after the dots.

fabpot: fixed

PR #2316, opened 2016-12-23
moved Twig_Template::getAttribute() to a function

fabpot:


lib/Twig/Extension/Core.php
```diff
@@ -1337,3 +1337,188 @@ function twig_array_batch($items, $size, $fill = null)
...
+        if (null === $object) {
+            $message = sprintf('Impossible to invoke a method ("%s") on a null variable.', $item);
+        } else {
+            $message = sprintf('Impossible to invoke a method ("%s") on a %s variable ("%s").', $item, gettype($object), $object);
+        }
+
+        throw new Twig_Error_Runtime($message, -1, $source);
+    }
+
+    if ($object instanceof Twig_Template) {
```
stof: shouldn't it be check first ? No need to check for ArrayAccess on a Twig_Template

stof:
:+1:

lib/Twig/Extension/Core.php
fabpot: I've done the contrary, making the check as late as possible as this should happen rarely.

stof: Can you add a test covering it though ? The C extension currently misses this check but passes tests

PR #2318, opened 2016-12-24
removed remaining mentions of the C extension

fabpot:


stof:
shouldn't this wait to see the result of #2310 ?

fabpot:
Yes and no. I think it's going to take time to get feedback from users. As my first tests shows that the gains are quite minimal, I think I'm going to release 2.0 without a C extension and add it in 2.x if people prove that it helps significantly on their applications.

stof:
then most of these needs to be reverted in #2310, right ? Then, OK for this PR

fabpot:
Of course, that's what I will do when merged.

PR #2321, opened 2016-12-26
Remove mention of Twig_Autoloader() [deprecated]

samayo:
I could be wrong, but just started learning Symfony, and PHPStorm 9.0 says `Twig_Autoloader` in `Twig_Autoloader::register();` is deprecated!. if that is the case, might as well remove it. I didn't have issues removing it, since I installed twig via composer.


fabpot:
This has been changed in 2.x where the autoloader is gone.

PR #2323, opened 2016-12-29
Update umask before creating cache directory

shivamdixit:
The cache directory is not created with 0777 permissions because of the system umask. Therefore if the process is being triggered through different users (ex: web, CLI), it is unable to write to the directory and throws exception. This commit will ensure that cache directory is always created with 
0777 permissions irrespective of the system umask.


stof:
-1 for this. We precisely want to respect the umask, to avoid making things world-writable when your server does not need it.

lyrixx:
:-1: too. you should instead use the same user for your CLI and webserver.

stof:
Or use filesystem ACLs

shivamdixit:
Alright, thanks!

Issue #2329, opened 2017-01-05
[DOCUMENTATION] Argument name replace_pairs for filter replace is invalid


jean-pasqualini:
Hello,

In documentation of filter replace for branche 1.X, 
http://twig.sensiolabs.org/doc/1.x/filters/replace.html#arguments

The argument name **replace_pairs** is invalid due in code argument is named **from**
https://github.com/jean-pasqualini/Twig/blob/1.x/lib/Twig/Extension/Core.php#L475

I propose a correction of the documentation on this point.
See pull request.

- [X] Make pull request

PR #2330, opened 2017-01-05
Fix the argument name of filter replace

jean-pasqualini:
The argument name in the code is $from

See https://github.com/jean-pasqualini/Twig/blob/1.x/lib/Twig/Extension/Core.php#L475.

Issue : https://github.com/twigphp/Twig/issues/2329


fabpot:
Thank you @jean-pasqualini.

PR #2334, opened 2017-01-05
Improve Twig Error Messages

baileylo:
Removes an early exit from the `updateRepr` method, which allows for more informative error messages.

### DIFF
--- a/lib/Twig/Error.php
+++ b/lib/Twig/Error.php
@@ -250,8 +250,6 @@ class Twig_Error extends Exception
         if ($this->sourcePath && $this->lineno > 0) {
             $this->file = $this->sourcePath;
             $this->line = $this->lineno;
-
-            return;
         }
 
         $dot = false;
--- a/test/Twig/Tests/ErrorTest.php
+++ b/test/Twig/Tests/ErrorTest.php
@@ -94,7 +94,7 @@ EOHTML
 
             $this->fail();
         } catch (Twig_Error_Runtime $e) {
-            $this->assertEquals('Variable "foo" does not exist.', $e->getMessage());
+            $this->assertEquals('Variable "foo" does not exist in "index.html" at line 3.', $e->getMessage());
             $this->assertEquals(3, $e->getTemplateLine());
             $this->assertEquals('index.html', $e->getSourceContext()->getName());
             $this->assertEquals(3, $e->getLine());
@@ -113,7 +113,7 @@ EOHTML
 
             $this->fail();
         } catch (Twig_Error_Runtime $e) {
-            $this->assertEquals('An exception has been thrown during the rendering of a template ("Runtime error...").', $e->getMessage());
+            $this->assertEquals('An exception has been thrown during the rendering of a template ("Runtime error...") in "index.html" at line 3.', $e->getMessage());
             $this->assertEquals(3, $e->getTemplateLine());
             $this->assertEquals('index.html', $e->getSourceContext()->getName());
             $this->assertEquals(3, $e->getLine());


shashikreddy:
:thumbsup:

stof:
Well, the return was done because the location of the exception itself is updated in this case, meaning that the exception itself already has the file and line

stof:
of course, if you only display the exception message when there is an exception, without displaying the location of the exception, you misses the info. but this is true for any exception in your system and is an issue with your error reporting system then.

baileylo:
I don't understand the point of the method then. From reading the [previous implementation](https://github.com/twigphp/Twig/blob/v1.28.2/lib/Twig/Error.php#L198), the method injects the line and file into the exception message.

I don't know why the exception knowing the file and line should preclude it from performing that original action? From what you're saying, I feel like the entire method should be removed, along with the `rawMessage`.

stof:
no it should not. Loaders don't always provide a source path (as templates don't always come from a file).

and the original actions are about giving the original location of the exception when the exception cannot point to it directly (because the template comes from elsewhere than a file, or the loader did not say it was coming from a file) and so the exception file and line are pointing to the compiled template

baileylo:
Wouldn't those data points be better accessed through properties rather than having to grep them out of the exception message?

stof:
which data points are you talking about ?

baileylo:
Sorry, let me try and clarify my assumptions and how I read what you said.

My understanding of what you're saying is:
The point of changing the message in `updateRepr` is to give the original location of the exception when the exception cannot point to it directly. 

My Read of The Code:
This location information is found from `filename` and `lineno` properties on the exception.

My Question:
Can't the exception point to these two properties directly through, `getTemplateLine` and `getSourceContext()->getName()`?

stof:
@baileylo ``getSourceContext()->getName()`` is not a filename. It is a template name. So we can set the filename of the exception to the source filename only when we have a value in ``getSourceContext()->getPath()`` (which is not always the case, as said previously, and so we cannot always use it).

But when we have the real source filename and line number, we set them in the dedicated properties of the Exception, which means that a tool reporting the exception would directly point to the source rather than the cache file or the eval'd code (depending of how the template was loaded). This allows to have the proper location even when the tool does not know about Twig (avoiding to require each to make a special handling for Twig_Error).

baileylo:
So why not have that in the error message?

baileylo:
When I said file name, I meant `filename` property. I came to my conclusion that `$ex->filename == $ex->getSourceContext->getName()` from the following code function implementation.

```
    /**
     * Gets the source context of the Twig template where the error occurred.
     *
     * @return Twig_Source|null
     */
    public function getSourceContext()
    {
        return $this->filename ? new Twig_Source($this->sourceCode, $this->filename, $this->sourcePath) : null;
    }
```

By the way, I'm no trying to be rude, I'm just trying to understand.

> But when we have the real source filename and line number, we set them in the dedicated properties of the Exception, which means that a tool reporting the exception would directly point to the source rather than the cache file or the eval'd code (depending of how the template was loaded).

My understanding of this, is that previously the exception message may have contained the path to the cached version of the template?

stof:
@baileylo ``$this->filename`` in the Twig_Template class is actually the template name. The name of the property was wrong in Twig 1.x, but renaming the property was impossible without BC break, because it is protected.

And if you look at the code you modified, ``$ex->filename`` is set based on the source path, not on the source name.

> My understanding of this, is that previously the exception message may have contained the path to the cached version of the template, is that correct?

Not the message. The exception **location** (i.e. ``$ex->filename`` and ``$ex-line``), which are initially filled by PHP based on the location where the exception is thrown (which is generally in the compiled template for runtime errors, and in the parser for syntax errors).
When we have the path to the source template (when there is such a path), we update the exception location to the source location (think about source maps in frontend stacks...) as it is the place where the error really is.
when we cannot do it (because we don't have this path), we keep the old way of leaving the location pointing to the place where the exception was thrown (which is often useless to fix it in this case, as it is inside Twig itself) and we add the info we have in the message (as we cannot do anything better)

PR #2335, opened 2017-01-06
Added Twig_Simple* methods to deprecated doc

martijnpieters:
The documentation of these deprecations was available in the 1.0 branch and is still useful in 2.0.


stof:
This should be part of #2181 only, as we decided not to deprecate them yet (as we have to live with packages supporting both 1.x and 2.x in parallel for some time)

stof:
@fabpot fabbot.io seems broken

stof:
Actually, it is *already* part of #2181 (in a more compact way) so this PR should be closed

SpacePossum:
>fabbot.io seems broken

it is configured for a feature that is not released yet (https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2421)

PR #2339, opened 2017-01-06
Specifying PHPUnit version

Sam-Burns:


Sam-Burns:
Sorry - new to the project. Can anyone tell me what 'braces is not configurable' means?

fabpot:
fabbot is broken for now, so you can ignore it.
For the phpunit change, I don't want to add it to the composer.json as it might cause many issues. We don't want our own dependencies to interfere with the one of phpunit. phpunit should be considered as an external and opaque binary.

ciaranmcnulty:
@fabpot you still have the issue that Twig deps can conflict with PHPunit ones, just you'll only discover it when autoloading picks the version in the PHPUnit PHAR instead of in your vendor folder and you get a runtime error.

Or do I misunderstand what you mean?

fabpot:
But for this case, we have the Symfony PHPUnit bridge, which "solves" this for us.

PR #2345, opened 2017-01-09
Delay marking the environment as initialized until it is done

stof:
This avoids breaking the symfony exception page rendering saying that filters are not available due to a partial initialization: https://github.com/symfony/symfony/issues/21212
It would report the initialization error instead.

### DIFF
--- a/lib/Twig/Environment.php
+++ b/lib/Twig/Environment.php
@@ -1439,7 +1439,6 @@ class Twig_Environment
             return;
         }
 
-        $this->extensionInitialized = true;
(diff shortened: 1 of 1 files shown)


fabpot:
Is there a way to add a test for this?

stof:
Well, I don't have an idea to ensure that this works fine, except maybe triggering an exception in an extension, catching it in the test, and then ensuring that the environment has not registered itself as initialized. But this would be testing the fix by asserting that the fix is there, not that the issue is gone.

fabpot:
Thank you @stof.

PR #2346, opened 2017-01-09
Typehint the function, filter and test names as string

stof:
This gives faster feedback when using the class in a wrong way, instead of getting a ``Warning: Illegal offset type in ExtensionSet.php line 434`` during the function registration (see https://github.com/symfony/symfony/issues/21212 for a case getting this message)


fabpot:
Thank you @stof.

PR #2348, opened 2017-01-10
remove condition on PHP version

MatTheCat:
Twig now requires PHP >= 7 so there is no need to check for prior version.


fabpot:
Thank you @MatTheCat.

PR #2354, opened 2017-01-11
Turned fatal error into exception when a previously generated cache prevents loading its newly compiled version

nicolas-grekas:
Not sure if this is testable, yet this should save some hours to people that end up in the same trap we did on our project.
It turned out we generated some twig template cache in two separate commands, in a situation where "index" tracking for embeds where lost.

This PR detects the situation and invites to clear the cache properly, because that's the only way to fix the issue.

It also fixes a mistake in the cache $key generation, that should not take $index into account.
Doing so were making offline cache warmup useless in some situations.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -3,6 +3,8 @@
  * added Twig_NodeCaptureInterface for nodes that capture all output
  * fixed marking the environment as initialized too early
  * fixed C89 compat for the C extension
+ * turned fatal error into exception when a previously generated cache is corrupted
+ * fixed offline cache warm-ups for embedded templates
 
 * 1.30.0 (2016-12-23)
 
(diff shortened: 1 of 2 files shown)


nicolas-grekas:
Note that this does not solve the root issue, which in unknown.
To help understand it for someone familiar with twig:
- our first command uses a Twig file that extends the `app/Resources/views/layout.html.twig` file using the `::layout.html.twig` syntax
- the second command was a regular `cache:warmup`, which loaded other twig files that all used the `layout.html.twig` syntax when extending it.

Doing so ended up in a fatal error in the front, because of some "embed" index that did not match between the cache generated by our first command in layout.html.twig, and the way the `cache:warmup` generated indexes afterwards.

nicolas-grekas:
Root cause found: https://github.com/symfony/symfony/pull/21237

fabpot:
Thank you @nicolas-grekas.

PR #2356, opened 2017-01-11
do not overridde case-insentive cache entries

### LINKED ISSUE #2343: Twig 2.0 is calling a protected variable rather than a method
I have an object that i'm assigning to a twig variable called `assets`.  In that Object i have a protected variable called `$css`, and a method called `css()`.

With Twig 1.X I have been outputting the result of the method via:

```
{{ assets.css() }}
```

This has worked great, but when I update to Twig 2.0, I get an error:

```
Twig_Error_Runtime
An exception has been thrown during the rendering of a template ("Array to string conversion").
```

When I dump `{{ assets.css() }}` I do in fact get an array, but it's the protected variable $css.

I changed the method of the function to `render_css` and then called it via:

```
{{ assets.render_css() }}
```

and that does work, but that is not an acceptable solution.  Twig should definitely not be using a protected variable ahead of a public method of the same name.  Also I tried changing the access to `private` on the `$css` variable but it had no effect.

Incidentally I have a similarly named protected variable called `$js` with a corresponding function `js()` and a similar call to:

```
{{ assets.js() }}
```

And that works 100% fine, so not sure what's going on here.

xabbuh:
This will fix #2343.

### DIFF
--- a/lib/Twig/Extension/Core.php
+++ b/lib/Twig/Extension/Core.php
@@ -1486,10 +1486,10 @@ function twig_get_attribute(Twig_Environment $env, Twig_Source $source, $object,
         $methods = get_class_methods($object);
         sort($methods);
         $lcMethods = array_map('strtolower', $methods);
-        $cache = array();
+        $classCache = array();
         foreach ($methods as $i => $method) {
-            $cache[$method] = $method;
-            $cache[$lcName = $lcMethods[$i]] = $method;
+            $classCache[$method] = $method;
+            $classCache[$lcName = $lcMethods[$i]] = $method;
 
             if ('g' === $lcName[0] && 0 === strpos($lcName, 'get')) {
                 $name = substr($method, 3);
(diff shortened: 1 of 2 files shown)


lib/Twig/Extension/Core.php
```diff
@@ -1507,12 +1507,15 @@ function twig_get_attribute(Twig_Environment $env, Twig_Source $source, $object,
...
             }
 
-            if (!isset($cache[$name])) {
-                $cache[$name] = $method;
-                $cache[$lcName] = $method;
+            if (!isset($classCache[$name])) {
+                $classCache[$name] = $method;
+            }
+
+            if (!isset($classCache[$lcName])) {
```
xabbuh: @fabpot Since #2316 checks for existing cache entries have been separated as I did it here. Was it intended to merge them for Twig 2.0?

stof: @xabbuh I don't understand what you mean

xabbuh: @stof In Twig 1.x the code was as it is in this PR. In #2316 this was changed to the following:

```
if (!isset($classCache[$name])) {
    $classCache[$name] = $method;
    $classCache[$lcName] = $method;
}
```

However, this leads to an issue when you have two methods `bar()` and `getBar()`. When `bar()` has been processed, `$classCache['bar']` has the value `bar`. However, when `getBar()` is processed, `$name` is `Bar`. Since this key is not set in `$classCache`, the `bar` key would have been overwritten too.

fabpot: I messed up when moving the method to a function as there is now a clash between a local `$cache` var and the global one. Your change fixes this in a good way, but we could have used the same var directly.

xabbuh: @fabpot Actually, using a local `$classCache` variable does not fix the issue reported in #2343. It only prevents wrong "global" cache entries. The actual fix is the additional `isset()` call for the lowercased name.

fabpot:
👍

fabpot:
Thank you @xabbuh.

PR #2359, opened 2017-01-11
Optimise performance of getAttribute() calls by caching status of sandbox for the current environment

### LINKED ISSUE #2326: Performance Query - getAttribute()
Hi All,

I think I've noticed a possible performance optimization - the template i'm currently including makes 4212 calls to getAttribute(). (if I'm reading the profiler output right - 5.5% of total time)

and 1.2% of the page time was being used by the call to hasExtension() - which was being call 1934 times.

As i'm sure there's people more familiar with twig then myself ;) - is there a minor optimization here to add a:
` 
static $s_sandbox = null;
if ( $s_sandbox === null ) {
$s_sandbox = $this->env->hasExtension( 'Twig_Extension_Sandbox' );
}
`

above the // object property block in template.php.

Then have the 2 checks for the sandbox extension check if s_sandbox is true?

As the hasExtension code does a ltrim, on the string then checks for the class name exists, there would appear to be a optimization here. On the test page i've noticed this on, the size of the xdebug callgrind file froms from 7.164MB to 6.933MB - which seems a reasonable size.

If my logic here is sound, I can generate a pull request - i'd also appreciate suggestions for naming a static variable.

mantis:
Fixes #2326

### DIFF
--- a/lib/Twig/Extension/Core.php
+++ b/lib/Twig/Extension/Core.php
@@ -1463,6 +1463,16 @@ function twig_get_attribute(Twig_Environment $env, Twig_Source $source, $object,
         throw new Twig_Error_Runtime('Accessing Twig_Template attributes is forbidden.');
     }
 
+    static $extensionSandbox = array();
+    $envHash = spl_object_hash($env);
+    if (!isset($extensionSandbox[$envHash])) {
+        if ($env->hasExtension('Twig_Extension_Sandbox')) {
+            $extensionSandbox[$envHash] = $env->getExtension('Twig_Extension_Sandbox');
+        } else {
+            $extensionSandbox[$envHash] = false;
+        }
+    }
+
     // object property
     if (Twig_Template::METHOD_CALL !== $type) {
         if (isset($object->$item) || array_key_exists((string) $item, $object)) {
@@ -1470,8 +1480,8 @@ function twig_get_attribute(Twig_Environment $env, Twig_Source $source, $object,
                 return true;
             }
 
-            if ($env->hasExtension('Twig_Extension_Sandbox')) {
-                $env->getExtension('Twig_Extension_Sandbox')->checkPropertyAllowed($object, $item);
+            if ($extensionSandbox[$envHash] !== false ) {
+                $extensionSandbox[$envHash]->checkPropertyAllowed($object, $item);
             }
 
             return $object->$item;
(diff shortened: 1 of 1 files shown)


stof:
The issue with tests is probably that the same object hash is reused later by a different Environment object

mantis:
I'll need to check what I've done later - the object hash was fabpot's suggestion and I ran phpunit before generating PR and had it passing... So I wonder at what point I've got confused!

lib/Twig/Extension/Core.php
```diff
@@ -1463,15 +1463,25 @@ function twig_get_attribute(Twig_Environment $env, Twig_Source $source, $object,
         throw new Twig_Error_Runtime('Accessing Twig_Template attributes is forbidden.');
     }
 
+    static $extensionSandbox = array();
```
Koc: would it better introduce private non-static class property `hasSandboxExtension`?

mantis:
Regarding the phpunit test failures.... I'm not sure I understand:

PHPUnit 4.2.6 by Sebastian Bergmann.

Configuration read from twig\Twig\phpunit.xml.dist

Testing Twig Test Suite
...SS........................................................   61 / 1405 (  4%)
.............................................................  122 / 1405 (  8%)
.............................................................  183 / 1405 ( 13%)
.............................................................  244 / 1405 ( 17%)
.........................SS..................................  305 / 1405 ( 21%)
.............................................................  366 / 1405 ( 26%)
.............................................................  427 / 1405 ( 30%)
.............................................................  488 / 1405 ( 34%)
.............................................................  549 / 1405 ( 39%)
.............................................................  610 / 1405 ( 43%)
.............................................................  671 / 1405 ( 47%)
.............................................................  732 / 1405 ( 52%)
.............................................................  793 / 1405 ( 56%)
.............................................................  854 / 1405 ( 60%)
.............................................................  915 / 1405 ( 65%)
.............................................................  976 / 1405 ( 69%)
............................................................. 1037 / 1405 ( 73%)
............................................................. 1098 / 1405 ( 78%)
............................................................. 1159 / 1405 ( 82%)
............................................................. 1220 / 1405 ( 86%)
............................................................. 1281 / 1405 ( 91%)
............................................................. 1342 / 1405 ( 95%)
............................................................. 1403 / 1405 ( 99%)
..

Time: 46.29 seconds, Memory: 24.00MB

OK, but incomplete, skipped, or risky tests!
Tests: 1405, Assertions: 3308, Skipped: 4.

lib/Twig/Extension/Core.php
```diff
@@ -1463,15 +1463,25 @@ function twig_get_attribute(Twig_Environment $env, Twig_Source $source, $object,
...
+    $envHash = spl_object_hash($env);
```
mantis: in terms of the spl_object_hash - does it actually makes more sense to look at the value of $this->optionsHash on the environment? [that would appear to get updated whenever a new extension is added]

lib/Twig/Extension/Core.php
mantis: I've put up a different implementation in #2361 . Have also added what would appear to be a missing call to update the options hash. In terms of the non-static class property - the code I was originally modifying wasn't within a class..

fabpot:
@mantis Thanks for working on this. I've spent some time today to see how we could optimize it even more, and I've come up with something that is faster and cleaner IMHO. Closing in favor of #2635.

PR #2361, opened 2017-01-12
Optimise performance of getAttribute() calls by caching environment extension status

mantis:
Alternative implementation following @Koc comment. I'm guessing slightly slower. Passes phpunit locally, so lets see what travis does this time...

### DIFF
--- a/lib/Twig/Environment.php
+++ b/lib/Twig/Environment.php
@@ -38,6 +38,7 @@ class Twig_Environment
     private $strictVariables;
     private $templateClassPrefix = '__TwigTemplate_';
     private $originalCache;
+    private $extensionCache = array();
     private $extensionSet;
     private $runtimeLoaders = array();
     private $runtimes = array();
@@ -590,7 +591,10 @@ class Twig_Environment
      */
     public function hasExtension($class)
     {
-        return $this->extensionSet->hasExtension($class);
+        if( !isset( $this->extensionCache[$this->optionsHash][$class] ) ) {
+            return $this->extensionSet->hasExtension($class);
+        }
+        return $this->extensionCache[$this->optionsHash][$class];
     }
 
     /**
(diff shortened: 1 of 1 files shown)


stof:
Please show numbers to prove that it indeed make things faster

mantis:
@Stof is there a decent benchmark suite for 2.0 ? (the twig-perf suite currently doesn't work on 2.x without patching). In terms of numbers - on windows qcachegrind/xdebug seemed to be showing an increase in performance for my use case.  I'm not replicating much of a difference under linux having patched the testsuite

mantis:
I've also just noticed that a fix went into 2.1 for overriding the static cache object - that actually makes more of a difference. In terms of my local setup, if I do 5 request with and without this change:

image

As you can see, I do get lower scores for the 2nd batch of requests - but there's quite a bit of variance (db calls).

I'm wondering if I should do a seperate patch to add a updateOptionsHash call to setExtensions and come back to this...

mantis:
fwiw - in terms of numbers, between this and the previous attempt, if i set up a run using paesslers webserver stress tool (yes, windows i know), I'm seeing following response times for a 1 minute run for my use case:

WITHOUT:

Total Number of Clicks: 363 (0 Errors)
Average Click Time of all URLs: 722 ms 

WITH:

Total Number of Clicks: 487 (0 Errors)
Average Click Time of all URLs: 521 ms

stof:
Well, twig-perf has not been updated since 2014. I don't know whether @fabpot has another benchmark suite.
AFAIK, we mostly check performance impact on real projects using Twig by profiling them.

lib/Twig/Environment.php
```diff
@@ -590,7 +591,10 @@ public function getCharset()
      */
     public function hasExtension($class)
     {
-        return $this->extensionSet->hasExtension($class);
+        if( !isset( $this->extensionCache[$this->optionsHash][$class] ) ) {
+            return $this->extensionSet->hasExtension($class);
+        }
+        return $this->extensionCache[$this->optionsHash][$class];
```
gharlan: You do not write to `extensionCache` in `hasExtension`, so it does improve the performance when calling `hasExtension` multiple times.

stof:
@fabpot do you have any uptodate Twig performance benchmark ?

fabpot:
@stof just pushed an updated version.

stof:
hmm, we will need to add some sandboxed tests in it though

fabpot:
Closing in favor of #2635.

PR #2364, opened 2017-01-16
fixed getNode in Expression_Array

pedro151:


stof:
This is a BC break as it totally changes the behavior of these methods

stof:
Btw, you don't even describe what you want to fix

pedro151:
The way that data is retrieved from a given node in the class is best having an associative index

stof:
You are changing totally the behavior of these methods.

And this means that the behavior of these methods does not correspond to the Twig_Node contract anymore.
Btw, the testsuite shows that this breaks things.

``getNode`` and ``hasNode`` is not meant to access properties of the array by key, but accessing Twig nodes in the AST.

Thus, the API you suggest is broken. Keys in pairs are **not** strings. They may be very complex Twig expressions. There is no way to get the value of an array for a given key using just the AST, as actual keys are not known until the execution of the template.

stof:
Please revert all permission changes. Files must not be executable

composer.json
```diff
@@ -1,28 +1,10 @@
 {
-    "name": "twig/twig",
+    "name": "pedro151/twig",
```
stof: this is a no-go

lib/Twig/Node/Expression/Array.php
```diff
@@ -8,74 +8,140 @@
...
+        {
+            $key = new Twig_Node_Expression_Constant( ++ $this->index , $value->getTemplateLine () );
         }
 
-        array_push($this->nodes, $key, $value);
+        array_push ( $this->nodes , $key , $value );
     }
 
-    public function compile(Twig_Compiler $compiler)
+    public function compile ( Twig_Compiler $compiler )
```
stof: why are you changing coding standards ? If you want it to be merged, this must be reverted (and if you don't want it to be merged, you should close the PR)

pedro151:
Sorry, I sent files that should not be loaded and merged by themselves. Modifications already removed

PR #2370, opened 2017-01-19
Update striptags.rst

peter17:
Add information about supported second param.

### DIFF
--- a/doc/filters/striptags.rst
+++ b/doc/filters/striptags.rst
@@ -12,12 +12,10 @@ by one space:
 
     Internally, Twig uses the PHP `strip_tags`_ function.
 
-.. _`strip_tags`: http://php.net/strip_tags
-
-    As a consequence, a second parameter, `$allowable_tags`, can be provided. In this example, the tags like `<br/>`, `<br>`, `<p>` and `</p>` will remain in the string:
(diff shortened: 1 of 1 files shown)


doc/filters/striptags.rst
```diff
@@ -13,3 +13,11 @@ by one space:
     Internally, Twig uses the PHP `strip_tags`_ function.
 
 .. _`strip_tags`: http://php.net/strip_tags
+
+    As a consequence, a second parameter, `$allowable_tags`, can be provided. In this example, the tags like `<br/>`, `<br>`, `<p>` and `</p>` will remain in the string:
```
stof: this should be above the list of links (which should stay at the bottom of the file)

doc/filters/striptags.rst
```diff
@@ -13,3 +13,11 @@ by one space:
...
+
+.. code-block:: jinja
+
+    {{ some_html|striptags('<br><p>') }}
+
+.. note::
```
stof: why is there an empty note here

doc/filters/striptags.rst
stof: I would not say ``a second parameter`` but ``a parameter``, as the first argument of a filter does not look like a parameter in the Twig syntax

stof: inline code blocks must used 2 backticks, not 1. This is a rST file, not a markdown one. A single backtick creates a rST reference

peter17:
Replaced by #2371

PR #2374, opened 2017-01-23
Remove legacy exception catch blocks from 2.x

chalasr:
These `catch` blocks were only useful for compatibility with PHP versions prior to 7.0.


fabpot:
Good catch, thanks @chalasr.

PR #2378, opened 2017-01-27
Update Twig_Filter references to Twig_SimpleFilter

Mark-H:
It appears that at some point in the past (the `Twig_filter` class says v1.12) what used to be a `Twig_Filter` turned into a `Twig_SimpleFilter`. This PR updates the Extension documentation to that. It looks like this was updated in the 1.x documentation, but not the 2.x documentation which is served up  by default when browsing http://twig.sensiolabs.org/

### DIFF
--- a/doc/advanced.rst
+++ b/doc/advanced.rst
@@ -123,23 +123,23 @@ Filters
 Creating a filter is as simple as associating a name with a PHP callable::
 
     // an anonymous function
-    $filter = new Twig_Filter('rot13', function ($string) {
+    $filter = new Twig_SimpleFilter('rot13', function ($string) {
         return str_rot13($string);
     });
 
     // or a simple PHP function
-    $filter = new Twig_Filter('rot13', 'str_rot13');
+    $filter = new Twig_SimpleFilter('rot13', 'str_rot13');
 
     // or a class static method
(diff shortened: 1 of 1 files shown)


SpacePossum:
Are you sure it not the other way around?
It seems that `Twig_SimpleFilter` is deprecated in fav. of `Twig_Filter` https://github.com/twigphp/Twig/blob/2.x/lib/Twig/SimpleFilter.php#L13

stof:
In 2.x, Twig_Filter is the same than Twig_SimpleFilter. And Twig_SimpleFilter is expected to disappear in Twig 3

Mark-H:
Whoops! I verified with the code in my project, which I thought was on latest, but that appears to be latest 1.x. The docblock in 1.x [claimed Twig_Filter was to be removed in 2.0](https://github.com/twigphp/Twig/blob/1.x/lib/Twig/Filter.php#L21) in favour of Twig_SimpleFilter, but that seems to have changed since the deprecation notice. 

Sorry, I should have looked better before sending a PR.

stof:
Well, the 1.x class has been removed, replacing it with another one (as @fabpot wanted to have the simpler API named Twig_Filter from the start but the name was not available).

Looking back, reusing the same name in 2.x for a different class was probably a bad decision for the upgrade path.

Issue #2379, opened 2017-01-28
Twig formatters "date" and "date_modify" works on null values and implicitly treats that as NOW()


antanas-arvasevicius:
Hi,
I've bug in my project and found out that twig date formatters can be used on null values which implicitly will be converted to current date time for calculations.
I think it would be better to throw exception because formatting a null is non sense. And implicit null to "now" conversion is hard to remember.

SpacePossum:
This cannot be done because if BC reasons, @ see the discussion here https://github.com/twigphp/Twig/issues/1983

antanas-arvasevicius:
I think this BC will help people to find a bugs in their codes. And leaving that kind of "documented features" fearing of BC do not helps anyone. At least for me where I've debugged 2 hours of what's happened in my code, so I hope this fix will help to save someone's time.
Super strict BC fame is not good for developers, look at Java or PHP how much non rationale things left in case of BC. Libraries should not follow that path.

xabbuh:
If this change is to be done, it has to be made in Twig 3.0 while the current behaviour would have to be deprecated in 2.x.

antanas-arvasevicius:
OK, I'm leaving PR open for twig 3, but don't understand the reason why we should not do BC for a minority of people who uses that illogical "feature" (where probably there is a bug inside their system as I can't imagine where intentionally could be used this null | date() formatting feature) and help them fix their systems if they do updates. And waiting for twig 3 is not an option, because where will be also illogical "features" where would be fixed in twig 4 and so on.. so I think the problem is in our fears of BC which probably breaks <1% buggy systems out there. But, wait, maybe we could add a deprecation warning instead of exception now in 1.x  to help other spot problems in their systems?

fabpot:
We made a conscious choice of allowing `null` here to be converted to `now`. So, I'm closing the issue and the related PR.

antanas-arvasevicius:
And how to test and mock these consciously chosen values? How about timezones? These implicit "features" just hides bugs most of the time.

PR #2380, opened 2017-01-28
bugfix #2379 Twig formatters "date" and "date_modify" works on null values

antanas-arvasevicius:
see https://github.com/twigphp/Twig/issues/2379


stof:
This is a BC break, so it is totally impossible to merge this in 1.x or 2.x as a bugfix.

stof:
What can be done (and in 2.x, not in 1.x), is triggering a deprecation warning when null is passed to the filter.

Issue #2383, opened 2017-01-30
Undefined variable 'source' in lib/Twig/Environment.php


charlieman:
In [line 377](https://github.com/twigphp/Twig/blob/465689d/lib/Twig/Environment.php#L377). I'm not sure if this should simply be null or the result of `$this->getLoader()->getSourceContext($name)`.

PR #2384, opened 2017-01-31
Fix undef. source in exception.

SpacePossum:
closes https://github.com/twigphp/Twig/issues/2383

Two things:
- the tests fails without a corrupt cache, so maybe the message is of or the cache generation is off (i.e. not respecting the `$index`)
- passing the `$source` seems superfluous (but maybe that is only the case in the exception message)


stof:
👍

fabpot:
Thank you @SpacePossum.

Issue #2358, opened 2017-01-11
Support ===


dawehner:
I just wrote some twig template which contained ```===``` and the parser failed:

```
Twig_Error_Syntax: Unexpected token "operator" of value "=" in "foo.html.twig" at line 19. in Twig_ExpressionParser->parsePrimaryExpression() (line 208 of vendor/twig/twig/lib/Twig/ExpressionParser.php).
```

I wonder whether twig should simply treat ```===``` as ```==``` or even provide some form of strict comparison.

fabpot:
`===` does exist in Twig, but the name of the test is `same as`.

dawehner:
Right, I'm simply wondering whether === should be treated as "same as" automatically. I'm wondering whether its just me who is so stupid falling into that trap :)

dawehner:
This is really not like a feature request or so, I'm really just wondering

fabpot:
That's a good question and one for which we had a discussion in the past. The conclusion was that `===` is confusing for anyone not familiar with PHP. So, we decided to support it via `same as` instead.

dawehner:
One thing which could be done is to not support "===" but instead throw a helpful exception telling people about "same as"

fabpot:
That makes sense. Anyone willing to submit a PR for that? Could be done in the 1.x branch.

dawehner:
Absolutely

dawehner:
:+1: for closing PRs.

fabpot:
@dawehner It's not closed but merged ;)

dawehner:
Oh haha. Well even better :)

fabpot:
See #4736

PR #2389, opened 2017-02-04
Add support for === and !==

dawehner:
Solves #2358

### DIFF
--- a/lib/Twig/Node/Expression/Binary/StrictEqual.php
+++ b/lib/Twig/Node/Expression/Binary/StrictEqual.php
@@ -1,0 +1,17 @@
+<?php
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+class Twig_Node_Expression_Binary_StrictEqual extends Twig_Node_Expression_Binary
+{
+    public function operator(Twig_Compiler $compiler)
+    {
+        return $compiler->raw('===');
+    }
+}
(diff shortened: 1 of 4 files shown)


lib/Twig/Node/Expression/Binary/StrictEqual.php
```diff
@@ -0,0 +1,17 @@
...
+ * (c) Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+class Twig_Node_Expression_Binary_StrictEqual extends Twig_Node_Expression_Binary
+{
+    public function operator(Twig_Compiler $compiler)
+    {
+        return $compiler->raw('===');
```
SpacePossum: I'm confused, this is causing an exception being thrown with a message informing that `same as` should be used right? (https://github.com/twigphp/Twig/issues/2358#issuecomment-271995283)

dawehner: Oh you are totally right. I was just not really awake this morning.

robfrawley:
@fabpot I absolutely understand why you chose to support `same as` - to allow easy adoption by those unfamiliar with PHP. That said, I can't understand the number of times I've typed in `===` only to compile the template and go "oh, right, I can't use that" - it would be nice to document the preferred syntax of `same as` while *also* allow the use of `===` for those more accustomed to it.

stof:
What @fabpot accepted in the issue discussion is *not* adding support for it, but throwing an helpful exception.

fabpot:
Closing this one as this is not what needs to be done.

robfrawley:
Disappointed, but fair enough @fabpot. I understand why you don't want to pollute the language with multiple syntaxes.

PR #2391, opened 2017-02-07
Change ~ with ^ for version in composer require

bocharsky-bw:
Caret symbol sticks closer to semantic versioning, and will always allow non-breaking updates

### DIFF
--- a/doc/intro.rst
+++ b/doc/intro.rst
@@ -37,7 +37,7 @@ The recommended way to install Twig is via Composer:
 
 .. code-block:: bash
 
-    composer require "twig/twig:~2.0"
+    composer require "twig/twig:^2.0"
 
 .. note::
 


robfrawley:
The caret symbol only behaves differently when a patch number is specified or a `>1.0` version is specified, neither of which are the case here. In the given example, the two version constraints behave exactly the same. Why change it?

bocharsky-bw:
@robfrawley Yeah, I know... but the original idea is that caret symbol sticks closer to semantic versioning and IMO libraries should suggest it in the first place. `~` symbol should be used for some edge cases where it makes sense, but I don't see them in this line.

This version notation stores in user's composer.json and if user change it slightly to something with patch number, it will behave differently. I bet many people don't know the difference between `^` and `~`.

fabpot:
Thank you @bocharsky-bw.

PR #2394, opened 2017-02-13
Deprecate warnings and upgrade path.

SpacePossum:
The classes
- Twig_SimpleFilter
- Twig_SimpleFunction
 -Twig_SimpleTest

were left in the 2.x release `for Twig 1.x compatibility`. However since the classes are now declared as `final` it won't help for projects extending those classes ([example](https://github.com/GeckoPackages/GeckoTwig/blob/v1.3/src/Twig/Filters/BytesFilter.php#L53)).

### DIFF
--- a/lib/Twig/SimpleFilter.php
+++ b/lib/Twig/SimpleFilter.php
@@ -12,6 +12,15 @@
 /**
  * Empty class for Twig 1.x compatibility.
  */
-final class Twig_SimpleFilter extends Twig_Filter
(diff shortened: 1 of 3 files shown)


stof:
we added ``@final`` in 1.x, so the Symfony DebugClassLoader warns you when extending them. Marking them as final was done on purpose.

and regarding deprecation warnings, see the discussion in https://github.com/twigphp/Twig/pull/2181
We don't want to add them yet.

SpacePossum:
alright, thank you

PR #2401, opened 2017-02-17
Add ContainerRuntimeLoader

chalasr:
Next to https://github.com/symfony/symfony/pull/21625#discussion_r101715445

### DIFF
--- a/lib/Twig/ContainerRuntimeLoader.php
+++ b/lib/Twig/ContainerRuntimeLoader.php
@@ -1,0 +1,37 @@
+<?php
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Psr\Container\ContainerInterface;
+
+/**
+ * Lazily loads Twig runtime implementations from a PSR-11 container.
+ *
+ * The runtime services must use their class name as identifier in the container.
+ *
+ * @author Fabien Potencier <fabien@symfony.com>
+ * @author Robin Chalas <robin.chalas@gmail.com>
+ */
+class Twig_ContainerRuntimeLoader implements Twig_RuntimeLoaderInterface
+{
+    private $container;
+
+    public function __construct(ContainerInterface $container)
+    {
+        $this->container = $container;
+    }
+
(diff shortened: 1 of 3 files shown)


lib/Twig/ContainerRuntimeLoader.php
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    public function load($class)
+    {
+        if ($this->container->has($class)) {
+            return $this->container->get($class);
+        }
+
+        if (null !== $this->logger) {
+            $this->logger->warning(sprintf('Class "%s" is not configured as a Twig runtime service.', $class));
```
stof: I'm not sure we should keep this logging call. If another loader is able to load the runtime, this is just spamming logs (especially at warning level). and if no loader can load it, ``Twig_Environment::getRuntime`` throws an exception with a meaningful message already

chalasr: Indeed, removed the logger dependency.

lib/Twig/ContainerRuntimeLoader.php
```diff
@@ -0,0 +1,35 @@
...
+ * (c) Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Psr\Container\ContainerInterface;
+
+/**
+ * Lazily loads Twig runtime implementations from a container.
```
stof: I would say ``from a PSR-11 container``.

And the long description of the class should explain that it expects being able to use the runtime class as an identifier in the container

chalasr: done

fabpot:
Should be for the 1.x branch in which we still can add new features (as long as they don't break BC of course as we cannot add new deprecation notices).

chalasr:
Rebased!

CHANGELOG
```diff
@@ -1,4 +1,8 @@
-* 1.31.0 (2017-XX-XX)
+* 1.31.1 (2017-XX-XX)
```
fabpot: That's a new feature, so we need to bump to 1.32

fabpot:
I will take care of bumping the version when merging.

fabpot:
Thank you @chalasr.

Issue #1802, opened 2015-09-01
left/right trim


abhibeckert:
When the `trim` filter was added, there was some discussion about extending it to include `ltrim`/`rtrim` functionality: https://github.com/twigphp/Twig/pull/662

Should this be done? I think so, in particular when working with `<a href>` tags I often need to trim a leading slash.

If I do submit a pull request adding this to twig, what syntax should I go with?
- `<a href="{{ foo|trim('/', 'left') }}">`
- `<a href="{{ foo|ltrim('/') }}">`
- `<a href="{{ foo|left_trim('/') }}">`

ninsuo:
`ltrim` and `rtrim` seems easier to use IMO

But to avoid adding too much filters in Twig Core, those may be implemented in the [Text extension](http://twig.sensiolabs.org/doc/extensions/text.html).

fabpot:
As this can be done easily outside of core and because not that many people ask for this, I'm closing this feature request.

fabpot:
Thinking again, back when we introduced the trim filter, we didn't have the named parameters, which is now the case, so adding a way to only trim on the left or on the right might be easier to include in the existing trim filter.

antonydandrea:
I would really like to contribute some code to this project, can I have a go at implementing this?

ninsuo:
Everybody is welcome to contribute, don't ask for permission 😄

antonydandrea:
Just didn't want to step on any toes of someone was already working on it

antonydandrea:
Is there a reason why we would want them as parameters of trim rather than introducing rtrim and ltrim filters?

antonydandrea:
Ok, I have done a pull request #2403

PR #2403, opened 2017-02-18
#1802 left/right trim

antonydandrea:
It was said in this thread that there was a preference of adding parameters to the trim filter in order to implement this.

So this feature works like:

trim(character_mask, mode), where mode is either "both", "left" or "right". It is by default "both".

This is my first pull request of this project, so I apologise in advance if something is wrong.

### DIFF
--- a/doc/filters/trim.rst
+++ b/doc/filters/trim.rst
@@ -28,6 +28,6 @@ Arguments
 * ``character_mask``: The characters to strip
 
 * ``type``: The default is to strip from the start and the end (`both`), but `left`
-and `right` will strip from either the left side or right side only.
+  and `right` will strip from either the left side or right side only.
 
 .. _`trim`: http://php.net/trim
--- a/doc/filters/trim.rst
+++ b/doc/filters/trim.rst
@@ -1,5 +1,8 @@
 ``trim``
 ========
+.. versionadded:: 1.32
+
+    The ``type`` argument was added in Twig 1.32.
 
 The ``trim`` filter strips whitespace (or other characters) from the beginning
 and end of a string:
@@ -24,7 +27,7 @@ and end of a string:
 
 .. note::
 
-    Internally, Twig uses the PHP `trim`_, `ltrim`_ and `rtrim`_ functions.
+    Internally, Twig uses the PHP `trim`_, `ltrim`_, and `rtrim`_ functions.
 
 Arguments
 ---------
@@ -32,7 +35,7 @@ Arguments
 * ``character_mask``: The characters to strip
 
 * ``type``: The default is to strip from the start and the end (`both`), but `left`
-  and `right` will strip from either the left side or right side only.
+  and `right` will strip from either the left side or right side only
 
 .. _`trim`: http://php.net/trim
 .. _`ltrim`: http://php.net/ltrim
--- a/doc/filters/trim.rst
+++ b/doc/filters/trim.rst
@@ -26,6 +26,8 @@ Arguments
 ---------
 
 * ``character_mask``: The characters to strip
-* ``mode``: The default is to strip from the start and the end ('both'), but 'left'
-and 'right' will strip from either the left side or right side only.
+
+* ``mode``: The default is to strip from the start and the end (`both`), but `left`
+and `right` will strip from either the left side or right side only.
+
 .. _`trim`: http://php.net/trim
--- a/doc/filters/trim.rst
+++ b/doc/filters/trim.rst
@@ -35,3 +35,5 @@ Arguments
   and `right` will strip from either the left side or right side only.
 
 .. _`trim`: http://php.net/trim
+.. _`ltrim`: http://php.net/ltrim
+.. _`rtrim`: http://php.net/rtrim
--- a/doc/filters/trim.rst
+++ b/doc/filters/trim.rst
@@ -14,6 +14,10 @@ and end of a string:
 
     {# outputs '  I like Twig' #}
 
+    {{ '  I like Twig.  '|rtrim('', 'left') }}
+
+    {# outputs '  I like Twig. #}
+
 .. note::
 
     Internally, Twig uses the PHP `trim`_ function.
@@ -22,5 +26,6 @@ Arguments
 ---------
 
 * ``character_mask``: The characters to strip
-
+* ``mode``: The default is to strip from the start and the end ('both'), but 'left'
+and 'right' will strip from either the left side or right side only.
 .. _`trim`: http://php.net/trim
--- a/doc/filters/trim.rst
+++ b/doc/filters/trim.rst
@@ -14,7 +14,7 @@ and end of a string:
 
     {# outputs '  I like Twig' #}
 
-    {{ '  I like Twig.  '|rtrim('', 'left') }}
+    {{ '  I like Twig.  '|trim('', 'left') }}
 
     {# outputs '  I like Twig. #}
 
--- a/doc/filters/trim.rst
+++ b/doc/filters/trim.rst
@@ -14,7 +14,7 @@ and end of a string:
 
     {# outputs '  I like Twig' #}
 
-    {{ '  I like Twig.  '|trim('', 'left') }}
+    {{ '  I like Twig.  '|trim(type='left') }}
 
     {# outputs '  I like Twig. #}
 
@@ -27,7 +27,7 @@ Arguments
 
 * ``character_mask``: The characters to strip
 
-* ``mode``: The default is to strip from the start and the end (`both`), but `left`
+* ``type``: The default is to strip from the start and the end (`both`), but `left`
 and `right` will strip from either the left side or right side only.
 
 .. _`trim`: http://php.net/trim
--- a/doc/filters/trim.rst
+++ b/doc/filters/trim.rst
@@ -16,7 +16,7 @@ and end of a string:
 
     {{ '  I like Twig.  '|trim(type='left') }}
 
-    {# outputs '  I like Twig. #}
+    {# outputs 'I like Twig.  ' #}
 
 .. note::
 
--- a/doc/filters/trim.rst
+++ b/doc/filters/trim.rst
@@ -2,7 +2,7 @@
 ========
 .. versionadded:: 1.32
 
-    The ``type`` argument was added in Twig 1.32.
+    The ``side`` argument was added in Twig 1.32.
 
 The ``trim`` filter strips whitespace (or other characters) from the beginning
 and end of a string:
@@ -17,7 +17,7 @@ and end of a string:
 
     {# outputs '  I like Twig' #}
 
-    {{ '  I like Twig.  '|trim(type='left') }}
+    {{ '  I like Twig.  '|trim(side='left') }}
 
     {# outputs 'I like Twig.  ' #}
 
@@ -34,7 +34,7 @@ Arguments
 
 * ``character_mask``: The characters to strip
 
-* ``type``: The default is to strip from the start and the end (`both`), but `left`
+* ``side``: The default is to strip from the left and the right (`both`) sides, but `left`
   and `right` will strip from either the left side or right side only
 
 .. _`trim`: http://php.net/trim
--- a/doc/filters/trim.rst
+++ b/doc/filters/trim.rst
@@ -18,9 +18,13 @@ and end of a string:
 
     {# outputs 'I like Twig.  ' #}
 
+    {{ '  I like Twig.  '|trim(' ', 'right') }}
+
+    {# outputs '  I like Twig.' #}
+
 .. note::
 
-    Internally, Twig uses the PHP `trim`_ function.
+    Internally, Twig uses the PHP `trim`_, `ltrim`_ and `rtrim`_ functions.
 
 Arguments
 ---------
--- a/lib/Twig/Extension/Core.php
+++ b/lib/Twig/Extension/Core.php
@@ -860,6 +860,8 @@ function twig_in_filter($value, $compare)
  * @param string $mode
  *
  * @return string
+ *
+ * @throws Twig_Error_Runtime When an invalid type is used (not 'left', 'right' or 'both')
  */
 function twig_trim_filter($string, $characterMask = null, $type = 'both')
 {
@@ -872,8 +874,9 @@ function twig_trim_filter($string, $characterMask = null, $type = 'both')
         case 'right':
             return rtrim($string, $characterMask);
         case 'both':
-        default:
             return trim($string, $characterMask);
+        default:
+            throw new Twig_Error_Runtime(sprintf('trim type must be "left", "right" or "both". Got %s.', $type));
     }          
 }
 
--- a/lib/Twig/Extension/Core.php
+++ b/lib/Twig/Extension/Core.php
@@ -860,8 +860,11 @@ function twig_in_filter($value, $compare)
  * @param string $mode
  * @return string
  */
-function twig_trim_filter($string, $characterMask = " \t\n\r\0\x0B", $mode = 'both')
+function twig_trim_filter($string, $characterMask = null, $mode = 'both')
 {
+    if (empty($characterMask)) {
+        $characterMask = " \t\n\r\0\x0B";
+    }
     switch ($mode) {
         case 'left':
             $trimmed = ltrim($string, $characterMask);
--- a/lib/Twig/Extension/Core.php
+++ b/lib/Twig/Extension/Core.php
@@ -860,19 +860,19 @@ function twig_in_filter($value, $compare)
  * @param string $mode
  * @return string
  */
-function twig_trim_filter($string, $character_mask = null, $type = 'both')
+function twig_trim_filter($string, $characterMask = null, $type = 'both')
 {
-    if (null === $character_mask) {
-        $character_mask = " \t\n\r\0\x0B";
+    if (null === $characterMask) {
+        $characterMask = " \t\n\r\0\x0B";
     }
     switch ($type) {
         case 'left':
-            return ltrim($string, $character_mask);
+            return ltrim($string, $characterMask);
         case 'right':
-            return rtrim($string, $character_mask);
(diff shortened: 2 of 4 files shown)


doc/filters/trim.rst
```diff
@@ -14,6 +14,10 @@ and end of a string:
 
     {# outputs '  I like Twig' #}
 
+    {{ '  I like Twig.  '|rtrim('', 'left') }}
```
SpacePossum: `rtrim` is not added as filter in this PR, so I guess maybe a typo?

antonydandrea: Yes that is a typo. I have updated with a fix.

doc/filters/trim.rst
```diff
@@ -22,5 +26,6 @@ Arguments
 ---------
 
 * ``character_mask``: The characters to strip
-
+* ``mode``: The default is to strip from the start and the end ('both'), but 'left'
+and 'right' will strip from either the left side or right side only.
```
xabbuh: I think you should keep a blank line before the next one.

xabbuh: And I would enclose `both`, `left` and `right` in double backticks to get proper syntax highlighting in the rendered docs.

lib/Twig/Extension/Core.php
```diff
@@ -852,6 +852,34 @@ function twig_in_filter($value, $compare)
...
+ * 
+ * @param string $string
+ * @param string $characterMask
+ * @param string $mode
+ * @return string
+ */
+function twig_trim_filter($string, $characterMask = null, $mode = 'both')
+{
+    if (empty($characterMask)) {
+        $characterMask = " \t\n\r\0\x0B";
```
xabbuh: Why not setting this as the parameter's default value instead?

lib/Twig/Extension/Core.php
```diff
@@ -852,6 +852,34 @@ function twig_in_filter($value, $compare)
...
+    }
+    switch ($mode) {
+        case 'left':
+            $trimmed = ltrim($string, $characterMask);
```
xabbuh: You can just return here instead of introducing the `$trimmed` variable (same below).

antonydandrea: Is it not better practice to have only one "Return" in a function rather than multiple?

lib/Twig/Extension/Core.php
antonydandrea: If someone passes "empty string" instead of null as the first parameter, I believe it will override the default mask. So this catches all of those possibilities.

jvasseur: An empty string should not make the function function work like passing no parameter/passing null.

And this would be a BC break since currently `|trim('')` doesn't trim anything.

antonydandrea: Understood. How should we handle this first parameter when people want to have the second parameter but use only the default mask?

jvasseur: Either `|trim(mode='left')` or `|trim(null, 'left')`.

lib/Twig/Extension/Core.php
```diff
@@ -860,8 +860,11 @@ function twig_in_filter($value, $compare)
  * @param string $mode
  * @return string
  */
-function twig_trim_filter($string, $characterMask = " \t\n\r\0\x0B", $mode = 'both')
+function twig_trim_filter($string, $characterMask = null, $mode = 'both')
 {
+    if (empty($characterMask)) {
```
SpacePossum: I think a check against `null` is better, as this way you cannot set `$characterMask` to a single space (' ' is empty).

lib/Twig/Extension/Core.php
```diff
@@ -852,6 +852,34 @@ function twig_in_filter($value, $compare)
...
 
+/** 
+ * Returns a trimmed string.
+ * 
+ * @param string $string
+ * @param string $characterMask
+ * @param string $mode
+ * @return string
+ */
+function twig_trim_filter($string, $characterMask = null, $mode = 'both')
```
javiereguiluz: If I'm right, in English they say *"trim from [left/right/both sides]"* ... so maybe we could rename the `$mode` parameter to either `from` or `side`:

```twig
{{ '...'|trim(mode='both') }}
{{ '...'|trim(mode='left') }}
{{ '...'|trim(mode='right') }}

{{ '...'|trim(from='both') }}
{{ '...'|trim(from='left') }}
{{ '...'|trim(from='right') }}

{{ '...'|trim(side='both') }}
{{ '...'|trim(side='left') }}
{{ '...'|trim(side='right') }}
```

antonydandrea: Also throwing out there, PHP's "str_pad" uses the word "_type_" for declaring left and right.

lib/Twig/Extension/Core.php
```diff
@@ -852,6 +852,33 @@ function twig_in_filter($value, $compare)
...
+ * Returns a trimmed string.
+ * 
+ * @param string $string
+ * @param string $character_mask
+ * @param string $mode
+ * @return string
+ */
+function twig_trim_filter($string, $character_mask = null, $type = 'both')
+{
+    if (is_null($character_mask)) {
```
SpacePossum: please use `null ===`  (slightly faster and cannot be overridden)

lib/Twig/Extension/Core.php
```diff
@@ -852,6 +852,33 @@ function twig_in_filter($value, $compare)
...
+        $character_mask = " \t\n\r\0\x0B";
+    }
+    switch ($type) {
+        case 'left':
+            return ltrim($string, $character_mask);
+            break;
```
SpacePossum: these `breaks` now have no function, not sure if these should stay?

lib/Twig/Extension/Core.php
```diff
@@ -852,6 +852,30 @@ function twig_in_filter($value, $compare)
...
+    if (null === $character_mask) {
+        $character_mask = " \t\n\r\0\x0B";
+    }
+    switch ($type) {
+        case 'left':
+            return ltrim($string, $character_mask);
+        case 'right':
+            return rtrim($string, $character_mask);
+        case 'both':
+        default:
```
julienfalque: Wouldn't it be better to throw an exception here instead of fallbacking to `both`, which is already the default value of `$type` parameter anyway?

antonydandrea: @julienfalque I did think this, but I wasn't sure which Exception. Is there a specific Twig Exception you would throw?

lib/Twig/Extension/Core.php
```diff
@@ -852,6 +852,30 @@ function twig_in_filter($value, $compare)
...
 
+/** 
+ * Returns a trimmed string.
+ * 
+ * @param string $string
+ * @param string $character_mask
+ * @param string $mode
+ * @return string
+ */
+function twig_trim_filter($string, $character_mask = null, $type = 'both')
```
julienfalque: Should be `$characterMask`.

antonydandrea: @julienfalque would that work for named parameters? As the documentation has "character_mask".

lib/Twig/Extension/Core.php
julienfalque: I guess `Twig_Error_Runtime` is fine.

lib/Twig/Extension/Core.php
antonydandrea: @julienfalque answered my own question: it does work.

doc/filters/trim.rst
```diff
@@ -14,6 +14,10 @@ and end of a string:
 
     {# outputs '  I like Twig' #}
 
+    {{ '  I like Twig.  '|trim(type='left') }}
+
+    {# outputs '  I like Twig. #}
```
lyrixx: It looks like a `'` is missing.

lib/Twig/Extension/Core.php
julienfalque: Great, I forgot about named arguments so I learnt something :)

lib/Twig/Extension/Core.php
```diff
@@ -852,6 +852,30 @@ function twig_in_filter($value, $compare)
     return false;
 }
 
+/** 
+ * Returns a trimmed string.
+ * 
+ * @param string $string
+ * @param string $character_mask
+ * @param string $mode
+ * @return string
```
julienfalque: Missing empty line before `@return` tag.

lib/Twig/Extension/Core.php
```diff
@@ -852,6 +852,30 @@ function twig_in_filter($value, $compare)
     return false;
 }
 
+/** 
+ * Returns a trimmed string.
+ * 
+ * @param string $string
+ * @param string $character_mask
```
julienfalque: Should be `$characterMask`.

doc/filters/trim.rst
```diff
@@ -23,4 +27,7 @@ Arguments
 
 * ``character_mask``: The characters to strip
 
+* ``type``: The default is to strip from the start and the end (`both`), but `left`
+and `right` will strip from either the left side or right side only.
```
julienfalque: This line should be indented with two spaces so it is rendered as part of the bullet point instead of a new line. See https://github.com/antonydandrea/Twig/blob/2.x_ltrim_rtrim/doc/filters/trim.rst.

lib/Twig/Extension/Core.php
```diff
@@ -852,6 +852,34 @@ function twig_in_filter($value, $compare)
...
+    }
+    switch ($type) {
+        case 'left':
+            return ltrim($string, $characterMask);
+        case 'right':
+            return rtrim($string, $characterMask);
+        case 'both':
+            return trim($string, $characterMask);
+        default:
+            throw new Twig_Error_Runtime(sprintf('trim type must be "left", "right" or "both". Got %s.', $type));
```
julienfalque: "Trimming type"?

Btw, I think this will break if `$type` is passed something that cannot be converted to a string, e.g. an array.

antonydandrea: Yes it would. Is there any value showing the given type?

julienfalque: The function receives the raw value, so you can use `gettype()`, `is_array()` or whatever you want to use.

SpacePossum:
```php
throw new Twig_Error_Runtime(sprintf('trim type must be "left", "right" or "both", got "%s".', is_object($type) ? get_class($type) : (gettype($type).'#'.$type)));
```
(this will fail `resources` though)

On a side note, as `both` is prop. used the most, please make it the first check of the `switch`

lib/Twig/Extension/Core.php
```diff
@@ -852,6 +852,39 @@ function twig_in_filter($value, $compare)
...
+ * Returns a trimmed string.
+ * 
+ * @param string $string
+ * @param string $characterMask
+ * @param string $mode
+ *
+ * @return string
+ *
+ * @throws Twig_Error_Runtime When an type is not a string.
+ * @throws Twig_Error_Runtime When an invalid type is used (not 'left', 'right' or 'both')
```
SpacePossum: can be merged into one line

lib/Twig/Extension/Core.php
```diff
@@ -852,6 +852,39 @@ function twig_in_filter($value, $compare)
...
+ */
+function twig_trim_filter($string, $characterMask = null, $type = 'both')
+{
+    if (!is_string($type)) {
+        throw new Twig_Error_Runtime(sprintf('Trimming type must be a string but got %s.', gettype($type)));
+    }
+    if (!in_array($type, ['both', 'left', 'right'])) {
```
SpacePossum: I would throw the exception in the `default` of the switch, as most of the time this will be false it is better for speed to not test it every time on forhand

lib/Twig/Extension/Core.php
```diff
@@ -852,6 +852,39 @@ function twig_in_filter($value, $compare)
...
+        throw new Twig_Error_Runtime(sprintf('Trimming type must be "left", "right" or "both" but got %s.', $type));
+    }
+    if (null === $characterMask) {
+        $characterMask = " \t\n\r\0\x0B";
+    }
```
SpacePossum: One last CS thingy, please add a blank line after a control structure.
Beside these few minors this looks good to me, thanks for adding this, much appreciated  :+1:  :)

lib/Twig/Extension/Core.php
```diff
@@ -852,6 +852,41 @@ function twig_in_filter($value, $compare)
...
+/** 
+ * Returns a trimmed string.
+ * 
+ * @param string $string
+ * @param string $characterMask
+ * @param string $mode
+ *
+ * @return string
+ *
+ * @throws Twig_Error_Runtime When an type is not a string or when an invalid trimming type is used (not 'left', 'right' or 'both')
```
julienfalque:
```diff
-When an type
+When trimming type
```

IMO a shorter sentence like _"When an invalid trimming type is used"_ would be enough, but I'm fine with this one.

antonydandrea: You're right, there is a typo there also, and it is a bit wordy

SpacePossum:
:+1: shippit :)

antonydandrea:
Thanks everyone for being so welcoming to a newbie and for helping make the PR meet the coding standards of the project. I hope to contribute again in the future!

lib/Twig/Extension/Core.php
```diff
@@ -852,6 +852,41 @@ function twig_in_filter($value, $compare)
...
+    switch ($type) {
+        case 'both':
+            return trim($string, $characterMask);
+        case 'left':
+            return ltrim($string, $characterMask);
+        case 'right':
+            return rtrim($string, $characterMask);
+        default:
+            if (!is_string($type)) {
+                throw new Twig_Error_Runtime(sprintf('Trimming type must be a string but got %s.', gettype($type)));
```
fabpot: I would remove this test a we are not doing that anywhere else and instead, change the condition that follows to: `if (!is_array($type) || !in_array($type, array('both', 'left', 'right'))) {`

doc/filters/trim.rst
```diff
@@ -14,13 +14,26 @@ and end of a string:
...
+    {# outputs 'I like Twig.  ' #}
+
+    {{ '  I like Twig.  '|trim(' ', 'right') }}
+
+    {# outputs '  I like Twig.' #}
+
 .. note::
 
-    Internally, Twig uses the PHP `trim`_ function.
+    Internally, Twig uses the PHP `trim`_, `ltrim`_ and `rtrim`_ functions.
```
fabpot: missing comma before `and`.

lib/Twig/Extension/Core.php
```diff
@@ -852,6 +852,41 @@ function twig_in_filter($value, $compare)
     return false;
 }
 
+/** 
+ * Returns a trimmed string.
+ * 
+ * @param string $string
+ * @param string $characterMask
+ * @param string $type
```
fabpot: The three `@param` can be removed

doc/filters/trim.rst
```diff
@@ -14,13 +14,26 @@ and end of a string:
...
 
 Arguments
 ---------
 
 * ``character_mask``: The characters to strip
 
+* ``type``: The default is to strip from the start and the end (`both`), but `left`
+  and `right` will strip from either the left side or right side only.
```
fabpot: Can you remove the dot at the end?

doc/filters/trim.rst
```diff
@@ -14,13 +14,26 @@ and end of a string:
 
     {# outputs '  I like Twig' #}
 
+    {{ '  I like Twig.  '|trim(type='left') }}
```
fabpot: You should add a `versionadded` tag at the beginning of the file (like done in `doc/filters/convert_encoding.rst` for instance):

```
.. versionadded:: 1.32
    The ``type`` argument was added in Twig 1.32.
```

fabpot:
This feature should be added in the 1.x branch instead. You can change the target here and rebase your changes (if you're not sure about this, keep it as is and I will take care of that when merging).

Can you add a note in the CHANGELOG file as well about this new addition?

antonydandrea:
@fabpot struggling with the rebase to be honest :-S

fabpot:
@antonydandrea No worries, I will take care of that.

I'm not a big fan of `type` as the word does not say much about the feature. What about trying to be more precise like `position` or `side`, or ...? Any other ideas?

robfrawley:
I rather like `side`, as it is short (if using named arguments, this can be important) and fairly precise. Good job on your first PR for `twig/twig` @antonydandrea!

antonydandrea:
I think "side" is best. 

When I originally picked "mode" and then "type", I did so because I thought it kept consistent with the terminology in PHP. But I realise now, that Twig is suppose to abstract away from "developer speak".

I think "side" is the most natural way of describing this: for example, 
"I want to trim the spaces." 
"Which side? The left side? The right side? Or both sides?"

Other words I considered I think have other connotations: "position" is more to point to a specific point; "end" (which end do you want to trim), is too ambiguous because it can be used in the context of "start" and "end".

lib/Twig/Extension/Core.php
```diff
@@ -852,6 +852,33 @@ function twig_in_filter($value, $compare)
...
+
+    switch ($side) {
+        case 'both':
+            return trim($string, $characterMask);
+        case 'left':
+            return ltrim($string, $characterMask);
+        case 'right':
+            return rtrim($string, $characterMask);
+        default:
+            if (!is_string($side) || !in_array($side, ['both', 'left', 'right'])) {
```
julienfalque: I think this `if` is unnecessary: if we are in the `default` case, then we already know that `$side` is not a string or has an invalid value.

antonydandrea: Ah true. This evolved from when I was capturing both conditions separately to give two different errors - which is no longer the case.

fabpot:
Thank you @antonydandrea.

PR #2406, opened 2017-02-21
Fix tokenize call in DeprecationCollector

vdechenaux:
Hi,

if fixes `Argument 1 passed to Twig_Environment::tokenize() must be an instance of Twig_Source, string given`


stof:
this should even be fixed in 1.x

fabpot:
Good catch, thanks @vdechenaux.

stof:
@fabpot it may make sense to write tests for this utility class.

fabpot:
see #2407

PR #2407, opened 2017-02-21
added some missing tests

fabpot:


doc/recipes.rst
```diff
@@ -30,7 +30,7 @@ catches deprecation notices, and return them.
 .. tip::
 
     If your templates are not stored on the filesystem, use the ``collect()``
-    method instead which takes an ``Iterator``; the iterator must return
+    method instead which takes a ``Traversable``; the iterator must return
```
SpacePossum: maybe:
`method instead which takes a ``Traversable``; the iterator must return`
to:
`method instead which takes a ``Traversable``; which must return`
as there is indeed no need for an iterator anymore.
with or without the change, :+1: nice changes :)

PR #2411, opened 2017-02-23
Exception for use of "===" instead of "same as"

rechl:
As mentioned in #2358
Hope this is the right location for this Exception


fabpot:
Thank you @sgotre.

Issue #2309, opened 2016-12-22
Silent display of undefined block is deprecated


enumag:
So I got this error:

```
Silent display of undefined block "help" in template "base.html.twig" is deprecated since version 1.29 and will throw an exception in 2.0
```

The thing is that the block simply may or may not exist. How can I check that without raising this error? The [answer](http://stackoverflow.com/questions/13774384/how-to-check-if-a-block-exists-in-a-twig-template-symfony2) on stackoverflow is outdated - it says to simply render the block which would raise the error.

Also I think the solution should be mentioned in the error message.

stof:
``{% if block('help') is defined`` is available for that in recent Twig versions

enumag:
Thank you. :-)

stof:
@fabpot should we mention the ``block() is defined`` feature in the deprecation message of the silent display of a block to ease the migration path ?

fabpot:
see 97a8c53

pierredup:
The `is defined` doesn't seem to always work as expected.
In my one project, I'm getting the following error:

![screen shot 2017-03-08 at 08 08 38](https://cloud.githubusercontent.com/assets/144858/23692337/805e7430-03d6-11e7-9c77-eb587a6de10a.png)

As you can see, the block rendering is inside a `is defined` check, but it still throws an exception (I tried with `{{ block('scripts') is defined ? block('scripts') }}` as well, but getting the same error).

This only seems to happen in certain cases. So far I've noticed it when using the block `javascripts` and `scripts`, I have several other blocks that works fine. Changing those to only `javascript` or `script` works.

fabpot:
@pierredup Can you provide a script that reproduces the error?

pierredup:
@fabpot I figured that it is because a child template was calling `parent()` when the block didn't exist in the parent template. I created a PR to add a better error message in this case #2415

ninze:
How about performance of this?

    {% if block('help') is defined %}
        {% set help = block('help') %}
    {% endif %}


Wouldn't this render the block twice internally - one for the check and another for the assignment?

jvasseur:
@ninze the syntax `bock(...) is defined` is treated specially by Twig and compiled into an `hasBlock` call allowing your example to do only one render of the block.

PR #2415, opened 2017-03-08
Add error message when calling parent() in a block that doesn't exist

pierredup:
Related: #2309

When calling `parent()` in a block that isn't defined in the parent template, you get an error message that states `Block "%s" on template "%s" does not exist`, even if the block is wrapped in an `is defined` check, which is confusing and makes debugging hard.

This fix adds a better error message indicating that you should not call `parent()` when the parent block is not defined.

Before:
screen shot 2017-03-08 at 08 08 38

After:
screen shot 2017-03-08 at 08 56 18

### DIFF
--- a/lib/Twig/Template.php
+++ b/lib/Twig/Template.php
@@ -204,12 +204,10 @@ abstract class Twig_Template
             }
         } elseif (false !== $parent = $this->getParent($context)) {
             $parent->displayBlock($name, $context, array_merge($this->blocks, $blocks), false);
+        } elseif (isset($blocks[$name])) {
+            throw new Twig_Error_Runtime(sprintf('Block "%s" should not call parent() in %s as the block does not exist in the parent template %s.', $name, $blocks[$name][0]->getTemplateName(), $this->getTemplateName()), -1, $blocks[$name][0]->getTemplateName());
         } else {
-            if (isset($blocks[$name])) {
-                throw new Twig_Error_Runtime(sprintf('Block "%s" should not call parent() in %s as the block does not exist in the parent template %s.', $name, $blocks[$name][0]->getTemplateName(), $this->getTemplateName()), -1, $blocks[$name][0]->getTemplateName());
-            } else {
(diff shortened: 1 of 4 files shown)


fabpot:
Can you try to add a test for that?

pierredup:
Test added

lib/Twig/Template.php
```diff
@@ -205,7 +205,11 @@ public function displayBlock($name, array $context, array $blocks = array(), $us
         } elseif (false !== $parent = $this->getParent($context)) {
             $parent->displayBlock($name, $context, array_merge($this->blocks, $blocks), false);
         } else {
-            throw new Twig_Error_Runtime(sprintf('Block "%s" on template "%s" does not exist.', $name, $this->getTemplateName()), -1, $this->getTemplateName());
+            if (isset($blocks[$name])) {
```
xabbuh: this could be an `elseif` branch of the surrounding condition thus avoiding another level of nesting

lib/Twig/Template.php
```diff
@@ -204,6 +204,8 @@ public function displayBlock($name, array $context, array $blocks = array(), $us
             }
         } elseif (false !== $parent = $this->getParent($context)) {
             $parent->displayBlock($name, $context, array_merge($this->blocks, $blocks), false);
+        } elseif (isset($blocks[$name])) {
+            throw new Twig_Error_Runtime(sprintf('Block "%s" should not call parent() in %s as the block does not exist in the parent template %s.', $name, $blocks[$name][0]->getTemplateName(), $this->getTemplateName()), -1, $blocks[$name][0]->getTemplateName());
```
xabbuh: Oh, didn't notice that before, but I would enclose the `%s` placeholders for the template names with double quotes for consistency with other exception messages.

fabpot:
LGTM. Can you also add an integration test (under `test/Twig/Tests/Fixtures`) and a note in the CHANGELOG?

fabpot:
Thank you @pierredup.

Issue #2416, opened 2017-03-08
Warning: setlocale()...


BerkhanBerkdemir:
PHP 7.1.2, Twig 1.32 (ctype, xdebug enable)

=output=
Warning: setlocale() expects parameter 1 to be integer, string given in 
=/output=

I follow the documents and twig throw this error. What should i do?

If you say "yes we need see this code" i will add code.

stof:
Please give the end of the error message. you cut it just before the location of the error

BerkhanBerkdemir:
OK. This error reason php 7.1 , i m turn back php 5.6 and now i have not a problem. But you want my code?

stof:
I don't need your code, I need the end of the error message (the part after ``in``, where it gives the name of the file in which there is the error)

stof:
Btw, giving the error location part of the error message is good for *any* bug report. It generally allows to identify the issue much faster.

BerkhanBerkdemir:
https://ibb.co/gXVxdv

PR #2417, opened 2017-03-08
Compiler - setlocale must be a string

SpacePossum:
Following the docs of [setlocale](http://php.net/manual/en/function.setlocale.php) the second argument must be a `string`, `null` or an `array`. Special note about "0" is there as well.

closes https://github.com/twigphp/Twig/issues/2416


stof:
@fabpot @lyrixx this could be a good addition to the static analysis of SensioLabsInsight

fabpot:
Thank you @SpacePossum.

SpacePossum:
btw. just checking a bit more, HHVM is off on the return values;
https://3v4l.org/fkKJT
:(

PR #2418, opened 2017-03-09
Change Twig_Loader_Chain to accept Traversable

enumag:


lib/Twig/Loader/Chain.php
```diff
@@ -22,8 +22,12 @@
     /**
      * @param Twig_LoaderInterface[] $loaders
      */
-    public function __construct(array $loaders = array())
+    public function __construct($loaders = array())
     {
+        if (!is_array($loaders) && !$loaders instanceof \Traversable) {
```
linaori, line 27: What about `iterable`?

nicolas-grekas: PHP 7.1 only

enumag:
Can this be merged? The tests failure is not related to my change.

nicolas-grekas:
addLoader will fail when a non-array is injected I suppose, shouldn't it throw an exception in this case?

enumag:
Huh? It works fine with non-array iterable... You mean iterable with different items than Twig_LoaderInterface? In my opinion it's not necessary to add such exception. Useless balast code if you ask me.

nicolas-grekas:
`$this->loaders[] = ...` <- this doens't work for all `Traversable`, so there is no gurantee it will work at all with any Traversable, that's what I meant.

enumag:
I have no idea what you mean. `$this->loaders` is always array even with my change.

nicolas-grekas:
Oh, OK, got it sorry, the Traversable is turned to an array in the constructor with the foreach. I thought it would be kept as is for laziness purpose.

stof:
Iterating in the constructor removes the benefit of using a lazy-loading iterator (especially if your first loader is able to load most of your templates.
It would be better to keep lazyness, and turn the iterator into an array only when adding a new loader later (projects wanting to benefit from lazyness would then have to take care to provide all loaders in the constructor, but this is fine IMO)

enumag:
@stof Possibly but that's not really the aim of this PR.

stof:
Well, you are accepting Traversable, but without any benefit for passing a Traversable rather than an array.

enumag:
The ability to pass Traversable is the benefit for me. But alright, I'll see what I can do about the laziness.

fabpot:
@enumag Have you made any progress on the laziness? Are you still interested in finishing this pull request?

enumag:
@fabpot I solved it differently so not interested anymore.

PR #2419, opened 2017-03-10
Allow "length" filter to work on objects with a __toString method that are not \Countable

mpdude:
**The Twig 1.x version of this PR is at #2420, feel free to close either one.**

Use case: When you have variables in your views that are actually objects but implement `__toString`, they feel like strings: For example, `{{ something }}` will make use of that to-string-conversion.

What does *not* work is `{{ something | length }}`, because that will only have a meaningful return value for objects implementing `\Countable`. This interface, however, may have a totally different semantic/purpose for the object in question.

For template designers, this may be surprising if they don't actually care about the object-or-string difference, they just "use" the variable.

This change tries to address this as it changes the behavior for  such objects that have a `__toString` method and are *not* `\Countable`.

*Yes*, it's a BC break in an edge case: Objects that implement a `__toString` but not `\Countable` would previously yield `1` for `{{ object | length }}`, and now would return the length of the string returned by `__toString`.

### DIFF
--- a/lib/Twig/Extension/Core.php
+++ b/lib/Twig/Extension/Core.php
@@ -1139,7 +1139,15 @@ function twig_convert_encoding($string, $to, $from)
  */
 function twig_length_filter(Twig_Environment $env, $thing)
 {
-    return is_scalar($thing) ? mb_strlen($thing, $env->getCharset()) : count($thing);
+    if (is_scalar($thing)) {
+        return mb_strlen($thing, $env->getCharset());
+    }
+    
+    if (is_object($thing) && !($thing instanceof \Countable) && is_callable([$thing, '__toString'])) {
+        return mb_strlen((string) $thing, $env->getCharset());
+    }
+    
+    return count($thing);
 }
 
 /**
--- a/lib/Twig/Extension/Core.php
+++ b/lib/Twig/Extension/Core.php
@@ -1143,7 +1143,7 @@ function twig_length_filter(Twig_Environment $env, $thing)
         return mb_strlen($thing, $env->getCharset());
     }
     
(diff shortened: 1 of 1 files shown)


lib/Twig/Extension/Core.php
```diff
@@ -1139,7 +1139,15 @@ function twig_convert_encoding($string, $to, $from)
  */
 function twig_length_filter(Twig_Environment $env, $thing)
 {
-    return is_scalar($thing) ? mb_strlen($thing, $env->getCharset()) : count($thing);
+    if (is_scalar($thing)) {
+        return mb_strlen($thing, $env->getCharset());
+    }
+    
+    if (is_object($thing) && !($thing instanceof \Countable) && is_callable([$thing, '__toString'])) {
```
stof: short array syntax is a no-go in the Twig codebase. We stay consistent with the 1.x branch as it is maintained.

stof: parenthesis around the instance are useless too

SpacePossum: array syntax is [configured](https://github.com/twigphp/Twig/blob/v2.2.0/.php_cs.dist#L7) but fabbot.io is happy, maybe the service is not using the configuration?

stof: I'm not sure fabbot.io uses the config in the repo. I think it uses its own config.

mpdude:
What is the release policy – would you accept this against 1.x as well?

stof:
@mpdude new features are accepted in 1.x, as long as they don't involve deprecating something. Deprecations are not accepted in 1.x anymore.

mpdude:
@stof So this would work for 1.x?

fabpot:
1.x is fine.

PR #2424, opened 2017-03-11
fix explanation according to the example

lazyants:


fabpot:
The current docs looks good to me. The sentence above says that this is the defaults when args are not provided, so it's not related to the example.

PR #2420, opened 2017-03-10
Make "length" filter and "empty" test consider __toString [Twig 1.x]

mpdude:
Use case: When you have variables in your views that are actually objects but implement `__toString`, they feel like strings: For example, `{{ something }}` will make use of that to-string-conversion.

What does *not* work is 

**a)** `{{ something | length }}`, because that will only have a meaningful return value for objects implementing `\Countable`. This interface, however, may have a totally different semantic/purpose for the object in question.

**b)** `{% if something is empty %}`, because

> empty checks if a variable is an empty string, an empty array, an empty hash, exactly false, or exactly null  `[http://twig.sensiolabs.org/doc/2.x/tests/empty.html]`

... and obviously `something !== null` in this case.

For template designers, this may be surprising if they don't actually care about the object-or-string difference, they just "use" the variable.

This change tries to address this as it changes the behavior for  such objects that have a `__toString` method and are *not* `\Countable`.

*Yes*, it's a BC break in edge cases:

For a), objects that implement a `__toString` but not `\Countable` would previously yield `1` for `{{ object | length }}`, and now would return the length of the string returned by `__toString`.

For b), testing (defined) variables that are objects implementing  `__toString` and that return `''`,  the test now is `false`.

PR #2426, opened 2017-03-14
Consider __toString() implementations in "empty" test [Twig 1.x]

mpdude:
This follows up on #2420:

When you have variables in your views that are actually objects but implement `__toString`, they feel like strings: For example, `{{ something }}` will make use of that to-string-conversion.

What does *not* work is `{% if something is empty %}`, because

> empty checks if a variable is an empty string, an empty array, an empty hash, exactly false, or exactly null  `[http://twig.sensiolabs.org/doc/2.x/tests/empty.html]`

... and obviously `something !== null` in this case.

For template designers, this may be surprising if they don't actually care about the object-or-string difference, they just "use" the variable.

This PR changes the test to consider the return value of the `__toString` method when the value under test is an object, *not* `\Countable` *and* implements `__toString`.

*Yes*, it's a BC break in the edge case of testing (defined) variables that are objects implementing  `__toString` but return the empty string `''`.

### DIFF
--- a/lib/Twig/Extension/Core.php
+++ b/lib/Twig/Extension/Core.php
@@ -1400,8 +1400,16 @@ function twig_ensure_traversable($seq)
  */
 function twig_test_empty($value)
 {
-    if ($value instanceof Countable) {
-        return 0 == count($value);
+    if (is_object($value)) {
+
+        if ($value instanceof Countable) {
+            return 0 == count($value);
+        }
+
+        if (is_callable(array($value, '__toString'))) {
+            return '' === (string) $value;
+        }
+
     }
 
     return '' === $value || false === $value || null === $value || array() === $value;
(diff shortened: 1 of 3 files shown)


lib/Twig/Extension/Core.php
```diff
@@ -1400,8 +1400,14 @@ function twig_ensure_traversable($seq)
  */
 function twig_test_empty($value)
 {
-    if ($value instanceof Countable) {
-        return 0 == count($value);
+    if (is_object($value)) {
```
fabpot: I think the `is_object()` test is not needed.

lib/Twig/Extension/Core.php
```diff
@@ -1400,8 +1400,14 @@ function twig_ensure_traversable($seq)
...
+        if ($value instanceof Countable) {
+            return 0 == count($value);
+        }
+
+        if (is_callable(array($value, '__toString'))) {
```
fabpot: Same as for the other PR, `method_exists` seems more appropriate in this case.

fabpot:
@mpdude Can you merge this with the other PR as this is really just the same topic? Tests could then be consolidated.

mpdude:
Closing this as changes have been merged into #2420

PR #2427, opened 2017-03-17
Fixed JS escaping for unicode characters with higher code points

mikealmond:
Unicode characters with higher code points were being escaped incorrectly. When these characters are escaped, they should maintain their surrogate halves. Previously, Twig was dropping the first surrogate half. 

https://mathiasbynens.be/notes/javascript-escapes#unicode-code-point
> The tetragram for centre symbol (𝌆) has code point U+1D306, so you could write it as \u{1D306}. For comparison, if you were to use simple Unicode escapes to represent this symbol, you’d have to write out the surrogate halves separately: '\uD834\uDF06'.

### DIFF
--- a/lib/Twig/Extension/Core.php
+++ b/lib/Twig/Extension/Core.php
@@ -975,13 +975,13 @@ function twig_escape_filter(Twig_Environment $env, $string, $strategy = 'html', 
 
                 // \uHHHH
                 $char = twig_convert_encoding($char, 'UTF-16BE', 'UTF-8');
+                $char = strtoupper(bin2hex($char));
 
-                return implode('', array_map(
-                    function ($surrogate) {
-                        return '\\u'.strtoupper(substr('0000'.$surrogate, -4));
-                    },
-                    str_split(bin2hex($char), 4)
-                ));
+                if (4 >= strlen($char)) {
+                    return sprintf('\u%04s', $char);
+                }
+
+                return sprintf('\u%04s\u%04s', substr($char, 0, -4), substr($char, -4));
             }, $string);
 
             if ('UTF-8' !== $charset) {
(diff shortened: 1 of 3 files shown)


lib/Twig/Extension/Core.php
```diff
@@ -976,7 +976,12 @@ function twig_escape_filter(Twig_Environment $env, $string, $strategy = 'html',
                 // \uHHHH
                 $char = twig_convert_encoding($char, 'UTF-16BE', 'UTF-8');
 
-                return '\\u'.strtoupper(substr('0000'.bin2hex($char), -4));
+                return implode('', array_map(
```
nicolas-grekas: the fix looks valid to me, I'd just suggest an alternative implementation:
```php
    $char = strtoupper(bin2hex($char));

    if (4 >= strlen($char)) {
        return sprintf('\u%04s', $char);
    }

    return sprintf('\u%04s\u%s', substr($char, 0, -4), substr($char, -4));
```

bburnichon: I do agree with @nicolas-grekas although I think `str_pad` is more efficient than `sprintf` in this. Or use `substr` tricks:

```php
$char = strtoupper(bin2hex($char));

if (4 >= strlen($char)) {
  return '\u' . substr('0000' . $char, -4);
}

return '\u' . substr('0000' . $char, 0, -4) . '\u' . substr($char, -4);
```

mikealmond: Based on http://ww.ephestione.altervista.org/ and local testing on PHP7.1 `sprintf()` was marginally faster. Updated in 198f4d37c5fe04dfde7eb0d4d4c32772616915db

nicolas-grekas:
:+1:

fabpot:
Thank you @mikealmond.

PR #2429, opened 2017-03-20
fixed risky tests

fabpot:


test/Twig/Tests/NodeVisitor/OptimizerTest.php
```diff
@@ -37,6 +37,8 @@ public function testRenderParentBlockOptimizer()
     public function testRenderVariableBlockOptimizer()
     {
         if (PHP_VERSION_ID >= 50400) {
+            $this->markTestSkipped('not needed on PHP >= 5.4');
+
             return;
```
stof: this return is now dead code

test/Twig/Tests/TemplateTest.php
```diff
@@ -262,6 +266,8 @@ public function testGetAttributeStrict($defined, $value, $object, $item, $argume
             } catch (Twig_Error_Runtime $e) {
                 if (null !== $exceptionMessage) {
                     $this->assertSame($exceptionMessage, $e->getMessage());
+                } else {
+                    $this->addToAssertionCount(1);
```
stof: much better would be to change this to use ``expectException`` and ``expectExpectionMessage`` (or ``setExpectedException('Twig_Error_Runtime', $exceptionMessage)`` on older versions) and remove the try/catch entirely

test/Twig/Tests/NodeVisitor/OptimizerTest.php
fabpot: fixed

lib/Twig/Test/IntegrationTestCase.php
```diff
@@ -121,6 +121,14 @@ public function getLegacyTests()
 
     protected function doIntegrationTest($file, $message, $condition, $templates, $exception, $outputs)
     {
+        if (!$outputs) {
+            // add a dummy assertion here to satisfy PHPUnit, it happens when
+            // we return a dummy test from getTests() when there is no legacy tests
+            $this->addToAssertionCount(1);
```
xabbuh: Doesn't `$this->markTestSkipped()` fit better here?

stof: yeah, probably

fabpot: much better indeed, fixed

test/Twig/Tests/TemplateTest.php
fabpot: Not that easy as we have some tests that throw exceptions without a message. Will see how to fix that later on.

fabpot: fixed now

test/Twig/Tests/TemplateTest.php
```diff
@@ -255,15 +259,17 @@ public function testGetAttributeStrict($defined, $value, $object, $item, $argume
...
             $this->assertEquals($value, $template->getAttribute($object, $item, $arguments, $type));
         } else {
-            try {
-                $this->assertEquals($value, $template->getAttribute($object, $item, $arguments, $type));
-
-                throw new Exception('Expected Twig_Error_Runtime exception.');
-            } catch (Twig_Error_Runtime $e) {
-                if (null !== $exceptionMessage) {
-                    $this->assertSame($exceptionMessage, $e->getMessage());
+            if (!$defined) {
```
stof: this ``if`` will always be true, as it is inside the else clause of the inverse condition

fabpot: indeed

PR #2432, opened 2017-03-24
Update composer.json

taras-m0:
fix bug


stof:
This is totally wrong. It breaks the autoloading

sstok:
This is only possible with PSR-4 (which no longer uses `_` for namespace separator), but Twig doesn't use namespaces (yet). And therefor this is not possible.

PR #2433, opened 2017-03-24
Fix #2431

### LINKED ISSUE #2431: 1.33.0 and "empty" test will try to load random strings as classes
Hi,

After upgrading to 1.33.0 I started getting errors from class loader and found that release notes had:
`* "empty" test will now consider the return value of the __toString() method for`

Which is in code: twig/twig/lib/Twig/Extension/Core.php:
```
function twig_test_empty($value)
{
    if ($value instanceof Countable) {
        return 0 == count($value);
    }
    if (method_exists($value, '__toString')) {
        return '' === (string) $value;
    }
    return '' === $value || false === $value || null === $value || array() === $value;
}
```

Which is the reason, method_exists without any previous checks.. 
http://www.php.net/manual/en/function.method-exists.php
`Note:
Using this function will use any registered autoloaders if the class is not already known.`

And that's not nice! In many other cases in the same file there is is_object check first, this should also have it! So next version should have:
```
    if (is_object($value) && method_exists($value, '__toString')) {
        return '' === (string) $value;
    }
```

mpdude:
See title

### DIFF
--- a/lib/Twig/Extension/Core.php
+++ b/lib/Twig/Extension/Core.php
@@ -1425,7 +1425,7 @@ function twig_test_empty($value)
         return 0 == count($value);
     }
 
-    if (method_exists($value, '__toString')) {
+    if (is_object($value) && method_exists($value, '__toString')) {
         return '' === (string) $value;
     }
 


rellect:
What if $value is an object but does not implements toString()?

mpdude:
The `method_exists` call is supposed to check exactly that, so I probably don't understand the question...?

rellect:
I'm thinking of an edge case where $value is an object, which is not Countable and does not implements __toString(). In such case it will continue to
`return '' === $value || false === $value || null === $value || array() === $value;`
Which always be false (value is not empty)..
Not sure what would be expected in such case.

mpdude:
... which is exactly how `empty` behaved before #2420. `{% if someobj is empty ... %}` does not hold if `someobj` is a plain object.

SpacePossum:
so old + 2420 + this PR should be something like:

```php
function twig_test_empty($value)
{
    if (is_object($value)) {    
        if ($value instanceof Countable) {
            return 0 == count($value);
        }

        if method_exists($value, '__toString')) {
            return '' === (string) $value;
        }

        return false;
    }

    return '' === $value || false === $value || null === $value || array() === $value;
}
```

?

mpdude:
@SpacePossum I cannot tell you why it was designed this way. But probably this PR is not the best place to discuss this. I'd suggest you open a new issue if you think that should be improved.

fabpot:
Thank you @mpdude.

PR #2434, opened 2017-03-27
Autoescape test addition; collision fix in Array Loader.

dave-newson:
This PR makes two changes. It was created while testing a third issue that was discovered to be fixed in the latest version.

## Added JS to Autoescape: name test

This adds "index.js.html" to the autoescape:name test, proving the functionality of the JS escape system and the name guessing strategy.

## Fixed collision in Array Loader

While trying to run the above test, it was discovered that the Array Loader uses the content of the template file as the "key" for determining which compiled template to load.
As the above test uses identical content for both HTML and JS templates this led to the compiled HTML template always being loaded.

This change prepends the content with the filename, which vaguely follows the same mechanism as the "options" (which are appended after `getCacheKey` is called).

**This change might be a backwards-incompatible change** depending on your interpretation of how `Loader/Array` is intended to be used.
The class itself states:
> This loader should only be used for unit testing.

The assumption is it's OK for this to occur.

Test cases have been updated to take this key change into account, and an additional test has been added to confirm `Loader/Array` specifically has the capability to resolve two files with the same content to different keys.

### DIFF
--- a/lib/Twig/Loader/Array.php
+++ b/lib/Twig/Loader/Array.php
@@ -80,7 +80,7 @@ class Twig_Loader_Array implements Twig_LoaderInterface, Twig_ExistsLoaderInterf
             throw new Twig_Error_Loader(sprintf('Template "%s" is not defined.', $name));
         }
 
-        return $this->templates[$name];
+        return $name . $this->templates[$name];
     }
 
     public function isFresh($name, $time)
--- a/lib/Twig/Loader/Array.php
+++ b/lib/Twig/Loader/Array.php
@@ -80,7 +80,7 @@ class Twig_Loader_Array implements Twig_LoaderInterface, Twig_ExistsLoaderInterf
             throw new Twig_Error_Loader(sprintf('Template "%s" is not defined.', $name));
         }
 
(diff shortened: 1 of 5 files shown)


lib/Twig/Loader/Array.php
```diff
@@ -80,7 +80,7 @@ public function getCacheKey($name)
             throw new Twig_Error_Loader(sprintf('Template "%s" is not defined.', $name));
         }
 
-        return $this->templates[$name];
+        return $name.$this->templates[$name];
```
stof: you should add a delimiter between them, to avoid collision based on the beginning of the source and the end of the name

dave-newson: @stof Dangit, I thought about that, but `\Twig_Environment::getTemplateClass` doesn't make this distinction when it appends the optionsHash (https://github.com/twigphp/Twig/blob/2.x/lib/Twig/Environment.php#L269), so I skipped it.

It's definitely possible to get a collision in the method I provided though; I'll add something in.

fabpot:
Thank you @dave-newson.

PR #2436, opened 2017-03-31
Prepare for php 7.2

hboomsma:
For php >= 7.2 (nightly)
* allow `|length` of `null`
* allow `|length` of objects not implementing `\Countable`

see https://wiki.php.net/rfc/counting_non_countables for more information.

Ran all unit tests with php 7.2 nightly


Tobion:
> allow |length of objects not implementing \Countable

Why should twig still support that and hide the php warning? Counting non-countable things in twig does not make more sense than in php.

hboomsma:
The change in behaviour in Twig when running on php >=7.2 will break projects (found out with composer/satis). So it would be nice to at least maintain BC until 3.0.

That said, there might still be good reasons why to stick with this behaviour. Having `|length` on a `null` value return `0` and for a single object `1` is not very strange but helpful in some cases. In php this was abandoned because it can hide bugs. In a template one could argue this would happen less often and the convenience to not write checks to see if something is countable, could be worth keeping the current behaviour for the future.

stof:
I would deprecate this behavior though

hboomsma:
@stof, depricate and remove in 3.0 seems a good path to me. Do you want me to implement it? What about version 1.x?

stof:
no deprecations in 1.x (we cannot add new ones)

hboomsma:
@stof are the messages sufficient? Should I create a pull with the BC part for version 1?

fabpot:
@hboomsma Can you create a PR on the 1.x branch without the deprecations? Thanks.

hboomsma:
@fabpot, Sure I will look into it start of next week.

hboomsma:
@fabpot #2462

fabpot:
I've merged the other PR on 1.x and then merged 1.x into 2.x, without the deprecation notices. Not sure they make sense.

PR #2438, opened 2017-04-06
Fix twig_test_empty and twig_length_filter to not trigger autoloading

enumag:
Introduced in #2420, partially fixed in #2433.

cc @mpdude

Also because of this I found another bug in both Symfony and Composer - autoloading fails for an empty string. I'll send pull requests to both shortly.

### DIFF
--- a/lib/Twig/Extension/Core.php
+++ b/lib/Twig/Extension/Core.php
@@ -1148,7 +1148,7 @@ function twig_length_filter(Twig_Environment $env, $thing)
         return mb_strlen($thing, $env->getCharset());
     }
 
-    if (method_exists($thing, '__toString') && !$thing instanceof \Countable) {
+    if (is_object($thing) && method_exists($thing, '__toString') && !$thing instanceof \Countable) {
         return mb_strlen((string) $thing, $env->getCharset());
     }
 
@@ -1241,7 +1241,7 @@ function twig_test_empty($value)
         return 0 == count($value);
     }
 
-    if (method_exists($value, '__toString')) {
+    if (is_object($value) && method_exists($value, '__toString')) {
         return '' === (string) $value;
     }
 


lib/Twig/Extension/Core.php
```diff
@@ -1241,7 +1241,7 @@ function twig_test_empty($value)
         return 0 == count($value);
     }
 
-    if (method_exists($value, '__toString')) {
+    if (is_object($value) && method_exists($value, '__toString')) {
```
SpacePossum: can be optimized, as the `is_object` test is now always done there is not point for the test at line 1280 (`return '' === $value || false === $value || null === $value || array() === $value;`) if `is_object` returned `true`

enumag: You mean like this?

```php
if (is_object($value)) {
    if (method_exists($value, '__toString')) {
        return '' === (string) $value;
    } else {
        return false;
    }
}

return '' === $value || false === $value || null === $value || array() === $value;
```

Should I use ternary operator?


```php
if (is_object($value)) {
    return method_exists($value, '__toString') ? '' === (string) $value : false;
}

return '' === $value || false === $value || null === $value || array() === $value;
```

mpdude:
What do you mean with "partially fixed"? Isn't this change exactly the same as in #2433?

Note that #2433 was against 1.x which hasn't yet been merged into 2.x.

enumag:
@mpdude Right... twig_length_filter doesn't need to be changed because of is_scalar check. So the only thing left is the optimization that @SpacePossum suggested.

SpacePossum:
the optimize might not be worth a PR I guess

enumag:
Alright. In that case I'll just wait for the 1.x branch to be merged into 2.x. Thank you both.

Issue #1689, opened 2015-05-12
Double escaping when using custom esaper


Koc:
I need create custom escaper which would replace `'` with `\'` and call `html_attr` escaping for processed string.

``` php
<?php

namespace Metal\ProjectBundle\Twig;

class AngularEscaping extends \Twig_Extension
{
    public function initRuntime(\Twig_Environment $environment)
    {
        $environment->getExtension('core')->setEscaper('angular', array($this, 'angularEscaping'));
    }

    public function angularEscaping(\Twig_Environment $environment, $string, $charset)
    {
        $string = str_replace("'", "\\'", $string);

        return twig_escape_filter($environment, $string, 'html_attr', $charset);
    }

    public function getName()
    {
        return 'angular_escaping';
    }
}
```

``` php
`$a = "'4"`
```

``` twig
{{ a|replace({"'": "\\'"})|escape('html_attr') }}
{# &#x5C;&#x27;4 - as expected #}

{{ a|escape('angular') }}
{# &amp;#x5C;&amp;#x27;4 - automatic html escaping goes after my custom escaper #}
```

Koc:
As workaround I can replace escaping callable with
```php
    public function angularEscaping(\Twig_Environment $environment, $string, $charset)
    {
        $string = str_replace("'", "\\'", $string);

        return new \Twig_Markup(twig_escape_filter($environment, $string, 'html_attr', $charset), $charset);
    }
```
buy it looks buggy. As I see we have similar problem some time ago https://github.com/twigphp/Twig/pull/1211

stof:
@Koc this is because the code escaped as ``angular`` is not considered as safe for the ``html`` context, while the ``html_attr`` escaping is (as we know it is a stricter escaping than ``html``).

A simple solution for your case would be to use a dedicated filter instead:

```php
<?php

namespace Metal\ProjectBundle\Twig;

class AngularEscaping extends \Twig_Extension
{
    public function getFilters()
    {
        return array(
            new \Twig_SimpleFilter('escape_angular', array($this, 'angularEscaping'), array('is_safe' => array('html_attr'), 'needs_environment' => true)),
        );
    }

    public function angularEscaping(\Twig_Environment $environment, $string)
    {
        $string = str_replace("'", "\\'", $string);

        return twig_escape_filter($environment, $string, 'html_attr');
    }

    public function getName()
    {
        return 'angular_escaping';
    }
}
```

```jinja
{{ a|escape_angular }}
[...]

PR #2441, opened 2017-04-07
Add ability to set safety for custom escapers

0b10011:
Currently, there's no way to set "is_safe" for custom escapers, so it's necessary to use a filter instead (see #1689).

This adds the ability to specify "is_safe" for custom escapers:

```php
$core->setEscaper('url_query', 'esc_url_query');
$core->setEscaper('url_query_key', 'esc_url_query_key', ['url_query']);
$core->setEscaper('url_query_value', 'esc_url_query_value', ['url_query']);
```

I've also added a way to specify that an already specified escaper is safe for the new escaper (useful for adding missing escapers to existing groups of escapers without having to recreate those escapers):

```php
$core->setEscaper('url_query_key', 'esc_url_query_key');
$core->setEscaper('url_query_value', 'esc_url_query_value');

/* Later on, perhaps in a separate extension */
$core->setEscaper('url_query', 'esc_url_query', [], ['url_query_key', 'url_query_value']);
```

---

For the most part, the changes are rather straightforward:

### `SafeAnalysis.php`

- Pass `$env` to filter's and function's `getSafe()`

### `Filter.php` and `Function.php`

- Receive `$env` and pass to `is_safe_callback` as second (currently unused) argument

### `Core.php`

The diff is messy from static to dynamic change; mostly just moved static checks down to their own functions.

- Add default escapers to `Twig_Extension_Core::$escapers`
- Add `Twig_Extension_Core::$escapers_safe` to track escaper safety
- Add `$is_safe_for` and `$is_safe` parameters to `setEscaper()`

- Remove `switch` check for default escapers (moved to own functions at bottom to be handled dynamically)
- Remove `static $escapers` and call `$core->getEscapers()` each time (otherwise different instantiations can affect eachother)
- Remove `array_merge($escapers, ['html', ...])`
- Remove hardcoded check for `html_attr` to add `html` as safe (handled by `$escapers_safe`)

### `escape.rst`
- Updated with an example marking escapers as safe

---

### DIFF
--- a/lib/Twig/Extension/Core.php
+++ b/lib/Twig/Extension/Core.php
@@ -21,13 +21,6 @@ final class Twig_Extension_Core extends Twig_Extension
         'html_attr' => 'twig_escape_html_attr',
         'url' => 'twig_escape_url',
     );
-    private $escapers_safe = array(
-        'html' => array('html'),
-        'js' => array('js'),
-        'css' => array('css'),
-        'html_attr' => array('html', 'html_attr'),
-        'url' => array('url'),
-    );
 
     /**
      * Defines a new escaper to be used via the escape filter.
--- a/lib/Twig/Extension/Core.php
+++ b/lib/Twig/Extension/Core.php
@@ -28,9 +28,9 @@ final class Twig_Extension_Core extends Twig_Extension
      * @param string   $strategy The strategy name that should be used as a strategy in the escape call
      * @param callable $callable A valid PHP callable
      */
-    public function setEscaper($strategy, callable $callable, $ignore_deprecation = false)
+    public function setEscaper($strategy, callable $callable, $ignoreDeprecation = false)
     {
-        if (!$ignore_deprecation) {
+        if (!$ignoreDeprecation) {
             @trigger_error("Twig_Extension_Core::setEscaper() is deprecated. Used Twig_Extension_Escaper::setEscaper() instead", E_USER_DEPRECATED);
         }
 
@@ -42,9 +42,9 @@ final class Twig_Extension_Core extends Twig_Extension
      *
      * @return callable[] An array of escapers
      */
-    public function getEscapers($ignore_deprecation = false)
+    public function getEscapers($ignoreDeprecation = false)
     {
-        if (!$ignore_deprecation) {
+        if (!$ignoreDeprecation) {
             @trigger_error("Twig_Extension_Core::getEscapers() is deprecated. Used Twig_Extension_Escaper::getEscapers() instead", E_USER_DEPRECATED);
         }
 
--- a/lib/Twig/Extension/Core.php
+++ b/lib/Twig/Extension/Core.php
@@ -34,24 +34,14 @@ final class Twig_Extension_Core extends Twig_Extension
      *
      * @param string   $strategy The strategy name that should be used as a strategy in the escape call
      * @param callable $callable A valid PHP callable
-     * @param array $is_safe_for Strategies this strategy should be marked safe for. For example, 'html_attr' lists 'html' as being safe since 'html_attr' escapes everything 'html' does, plus more.
-     * @param array $is_safe Strategies that should be marked safe for this strategy. Useful for adding strategies compatible to existing strategies. For example, an extension that adds a competing html escaper that escapes fewer characters could list 'html' and 'html_attr' as being safe.
      */
-    public function setEscaper($strategy, callable $callable, array $is_safe_for = array(), array $is_safe = array())
+    public function setEscaper($strategy, callable $callable, $ignore_deprecation = false)
     {
+        if (!$ignore_deprecation) {
+            @trigger_error("Twig_Extension_Core::setEscaper() is deprecated. Used Twig_Extension_Escaper::setEscaper() instead", E_USER_DEPRECATED);
+        }
+
         $this->escapers[$strategy] = $callable;
-
-        if (!empty($is_safe_for)) {
-            $is_safe_for = array();
-        }
-        foreach ($is_safe_for as $safe_strategy) {
-            $this->escapers_safe[$safe_strategy][] = $strategy;
-        }
-
-        if (!empty($is_safe)) {
-            $is_safe = array();
-        }
-        $this->escapers_safe[$strategy] = array_merge(array($strategy), $is_safe);
     }
 
     /**
@@ -59,19 +49,13 @@ final class Twig_Extension_Core extends Twig_Extension
      *
      * @return callable[] An array of escapers
      */
-    public function getEscapers()
+    public function getEscapers($ignore_deprecation = false)
     {
-        return $this->escapers;
-    }
+        if (!$ignore_deprecation) {
+            @trigger_error("Twig_Extension_Core::getEscapers() is deprecated. Used Twig_Extension_Escaper::getEscapers() instead", E_USER_DEPRECATED);
+        }
 
-    /**
-     * Gets safe escapers for all escapers.
-     *
-     * @return callable[] An array of escapers safe for each escaper
-     */
-    public function getEscapersSafe()
-    {
-        return $this->escapers_safe;
+        return $this->escapers;
     }
 
     /**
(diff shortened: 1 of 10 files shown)


0b10011:
Note: I haven't added or changed any tests to test the additional functionality of `setEscaper`, but the existing tests cover `$escapers` and `$escapers_safe` (due to the previously hard-coded "html_attr" check). I'll try to get to this in the next couple of days, but if it's an easy thing for you to add real quick, feel free!

Also, it looks like this sets escaping up to moved to an external extension fairly easily. Or, at the very least, moved from `Twig_Extension_Core` to `Twig_Extension_Escaper`. If you'd like, I can look into making that change in a followup commit.

0b10011:
I've added a second commit that deprecates `Twig_Extension_Core::setEscaper()` and `Twig_Extension_Core::getEscapers()`. In order to deprecate the methods, but keep them around (and working), the `Twig_Environment` has to be passed to `Twig_Extension_Escaper::__construct()`. Once the deprecation ends (I'm not sure of how long methods are kept around), the little bit of code still in the core extension can be properly moved over to the escaper extension.

0b10011:
I've pushed some new tests (and fixes).

It looks like the caching wasn't taking into account details specific to the escaper (like new escapers), so I've added a method for extensions to specify a signature. If they don't provide one, it will fall back to just relying on the class name (which is what it was doing already).

The environment signature was being updated every time an extension was added (along with several other actions). I've added a flag to track changes so the signature only has to be generated once.

With the new functionality, escapers can now be overwritten.

Also cleaned up my code a bit in this latest commit.

stof:
@0b10011 deprecation period ends with the next major release.

lib/Twig/Extension/Escaper.php
```diff
@@ -11,16 +11,95 @@
...
      * @param string|false|callable $defaultStrategy An escaping strategy
+     * @param Twig_Environment      $env             Environment this extension is added to. Used temporarily to deprecate Twig_Extension_Core::setEscaper()/getEscapers() instead of removing them without notice.
      *
      * @see setDefaultStrategy()
      */
-    public function __construct($defaultStrategy = 'html')
+    public function __construct($defaultStrategy = 'html', Twig_Environment $env = null)
     {
         $this->setDefaultStrategy($defaultStrategy);
+        $this->environment = $env;
```
stof: can we avoid storing a reference to the environment here ? This creates a circular object graph, which makes it more painful to handle

lib/Twig/Extension/Escaper.php
```diff
@@ -11,16 +11,95 @@
 
 final class Twig_Extension_Escaper extends Twig_Extension
 {
+    private $environment;
+    private $core;
     private $defaultStrategy;
+    private $escapersSafe = array(
```
stof: ``escaperSafety`` looks better

lib/Twig/Extension/Escaper.php
```diff
@@ -74,6 +155,39 @@ public function getDefaultStrategy($name)
...
+            $this->escapersSafe,
+            $this->getEscapers(),
+        )));
+    }
+
+    private function getStrategySignature($strategy)
+    {
+        if (false === $strategy) {
+            return 'none';
+        } elseif (is_string($strategy)) {
```
stof: just ``if`` as the previous if returns

lib/Twig/Extension/Escaper.php
```diff
@@ -87,3 +201,277 @@ function twig_raw_filter($string)
...
+/**
+ * @internal
+ */
+function twig_escape_html(Twig_Environment $env, $string, $charset)
+{
+    // see http://php.net/htmlspecialchars
+
+    // Using a static variable to avoid initializing the array
+    // each time the function is called. Moving the declaration on the
+    // top of the function slow downs other escaping strategies.
```
stof: this comment is irrelevant as it is a separate function now, and this *is* at the top

lib/Twig/Extension/Escaper.php
```diff
@@ -87,3 +201,277 @@ function twig_raw_filter($string)
...
+
+/**
+ * @internal
+ */
+function twig_escape_filter_is_safe(Twig_Node $filterArgs, Twig_Environment $env = null)
+{
+    foreach ($filterArgs as $arg) {
+        if ($arg instanceof Twig_Node_Expression_Constant) {
+            if ($env) {
+                return $env->getExtension('Twig_Extension_Escaper')->getEscapersSafe()[$arg->getAttribute('value')];
```
stof: what if it is not defined ? This will trigger a warning, which is a no-go.

lib/Twig/Extension/Escaper.php
```diff
@@ -11,16 +11,95 @@
...
+            );
+        }
+
+        return $this->core->getEscapers(true);
+    }
+
+    /**
+     * Gets safe escapers for all escapers.
+     *
+     * @return callable[] An array of escapers safe for each escaper
```
stof: wrong phpdoc

lib/Twig/Extension/Core.php
```diff
@@ -32,8 +42,12 @@ public function setEscaper($strategy, callable $callable)
      *
      * @return callable[] An array of escapers
      */
-    public function getEscapers()
+    public function getEscapers($ignore_deprecation = false)
     {
+        if (!$ignore_deprecation) {
```
stof: variable name should be camelCased

stof (changes requested):

lib/Twig/Extension/Escaper.php
0b10011: I've moved the signature code to methods on Twig_Extension. It takes a bit more work to check if the signature needs update (loops through extensions and checks each individually), but it gets rid of the circular reference for the signature. (Committed separately from the other changes in case you'd rather keep the circular reference.)

There's still a dependency on Environment in Escaper, but it's to pull in Core for `setEscaper()` and `getEscaper()`. As far as I can tell, there's no way around this, except to wait to push this through until 3.x (which would allow adding the new methods on Twig_Extension to Twig_ExtensionInterface).

0b10011:
Updated per review. You may want to look at the comment regarding the circular reference.

Koc:
@0b10011 cool, but IMHO it is better split this PR in two:
1. changes in option hash calculations
2. changes in custom escapers

0b10011:
> @0b10011 cool, but IMHO it is better split this PR in two:
> 
> 1. changes in option hash calculations
> 2. changes in custom escapers

This PR won't pass tests without the hash calc changes. What would be the best way to handle this?

1. `hash` branch in `PR#1` (pass) and `escaper` branch in `PR#2` (fail); Could be rebased after `PR#1` is merged.
2. `hash` branch in `PR#1` (pass) and `escaper` branch merged into `hash` branch in `PR#2` (pass)
3. `hash` changes pulled into separate commit in the `escaper` branch in `PR#1` (pass)
4. Something else?

Koc:
maybe 1 or 2

lib/Twig/Extension/Escaper.php
```diff
@@ -87,3 +206,281 @@ function twig_raw_filter($string)
...
+        } elseif (in_array($strategy, array('html', 'js', 'css', 'html_attr', 'url'))) {
+            return $string;
+        }
+    }
+
+    if (null === $charset) {
+        $charset = $env->getCharset();
+    }
+
+    $escapers = $env->getExtension('Twig_Extension_Escaper')->getEscapers();
```
fabpot: Current code is optimized for speed. It looks like the new one is not. That's a very sensitive part of the code as it might be called thousand of times for a page, and small changes in the past had great performance impact.

0b10011: I only noticed a small change in my own tests, but after setting up a proper benchmark specific to escaping, it looks like a ~50% slowdown. I've made some progress on speeding this up already, and will continue to investigate. Will update with more info/fix in the next few days.

fabpot:
@0b10011 Did you made any progress on this one. Are you still willing to finish it?

fabpot:
Closing as there is no more activity on this pull request. The performance impact is too important for it to be merged as is.

PR #2448, opened 2017-04-16
Wrong stdClass class name generates warning

chrisCheeseBurgames:


fabpot:
Thank you @chrisCheeseBurgames.

PR #2452, opened 2017-04-22
Support PHPUnit 6 for testing extensions

hkdobrev:
When testing a Twig extension via the provided test cases [according to the documentation](https://twig.sensiolabs.org/doc/2.x/advanced.html#testing-an-extension), PHPUnit 6 is not supported.

I think we should force people who haven't upgraded to upgrade to PHPUnit 6 instead of limiting the ones who did.

Also, Twig 2.x runs only on PHP 7. People using Twig 2.x would most likely also test their code on PHP 7. Even though PHPUnit 5 supports PHP 7, I think most people who have updated PHP could update or would have already updated PHPUnit.

### DIFF
--- a/lib/Twig/Test/IntegrationTestCase.php
+++ b/lib/Twig/Test/IntegrationTestCase.php
@@ -9,13 +9,16 @@
  * file that was distributed with this source code.
  */
 
+use PHPUnit\Framework\TestCase;
+use PHPUnit\Framework\Constraint\Exception as ExceptionConstraint;
+
 /**
  * Integration test helper.
  *
  * @author Fabien Potencier <fabien@symfony.com>
  * @author Karma Dordrak <drak@zikula.org>
  */
-abstract class Twig_Test_IntegrationTestCase extends PHPUnit_Framework_TestCase
+abstract class Twig_Test_IntegrationTestCase extends TestCase
 {
     /**
      * @return string
@@ -196,7 +199,8 @@ abstract class Twig_Test_IntegrationTestCase extends PHPUnit_Framework_TestCase
 
             if (false !== $exception) {
                 list($class) = explode(':', $exception);
-                $this->assertThat(null, new PHPUnit_Framework_Constraint_Exception($class));
+                $constraintClass = class_exists(ExceptionConstraint::class) ? ExceptionConstraint::class : \PHPUnit_Framework_Constraint_Exception::class;
+                $this->assertThat(null, new $constraintClass($class));
             }
 
             $expected = trim($match[3], "\n ");
--- a/lib/Twig/Test/NodeTestCase.php
+++ b/lib/Twig/Test/NodeTestCase.php
@@ -1,5 +1,7 @@
 <?php
 
+use PHPUnit\Framework\TestCase;
+
 /*
  * This file is part of Twig.
  *
@@ -8,7 +10,7 @@
  * For the full copyright and license information, please view the LICENSE
  * file that was distributed with this source code.
  */
-abstract class Twig_Test_NodeTestCase extends PHPUnit_Framework_TestCase
+abstract class Twig_Test_NodeTestCase extends TestCase
 {
     abstract public function getTests();
 


hkdobrev:
@fabpot Any thoughts on that? Do you think this is the correct way to go? We're just upgrading from Twig 1.x to Twig 2.x together with upgrading to PHPUnit 6 and this would have helped with a seamless transition for our tests on Twig extensions. For now, we've fallen back to a couple of class aliases in our test bootstrap.

Cheers! 🍻

lib/Twig/Test/IntegrationTestCase.php
```diff
@@ -193,7 +193,7 @@ protected function doIntegrationTest($file, $message, $condition, $templates, $e
 
             if (false !== $exception) {
                 list($class) = explode(':', $exception);
-                $this->assertThat(null, new PHPUnit_Framework_Constraint_Exception($class));
+                $this->assertThat(null, new \PHPUnit\Framework\Constraint\Exception($class));
```
stof: this one should check whether the class exists, and fallback to the old one. The namespaced TestCase is available in PHPUnit 4.8.35 and 5.4+, but the exception is not

lib/Twig/Test/IntegrationTestCase.php
```diff
@@ -15,7 +15,7 @@
  * @author Fabien Potencier <fabien@symfony.com>
  * @author Karma Dordrak <drak@zikula.org>
  */
-abstract class Twig_Test_IntegrationTestCase extends PHPUnit_Framework_TestCase
+abstract class Twig_Test_IntegrationTestCase extends \PHPUnit\Framework\TestCase
```
stof: please add a use statement for this class instead of using the FQCN inline

lib/Twig/Test/IntegrationTestCase.php
hkdobrev: Thanks! I didn't know the TestCase was available from before! Will do that later today.

lib/Twig/Test/IntegrationTestCase.php
hkdobrev: Done.

hkdobrev:
@stof Thanks for the feedback! This way this change is not breaking anything at all! Let me know if you want me to squash commits!

stof:
@hkdobrev you can squash them if you want, but this is not mandatory. @fabpot has a tool to sqaush things automatically before merging (the drawback is that the PR is considered closed rather than merged, as github cannot detect it as a merge)

hkdobrev:
Hmm, GitHub has now squashing functionality which if used make the PR be considered as merged :)

stof:
@hkdobrev we don't use the github interface to make merges.

emodric:
Any reason why this doesn't land in 1.x? Especially since namespaced variant of `TestCase` is available in both PHPUnit 4.8 and 5.4+

stof:
@emodric min supported version in Twig 1.x was PHP 5.2, so using the namespaced classes would break it for testcases.
But I would actually be in favor of it. This is up to @fabpot

hkdobrev:
Anything else needed here to merge? Cheers!

lib/Twig/Test/IntegrationTestCase.php
```diff
@@ -193,7 +196,10 @@ protected function doIntegrationTest($file, $message, $condition, $templates, $e
 
             if (false !== $exception) {
                 list($class) = explode(':', $exception);
-                $this->assertThat(null, new PHPUnit_Framework_Constraint_Exception($class));
+                $constraintClass = class_exists(ExceptionConstraint::class)
+                    ? ExceptionConstraint::class
+                    : \PHPUnit_Framework_Constraint_Exception::class;
```
fabpot: This should be on one line.

fabpot:
I think doing that on 1.x is fine. We just need to warn people and bump version to the next min one.

lib/Twig/Test/IntegrationTestCase.php
hkdobrev: Updated.

hkdobrev:
Rebased on top of `1.x` and changed base branch of the PR.

fabpot:
Thank you @hkdobrev.

Issue #2454, opened 2017-04-24
Optimising of escape output for concat operation


frol-kr:
When you try to output a literal `{{ 'constant one' }}` without twig compiler will produce:

```php
protected function doDisplay(array $context, array $blocks = array())
{
    // line 1
    echo "constant one ";
}
```
However if you try to concat two literals, it will produce the following:
```twig
{{ 'constant one ' ~ 'constant two' }}
```
```php
protected function doDisplay(array $context, array $blocks = array())
{
    // line 1
    echo twig_escape_filter($this->env, ("constant one " . "constant two"), "html", null, true);
}
```

PR #2455, opened 2017-04-24
Escapting of concatenation optimized

frol-kr:
This request provides an optimization for an issue [2454](https://github.com/twigphp/Twig/issues/2454)
With provided optimisation the template `{{ 'constant one ' ~ 'constant two' }}` will produce the following compiled template:
```php
{
     // line 1
     echo ("constant one " . "constant two");
}
```
The changes for this optimisation include simply checking the node `Twig_Node_Expression_Binary_Concat` during the save analysis phase and checks whether the both operands are save if they are we mark a `Twig_Node_Expression_Binary_Concat` node as safe as well.

### DIFF
--- a/lib/Twig/NodeVisitor/Optimizer.php
+++ b/lib/Twig/NodeVisitor/Optimizer.php
@@ -25,6 +25,8 @@ final class Twig_NodeVisitor_Optimizer extends Twig_BaseNodeVisitor
     const OPTIMIZE_NONE = 0;
     const OPTIMIZE_FOR = 2;
     const OPTIMIZE_RAW_FILTER = 4;
+    const OPTIMIZE_CONCAT_CONST = 16;
+
     // obsolete, does not do anything
     const OPTIMIZE_VAR_ACCESS = 8;
 
@@ -37,7 +39,7 @@ final class Twig_NodeVisitor_Optimizer extends Twig_BaseNodeVisitor
      */
     public function __construct($optimizers = -1)
     {
-        if (!is_int($optimizers) || $optimizers > (self::OPTIMIZE_FOR | self::OPTIMIZE_RAW_FILTER | self::OPTIMIZE_VAR_ACCESS)) {
+        if (!is_int($optimizers) || $optimizers > (self::OPTIMIZE_FOR | self::OPTIMIZE_RAW_FILTER | self::OPTIMIZE_VAR_ACCESS | self::OPTIMIZE_CONCAT_CONST)) {
             throw new InvalidArgumentException(sprintf('Optimizer mode "%s" is not valid.', $optimizers));
         }
 
@@ -48,6 +50,10 @@ final class Twig_NodeVisitor_Optimizer extends Twig_BaseNodeVisitor
     {
         if (self::OPTIMIZE_FOR === (self::OPTIMIZE_FOR & $this->optimizers)) {
             $this->enterOptimizeFor($node, $env);
+        }
+
+        if (self::OPTIMIZE_CONCAT_CONST === (self::OPTIMIZE_CONCAT_CONST & $this->optimizers)) {
+            $node = $this->optimizeConcatOfTwoConsts($node, $env);
         }
 
         return $node;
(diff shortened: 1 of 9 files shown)


stof:
a better option would be to optimize the concatenation itself at compile time, and running this optimization earlier than the autoescaping (so that we end up in the existing case)

frol-kr:
@stof, thanks for the fast reply
I am new to twig internals so it would very helpful if you provide me with some idea about optimisation at compile time or provide a case where we end up with the existing case with current solution.

frol-kr:
@stof you mean convert three nodes (Concat, Constant, Constant) convert into one Constant node. Is that what you meant?

stof:
@frol-kr yes, having an optimizer turning a Concat node containing 2 Constant nodes into a single Constant node with the concatenated string

stof:
this would avoid having to do special things for the safe analysis (it would not really optimize the concatenation itself, as OPCache is probably optimizing it anyway)

frol-kr:
@stof 
> OPCache is probably optimizing it anyway
You are right. I just was thinking  towards the second phase where you have a constant node + a variable node, where variable node is escaped and there is no point in doing double escaping of the whole expression and a variable node inside of the expression

stof:
@frol-kr but the variable not is *not* escaped. The auto-escaping is applied at the display level.

And concatenation of 2 string can become unsafe while the original strings were safe. the concatenation could create HTML entities, even if the right side of the concatenation contains *escaped* user input. So `{{ 'something' ~ user_input|e }}` should not remove the escaping of the whole string

stof:
@frol-kr your change is wrong. As I said, concatenating 2 safe nodes can create a string requiring escaping

frol-kr:
@stof, push before reading =)
Do you thinл it's worth of doing optimizing of two constant nodes?

stof:
I'm not sure this case happens a lot (why using concatenation in this case)

frol-kr:
@stof, tried to apply an optimization as you said but I feel like that I should have added the optimization visitor somewhere around [here](https://github.com/twigphp/Twig/blob/2.x/lib/Twig/NodeVisitor/Escaper.php#L116) as well in order to get the optimization right. What do you think about that?

lib/Twig/NodeVisitor/Optimizer.php
```diff
@@ -25,6 +25,8 @@
     const OPTIMIZE_NONE = 0;
     const OPTIMIZE_FOR = 2;
     const OPTIMIZE_RAW_FILTER = 4;
+    const OPTIMIZE_CONCAT_CONST = 5;
```
stof: 5 is not good. Binary flags must use powers of 2 for the value of each flag. So it should use `16`

lib/Twig/NodeVisitor/Optimizer.php
```diff
@@ -97,6 +103,32 @@ private function optimizePrintNode(Twig_Node $node, Twig_Environment $env)
...
+    private function optimizeConcatConst(Twig_Node $node, Twig_Environment $env)
+    {
+        if ($node instanceof Twig_Node_Expression_Binary_Concat) {
+            $leftNode = $node->getNode('left');
+            $rightNode = $node->getNode('right');
+            if (
+                $leftNode instanceof Twig_Node_Expression_Constant
+                && $rightNode instanceof Twig_Node_Expression_Constant
+            ) {
+                $leftNode->setAttribute(
```
stof: I would not mutate the left node in case it is shared in other nodes (which can happen in some cases). Create a new `Twig_Node_Expression_Constant` instead

stof:
@frol-kr no you should not add it there. The auto-escaping uses a separate node traverser to perform the safety analysis. This is not the traverser optimizing things.

fabpot:
I don't think it's worth it. Why would one use concatenation for 2 static strings in a template?

frol-kr:
Sorry messed up with a fork.
Made the other [pr](https://github.com/twigphp/Twig/pull/2460)

frol-kr:
I agree with you guys that this is not worth it. I just wanted to complete it anyway just for learning purpose though.

fabpot:
ok, closing then.

PR #2462, opened 2017-05-08
Prepare for php 7.2

hboomsma:
* allow `|length` of `null`
* allow `|length` of objects not implementing `\Countable`

*To get no errors when running with PHP 7.2 You need to use phpunit
^6.1, otherwise the `each` function, used by phpunit, triggers errors.
You can do so by running phpunit like:
`SYMFONY_PHPUNIT_VERSION=6.1 phpnightly vendor/bin/simple-phpunit`*

### DIFF
--- a/lib/Twig/Extension/Core.php
+++ b/lib/Twig/Extension/Core.php
@@ -1264,6 +1264,10 @@ if (function_exists('mb_get_info')) {
      */
     function twig_length_filter(Twig_Environment $env, $thing)
     {
+        if (null === $thing) {
+            return 0;
+        }
+
         if (is_scalar($thing)) {
             return mb_strlen($thing, $env->getCharset());
         }
@@ -1272,7 +1276,11 @@ if (function_exists('mb_get_info')) {
             return mb_strlen((string) $thing, $env->getCharset());
         }
 
-        return count($thing);
+        if ($thing instanceof \Countable || is_array($thing)) {
+            return count($thing);
+        }
+
+        return 1;
     }
 
     /**
--- a/test/Twig/Tests/Fixtures/filters/length.legacy.test
+++ b/test/Twig/Tests/Fixtures/filters/length.legacy.test
@@ -1,0 +1,16 @@
+--TEST--
+"length" filter
+--TEMPLATE--
+{{ null|length }}
+{{ magic|length }}
+{{ non_countable|length }}
+--DATA--
+return array(
+    'null'          => null,                 /* triggers deprecation error */
+    'magic'         => new MagicCallStub(),  /* used to assert we do *not* call __call, also triggers deprecation */
+    'non_countable' => new \StdClass(),      /* triggers deprecation error */
+);
+--EXPECT--
+0
+1
+1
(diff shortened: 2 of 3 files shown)


xabbuh:
IMO it would make sense to then enable a PHP 7.2 build on Travis too.

SpacePossum:
that would be `php: nightly` as PHP 7.2 is not available yet on travis I think?

hboomsma:
@xabbuh , @SpacePossum, yes it would, yes `php: nightly`. But... that build will fail because it uses phpunit 5 (still uses `each()`) in combination with the phpunit-bridge. We could force it to use 6.1 for the built with `php: nightly` using an environment variable. Should we do so?

lib/Twig/Extension/Core.php
```diff
@@ -1264,6 +1264,10 @@ function _twig_escape_html_attr_callback($matches)
      */
     function twig_length_filter(Twig_Environment $env, $thing)
     {
+        if (null === $thing) {
```
rxu: If the var passed to the filter is dynamic and not set at the moment, this may cause ```Undefined variable``` PHP notice I guess. Probably ```isset()``` would be safer.

fabpot: `$thing` is always defined.

rxu: Fine, thanks.

lib/Twig/Extension/Core.php
```diff
@@ -1272,7 +1276,11 @@ function twig_length_filter(Twig_Environment $env, $thing)
             return mb_strlen((string) $thing, $env->getCharset());
         }
 
-        return count($thing);
+        if ($thing instanceof \Countable || is_array($thing)) {
+            return count($thing);
+        }
+
+        return 1;
```
SpacePossum: Wouldn't be better to throw an exception and not return a magic `1`?
Something like;
```php
throw new Twig_Error_Runtime(sprintf('Cannot determine length of "%s", expect scalar, array or Countable', get_class($thing)));
```

test/Twig/Tests/Fixtures/filters/length.legacy.test
```diff
@@ -0,0 +1,16 @@
...
+"length" filter
+--TEMPLATE--
+{{ null|length }}
+{{ magic|length }}
+{{ non_countable|length }}
+--DATA--
+return array(
+    'null'          => null,                 /* triggers deprecation error */
+    'magic'         => new MagicCallStub(),  /* used to assert we do *not* call __call, also triggers deprecation */
+    'non_countable' => new \StdClass(),      /* triggers deprecation error */
```
SpacePossum: does it trigger such error on the branch?

fabpot:
Thank you @hboomsma.

test/Twig/Tests/Fixtures/filters/length.legacy.test
fabpot: fixed in f4adaef9f160f541b6f21bb4fe2f0c5af16196ab

lib/Twig/Extension/Core.php
stof: no, because this would be a BC break (2.x will trigger a deprecation warning there)

PR #2465, opened 2017-05-11
Fix generating template name in createTemplate() method

pulzarraider:
Before this change the name for the template was generated randomly based on uniqid(mt_rand()). It generated new file whenever the template was loaded as string using the `template_from_string` function even the template was the same. The generated file wasn't used anymore.

Now the cache file is generated by the content of template so using the `template_from_string` is much faster and doesn't pollute the cache folder with unneeded files.


fabpot:
Thank you @pulzarraider.

PR #2467, opened 2017-05-11
added PHP 7.2 to the test matrix

fabpot:


stof (approved):

.travis.yml
```diff
@@ -41,3 +42,5 @@ matrix:
           env: TWIG_EXT=yes
         - php: 7.1
           env: TWIG_EXT=yes
+        - php: nightly
+          env: TWIG_EXT=yes
```
xabbuh: Does this make sense on PHP 7?

stof: no, which is precisely why @fabpot added it in the exclude section

stof (approved):

.travis.yml
xabbuh: Oops, I did not notice that this was the `exclude` section.

xabbuh (approved):

PR #2468, opened 2017-05-11
Fix code for older versions of PHP

fabpot:


stof (approved):

stof:
@fabpot can you merge this ? I think it is needed to fix issues like the one in https://travis-ci.org/twigphp/Twig/jobs/231472203

Issue #2464, opened 2017-05-11
Tests fail on PHP nightly (7.2-dev)


rxu:
Early tests against ```- php: nightly``` fail.
1st, the ```length``` filter own Twig's repo test fails (like here https://travis-ci.org/rxu/Twig/jobs/230340700) with error:
```
Twig_Error: Twig_Error_Runtime: An exception has been thrown during the rendering of a template ("count(): Parameter must be an array or an object that implements Countable") in "index.twig" at line 7 in "filters/length.test".)
```

2nd, ```include``` tag test fails as well with error:
```
Twig_Error_Syntax: An exception has been thrown during the compilation of a template ("count(): Parameter must be an array or an object that implements Countable") in "include.html".
```
The scenario is: create template file (say, ```include.html```) with the code 
```twig
{% include 'variable.html' %}
```
The ```variable.html``` itself can be either empty or containing some code which is doesn't matter.
Example: https://travis-ci.org/phpbb/phpbb/jobs/231046144#L790
While further debugging the 2nd issue, it turned out that in this scenario, here https://github.com/twigphp/Twig/blob/2.x/lib/Twig/Node/Module.php#L369 the ```$nodes``` consists of 2 elements: ```object(Twig_Node_Expression_Constant)``` and ```NULL```. Performing count on ```NULL``` causes the above PHP 7.2 error.

xabbuh:
see #2462

rxu:
@xabbuh You probably remember I was suggesting the same a while ago here https://github.com/twigphp/Twig/issues/2382 😉 
Nice to see it's going to be fixed right way.

fabpot:
I've merged #2462 and added PHP 7.2 on Travis.

rxu:
Actually, for phpbb we can't move above v.1.26 because of https://github.com/twigphp/Twig/issues/2218, but I'll find a way to test (probably via filter overriding).

rxu:
@fabpot Can't see PHP 7.2 tests for twig though..

rxu:
Also, you forgot to update the non-mbstring fallback part here https://github.com/twigphp/Twig/blob/1.x/lib/Twig/Extension/Core.php#L1364-L1375

SpacePossum:
those can be replaced with `mb_` counterparts as it is polyfilled https://github.com/twigphp/Twig/blob/2.x/composer.json#L31

rxu:
[...]

PR #2469, opened 2017-05-12
Length filter non-mb_string on 7.x fix.

SpacePossum:
follow up on https://github.com/twigphp/Twig/pull/2462
after https://github.com/twigphp/Twig/issues/2464#issuecomment-301014241


SpacePossum:
Maybe it would be a good idea to identify tests that depend on `mb_string` _not_ being enabled and run those test again after disabling the extension?

fabpot:
Thank you @SpacePossum.

Issue #2471, opened 2017-05-12
Twig Template Caching does not work across PHP versions


bshaffer:
If you want to prime the cache in one environment and push to another environment, you have to be on the same PHP minor version. Otherwise, `updateOptionsHash` returns a different result, and breaks the cache key.

See [`Twig_Environment::updateOptionsHash`](https://github.com/twigphp/Twig/blob/2.x/lib/Twig/Environment.php#L934).

The generated template code **should** be the same across supported PHP versions, and not change as a result of the underlying PHP version. Only the twig version and other supplied options should invalidate the cache. Adding the PHP version to the options hash seems like an extra precaution, unless there is a case I am missing.

The use case here is to warm the cache in one environment and deploy to another environment. This would be theoretically possible if we removed the PHP version from the options hash.

stof:
Changing the compiled code based on the PHP version allows to optimize the performance of the compiled code on newer PHP versions.

Removing the PHP version from the hash is not an option if you don't remove all these optimizations first

Tobion:
Or just use the same PHP version locally that you are deploying to. It's usually the best practice you should go for.

fabpot:
Closing as the current behavior is the right one. We want to be able to optimize the generator code based on the version of PHP. I agree with @Tobion here, you should be using the same PHP version when generating the cache and in production.

PR #2472, opened 2017-05-12
Remove PHP version from Options Hash

bshaffer:
Addresses #2471

- [ ] It would be great to have tests for this, but automating cached code across PHP versions will take some consideration


stof:
-1 on this. The output of the compilation depends on the PHP version

bshaffer:
Closing, but hoping we can solve this problem a different way.

PR #2478, opened 2017-05-19
Drop support for HHVM

fabpot:


.travis.yml
```diff
@@ -14,15 +14,13 @@ php:
     - 5.6
     - 7.0
     - 7.1
-    - nightly
```
stof: you should not remove testing on PHP nightly

stof (changes requested):

robfrawley:
Shouldn't this be done aginst `2.x`, as this is basically a deprecation?

fabpot:
Doing it in 2.x would be the same. If we really want to "deprecate" HHVM, we would have to do it in 3.x, which is a no-go for me.

robfrawley:
Fair enough @fabpot. We've done the same in a few other projects I maintain because the situation was untenable, so I understand the motivation.

PR #2481, opened 2017-05-22
Mark Twig_Node_Module as final via an annotation (will be final in 3.0)

fabpot:


lib/Twig/Filter.php
```diff
@@ -34,6 +34,10 @@ class Twig_Filter
      */
     public function __construct(string $name, $callable = null, array $options = array())
     {
+        if (__CLASS__ !== get_class($this)) {
+            @trigger_error('Overriding '.__CLASS__.' is deprecated since version 2.4.0 and the class will be final in 3.0.', E_USER_DEPRECATED);
```
stof: wouldn't this make it painful to use Twig_SimpleFilter for now, to keep support for Twig 1.x and 2.x in parallel ? I suggest whitelisting Twig_SimpleFilter for now here.
In the past, we agreed on delaying the deprecation of Twig_SimpleFilter until soon before 3.x to make it easier for external libraries not able to drop support for Twig 1.x (for instance Symfony 3.x due to the min PHP version)

fabpot: I think time has come. As we will soon have namespace support, I would want to release 3.0 soon after.

stof: then, we will make it a pain for the whole Symfony community or they will just drop support for PHP 5.x in bundles to be able to avoid having to do checks on which class to use for each function and filter they expose. Have you tried to update Symfony itself to avoid these deprecations ?
Thus, the check for which class to use even has to be done by detecting the Twig version rather than the availability of features, as Twig_Filter also exists in 1.x (but is a different class, which is why we suffer all this pain btw).
At least, making them migrate to `Twig\Filter` directly would avoid doing 2 migrations (and can be detected by `class_exists). But it does not solve the issue regarding the fact that Symfony 3 LTS is stuck with Twig 1.x support

PR #2483, opened 2017-05-24
Fix `@final`-related deprecations

nicolas-grekas:
The run-time notices make Symfony fail because the check should exclude e.g. `Twig_SimpleFilter`.
But IMHO, we may not really need these run-time checks, if we say that relying on the DebugClassLoader + CHANGELOG in enough (which we did in the past).

### DIFF
--- a/lib/Twig/SimpleFilter.php
+++ b/lib/Twig/SimpleFilter.php
@@ -10,8 +10,12 @@
  */
 
 /**
- * Empty class for Twig 1.x compatibility.
+ * For Twig 1.x compatibility.
  */
-final class Twig_SimpleFilter extends Twig_Filter
-{
+require __DIR__.'/Filter.php';
+
+if (false) {
+    final class Twig_SimpleFilter extends Twig_Filter
+    {
+    }
 }
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,6 @@
 * 2.4.0 (2017-XX-XX)
 
- * marked Twig_Node_Module as final via the @final annotation
+ * marked Twig_Filter, Twig_Function, Twig_Test, Twig_Node_Module and Twig_Profiler_Profile as final via the @final annotation
 
 * 2.3.2 (2017-04-20)
 
--- a/lib/Twig/Filter.php
+++ b/lib/Twig/Filter.php
@@ -12,7 +12,7 @@
 /**
  * Represents a template filter.
  *
- * @final
+ * @final since version 2.4.0
  *
  * @author Fabien Potencier <fabien@symfony.com>
  *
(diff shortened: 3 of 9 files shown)


stof:
@nicolas-grekas see the discussion in https://github.com/twigphp/Twig/pull/2481#discussion_r118184447

nicolas-grekas:
PR updated to use class aliases

lib/Twig/SimpleFilter.php
```diff
@@ -10,8 +10,12 @@
...
 
 /**
- * Empty class for Twig 1.x compatibility.
+ * For Twig 1.x compatibility.
  */
-final class Twig_SimpleFilter extends Twig_Filter
-{
+require __DIR__.'/Filter.php';
+
+if (false) {
```
stof: Why ? AFAIK, IDEs understand `class_alias` (at least, PHPStorm does)

lib/Twig/SimpleFilter.php
```diff
@@ -10,8 +10,12 @@
  */
 
 /**
- * Empty class for Twig 1.x compatibility.
+ * For Twig 1.x compatibility.
  */
-final class Twig_SimpleFilter extends Twig_Filter
-{
+require __DIR__.'/Filter.php';
```
stof: you should use `class_exists` instead, to ensure there is no double registration

stof:
This will make it harder to deprecate the `Simple*` versions when preparing 3.0, right ?

lib/Twig/SimpleFilter.php
nicolas-grekas: for composer class maps

lib/Twig/SimpleFilter.php
nicolas-grekas: this cannot happen - we don't enclose classes into if blocks for the same reason

nicolas-grekas:
for 3, it's easy: we'll move to namespaces, so everything will be implicitly deprecated, and the namespaced version will only ship the correct name

fabpot:
@stof was talking about how to emit deprecation notices in 2.x.

nicolas-grekas:
there is no need to do so IMHO
we should just tell in the doc that people have to use namespaces
and that will be enough

fabpot:
Thank you @nicolas-grekas.

nicolas-grekas:
3.2 now merged into 3.3 so that next steps can happen on 3.3
3.2 is good as is now to me.

prisis:
@nicolas-grekas didnt you forget Twig_Profiler_Dumper_Text?

im getting this error
```php
The "Twig_Profiler_Dumper_Text" class is considered final. It may change without further notice as of its next major version. You should not extend it from "Twig_Profiler_Dumper_Html".
```

fabpot:
@prisis see #2513

PR #2486, opened 2017-05-25
[2.x] Add namespaced aliases

nicolas-grekas:
The 2.x-only part of #2484

### DIFF
--- a/lib/Twig/ExtensionSet.php
+++ b/lib/Twig/ExtensionSet.php
@@ -469,3 +469,5 @@ final class Twig_ExtensionSet
         }
     }
 }
+
+class_alias('Twig_ExtensionSet', 'Twig\ExtensionSet', false);
--- a/lib/Twig/Filter.php
+++ b/lib/Twig/Filter.php
@@ -138,3 +138,5 @@ class Twig_Filter
 
 // For Twig 1.x compatibility
 class_alias('Twig_Filter', 'Twig_SimpleFilter', false);
+
+class_alias('Twig_Filter', 'Twig\TwigFilter', false);
--- a/lib/Twig/Function.php
+++ b/lib/Twig/Function.php
@@ -128,3 +128,5 @@ class Twig_Function
 
 // For Twig 1.x compatibility
 class_alias('Twig_Function', 'Twig_SimpleFunction', false);
+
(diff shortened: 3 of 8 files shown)


fabpot:
Thank you @nicolas-grekas.

robfrawley:
@nicolas-grekas After this change was merged, when running our test suite for the `liip/imagine-bundle` package using `twig@2.x-dev`, [a PHP error is being thrown](https://travis-ci.org/liip/LiipImagineBundle/jobs/238667705#L373) with the following message when defining a [custom extension that returns a single Twig filter](https://github.com/liip/LiipImagineBundle/blob/1.0/Templating/ImagineExtension.php#L39) (nothing special about it).

> Cannot declare class Twig\TwigFilter, because the name is already in use

We are referencing `TwigFilter` using the `Twig_SimpleFilter` class name, but I can also confirm that when using the new namespaced class name `Twig\TwigFilter` the error is still thrown. Downgrading to `1.x-dev` *fixes the issue* so it appears only the `2.x` branch contains this problem.

Removing the following line from `lib/Twig/SimpleFilter.php` *fixes this issue*.

```php
class_alias('Twig_SimpleFilter', 'Twig\TwigFilter', false);
```

But it I'm not sure it's that simple. Based on a cursory examination, it appears like [Twig/SimpleFilter.php:23](https://github.com/twigphp/Twig/blob/2.x/lib/Twig/SimpleFilter.php#L23) and [Twig/Filter.php:142](https://github.com/twigphp/Twig/blob/2.x/lib/Twig/Filter.php#L142) are the cause of the error, because one includes the other and they both declare an alias for `Twig\TwigFilter`. Not positive on how to overcome this, but it doesn't appear they can both alias the same class without a class exists check or another mechanism in place.

If you'd like me to open a proper issue instead of bringing this up on a closed PR, let me know and I'll make it happen.

robfrawley:
Also, it seems odd that the test suite for this bundle didn't fail; shouldn't there be unit tests somewhere instantiating a filter to test it *can be instantiated*?

nicolas-grekas:
Please don't comment on closed PR, it makes tracking harder for everyone.
Can you please check https://github.com/twigphp/Twig/issues/2495 a report there, or open an issue here if more sensible to you?

PR #2488, opened 2017-05-26
Twig_TokenParser_Include cannot be final

ro0NL:
Fixes `User Deprecated: The "Twig_TokenParser_Include" class is considered final. It may change without further notice as of its next major version. You should not extend it from "Twig_TokenParser_Embed".` on 1.x

See https://github.com/twigphp/Twig/pull/2288#issuecomment-304039530

cc @fabpot


ro0NL:
Never mind.. fixed somewhere along the way :)

PR #2490, opened 2017-05-30
[1.x] Allow using namespaced aliases as extension names

nicolas-grekas:


stof:
Actually, can you add a unit test covering this ?

nicolas-grekas:
test added

nicolas-grekas:
Now with some `require_once` so that PHP doesn't cry because of incompatible type hints (because it doesn't know about the aliases yet.)

test/Twig/Tests/EnvironmentTest.php
```diff
@@ -9,7 +9,7 @@
  * file that was distributed with this source code.
  */
 
-require_once dirname(__FILE__).'/FilesystemHelper.php';
+require_once __DIR__.'/FilesystemHelper.php';
```
fabpot: Should be reverted. Even if you're right here, I prefer to not upgrade the code to 5.3+ in 1.x. It has been done in 2.x and I don't want people starting to send random PR about such changes :)

nicolas-grekas: reverted

nicolas-grekas: haha, who should I fallow? fabpot or fabbot?

stof:
Do we actually need these require_once everywhere ?

nicolas-grekas:
Yes. That's what https://github.com/symfony/symfony/pull/22962 allowed me to spot.

nicolas-grekas:
renamed `Twig\Environment` to `Twig\Twig`

test/Twig/Tests/EnvironmentTest.php
SpacePossum: (https://github.com/twigphp/Twig/pull/2493 ;) )

lib/Twig/Environment.php
```diff
@@ -758,7 +758,7 @@ public function compileSource($source, $name = null)
 
     public function setLoader(Twig_LoaderInterface $loader)
     {
-        if (!$loader instanceof Twig_SourceContextLoaderInterface && 0 !== strpos(get_class($loader), 'Mock_Twig_LoaderInterface')) {
+        if (!$loader instanceof Twig_SourceContextLoaderInterface && 0 !== strpos(get_class($loader), 'Mock_')) {
```
nicolas-grekas: Mock_Twig_LoaderInterface vs Mock_LoaderInterface when sing the namespaced variant
looking for mock_ is enough IMHO

nicolas-grekas:
Now with aliases that were missing for Twig_Extension_InitRuntimeInterface, Twig_Extension_GlobalsInterface, Twig_SourceContextLoaderInterface and Twig_ExistsLoaderInterface

PR should be ready.

fabpot:
Thank you @nicolas-grekas.

Koc:
@fabpot why reverted `\Twig\Twig`?

fabpot:
After trying to use the new class, it felt wrong. `Twig` feels like a singleton, which is wrong. `Environment` conveys the idea that you can have more than one, which is better.

PR #2494, opened 2017-06-01
Fix Twig\Environment alias

nicolas-grekas:


fabpot:
Thank you @nicolas-grekas.

apfelbox:
Why not name the class just `Twig\TwigEnvironment`? Both arguments are understandable, that

```php
function a (Environment $twig)
{
}
```

is unclear and that `Twig` by itself looks like as singleton.
Why not a third alternative?

lyrixx:
ahah, funny. I just said the same things to @nicolas-grekas 3 minutes ago (not the part about the singleton, only about the naming)

javiereguiluz:
`TwigEnvironment` sounds like a "context object" that stores some config options about the environment where "the real Twig" is executed. But here we are talking about "the real Twig object that you must use", so anything different than Twig looks confusing in my opinion.

nicolas-grekas:
You can all have what you want by using aliases:
`use Twig\Environment as TwigEnvironment;`
or
`use Twig\Environment as Twig;`

Since we even don't agree on a new name, let's keep the historical one, it can only help migrations.

lyrixx:
Who is not agree?

apfelbox:
> `TwigEnvironment` sounds like a "context object" that stores some config options about the environment where "the real Twig" is executed.

Yes, but it was like that quite some years now.


> Since we even don't agree on a new name, let's keep the historical one, it can only help migrations.

Don't want to bikeshed, but I would argue that

```php
function a (TwigEnvironment $twig)
{
}
```

would have more historical relevance than

```php
function a (Environment $twig)
{
}
```

The argument about renaming is correct, but when I write the code in my IDE I always auto-import Twig and never write the `use` by hand. So one would need to remember that `Environment` is the main class to import.

Also, keep in mind, that `Environment` will be the class name that is mainly associated with twig from now on for the regular user.

In the end, if we don't find a obvious consensus it probably isn't worth the time discussing it, though.

fabpot:
The argument about `Environment` and IDE is the same for almost all classes in Twig. And we won't add the Twig prefix for all classes, that would not make sense.

lyrixx:
Actually, I don't want to name this class `TwigEnvironment`, just `Twig`.
What is the issue with `Twig\Twig`?

PR #2502, opened 2017-06-05
Use class_exists instead of require

ogizanagi:
I think this is required for https://github.com/symfony/symfony/pull/23073 to pass. It's probably very similar to https://github.com/symfony/symfony/pull/22657.

A simple reproducer:

```php
<?php

use Twig\Node\Node;

require_once __DIR__ . '/vendor/autoload.php';

new Node();
new \Twig_Node();
```

[Without this patch](https://travis-ci.org/symfony/symfony/jobs/239712676#L2100):

```php
PHP Fatal error:  Cannot declare class Twig_Node, because the name is already in use in vendor/twig/twig/lib/Twig/Node.php on line 20
```

With, everything works fine, whatever the two calls order.

### DIFF
--- a/src/Cache/FilesystemCache.php
+++ b/src/Cache/FilesystemCache.php
@@ -2,7 +2,7 @@
 
 namespace Twig\Cache;
 
-require __DIR__.'/../../lib/Twig/Cache/Filesystem.php';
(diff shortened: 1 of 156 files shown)


fabpot:
Thank you @ogizanagi.

PR #2505, opened 2017-06-07
Support absolute paths when rendering a template

FrancescoBorzi:


stof:
@fabpot we should work on deprecating this pattern IMO, which was allowed due to internal implementation details of the filesystem loader but not intentionally supported (we actually even broke it in the past and reverted some changes to avoid BC breaks for people relying on this behavior).

lib/Twig/Loader/Filesystem.php
```diff
@@ -181,6 +181,12 @@ protected function findTemplate($name, $throw = true)
             return $this->cache[$name];
         }
 
+        if (is_file($name)) {
+            $this->cache[$name] = $name;
+
+            return $name;
```
stof: returning a name without the validation is a no-go. It is a security issue. This logic must go after the validation

stof (changes requested):

lib/Twig/Loader/Filesystem.php
ShinDarth: done

fabpot:
Can you add a test?

ShinDarth:
currently I'm on holiday but I can add the test as soon as I'll be back if my proposed change is acceptable

stof:
Btw, this also breaks the promise that Twig does not allow rendering things outside the configured template folders, which could introduce security issues. So I don't like this change.

fabpot:
I'm closing this pull request as I agree with @stof, we don't want to support absolute paths. It worked by chance but was never supposed to be supported.

Issue #2503, opened 2017-06-05
 Cannot redeclare class Twig_Node


anacona16:
I'm getting this error after upgrade from Symfony 3.3.0 to Symfony 3.3.1

```
 (1/1) FatalErrorException

Compile Error: Cannot redeclare class Twig_Node
in classes.php (line 21780)
```
Here the versions:

```
symfony/symfony                               v3.3.1             The Symfony PHP framework
twig/extensions                               v1.5.0             Common additional features for Twig that do not directly belong in core
twig/twig                                     v1.34.1            Twig, the flexible, fast, and secure template language for PHP

```

I cleared cache, I reinstalled all vendors, I deleted the bootstrap.app.cache file, but the error still exits.

anacona16:
I executed 
```
opcache_reset();
```

And everything is ok now.

anacona16:
-- Update

If I comment this line `$kernel->loadClassCache();` in web/app.php or web/app_dev.php it works.

xabbuh:
Can you try Twig 1.34.2? I think this should be solved by #2502.

gremo:
For me the only way to fix the problem was:

    "conflict": {
        "twig/twig": ">=1.34"
    }

stof:
@gremo have you tried 1.34.2 ?

gremo:
@stof yes and it doesn't work.

xabbuh:
@gremo What is the exact error that you get?

anacona16:
I updated to 

```
symfony/symfony                               v3.3.2             The Symfony PHP framework
twig/extensions                               v1.5.0             Common additional features for Twig that do not directly belong in core
twig/twig                                     v1.34.2            Twig, the flexible, fast, and secure template language for PHP
      
```

And isn't working.

gremo:
@xabbuh same error

> Compile Error: Cannot redeclare class Twig_Node in classes.php (line 19636)

GuigZ-:
Hey,

I've a same error before #2495 and now i've this.

I'm in symfony 2.8.

nicolas-grekas:
Why is Twig_Node inlined in classes.php?
What is on line 19636 in this file?

nicolas-grekas:
Can you share that file in a gist?

GuigZ-:
Personnaly is just a declaration class.

anacona16:
[...]

PR #2506, opened 2017-06-07
Fix can not redeclare Twig_Node

GuigZ-:
Hey,

I fix the bug #2503 with this fix on my symfony 2.8.

### DIFF
--- a/lib/Twig/Compiler.php
+++ b/lib/Twig/Compiler.php
@@ -10,8 +10,6 @@
  * file that was distributed with this source code.
  */
 
-class_exists('Twig_Node');
-
 /**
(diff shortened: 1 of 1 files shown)


stof:
This looks wrong to me. We need to have the class loaded to have the alias properly defined.

Thus, a `class_exists` call should not be triggering a double class definition, as it will *not* call the autoloader if the class already exists. So it would imply we have some weirdness somewhere

GuigZ-:
In my case, since this fix, i've not a problem.

The first time, my website load correctly, but after, doesn't.
When i set debug_backtrace in Node.php, we found that Twig_Compiler load Twig_Node before the Symfony loader.

PR #2509, opened 2017-06-09
Add support for runtime loaders to IntegrationTestCase

### LINKED ISSUE #2492: IntegrationTestCase does not support runtime loaders
As far as I can see, it is not possible to write integration tests for Twig extensions using `Twig_Test_IntegrationTestCase` because the class does not support adding runtime loaders.

Is there another way to write tests for extensions that use runtime loader?

emodric:
Adds the ability to provide runtime loaders in `IntegrationTestCase`. Otherwise, there's no way to test extensions that use them.

Fixes #2492

### DIFF
--- a/lib/Twig/Test/IntegrationTestCase.php
+++ b/lib/Twig/Test/IntegrationTestCase.php
@@ -25,6 +25,14 @@ abstract class Twig_Test_IntegrationTestCase extends TestCase
     abstract protected function getFixturesDir();
 
     /**
+     * @return Twig_RuntimeLoaderInterface[]
+     */
+    protected function getRuntimeLoaders()
+    {
+        return array();
+    }
+
+    /**
      * @return Twig_ExtensionInterface[]
      */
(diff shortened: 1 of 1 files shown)


lib/Twig/Test/IntegrationTestCase.php
```diff
@@ -25,6 +25,14 @@
     abstract protected function getFixturesDir();
 
     /**
+     * @return Twig_RuntimeLoaderInterface[]
+     */
+    protected function getRuntimeLoaders()
```
SpacePossum: in theory this is a BC break, but I do like the change :)

keradus: I think that Twig is following symfony BC promise.

| If you extend the class and... | Then we guarantee BC... |
|---|---|
| Add a new method | No |

http://symfony.com/doc/current/contributing/code/bc.html

fabpot:
Thank you @emodric.

PR #2511, opened 2017-06-09
Create path.rst

flatline-studios:
Several core functions are not documented.


jvasseur:
`path` in not part of the Twig core. It is part of Symfony and is documented in the symfony doc : http://symfony.com/doc/current/templating.html#linking-to-pages

PR #2512, opened 2017-06-12
Issue in documentation for twig number formatter

simsim2006:
The explaination below my change is correct. Just the output is not correct


fabpot:
Good catch, thanks @simsim2006.

fabpot:
merged into 1.x

PR #2513, opened 2017-06-12
Fix deprecation when using Twig_Profiler_Dumper_Html

fabpot:


lib/Twig/Profiler/Dumper/Base.php
```diff
@@ -0,0 +1,61 @@
...
+        $nCount = count($profile->getProfiles());
+        foreach ($profile as $i => $p) {
+            $str .= $this->dumpProfile($p, $prefix, $i + 1 !== $nCount);
+        }
+
+        return $str;
+    }
+}
+
+class_alias('Twig_Profiler_Dumper_Base', 'Twig\Profiler\Dumper\BaseDumper', false);
```
nicolas-grekas: because the class is extensible, it should load the Twig_Profiler_Profile class using class_exists
OR the class should me made `@internal`

fabpot: Fixed. Would it make sense to write some automatic tests to check that we don't break something inadvertently?

nicolas-grekas: it would, but this going to be hard to do :)

nicolas-grekas (approved):

PR #2515, opened 2017-06-15
Update internals.rst

pribilinskiy:
Missing an opening paren


fabpot:
Good catch, thanks @pribilinskiy.

PR #2519, opened 2017-06-22
Added RuntimeExtensionInterface

lyrixx:


nicolas-grekas:
rebase needed

lyrixx:
rebased

fabpot:
This PR should be done on the 1.x branch.

nicolas-grekas:
On 1.x, this means it needs to be an "underscore" class, with namespaced FC layer, isn't it?

lyrixx:
I'll wait for @fabpot's feedback before updating the PR

fabpot:
I have no problem in adding new feature that would be using namespaced class names instead of _ ones. The changelog should just mention that this is restricted to PHP 5.3 only.

fabpot:
@lyrixx?

lyrixx:
PR updated. Sorry for the delay.

PR #2525, opened 2017-06-27
Support ArrayAccess in containment operator

hkdobrev:
`ArrayAccess` could easily be supported via its `offsetExists` method.

It is important `ArrayAccess` to be checked before `Traversable`, because it is common objects to implement both interfaces, but `ArrayAccess` may be implemented without the need of traversing.

### DIFF
--- a/doc/templates.rst
+++ b/doc/templates.rst
@@ -716,7 +716,7 @@ It returns ``true`` if the left operand is contained in the right:
 .. tip::
 
     You can use this filter to perform a containment test on strings, arrays,
-    or objects implementing the ``Traversable`` interface.
+    or objects implementing the ``ArrayAccess` or ``Traversable`` interfaces.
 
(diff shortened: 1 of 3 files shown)


hkdobrev:
The failing coding standard checks seem unrelated to this PR. There are few others in other files as well.

stof:
This implementation is flawed. `in_array` in PHP checks for **values** while `offsetExists` expects a **key** (like `array_key_exists`). So your PR is breaking the expectations about the `in` operator as it changes its meaning. and it also breaks BC.

so a huge -1 from me.

hkdobrev:
Hmm, indeed. It seems I was mistaken, because we are using it `foo in bar|keys`. Thanks for your reply!

Do you think there is or there should be another operator doing that for keys? Would you just use `is defined` on the key?

stof:
yeah, just use `is defined`

PR #2530, opened 2017-07-18
Throw exception on circular reference detection

yceruto:
Out-the-box this isn't common... but into Symfony framework this kind of issue can happen frequently when we're overriding a template from third-party bundle (see main issue https://github.com/symfony/symfony/issues/17557)

```twig
{# app/Resources/AcmeBlogBundle/views/Blog/index.html.twig #}
{% extends '@AcmeBlog/Blog/index.html.twig' %}
    
{% block title 'New Title' %}
```

This exception avoids inifite loop or maximum execution time for this case, by detecting a recursive loading when a template extends from itself, thinking to override the parent template which has the same namespace + template name.

### DIFF
--- a/lib/Twig/Environment.php
+++ b/lib/Twig/Environment.php
@@ -58,6 +58,7 @@ class Twig_Environment
     private $runtimeLoaders = array();
     private $runtimes = array();
     private $optionsHash;
+    private $loading = array();
 
     /**
      * Constructor.
@@ -472,7 +473,22 @@ class Twig_Environment
             $this->initRuntime();
         }
 
-        return $this->loadedTemplates[$cls] = new $cls($this);
+        if (isset($this->loading[$cls])) {
+            throw new Twig_Error_Runtime(sprintf('Circular reference detected for Twig template "%s", path: %s.', $name, implode(' -> ', array_merge($this->loading, array($name)))));
+        }
+
+        $this->loading[$cls] = $name;
+
+        try {
+            $this->loadedTemplates[$cls] = new $cls($this);
+            unset($this->loading[$cls]);
+        } catch (\Exception $e) {
+            unset($this->loading[$cls]);
+
+            throw $e;
+        }
+
+        return $this->loadedTemplates[$cls];
     }
 
     /**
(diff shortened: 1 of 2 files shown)


yceruto:
Failure unrelated on `PHP: nightly`

yceruto:
I'm thinking again about this exception, what about to create a new exception class for this purpose? something like `Twig_Error_Template_Circular_Reference` ?

fabpot:
This should be rebased on the 1.x branch.

yceruto:
Rebase done and green.

lib/Twig/Environment.php
```diff
@@ -41,6 +41,7 @@ class Twig_Environment
     protected $globals;
     protected $runtimeInitialized = false;
     protected $extensionInitialized = false;
+    protected $loading;
```
fabpot: Better to make it private.

lib/Twig/Environment.php
```diff
@@ -472,7 +473,21 @@ public function loadTemplate($name, $index = null)
             $this->initRuntime();
         }
 
-        return $this->loadedTemplates[$cls] = new $cls($this);
+        if (isset($this->loading[$cls])) {
+            throw new Twig_Error_Runtime(sprintf('Circular reference detected for Twig template "%s", path: %s', $name, implode(' -> ', array_merge($this->loading, array($name)))));
```
fabpot: There is a missing dot a the end of the exception message.

lib/Twig/Environment.php
```diff
@@ -472,7 +473,21 @@ public function loadTemplate($name, $index = null)
...
+        }
+
+        $this->loading[$cls] = $name;
+
+        try {
+            $this->loadedTemplates[$cls] = new $cls($this);
+        } catch (\Exception $e) {
+            unset($this->loading[$cls]);
```
fabpot: you need to unset at some point when the template is loaded for real, not just in case of an error, right?

yceruto: Of course, that was before rebase, my bad

yceruto: Fixed.

lib/Twig/Environment.php
```diff
@@ -58,6 +58,7 @@ class Twig_Environment
     private $runtimeLoaders = array();
     private $runtimes = array();
     private $optionsHash;
+    private $loading;
```
fabpot: could be initialized to `= array();`

yceruto: Fixed.

fabpot:
Thank you @yceruto.

PR #2535, opened 2017-07-27
Instantiate datetime without timezone

xurshid29:
Do we need direct call of `getTimezone()` on `new DateTime($date)` when `$date` is string? As timezone has already been set if it's not `false`


stof:
where has it already been set ? The object does not exist before the line you changed.

fabpot:
Closing as there is no feedback.

PR #2536, opened 2017-07-29
show how to use a variable as hash map key

tolry:
currently the documentation does not explain how it's possible to use a variable as key in a hash map

### DIFF
--- a/doc/templates.rst
+++ b/doc/templates.rst
@@ -603,11 +603,11 @@ exist:
     { 2: 'foo', 4: 'bar' }
 
     {# keys as expressions (the expression must be enclosed into parentheses) #}
-    { (1 + 1): 'foo', (a ~ 'b'): 'bar' }
-    
-    {# keys as variables #}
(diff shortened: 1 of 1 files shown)


doc/templates.rst
```diff
@@ -604,6 +604,10 @@ exist:
 
     {# keys as expressions (the expression must be enclosed into parentheses) #}
     { (1 + 1): 'foo', (a ~ 'b'): 'bar' }
+    
+    {# keys as variables #}
+    {% set dynamic_key = 'foo' %}
+    { (dynamic_key): 'foo' }
```
fabpot: That's the exact same example as the one above. What about merging the two with 3 examples: a "simple" var, an operation (1 + 1) and a concatenation (a ~ 'b')?

tolry: sounds good, will get onto that

tolry:
@fabpot updated, any thoughts? The second comment line might be a bit redundant, since expressions have their own chapter.

tolry:
found a more compact solution

fabpot:
Thank you @tolry.

PR #2539, opened 2017-08-11
Fix UseTokenParser::parse() not adhering to interface.

### LINKED ISSUE #2538: UseTokenParser::parse() does not adhere to interface
I was doing a static code analysis to see if there are type issues in Twig worth fixing and stumbled upon the following problem. I'm not sure if it's worth fixing in the first place and if so, what would be the best way to approach this.

The [TokenParserInterface::parse specifies a return type Twig_Node in it's docblock](https://github.com/twigphp/Twig/blob/2.x/lib/Twig/TokenParserInterface.php#L27). The [UseTokenParser::parse()](https://github.com/twigphp/Twig/blob/2.x/lib/Twig/TokenParser/Use.php#L28) does not adhere to that interface as it returns void.

Since moving to type hints in Twig is not a priority I think this might not be worth dealing with in the first place, but should at any point the return type on the interface be enforced this parser will break. I therefore suggest making the return type (in the docblock) on the interface optionally null or (less desirable) just add a `return new Twig_Node()` to the UseTokenParser. As a quick fix I tried to return the Node that is generated and passed to `parser::addTrait()` instead, but unlike with the empty dummy node, that will break a lot of tests.

Feel free to just close the issue, if you think it's not worth discussing/fixing at the moment.

dbrumann:
Fixes #2538

### DIFF
--- a/lib/Twig/Node/Empty.php
+++ b/lib/Twig/Node/Empty.php
@@ -15,7 +15,7 @@
  *
  * @author Denis Brumann <denis.brumann@sensiolabs.de>
  */
-class Twig_Node_Empty extends Twig_Node
+final class Twig_Node_Empty extends Twig_Node implements Twig_NodeCaptureInterface
 {
     public function __construct($lineno = 0, $tag = null)
     {
--- a/lib/Twig/Node/Empty.php
+++ b/lib/Twig/Node/Empty.php
@@ -19,7 +19,7 @@ class Twig_Node_Empty extends Twig_Node
 {
     public function __construct($lineno = 0, $tag = null)
     {
-        parent::__construct([], [], $lineno, $tag);
+        parent::__construct(array(), array(), $lineno, $tag);
     }
 
     public function __toString()
--- a/lib/Twig/Node/Empty.php
+++ b/lib/Twig/Node/Empty.php
@@ -1,0 +1,31 @@
+<?php
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) Fabien Potencier
+ * (c) Armin Ronacher
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+/**
+ * Dummy node that represents an empty node, not producing any output.
+ *
+ * @author Denis Brumann <denis.brumann@sensiolabs.de>
+ */
+class Twig_Node_Empty extends Twig_Node
+{
+    public function __construct($lineno = 0, $tag = null)
+    {
+        parent::__construct([], [], $lineno, $tag);
+    }
+
+    public function __toString()
+    {
+        return '';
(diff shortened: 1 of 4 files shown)


lib/Twig/Node/Empty.php
```diff
@@ -0,0 +1,31 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+/**
+ * Dummy node that represents an empty node, not producing any output.
+ *
+ * @author Denis Brumann <denis.brumann@sensiolabs.de>
+ */
+class Twig_Node_Empty extends Twig_Node
```
stof: This node must implement `Twig_NodeCaptureInterface`, otherwise `{% use %}` will be forbidden at top-level in templates extending another template.

And the class should be final to forbid turning an empty node into a child class not being empty

test/Twig/Tests/Node/EmptyTest.php
```diff
@@ -0,0 +1,36 @@
...
+     */
+    public function testAccessingNameThrowsException()
+    {
+        $body = new Twig_Node_Empty();
+        $node = new Twig_Node_Block('foo', $body, 1);
+
+        $this->assertEquals($body, $node->getNode('body'));
+        $body->getAttribute('name');
+
+        $this->fail('Node should not have a "name"-attribute');
```
stof: should be removed. The failure will be triggered by `@expectedException`

test/Twig/Tests/Node/EmptyTest.php
```diff
@@ -0,0 +1,36 @@
...
+
+class Twig_Tests_Node_EmptyTest extends Twig_Test_NodeTestCase
+{
+    /**
+     * @expectedException LogicException
+     */
+    public function testAccessingNameThrowsException()
+    {
+        $body = new Twig_Node_Empty();
+        $node = new Twig_Node_Block('foo', $body, 1);
```
stof: why is this Twig_Node_Block` involved in the testing ? It does not make sense

stof:
@fabpot I think this should even be merged in 1.x

test/Twig/Tests/Node/EmptyTest.php
dbrumann: I saw this in some of the other tests I opened, that's why I copied it. I will remove it, if it's not necessary.

dbrumann:
I rebased against 1.x and changed the base branch for this PR

fabpot:
I don't think we need to create another Node class for that. See #2543, where I just return an empty Node. I've added a test to proves that it works well. We don't need to implement `Twig_NodeCaptureInterface` as empty nodes are allowed anywhere.

dbrumann:
Closed in favor of #2543

PR #2555, opened 2017-09-12
Compile to PHP code calling function directly from default namespace.

SpacePossum:
Calling functions directly in the default namespace has 2 effects:
- sometimes it's faster (but never slower), @ see https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3048
- it makes Twig always behave the same (i.e. not effected if function is overloaded)

Typical downside is that it doesn't look very nice in the code. However since people, besides debuggers and maintainers of Twig, usually don't look into the compiled PHP it is not much of an issue.

I only checked the code generating PHP because:
- SCA/PHP CS Fixer will not find these cases
- change the regular PHP code is probably best when the fixer has been adjusted to do so, rather than doing it by hand (also keeps the # changes in this PR down)


nicolas-grekas:
Are compiled templates namespaced? If not, then this is useless :)

SpacePossum:
AFAIK not, typical cache file looks like on one of my systems:
```
<?php

/* TwigBundle:Exception:error.html.twig */
class __TwigTemplate_b3455b49a9ae55d633b430813e5acc0ac659175029320044d17091682e48fcb6 extends Twig_Template
{
```

 So no namespace declarations there, `abstract class Twig_Template` isn't namespace either.

nicolas-grekas:
OK, so I confirm, all functions are already in the root namespace, so this PR won't provide anything at all :)

SpacePossum:
:(
we'll still have `it makes Twig always behave the same (i.e. not effected if function is overloaded)`,
lets see where this goes.

nicolas-grekas:
you cannot overload anything in the root namespace :)

SpacePossum:
OK, first beer is on me :)

PR #2557, opened 2017-09-18
add throws tags doc block for Twig_Environment:load

Haehnchen:
On an internal project review we noticed that `\Twig_Environment::load` is not providing throw tags. As this method is just a proxy for `\Twig_Environment::loadTemplate` which itself is _internal_ we had some discussion around.

this PR provides same throw tags as _loadTemplate_. Maybe this helps to understand code usage


fabpot:
Thank you @Haehnchen.

PR #2560, opened 2017-09-22
Added pssibility to reset the profiler

derrabus:
In order to implement symfony/symfony#18244 properly, I needed the ability to reset a Twig profile.

### DIFF
--- a/lib/Twig/Profiler/Profile.php
+++ b/lib/Twig/Profiler/Profile.php
@@ -145,6 +145,12 @@ class Twig_Profiler_Profile implements IteratorAggregate, Serializable
         );
     }
 
+    public function reset()
+    {
+        $this->starts = $this->ends = $this->profiles = array();
+        $this->enter();
+    }
+
     public function getIterator()
     {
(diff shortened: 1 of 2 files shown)


nicolas-grekas (approved):

xabbuh (approved):

stof:
Does resetting the Profile object actually work fine ? This object is the data object serialized by the TwigDataCollector in `lateCollect`.
If resetting happens before this `lateCollect` (I don't remember in which order they are called), the symfony profiler would store an empty profile.

derrabus:
@stof This applies to any data collector or tracable service used in Symfony. Because of this, the listener that resets the profiler and other resettable services is triggered very late on the `kernel.terminate` event.

https://github.com/symfony/symfony/blob/8136fa5050761ec375278fe9273d4936a9314ec0/src/Symfony/Component/HttpKernel/EventListener/ServiceResetListener.php#L47

fabpot:
Thank you @derrabus.

PR #2561, opened 2017-09-26
add @throws in method findTemplate comment

ebuildy:
Not a big one ^^


lib/Twig/Loader/Filesystem.php
```diff
@@ -171,6 +171,8 @@ public function isFresh($name, $time)
      * @param string $name  The template name
      * @param bool   $throw Whether to throw an exception when an error occurs
      *
+     * @throws Twig_Error_Loader
```
fabpot: To be useful, it should describe when such an exception could be thrown.

fabpot (changes requested):

PR #2562, opened 2017-09-26
Change typo in documentation

uthopiko:
In example of use for test_iterable function there is a typo

### DIFF
--- a/lib/Twig/Extension/Core.php
+++ b/lib/Twig/Extension/Core.php
@@ -1257,7 +1257,7 @@ function twig_test_empty($value)
  *
  * <pre>
  * {# evaluates to true if the foo variable is an array or a traversable object #}
- * {% if foo is traversable %}
+ * {% if foo is iterable %}
  *     {# ... #}
  * {% endif %}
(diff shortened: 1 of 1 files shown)


javiereguiluz (approved):

lib/Twig/Extension/Core.php
```diff
@@ -1257,7 +1257,7 @@ function twig_test_empty($value)
  *
  * <pre>
  * {# evaluates to true if the foo variable is an array or a traversable object #}
```
fabpot: Should this be updated as well then?

fabpot: Is it the only occurrence that needs to be changed in the doc?

stof: Well, we should update the short description of the function 3 lines above too (what it checks is for being iterable, not traversable, according to PHP's naming)

stof: The doc itself uses the right name

fabpot:
Thank you @uthopiko.

PR #2570, opened 2017-10-20
fix grammar

redbrickone:


fabpot:
Good catch, thanks @redbrickone.

PR #2578, opened 2017-11-02
Clarify replace documentation and example

lftl:
At a quick glance the description and example for the replace filter make it seem like the filter only replaces using a special %foo% syntax rather than working on any arbitrary search and replace. See some of the confusion here:

https://stackoverflow.com/questions/4943880/str-replace-in-twig

I tweaked the description of the filter to more closely match the PHP documentation's description of str_replace, and dropped the % formatting from the example.

### DIFF
--- a/doc/filters/replace.rst
+++ b/doc/filters/replace.rst
@@ -1,12 +1,11 @@
 ``replace``
 ===========
 
-The ``replace`` filter formats a given string by replacing the placeholders
-(placeholders are free-form):
+The ``replace`` filter formats a given string by replacing the occurences of the search string with the replacement string:
 
 .. code-block:: jinja
 
-    {{ "I like %this% and %that%."|replace({'%this%': foo, '%that%': "bar"}) }}
+    {{ "I like this and that."|replace({'this': foo, 'that': "bar"}) }}
 
     {# outputs I like foo and bar
        if the foo parameter equals to the foo string. #}


doc/filters/replace.rst
```diff
@@ -1,12 +1,11 @@
...
 ===========
 
-The ``replace`` filter formats a given string by replacing the placeholders
-(placeholders are free-form):
+The ``replace`` filter formats a given string by replacing the occurences of the search string with the replacement string:
 
 .. code-block:: jinja
 
-    {{ "I like %this% and %that%."|replace({'%this%': foo, '%that%': "bar"}) }}
+    {{ "I like this and that."|replace({'this': foo, 'that': "bar"}) }}
```
fabpot, line 8: Perhaps we can add a small paragraph recommending using some kind of conventions for placeholders to avoid wild replacements. Sometimes, the var where you want to do replacements comes from the DB or elsewhere.

fabpot (changes requested):

lftl:
That makes perfect sense. In my case I didn't have a ton of control over the string that I was doing the replacement on, and a quick glance at the docs made me think I HAD to use a %var% placeholder syntax. On closer view of the code it's clear you don't have to do that, but it threw me off at first. Recommending using some type of placeholder makes sense though.

doc/filters/replace.rst
```diff
@@ -1,12 +1,11 @@
 ``replace``
 ===========
 
-The ``replace`` filter formats a given string by replacing the placeholders
-(placeholders are free-form):
```
fabpot, line 5: Note that the removed parenthesis already explains how that works.

fabpot:
Done in f93530b4

PR #2581, opened 2017-11-06
Add JSON escape strategy

c960657:
The `js` escape strategy used `\xNN`-style escape sequences. This is not allowed in JSON, only `\uNNNN` is allowed.

This PR adds a new escape strategy, `json` that is similar to `js` except that it does not use `\xNN` but uses `\uNNNN` instead.

I know that I can use `json_decode()`, but it came as a surprise to me that `escape('js')` did not work. You probably shouldn't be generating JSON structures in your templates, but sometimes it comes in handy. My use-case was generating a piece of [JSON-LD](https://json-ld.org).

A different approach that would perhaps be less confusing to users would be to change `js` to not use `\xNN`. This output is still valid JavaScript, it just uses two more bytes per occurrence. Let me know what you think.

### DIFF
--- a/lib/Twig/Extension/Core.php
+++ b/lib/Twig/Extension/Core.php
@@ -1152,11 +1152,6 @@ function _twig_escape_js_callback($matches)
 {
     $char = $matches[0];
 
-    // \xHH
-    if (!isset($char[1])) {
-        return '\\x'.strtoupper(substr('00'.bin2hex($char), -2));
-    }
-
     // \uHHHH
     $char = twig_convert_encoding($char, 'UTF-16BE', 'UTF-8');
     $char = strtoupper(bin2hex($char));
--- a/lib/Twig/Extension/Core.php
+++ b/lib/Twig/Extension/Core.php
@@ -1040,7 +1040,7 @@ function twig_escape_filter(Twig_Environment $env, $string, $strategy = 'html', 
 
         case 'js':
             // escape all non-alphanumeric characters
-            // into their \xHH or \uHHHH representations
+            // into their \x or \uHHHH representations
             if ('UTF-8' !== $charset) {
                 $string = twig_convert_encoding($string, 'UTF-8', $charset);
             }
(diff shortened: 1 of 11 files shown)


fabpot:
1.x is accepting new features, so it should be done on the 1.x branch instead.

xabbuh:
And since this is a new feature it should be part of a minor version increment in the changelog (not part of a patch/bugfix release).

c960657:
Changed branch and moved to minor version.

The Fabbot failures are not in code I touched. The Travis failures on PHP 5.3 is a [separate issue](https://github.com/twigphp/Twig/pull/2582).

fabpot:
If that makes no difference, I would change the JS strategy instead of adding a new one, that would be less confusing. Also, if you rebase, tests would be ok and fabbot happy. Thanks.

c960657:
I have now modified the existing `js` filter to generate valid JSON.

The old approach is here: https://github.com/twigphp/Twig/compare/1.x...c960657:escape-json-bak

test/Twig/Tests/Fixtures/tags/autoescape/strategy.legacy.test
```diff
@@ -1,4 +1,4 @@
---TEST--
+  --TEST--
```
fabpot: This change looks wrong to me.

fabpot (changes requested):
@xabbuh  @stof Do you think this is acceptable? Having longer encoded strings? I think it's not an issue, but want to double-check with you :)

c960657:
If longer strings is an issue, we could use the shorter escape sequences for `\n`, `\t` etc.

fabpot:
@c960657 We need to balance output length and performance. If using `\n`/`\t` does not come with a performance penalty, go for it.

c960657:
I don't think there is any significant performance penalty.

I have added short codes for escape sequences supported in both JavaScript and JSON (JavaScript supports e.g. `\v` or `\.`, but JSON does not).

Also, inspired by [drupal_json_encode](https://api.drupal.org/api/drupal/includes%21common.inc/function/drupal_json_encode/7.x), `"` is still encoded using `\uHHHH` sequence.

lib/Twig/Extension/Core.php
```diff
@@ -1152,9 +1152,23 @@ function _twig_escape_js_callback($matches)
 {
     $char = $matches[0];
 
-    // \xHH
-    if (!isset($char[1])) {
-        return '\\x'.strtoupper(substr('00'.bin2hex($char), -2));
+    /*
+     * A few characters have short escape sequences in JSON and JavaScript.
+     * Escape sequences supported only by JavaScript, not JSON, are ommitted.
+     * \" is also supported but omitted, because the resulting string is not HTML safe.
```
fabpot, line 1158: Do you mean `\"` or `"` here?

c960657: Both, sort of :-)  I mean that the escape sequence for the character `"` is `\"`. Neither the raw character nor the escaped string is HTML safe.

fabpot:
Thank you @c960657.

PR #2585, opened 2017-11-09
Update spaceless to be more aggressive

pich:
Remove line breaks added.

### DIFF
--- a/lib/Twig/Node/Spaceless.php
+++ b/lib/Twig/Node/Spaceless.php
@@ -12,7 +12,7 @@
 /**
  * Represents a spaceless node.
  *
- * It removes spaces between HTML tags.
+ * It removes spaces and newlines in HTML.
  *
  * @author Fabien Potencier <fabien@symfony.com>
  */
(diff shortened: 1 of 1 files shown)


xabbuh:
IMO this must be a new tag or you need to implement a way to make this an opt-in feature. Otherwise, this will break existing application where people rely on line breaks being preserved.

stof:
this changes totally the behavior of the tag, as it now replaces any sequence of 2+ whitespace chars with a single space, even when this is not between HTML tags.

So this is a BC break. This cannot be added in the existing tag.

pich:
My point was to make html output minified. 
I usually do it partially by put spaceless tag outside html tag in main twig layout file. 
Maybe right approach would be to add it to base code and manage via configuration option minified: true/false (by default false to keep BC). As result output should by minify/pack. 
It's  worth to consider more advance features managed via configuration like strip_comments, prettify.

I would be nice to have prettify in dev env and minified on prod.  :)

fabpot:
Minification is not something I want to support in core. There is no point in doing so IMHO.

stof:
And anyway, it does not belong to the existing tag with a different behavior. You could create a separate tag in an extension instead.

thus, this implementation is not fool-proof (it will remove spaces even in places where they are necessary).

fabpot:
Closing.

PR #2586, opened 2017-11-10
Environment::resolveTemplate now accepts instances of TemplateWrapper

victorwelling:
The [documentation on the include function](https://twig.symfony.com/doc/2.x/functions/include.html) states (on the template expression):
> And if the expression evaluates to a Twig_Template or a Twig_TemplateWrapper instance, Twig will use it directly […]

This is currently broken, as `twig_include` [relies on `resolveTemplate`](https://github.com/twigphp/Twig/blob/31b577863b5ae8d19e2cefd174e95a433f16491f/lib/Twig/Extension/Core.php#L1304), which doesn't handle instances of `TemplateWrapper`. This PR fixes that.

For the sake of completeness: the `include` tag doesn't have this problem, as it relies on the `loadTemplate` method of the `Template` class which _does_ handle this case.

### DIFF
--- a/lib/Twig/Environment.php
+++ b/lib/Twig/Environment.php
@@ -453,12 +453,12 @@ class Twig_Environment
     /**
      * Tries to load a template consecutively from an array.
      *
-     * Similar to loadTemplate() but it also accepts Twig_Template instances and an array
-     * of templates where each is tried to be loaded.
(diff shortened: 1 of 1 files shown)


stof:
Please add a test covering this to prevent regressions

fabpot:
Thank you @victorwelling.

PR #2592, opened 2017-11-27
Modify the merge filter to allow merging any number of arrays as does php's array_merge

martinwittmann:
Modified the merge filter to allow merging any number of arrays|Traversables to align its behavior with that of php's array_merge function.

### DIFF
--- a/lib/Twig/Extension/Core.php
+++ b/lib/Twig/Extension/Core.php
@@ -537,26 +537,26 @@ function twig_urlencode_filter($url)
  *  {# items now contains { 'apple': 'fruit', 'orange': 'fruit', 'peugeot': 'car' } #}
  * </pre>
  *
- * @param array|Traversable $arr1 An array
- * @param array|Traversable $arr2 An array
+ * @param $arguments An array of array|Traversable which will be used to call array_merge.
  *
  * @return array The merged array
  */
-function twig_array_merge($arr1, $arr2)
+function twig_array_merge(...$arguments)
 {
-    if ($arr1 instanceof Traversable) {
-        $arr1 = iterator_to_array($arr1);
-    } elseif (!is_array($arr1)) {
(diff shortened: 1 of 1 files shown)


lib/Twig/Extension/Core.php
```diff
@@ -537,26 +537,26 @@ function twig_urlencode_filter($url)
...
  * </pre>
  *
- * @param array|Traversable $arr1 An array
- * @param array|Traversable $arr2 An array
+ * @param $arguments An array of array|Traversable which will be used to call array_merge.
  *
  * @return array The merged array
  */
-function twig_array_merge($arr1, $arr2)
+function twig_array_merge(...$arguments)
```
stof, line 544: This is not a good idea. It is breaking support for named arguments in Twig. You should keep the first 2 arguments

stof (commented):
Not having any test for that looks bad

lib/Twig/Extension/Core.php
```diff
@@ -537,26 +537,26 @@ function twig_urlencode_filter($url)
  *  {# items now contains { 'apple': 'fruit', 'orange': 'fruit', 'peugeot': 'car' } #}
  * </pre>
  *
- * @param array|Traversable $arr1 An array
- * @param array|Traversable $arr2 An array
+ * @param $arguments An array of array|Traversable which will be used to call array_merge.
```
stof, line 540: this documentation is wrong. The method signature does not have a `$arguments` argument (you have such variable internally, but the signature has several arguments instead)

fabpot:
I think I prefer to keep the method signature simple (as explained by @stof in his comments) and let user composer several calls in case they need to merge several arrays. Also, using array merges in templates is discouraged, so adding easier ways to do it is not something I'm willing to do. After thinking about this, I think we should not add support for this feature. Thanks for submitting.

PR #2598, opened 2017-12-07
Update advanced.rst

pfwd:
I noticed that the rot13Provider wasn't called correctly as it was trying to access a local variable instead of calling $this


fabpot:
Good catch, thanks @pfwd.

PR #2600, opened 2017-12-07
Be able to count on iterator

plandolt:
This PR allows to count on `\Traversable` Objects. Not sure if I made the change to the correct branch. Tell me please if I should target 2.x or master.

### DIFF
--- a/doc/filters/length.rst
+++ b/doc/filters/length.rst
@@ -14,6 +14,8 @@ return value of the ``count()`` method.
 For objects that implement the ``__toString()`` magic method (and not ``Countable``),
 it will return the length of the string provided by that method.
 
+For objects that implement the ``IteratorAggregate`` interface, ``length`` will use the return value of the ``iterator_count()`` method.
+
 .. code-block:: jinja
 
     {% if users|length > 10 %}
--- a/test/Twig/Tests/Fixtures/filters/length.test
+++ b/test/Twig/Tests/Fixtures/filters/length.test
@@ -6,6 +6,7 @@
 {{ number|length }}
 {{ to_string_able|length }}
 {{ countable|length }}
+{{ iterator_aggregate|length }}
 {{ null|length }}
 {{ magic|length }}
 {{ non_countable|length }}
@@ -16,6 +17,7 @@ return array(
     'number' => 1000,
     'to_string_able' => new ToStringStub('foobar'),
     'countable' => new CountableStub(42),       /* also asserts we do *not* call __toString() */
+    'iterator_aggregate' => new IteratorAggregateStub(array('a', 'b', 'c')),   /* also asserts we do *not* call __toString() */
     'null'          => null,
     'magic'         => new MagicCallStub(),     /* used to assert we do *not* call __call */
     'non_countable' => new \StdClass(),
@@ -26,6 +28,7 @@ return array(
 4
 6
 42
+3
 0
 1
 1
--- a/test/Twig/Tests/IntegrationTest.php
+++ b/test/Twig/Tests/IntegrationTest.php
@@ -307,3 +307,21 @@ class CountableStub implements \Countable
         throw new Exception('__toString shall not be called on \Countables');
     }
 }
+
+/**
+ * This class is used in tests for the length filter
+ */
+class IteratorAggregateStub implements \IteratorAggregate
+{
+    private $data;
+
+    public function __construct(array $data)
+    {
+        $this->data = $data;
+    }
+
+    public function getIterator()
+    {
+        return new ArrayIterator($this->data);
+    }
+}
--- a/lib/Twig/Extension/Core.php
+++ b/lib/Twig/Extension/Core.php
@@ -1270,6 +1270,10 @@ if (function_exists('mb_get_info')) {
         if ($thing instanceof \Countable || is_array($thing)) {
             return count($thing);
         }
+        
+        if ($thing instanceof \IteratorAggregate) {
+            return iterator_count($thing);
+        }
 
         return 1;
     }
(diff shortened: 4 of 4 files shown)


test/Twig/Tests/IntegrationTest.php
```diff
@@ -307,3 +307,43 @@ public function __toString()
         throw new Exception('__toString shall not be called on \Countables');
     }
 }
+
+/**
+ * This class is used in tests for the length filter
+ */
+class TraversableStub implements \Iterator
```
fabpot: Instead of creating a class, could we use the `ArrayIterator` PHP built-in class instead?

fabpot (changes requested):

plandolt:
@fabpot You were right. I updated the PR to just use php's built in ArrayIterator.

SpacePossum:
in the test it is better _not_ to use [ArrayIterator](http://php.net/manual/en/class.arrayiterator.php) as it implements `\Countable` so the unit test does not reach the new code (i.e. past `if ($thing instanceof \Countable || is_array($thing)) {`)

Looking at the PR itself one thing note:

```php
<?php

/**
 * This class is used in tests for the length filter
 */
class TraversableStub implements \Iterator
{
    private $position;
    private $data;

    public function __construct(array $data)
    {
        $this->position = 0;
        $this->data = $data;
    }

    public function current()
    {
        return $this->data[$this->position];
    }

    public function next()
    {
        ++$this->position;
    }

    public function key()
    {
        return $this->position;
    }

    public function valid()
    {
        return isset($this->data[$this->position]);
    }

    public function rewind()
    {
        $this->position = 0;
    }
}



$loader = new Twig_Loader_Array(array(
    'index' =>
    '{% for a in iterator %} {{ a }} {{ iterator|length }} {% endfor %}',
));
$twig = new Twig_Environment($loader);

$data = ['one', 'two', 'three'];

echo
    "\n------------\n".
    $twig->render('index', ['iterator' => new ArrayIterator($data)]).
    "\n------------\n".
    $twig->render('index', ['iterator' => $data]).
    "\n------------\n".
    $twig->render('index', ['iterator' => new TraversableStub($data)])
;
```

renders:

```
$ php test.php 

------------
 one 3  two 3  three 3 
------------
 one 3  two 3  three 3 
------------
 one 3 
```

This is the result of `iterator_count` modifying the current position of the iterator.
It is not (practical) possible to restore the position of the iterator, so this cannot be helped. Maybe something in the docs would be nice?

plandolt:
Good catch @SpacePossum!

I updated the PR and reintroduced the ``TraversableStub`` class to be used in the test. This will cover the new implementation. I also updated the doc with a warning section about the position might be changed (words from php.net) during counting on it. English is not my native language so please review that as well.

xabbuh:
Won't this break use cases where the provided `Traversable` implementation cannot be rewinded and someone tries to iterate the object after using the `length` filter?

nicolas-grekas:
> Won't this break use cases where the provided Traversable implementation cannot be rewinded and someone tries to iterate the object after using the length filter?

YES! the only safe interface is `IteratorAggregate`, so that's the one I'd recommend to use instead of `Traversable` in this PR.

plandolt:
@xabbuh yes it would break it. But also if you currently use ``length`` on a Traversable implementation you get ``1`` which is wrong too.

@nicolas-grekas && @xabbuh: I updated this PR to only use ``iterator_count`` on ``IteratorAggregate`` implementations. What about creating a PR to count on ``Iterator`` implementations (with the warning of not correct positions) or raising an exception if one tries to count on ``Iterator`` implementations in a next major?

test/Twig/Tests/IntegrationTest.php
```diff
@@ -307,3 +307,22 @@ public function __toString()
...
+    public function __construct(array $data)
+    {
+        $this->data = $data;
+    }
+
+    public function getIterator()
+    {
+        return new ArrayIterator($this->data);
+    }
+
```
fabpot: This blank line should be removed according to our CS standards.

test/Twig/Tests/Fixtures/filters/length.test
```diff
@@ -16,6 +17,7 @@ return array(
     'number' => 1000,
     'to_string_able' => new ToStringStub('foobar'),
     'countable' => new CountableStub(42),       /* also asserts we do *not* call __toString() */
+    'traversable' => new IteratorAggregateStub(array('a', 'b', 'c')),   /* also asserts we do *not* call __toString() */
```
fabpot: I would rename this variable to better reflect the class name.

fabpot (changes requested):

nicolas-grekas:
> raising an exception if one tries to count on Iterator

I think we should follow PHP 7.2, which raises a deprecation notice when `count()` is used on non-countable types - then throw in Twig 3 yes.

plandolt:
@fabpot I updated the PR with a more accurate naming.

@nicolas-grekas I can create such a PR targeting the master branch in the future.

nicolas-grekas (approved):

doc/filters/length.rst
```diff
@@ -14,6 +14,8 @@ return value of the ``count()`` method.
 For objects that implement the ``__toString()`` magic method (and not ``Countable``),
 it will return the length of the string provided by that method.
 
+For objects that implement the ``IteratorAggregate`` interface, ``length`` will use the return value of the ``interator_count()`` method.
```
xabbuh: typo: `iterator_count()`

plandolt: Thank you, fixed.

fabpot:
Thank you @scuben.

PR #2613, opened 2018-01-23
Fix possible array to string conversion concealing actual error

finwe:
Not sure this solution is the best as with larger objects/arrays it makes the message very long, but it works.


stof:
I would rather handle a specific display for arrays (not dumping it entirely) than printing the whole array in the exception message (which will be totally unreadable)

stof:
btw, this need to be covered by a test to prevent regressions

finwe:
Will elaborate.

finwe:
Updated and added tests for all 3 defined type cases.

stof (approved):

fabpot:
@finwe Looks good to me. Can you also submit a PR for the 1.x branch where the code is slightly different?

finwe:
Sure, hopefully within a week.

fabpot:
Closing in favor of #2624

PR #2616, opened 2018-01-29
Added Twig_Extension_WhiteSpaceCutter for cutting extra white spaces

KonstantinKuklin:
This is analog of 'spaceless' block, but without ob_* overhead and with global state on\off for all twig files.
Will be useful  to economy traffic and increase loading speed on devices which connected to internet by slow channels like mobile phones.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
 * 2.4.5 (2017-XX-XX)
 
+ * added Twig_Extension_WhiteSpaceCutter
  * fixed Environment::resolveTemplate to accept instances of TemplateWrapper
 
 * 2.4.4 (2017-09-27)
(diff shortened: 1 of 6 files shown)


fabpot:
Thanks for working on this. I'm worry about having such a global extension. Removing spaces in text is generally not safe (except perhaps in an HTML context, but Twig is output-agnostic as you can use it for more than just HTML generation). As this solution is not generic and will break code, I cannot accept it in core. You can of course publish it somewhere as a Twig extension.

KonstantinKuklin:
@fabpot I see your point.
Each Twig_Node have 'name', so I can add apply filter by that parameter and developers will get ability to configure list of nodes which should be cutted or not. (Like: *.html.twig, *.xml.twig and etc)
Does PR with such configuration rules have a chance to be accepted?

fabpot:
Only features that can be widely used are accepted in core. I don't think this is something that really needs to be in Twig itself. Also because HTTP compression means that such optimizations are mostly unneeded nowadays.

KonstantinKuklin:
@fabpot thank you for explanation.

PR #2617, opened 2018-01-30
Performance optimization

fabpot:
This PR avoids calling `$this->getSourceContext()` whenever we call `twig_get_attribute()`. That saves one method call and one object creation.

### DIFF
--- a/lib/Twig/Node/Module.php
+++ b/lib/Twig/Node/Module.php
@@ -80,16 +80,7 @@ class Twig_Node_Module extends Twig_Node
 
         $this->compileClassHeader($compiler);
 
-        if (
-            count($this->getNode('blocks'))
-            || count($this->getNode('traits'))
-            || !$this->hasNode('parent')
-            || $this->getNode('parent') instanceof Twig_Node_Expression_Constant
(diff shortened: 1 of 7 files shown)


lib/Twig/Node/Module.php
```diff
@@ -80,16 +80,7 @@ protected function compileTemplate(Twig_Compiler $compiler)
...
-            count($this->getNode('blocks'))
-            || count($this->getNode('traits'))
-            || !$this->hasNode('parent')
-            || $this->getNode('parent') instanceof Twig_Node_Expression_Constant
-            || count($this->getNode('constructor_start'))
-            || count($this->getNode('constructor_end'))
-        ) {
-            $this->compileConstructor($compiler);
-        }
+        $this->compileConstructor($compiler);
```
fabpot, line 83: Always calling `compileConstructor()` is actually another small performance optimization as it sets `$this->parent` to `false` when possibler, which avoids determining the parent during template evaluation.

fabpot:
It gives a nice 20% performance boost on my local tests (on simple `twig_get_attribute()` calls).

stof (approved):

Issue #2547, opened 2017-08-21
Suggestion: optimize use of getAttribute using additional syntax to use native code


inakivb:
Currently the use of `.` ends calling to `getAttribute()`, which is a function that's difficult to optimize. 

There's an option when working with arrays to use `[]` that is more optimized as it ignores all the object calls, but it fails when working with more than one depth of array as `a[b][c]` calls recursively to `getAttribute` returning a lot of data, instead of using native php doing `$context["a"]["b"]["c"];`
I suggest to extend the syntax so it's possible to compile to native php using:
- `[]` to access array variables
- `{}` to access object properties (not sure about this one)
- `->` to access object methods.

This will result in `a[b]{c}->getD` compiling into `$context["a"]["b"]->c->getD()` instead of calling multiple times getAttribute

This should not suppose a problem of backward compatibility, as the use of `[]` is only for arrays, and you could keep calling with `.`

inakivb:
I understand what you mean, but it's a big problem of performance for production environments where you hardly are going to use `strict_variables`

I've done a performance test using `.` and `[]` syntax and there isn't a lot of difference.

Using an average api response array, i've test
Twig . syntax (41.8 ms)
```twig
{% for i in 1..10000 %}
    {% set testvar = a.b.c.d.e %}
{% endfor %}
```
![using_dots](https://user-images.githubusercontent.com/8626495/29559404-50a74104-872f-11e7-9e83-4a50c57f4c09.png)
twig [] syntax (34.3 ms)
```twig
{% for i in 1..10000 %}
    {% set testvar = a[b][c][d][e] %}
{% endfor %}
```
![using_array](https://user-images.githubusercontent.com/8626495/29559403-509c254e-872f-11e7-8419-dca8461da3ec.png)

twig custom function (**1.58 ms** + 2.18 ms in getFunction + 1.06 ms in getCallable)
```twig
{% for i in 1..10000 %}
    {% set testvar = get_array(poiData) %}
{% endfor %}
```
```php
$this->getTwig()->addFunction(new \Twig_SimpleFunction('get_array', function ($array) {
    return $array['b']['c']['d']['e'];
}));
```
![using_custom_function](https://user-images.githubusercontent.com/8626495/29559402-509b8440-872f-11e7-9f24-c72d9c1e54da.png)

empty twig template + php code in controller (209 µs)
[...]

PR #2618, opened 2018-01-30
Optimized compilation for arrays when using a['b'] notation

fabpot:
Under certain conditions, and thanks to the `??` operator, it's possible to optimize array calls (and avoid the call to `twig_get_attribute()`).

On my local machine, code is almost twice as fast, which is nice if you are using a lot of `a["b"]` calls of nested ones.

refs #2547

### DIFF
--- a/lib/Twig/Node/Expression/GetAttr.php
+++ b/lib/Twig/Node/Expression/GetAttr.php
@@ -29,13 +29,16 @@ class Twig_Node_Expression_GetAttr extends Twig_Node_Expression
             && !$this->getAttribute('is_defined_test')
             && Twig_Template::ARRAY_CALL === $this->getAttribute('type')
         ) {
+            $var = '$'.$compiler->getVarName();
             $compiler
-                ->raw('(is_array(')
+                ->raw('(('.$var.' = ')
                 ->subcompile($this->getNode('node'))
+                ->raw(') && is_array(')
+                ->raw($var)
                 ->raw(') || ')
-                ->subcompile($this->getNode('node'))
+                ->raw($var)
                 ->raw(' instanceof ArrayAccess ? (')
-                ->subcompile($this->getNode('node'))
+                ->raw($var)
                 ->raw('[')
                 ->subcompile($this->getNode('attribute'))
                 ->raw('] ?? null) : null)')
--- a/lib/Twig/Node/Expression/GetAttr.php
+++ b/lib/Twig/Node/Expression/GetAttr.php
@@ -23,6 +23,27 @@ class Twig_Node_Expression_GetAttr extends Twig_Node_Expression
 
     public function compile(Twig_Compiler $compiler)
     {
+        // optimize array calls
+        if (
+            (!$compiler->getEnvironment()->isStrictVariables() || $this->getAttribute('ignore_strict_check'))
+            && !$this->getAttribute('is_defined_test')
+            && Twig_Template::ARRAY_CALL === $this->getAttribute('type')
+        ) {
+            $compiler
+                ->raw('(is_array(')
+                ->subcompile($this->getNode('node'))
+                ->raw(') || ')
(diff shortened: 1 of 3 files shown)


lib/Twig/Node/Expression/GetAttr.php
```diff
@@ -23,6 +23,25 @@ public function __construct(Twig_Node_Expression $node, Twig_Node_Expression $at
...
     public function compile(Twig_Compiler $compiler)
     {
+        // optimize array calls
+        if (
+            (!$compiler->getEnvironment()->isStrictVariables() || $this->getAttribute('ignore_strict_check'))
+            && !$this->getAttribute('is_defined_test')
+            && Twig_Template::ARRAY_CALL === $this->getAttribute('type')
+        ) {
+            $compiler
+                ->raw('(is_array(')
```
stof: shouldn't it be `is_array($node) || $node instanceof ArrayAccess` ?

stof (changes requested):

stof:
btw, the fact that Travis is green makes me wonder whether we have enough integration tests for the features implemented by GetAttrNode

lib/Twig/Node/Expression/GetAttr.php
fabpot: indeed, that's even better

jvasseur: Isn't this missing handling of objects implementing `ArrayAccess` ?

fabpot:
@stof I don't understand your comment about tests. We have plenty of tests on this part of Twig. A lot of edge cases are tested as well and we added a lot of them when we worked on the C ext.

stof:
@fabpot but your initial version of the PR was breaking support for ARRAY_CALL on an ArrayAccess object when strict checks are not enabled, and nothing was failing in the CI. So this is not covered by tests.

fabpot:
Indeed, there were no tests for compiled code when strict_variables was false. Tests added.

test/Twig/Tests/Node/Expression/GetAttrTest.php
```diff
@@ -37,7 +37,7 @@ public function getTests()
         $tests[] = array($node, sprintf('%s%s, "bar", array())', $this->getAttributeGetter(), $this->getVariableGetter('foo', 1)));
 
         $node = new Twig_Node_Expression_GetAttr($expr, $attr, $args, Twig_Template::ARRAY_CALL, 1);
-        $tests[] = array($node, sprintf('%s%s, "bar", array(), "array")', $this->getAttributeGetter(), $this->getVariableGetter('foo', 1)));
+        $tests[] = array($node, sprintf('(is_array(%s) || ($context["foo"] ?? null) instanceof ArrayAccess ? (($context["foo"] ?? null)["bar"] ?? null) : null)', $this->getVariableGetter('foo', 1), $this->getVariableGetter('foo', 1)));
```
stof: shouldn't `($context["foo"] ?? null)` use `$this->getVariableGetter` here too ?

stof (changes requested):

test/Twig/Tests/Node/Expression/GetAttrTest.php
fabpot: These abstractions were useful in 1.x, not so much now. I've kept them to ease merging, but no need to keep them for new code.

fabpot:
My first optimization for simple calls made performance worse in case of a deeply nested array.

I've just optimized calls to those deeply nested array calls (like `a["b"]["c"]["d"]`).  Now both are much faster than before.

I've updated the `twig/twig-perf` repo to add a new test for nested array calls so that we can track regressions.

Here are the results:

```
                    |      1.x |      2.x | array-perf-optim |
simple_array_access |    204.7 |    361.7 |             51.1 |
nested_array_access |    491.3 |    782.9 |             80.8 |
```

PR #2621, opened 2018-01-30
Make variable names deterministic

lstrojny:
Variables should be deterministic to support reproducible builds as outlined in https://github.com/symfony/symfony/issues/25958

### DIFF
--- a/lib/Twig/Compiler.php
+++ b/lib/Twig/Compiler.php
@@ -25,6 +25,7 @@ class Twig_Compiler implements Twig_CompilerInterface
     protected $sourceOffset;
     protected $sourceLine;
     protected $filename;
+    private $varNameSalt = 0;
 
     public function __construct(Twig_Environment $env)
     {
@@ -78,6 +79,7 @@ class Twig_Compiler implements Twig_CompilerInterface
         // source code starts at 1 (as we then increment it when we encounter new lines)
         $this->sourceLine = 1;
         $this->indentation = $indentation;
+        $this->varNameSalt = 0;
 
         if ($node instanceof Twig_Node_Module) {
             // to be removed in 2.0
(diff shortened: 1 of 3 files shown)


fabpot:
Could you submit this on 1.x instead?

lib/Twig/Compiler.php
```diff
@@ -233,7 +234,7 @@ public function outdent($step = 1)
 
     public function getVarName()
     {
-        return sprintf('__internal_%s', hash('sha256', uniqid(mt_rand(), true), false));
+        return sprintf('__internal_%s', hash('sha256', __METHOD__ . $this->varCounter++));
```
fabpot: We are not using spaces around dot.

lstrojny:
@fabpot sure

stof:
I'm not sure this makes them deterministic. Is the state reset properly between templates ? Otherwise, it is reproducible only if you compile templates in the same order.

stof:
I have my answer. this is not properly deterministic. Here is what needed to solve it:

- reset the state in `Compiler::compile`
- reset the state in `Parser::parse` near the other state resetting (**not** at the beginning of the method, as it has to save the previous state to reset it)
- use a static name for the profiler visitor. This visitor needs a single variable per compiled method (as blocks and macro nodes are compiled to separate methods). This will avoid the need to manage any state. You just need to make sure there is no conflicts with variable names generated by the parser or the compiler (but using `hash('sha256', __METHOD__)` would be fine then).

lstrojny:
@stof cool, that will indeed make it work across the more narrow use case of initially parsing the templates.

lstrojny:
Ready to be merged from my POV

fabpot:
Thank you @lstrojny.

PR #2624, opened 2018-02-07
Fix possible array to string conversion concealing actual error

finwe:
This is a 1.x branch backport of #2613


fabpot:
For next time, it's better to always submit on the lower branch where the problem occurs. We don't backport fixes.

fabpot:
Thank you @finwe.

PR #2635, opened 2018-03-02
Optimize the performance of the dot operator

### LINKED ISSUE #2326: Performance Query - getAttribute()

fabpot:
This is an alternative to #2359 and #2361, and closes #2326.

Basically, we don't need to call `hasExtension()` at runtime for the sandbox check when calling the dot operator as we know if it's enabled at compile time (the cache hash depends on enables extensions).

My benchmarks shows a *huge* performance improvement (**around 20%**) by doing so (especially when not using the sandbox, which is almost always the case).

### DIFF
--- a/lib/Twig/Environment.php
+++ b/lib/Twig/Environment.php
@@ -681,6 +681,7 @@ class Twig_Environment
     public function setExtensions(array $extensions)
     {
         $this->extensionSet->setExtensions($extensions);
+        $this->updateOptionsHash();
     }
 
     /**
--- a/lib/Twig/Extension/Core.php
+++ b/lib/Twig/Extension/Core.php
@@ -1432,7 +1432,7 @@ function twig_array_batch($items, $size, $fill = null)
  *
  * @internal
  */
-function twig_get_attribute(Twig_Environment $env, Twig_Source $source, $object, $item, array $arguments = array(), $type = Twig_Template::ANY_CALL, $isDefinedTest = false, $ignoreStrictCheck = false)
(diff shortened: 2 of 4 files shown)


lib/Twig/Environment.php
```diff
@@ -681,6 +681,7 @@ public function addExtension(Twig_ExtensionInterface $extension)
     public function setExtensions(array $extensions)
     {
         $this->extensionSet->setExtensions($extensions);
+        $this->updateOptionsHash();
```
fabpot, line 684: This is a big spotted by @mantis during their work on the same topic.

nicolas-grekas (approved):
nice!

gharlan:
Thanks @fabpot!

This is from a heavy twig page in a real project:

screenshot 2018-03-02 17 43 53

mantis:
Hi guys,
Nice to see what the experts (@fabpot) have done compared to my original patch attempt :)
Equally, nice to see to some real performance numbers. @gharlan 
:)

PR #2637, opened 2018-03-02
Fix C ext

fabpot:


.travis.yml
```diff
@@ -52,5 +52,7 @@ matrix:
           env: TWIG_EXT=yes
         - php: 7.1
           env: TWIG_EXT=yes
+        - php: 7.2
+          env: TWIG_EXT=yes
```
stof, line 56: @fabpot I think we should switch the Travis config to *include* relevant ext job instead, to avoid forgetting this each time we add testing for a new PHP 7.x version in Twig 1.x

PR #2639, opened 2018-03-02
Switch the PHP ext jobs to a whitelist rather than a blacklist

stof:
Any new PHP version added in the future will be a PHP 7.x one, not a 5.x one, and so should not have a job for the extension.

the whitelist is already shorter than the blacklist today (3 whitelisted vs 4 blacklisted before, as PHP 5.3 is already handled separately due to precise anyway)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,6 @@ php:
 
 env:
     - TWIG_EXT=no
-    - TWIG_EXT=yes
 
 before_install:
     # turn off XDebug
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -47,12 +46,9 @@ matrix:
         - php: 5.3
           dist: precise
           env: TWIG_EXT=no
```
fabpot, line 48: Could be removed as this is the default value, right?

fabpot:
Thank you @stof.

PR #2641, opened 2018-03-02
Optimize performance of calls to extensions at runtime

fabpot:


fabpot:
Here is an example of when it helps with Symfony (whenever you use `path()` or `url()`):

```php
echo $this->env->getExtension('Symfony\Bridge\Twig\Extension\RoutingExtension')->getPath("foobar", array()));
```

lib/Twig/Template.php
```diff
@@ -38,9 +38,19 @@
...
 
+    /**
+     * @internal
+     */
+    protected $extensions = array();
+
     public function __construct(Twig_Environment $env)
     {
         $this->env = $env;
+        foreach ($env->getExtensions() as $extension) {
```
sstok, line 49: Would it make sense to cache this in the Env instead? Because now it's done for every template load, and it doesn't change between loading right?

fabpot: And it's cleaner to do this logic only once at the right place. See #2649

Issue #2643, opened 2018-03-03
v2.4.5 An exception has been thrown during the rendering of a template ("Object not found"). in favicon.png.base64 (line 1)


crayner:
version 2.4.5 throws an error Twig_Error_Runtime
An exception has been thrown during the rendering of a template ("Object not found").
in favicon.png.base64 (line 1)

Log Dump:
[2018-03-03 12:58:56] php.CRITICAL: Uncaught Exception: An exception has been thrown during the rendering of a template ("Object not found"). {"exception":"[object] (Twig_Error_Runtime(code: 0): An exception has been thrown during the rendering of a template (\"Object not found\"). at /vendor/symfony/twig-bundle/Resources/views/images/favicon.png.base64:1, UnexpectedValueException(code: 0): Object not found at /vendor/symfony/twig-bridge/Extension/ProfilerExtension.php:48, Twig_Error_Runtime(code: 0): An exception has been thrown during the rendering of a template (\"Object not found\"). at /var/www/Busybee/templates/Default/style.html.twig:13, UnexpectedValueException(code: 0): Object not found at /vendor/symfony/twig-bridge/Extension/ProfilerExtension.php:48)"} []

Rolled back to 2.4.4 and no error.

jfcherng:
I have this issue too. And it only occurs when I set 

```
APP_ENV=dev
APP_DEBUG=1
```

in my Symfony 4 `.env` file.

fsaez-ix:
the problem is this update
Updating twig/twig (v2.4.4 => v2.4.5): Downloading (100%)
I installed 2.4.4 again and the problem disappear

derrabus:
`git bisect` tells me that the commit 6ab5fe9b8fd52f810fd902652659605b5fcf39a9 of #2621 has caused this issue. /cc @lstrojny

sylvaindeloux:
Confirmed, had to rollback to 2.4.4 to make it disappear.

abzolv:
V2.4.5 crashes for me as well. Symfony 3.4.5 on a local dev machine.

The error says:

> Twig_Error_Runtime
> An exception has been thrown during the rendering of a template ("Notice: Undefined offset: 0").

Rolling back to 2.4.4 fixes the problem.

smtlab:
Have the same issue with Symfony v3.4.5 and twig v2.4.5.

> (4/4) Twig_Error_Runtime
> An exception has been thrown during the rendering of a template ("Object not found").
[...]

Issue #2627, opened 2018-02-12
Undefined offset: 0


Soviann:
Hello, 

since f2e5f40 I get ``An exception has been thrown during the rendering of a template ("Notice: Undefined offset: 0").`` which makes it impossible to display any template.
The problem comes from the modification made in ``lib/Twig/Profiler/NodeVisitor/Profiler.php``.

I don't have a solution nor an idea as to why it crashes, I hope someone can fix this quickly.

stof:
can you give the stack trace of the exception ?

Soviann:
Found this (I removed the beginning of paths) : 
> request.CRITICAL: Exception thrown when handling an exception (Twig_Error_Runtime: An exception has been thrown during the rendering of a template ("Notice: Undefined offset: 0"). at Resources/views/Exception/exception_full.html.twig line 1) {"exception":"[object] (Twig_Error_Runtime(code: 0): An exception has been thrown during the rendering of a template (\"Notice: Undefined offset: 0\"). at Resources/views/Exception/exception_full.html.twig:1, Symfony\\Component\\Debug\\Exception\\ContextErrorException(code: 0): Notice: Undefined offset: 0 at vendor/twig/twig/lib/Twig/Extension/Profiler.php:23)"} []

Line 1 of the exception_full.html.twig is the ``{% extends 'TwigBundle::layout.html.twig' %}``

Is that what you needed ?

stof:
@fabpot looks like it is possible to put the Profiler extension in an invalid state by having 0 actives profiles in some cases

Vincz:
I ran into the same issue today after an update.  
Templating doesn't work anymore in dev. 
Tried to debug the enter / leave methods of the Profiler, didn't understand why but at some point it gets messy and as @stof said, the `$actives` property is empty.

Soviann:
Any news on this problem ? This is quite troublesome.

nicolas-grekas:
@NVasse can you provide a reproducer? or even better, debug the issue? That would help a lot.

Soviann:
Forgot to mention that I use the 1.x branch, I don't know if the problem occurs on the 2.x

@nicolas-grekas You only need to update 1.x branch to the latest and work on dev environment, nothing special. I'll try to take a look but I don't have any free time and I'm not even sure I could debug the problem. This would be quicker if someone familiar with Twig code debugs it.

xabbuh:
[...]

PR #2645, opened 2018-03-03
Include extensionName in the sha256 hash

PhilETaylor:
Proposed fix for #2627 and #2643 (same issue, same problem) implementing suggestion from @nieuwenhuisen at https://github.com/twigphp/Twig/issues/2627#issuecomment-370140438 which seems to resolve the issue

This issue was introduced with  6ab5fe9 of #2621 

After applying this PR for testing please remember to clear caches. 

The underlying issue is that Symfony 3.4.4 demands [Twig 2.4.4 in its composer.json](https://github.com/symfony/symfony/blob/master/composer.json#L23) but some developers have twig/twig in their own composer.json, unbounded to an exact version number, (or [symfony/twig-bundle which has a requirement of just "^2.4"](https://github.com/symfony/twig-bundle/blob/master/composer.json#L24)) and so when they composer upgrade they are getting Twig 2.4.5 which is not what Symfony itself wants or promotes. 

I guess one could say that this is not a Symfony bug currently, as Symfony 3.4.4 is EXPECTING you to use Twig 2.4.4...

### DIFF
--- a/lib/Twig/Profiler/NodeVisitor/Profiler.php
+++ b/lib/Twig/Profiler/NodeVisitor/Profiler.php
@@ -53,7 +53,7 @@ final class Twig_Profiler_NodeVisitor_Profiler extends Twig_BaseNodeVisitor
 
     private function getVarName()
     {
-        return sprintf('__internal_%s', hash('sha256', __METHOD__));
+        return sprintf('__internal_%s', hash('sha256', __METHOD__.$this->extensionName));
     }
 
     public function getPriority()


abzolv:
Doesn't the composer directive "twig/twig": "^1.35|^2.4.4", https://github.com/symfony/symfony/blob/master/composer.json#L23, also install v2.4.5?

fabpot:
Thanks @PhilETaylor for working on a fix. I've done it in a slightly different way in #2647 (as `__METHOD__` never changes) and I've also patched 1.x which has the issue as well.

PhilETaylor:
I bow out - this was a "fix on the sofa" - happy to let @fabpot  #2647 be the solution :)

PR #2647, opened 2018-03-03
Fix profiler unicity when several ones are registered

### LINKED ISSUE #2627: Undefined offset: 0

### LINKED ISSUE #2643: v2.4.5 An exception has been thrown during the rendering of a template ("Object not found"). in favicon.png.base64 (line 1)

fabpot:
Closes #2627, closes #2643, alternative to #2645

What happens is that both Twig and Symfony register a Twig profiler. Unfortunately, the current code does assign the same variable for both profiler, which means that trying to close it twice does not work.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,6 @@
 * 1.35.2 (2018-XX-XX)
 
- * n/a
+ * fixed a regression in the way the profiler is registered in templates
 
 * 1.35.1 (2018-03-02)
 
--- a/lib/Twig/Profiler/NodeVisitor/Profiler.php
+++ b/lib/Twig/Profiler/NodeVisitor/Profiler.php
@@ -55,7 +55,7 @@ class Twig_Profiler_NodeVisitor_Profiler extends Twig_BaseNodeVisitor
 
     private function getVarName()
     {
-        return sprintf('__internal_%s', hash('sha256', __METHOD__));
+        return sprintf('__internal_%s', hash('sha256', $this->extensionName));
     }
 
     public function getPriority()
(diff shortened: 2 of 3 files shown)


nibsirahsieu:
I can confirm this fix the issue. Thanks

PR #2649, opened 2018-03-04
Make a small performance improvement

fabpot:
And it also fixes some potential subtle bugs.


lib/Twig/ExtensionSet.php
```diff
@@ -160,6 +160,8 @@ public function addExtension(Twig_ExtensionInterface $extension)
             throw new LogicException(sprintf('Unable to register extension "%s" as it is already registered.', $class));
         }
 
+        // For BC/FC with namespaced aliases
+        $class = (new ReflectionClass($class))->name;
```
fabpot, line 164: This always returns the class without the leading `\\`, so no need to trim.

PR #2656, opened 2018-03-09
Silence errors coming from OpCache

ghost:
I noticed an error being thrown by OpCache:

```
Zend OPcache API is restricted by "restrict_api" configuration directive
```

This PR temporarily silences it until a solution can be found.

### DIFF
--- a/lib/Twig/Cache/Filesystem.php
+++ b/lib/Twig/Cache/Filesystem.php
@@ -66,7 +66,7 @@ class Twig_Cache_Filesystem implements Twig_CacheInterface
             if (self::FORCE_BYTECODE_INVALIDATION == ($this->options & self::FORCE_BYTECODE_INVALIDATION)) {
                 // Compile cached file into bytecode cache
                 if (function_exists('opcache_invalidate')) {
-                    opcache_invalidate($key, true);
+                    @opcache_invalidate($key, true);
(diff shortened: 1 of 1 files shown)


ghost:
It seems like this issue is only present on ServerPilot installations, I will contact them and report back.

ghost:
I have looked into it more - it is being thrown because `opcache.restrict_api`'s value is 1.

From [php.net](http://php.net/manual/en/opcache.configuration.php#ini.opcache.restrict-api), `opcache.restrict_api`...
> Allows calling OPcache API functions only from PHP scripts which path is started from specified string. The default "" means no restriction.

What would be a suitable permanent fix for this?

ghost:
I have discussed this issue with a ServerPilot support technician.
The conditional here is correctly checking if the function exists, but it should probably also be checking the value of opcache.restrict_api before calling opcache_invalidate().
If changing the code isn’t an option, the php.ini file can be changed to turn this feature on.

Tobion:
👎 for this change. You want to know when things don't work as you expect. Just surpressing it is not a solution. And it's also not twigs responsibility to check opcache.restrict_api. Either change your opcache.restrict_api or change the options in the constructor.

ghost:
@Tobion okay, if you're sure.

PR #2659, opened 2018-03-14
[v1] Fixed counting children of SimpleXMLElement object

michalbundyra:
It works correctly in v1.32 and since v1.33 it is not working correctly.
See: https://github.com/twigphp/Twig/compare/v1.32.0...v1.33.0#diff-d7378002f67a61c458c1de0468eef74fR1265

Test code:
```php
$xml = new SimpleXMLElement('<?xml version="1.0" encoding="UTF-8"?><doc><elem/><elem/></doc>');

var_dump($xml instanceof \Countable);
echo 'Count: ' . count($xml), PHP_EOL;
echo 'Strlen: ' . strlen($xml), PHP_EOL;
echo 'MB Strlen: ' . mb_strlen($xml), PHP_EOL;
```

Results:
```
bool(false)
Count: 2
Strlen: 0
MB Strlen: 0
```

https://3v4l.org/VvUXG

### DIFF
--- a/lib/Twig/Extension/Core.php
+++ b/lib/Twig/Extension/Core.php
@@ -1263,6 +1263,10 @@ if (function_exists('mb_get_info')) {
             return mb_strlen($thing, $env->getCharset());
         }
 
+        if ($thing instanceof \SimpleXMLElement) {
+            return count($thing);
+        }
+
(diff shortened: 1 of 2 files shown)


fabpot:
Thank you @webimpress.

Issue #2623, opened 2018-02-05
Autoescape does not escape javascript in a href


twoelevennorth:
I stumbled upon this, and I'm not sure if it's a bug or not.

It seems that Twig will not autoescape a "javascript:" URI in a href attribute. It _will_ escape if you explicitly tell it to escape js, however.

I have only tested this from Symfony, but I assume that shouldn't matter. Here's an example:

From a controller:
```
return $this->render('default/index.html.twig', [
    'foo' => 'javascript:alert("hello")',
]);
```

The view:
```
<a href="{{ foo }}">foo</a>
```

This will _not_ be escaped, and clicking the link will open an alert.

This does not work either:
```
{% autoescape %}
<a href="{{ foo }}">foo</a>
{% endautoescape %}
```

However, these _will_ escape properly:
```
<a href="{{ foo|e('js') }}">foo</a>
```

```
{% autoescape 'js' %}
<a href="{{ foo }}">foo</a>
{% endautoescape %}
```

umulmrum:
I stumbled across the same problem and am thankful for a solution.

Using "e('js')" will break legitimate/non-JS URLs, so this can't be used in my context. What is the best solution here? I could imagine that it's hard to solve in a transparent way, so introduce another escaper for URL-but-not-JS? Or an argument for html_attr to allow/disallow JS?

stof:
I think the `html_attr` escaper might escape things enough to cover this case (and without breaking things like the `js` escaper, which is not meant to be used in a HTML context at all)

umulmrum:
Please excuse my impatience, but I think it needs to be made clear that this is a security issue which opens space for XSS attacks. Supposedly a lot of developers might not be aware of this, as they assume that Twig handles output escaping completely, so a rather quick solution would be very appreciated.

I cannot see what the best solution is (don't know if people rely on JS NOT being escaped by html_attr so that simply adding escaping will break their use case), but can try an implementation if someone tells me what to do.

stof:
@umulmrum the default autoescaping strategy is `html`, not `html_attr`.

And escaping things entirely all the time would require doing contextual escaping (i.e. changing the escaping based on where the output is performed), which is not something done by Twig (it would be really hard to do that while keeping Twig agnostic of the output format)

umulmrum:
[...]

PR #2661, opened 2018-03-14
Add warning for possible XSS vulnerabilities when using html_attr esc…

umulmrum:
…aping.

Doc addition that handles #2623.
I think that many developers might miss this, which then leads to security issues.

(I'd still be happy to try a PR to actually handle the case instead of simply documenting it, but need a direction in which to go - see discussion in the ticket)


fabpot:
I don't want to add more specific information about escaping as this out of the scope of Twig, which only provide tools to help escape elements (see #2950 for a note I've just added).

PR #2668, opened 2018-03-26
Small optimization for Twig_NodeTraverser::traverseForVisitor

### LINKED ISSUE #2665: Perfomance of Twig_NodeTraverser::traverseForVisitor
I tried to find out why `cache:clear` command takes so much time. I created a project and then execuded a clear cache command with xdebug.profiler:
```
composer create-project symfony/skeleton . ^3.4
composer require sonata-project/doctrine-orm-admin-bundle doctrine/doctrine-bundle
php -d xdebug.profiler_enable=1 -d xdebug.profiler_output_dir=. bin\console cache:clear
```

I found, that [Twig_NodeTraverser::traverseForVisitor](https://github.com/twigphp/Twig/blob/2.x/lib/Twig/NodeTraverser.php#L62) was called 242594 times, and it took 15.5% of all execution time (on Windows, PHP 7.1)

Twig_NodeTraverser::traverseForVisitor was called 242594 times

Here is method code:

```php
private function traverseForVisitor(Twig_NodeVisitorInterface $visitor, Twig_Node $node)
{
    $node = $visitor->enterNode($node, $this->env);
    foreach ($node as $k => $n) {
        if (false !== $n = $this->traverseForVisitor($visitor, $n)) {
            $node->setNode($k, $n);
        } else {
            $node->removeNode($k);
        }
    }
    return $visitor->leaveNode($node, $this->env);
}
```

I found, that `traverseForVisitor` never returns `false` actualy, it always returns `Twig_Node`! So i guess main loop could be replaced with this:

```php
    foreach ($node as $k => $n) {
        $node->setNode($k, $this->traverseForVisitor($visitor, $n));
    }
```
Am i right? Or i missed something?
Even if i am wrong, this method should be optimized, 15% is to much =)

covex-nn:
During traversing a node tree `Twig_NodeTraverser::traverseForVisitor` (`Twig_NodeVisitorInterface::leaveNode`) often returns a same child node (a same object). So, there is no need to set it back to its parent.

This PR adds a check if a child node was changed. With this check a lot of calls to `Twig_Node::setNode` will be skipped.

Closes #2665

### DIFF
--- a/lib/Twig/NodeTraverser.php
+++ b/lib/Twig/NodeTraverser.php
@@ -70,8 +70,10 @@ class Twig_NodeTraverser
         $node = $visitor->enterNode($node, $this->env);
 
         foreach ($node as $k => $n) {
-            if (false !== $n = $this->traverseForVisitor($visitor, $n)) {
-                $node->setNode($k, $n);
+            if (false !== $m = $this->traverseForVisitor($visitor, $n)) {
+                if ($m !== $n) {
+                    $node->setNode($k, $m);
+                }
             } else {
                 $node->removeNode($k);
             }


fabpot:
Thank you @covex-nn.

PR #2670, opened 2018-03-31
Fix partial template caching after update when auto_reload is on

ay-git:
**Description of the bug:**
This bug manifests itself on production servers.
And with a high load the probability of getting on this bug tends to 100%
The bug results in the caching of a partial template and the failure of a part of the site working with this template. Can only be corrected by deleting the cache.

**What's happening:**
If in the process of uploading the template file, a render() is called, the partially uploaded file gets into the cache and after the template file is fully uploaded, the cache is not updated anymore.

**How to demonstrate:**
To demonstrate the bug, you can create 2 files:
File simulating the download of a file to the server (twig_test_write.php):

```
$classTwig = new Twig_Environment( new Twig_Loader_Filesystem(__DIR__."/../../site_templates/"), array(
	'cache' => __DIR__.'/../../site_templates_cache',
	'auto_reload'=>true
));

file_put_contents(__DIR__."/../../site_templates/"."test.twig","\n<br>Template start write to filesystem");

echo $classTwig->render("test.twig", array());

file_put_contents(__DIR__."/../../site_templates/"."test.twig"," - Template end write to filesystem",FILE_APPEND);

```

File for template rendering (twig_test_read.php):
```
$classTwig = new Twig_Environment( new Twig_Loader_Filesystem(__DIR__."/../../site_templates/"), array(
	'cache' => __DIR__.'/../../site_templates_cache',
	'auto_reload'=>true
));

echo $classTwig->render("test.twig", array());
```

Result executeing file twig_test_write.php:
```
Template start write to filesystem 
```

Result executeing file twig_test_read.php:
```
Template start write to filesystem 
```

**After this fix:**
Result executeing file twig_test_read.php:
```
Template start write to filesystem - Template end write to filesystem
```

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,6 @@
 * 1.35.4 (2018-XX-XX)
 
- * n/a
+ * "js" filter now produces valid JSON
 
 * 1.35.3 (2018-03-20)
 
--- a/lib/Twig/Extension/Core.php
+++ b/lib/Twig/Extension/Core.php
@@ -1040,7 +1040,7 @@ function twig_escape_filter(Twig_Environment $env, $string, $strategy = 'html', 
 
         case 'js':
             // escape all non-alphanumeric characters
-            // into their \xHH or \uHHHH representations
+            // into their \x or \uHHHH representations
             if ('UTF-8' !== $charset) {
                 $string = twig_convert_encoding($string, 'UTF-8', $charset);
             }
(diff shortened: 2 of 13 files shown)


stof:
A template cannot be reloaded inside the same PHP process (as the compiled template is already loaded in memory). So your use case where you render the template, then edit the file and then render it again will *never* work, even without any load.

For the `twig_test_read.php` case, this indeed comes from the precision of the mtime (the template is modified too soon after the cache generation so the timestamp is the same)

ay-git:
> So your use case where you render the template, then edit the file and then render it again will never work, even without any load.

I understand this is not a bug, it's a feature (removed this part from the code)

A bug occurs just in twig_test_read.php and to fix it I propose this patch.

>For the twig_test_read.php case, this indeed comes from the precision of the mtime (the template is modified too soon after the cache generation so the timestamp is the same)

I understand why this happens. And this patch corrected this error.

fabpot:
Thank you @ay-git.

PR #2674, opened 2018-04-09
update filesystem.php to fix OPcache warnings... causes crash in GRAV…

MattAppleton:
Currently php 7.2 / twig / GRAV  throws an error and crashes with this message:

> An exception has been thrown during the rendering of a template (“Zend OPcache API is restricted by “restrict_api” configuration directive”).

this is fixed by adding the @ prefix on line 71 of this file... suggest it would be a good idea to roll into next version?

### DIFF
--- a/lib/Twig/Cache/Filesystem.php
+++ b/lib/Twig/Cache/Filesystem.php
@@ -66,7 +66,7 @@ class Twig_Cache_Filesystem implements Twig_CacheInterface
             if (self::FORCE_BYTECODE_INVALIDATION == ($this->options & self::FORCE_BYTECODE_INVALIDATION)) {
                 // Compile cached file into bytecode cache
                 if (function_exists('opcache_invalidate')) {
-                    opcache_invalidate($key, true);
+                    @opcache_invalidate($key, true);
                 } elseif (function_exists('apc_compile_file')) {
                     apc_compile_file($key);
                 }


Tobion:
Duplicate of #2656

stof:
silencing this error has been rejected already.

forcing bytecode invalidation is a configurable behavior in Twig. If your system does not allow you to invalidate, you should disable the feature instead of making it failing silently (if your system expects invalidation to work, the silent failure will still break things)

stof:
thus, this feature is disabled by default in Twig, so having it enabled means you already have some code deciding it needs it.

MattAppleton:
fair enough, will take this up with Serverpilot :)

fabpot:
Closing then.

MattAppleton:
If anyone interested, this fixes the issue on Serverpilot which is caused by a configuration in php.ini for opcache: http://php.net/manual/en/opcache.configuration.php


(passing this on from the very helpful @rhukster)

To resolve, simply edit this setting in your php.ini (or relevant configuration file) for opcache settings. In my case I had an /etc/php7.2-sp/conf.d/opcache.ini file that contained:

opcache.restrict_api=1
I just changed this to:

opcache.restrict_api=
and restarted PHP FPM with: sudo service php7.2-fpm-sp restart

PR #2676, opened 2018-04-16
Fix links to repository in documentation

julienfalque:


fabpot:
Good catch, thanks @julienfalque.

Issue #2677, opened 2018-04-24
Documentation for `if` has innacurate/incomplete edge cases


ismael-miguel:
Currently, there's the following list of edge cases, on https://twig.symfony.com/doc/2.x/tags/if.html

![edge cases](https://user-images.githubusercontent.com/10207857/39183829-3e420954-47b9-11e8-8dad-e395398bf177.png)

This is innacurate/incomplete, as these edge cases have edge cases.

These are my findings:
- empty string: `false`
- numeric zero: `false`
- whitespace-only string: `false`?
    Running `echo !!'          ';` shows 1
    Running `echo '          ' == 0;` shows 1 
    Running `echo '          ' == '0';` doesn't display anything
    The whitespace-only string (or any string) will be casted to a number when compared for equality with a number. Otherwise, everything is casted to a string.
- empty array: `false`
    Everything is casted to an array when comparing to an empty array, except boolean values. In that case, the empty array is casted to the boolean value (`false`)
- null: `false`
    Null is a special value. Null will cast to a string when comparing with a string, to a number when comparing to a number **but** won't be cast to an empty array when comparing to an empty array.
    My guess is due to the fact that you can write `function xyz(array $arg = null){ [...] }`, to distinguish between an empty array and a function with an optional parameter.

Everything below is correct.

You can see it in action on http://sandbox.onlinephpfunctions.com/code/e4dc15a4f29cd6ef6a7498ad27ac202f5f874612

Should this information be added to the documentation? Is it fine the way it is?

xabbuh:
You are mixing two things here. The table is not about comparing the values with other arbitrary values, but how they will be evaluated in a context where a boolean is expected.

So basically in a construct like this:

```twig
{% if value %}
    value evaluates to true
{% else %}
    value evaluates to false
{% endif %}
```

stof:
your sandbox code is not covering the behavior of the boolean evaluation, but of the `==` operator, which is a totally different things (and is not what the `{% if %}` documentation needs)

ismael-miguel:
Then there are some edge cases missing.

The string `'0'` is a falsy value in PHP, and isn't mentioned in the table.
Also, `NAN` and `INF` are truthy, which may happen from some mathematical calculations.

xabbuh:
Would you like to contribute the missing edge cases? :) The file to edit would be https://github.com/twigphp/Twig/blob/1.x/doc/tags/if.rst.

PR #2678, opened 2018-04-25
Added missing edge-cases

ismael-miguel:
Discussion on: https://github.com/twigphp/Twig/issues/2677

### DIFF
--- a/doc/tags/if.rst
+++ b/doc/tags/if.rst
@@ -68,7 +68,10 @@ use more complex ``expressions`` there too:
     ====================== ====================
     empty string           false
     numeric zero           false
+    NAN (Not A Number)     true
+    INF (Infinity)         true
     whitespace-only string true
+    string "0" or '0'      false
     empty array            false
     null                   false
     non-empty array        true


doc/tags/if.rst
```diff
@@ -68,7 +68,10 @@ use more complex ``expressions`` there too:
     ====================== ====================
     empty string           false
     numeric zero           false
+    NAN (Not A Number)     true
+    INF (Infinity)         true
     whitespace-only string true
+    string "0" or '0'      false
```
stof, line 74: that's the same string. No need for this `or`

ismael-miguel: PHP's single-quoted and double-quoted strings behave differently.

Doing `"\t"` will show a tab while `'\t'` will show `\t`.
But in this case, yes, it's the same thing.
Just over-clarification.

stof: but that's about the *syntax* of the language. Once you have a string, there is no difference about how it was initially quoted. And for `0`, there is no difference

ismael-miguel: Quoting my previous comment:
```
But in this case, yes, it's the same thing.
Just over-clarification.
```

fabpot:
Thank you @ismael-miguel.

PR #2685, opened 2018-05-18
Deprecate using the spaceless tag at the root level of a child template (noop anyway)

fabpot:
~~WIP as I'd like to trigger a deprecation notice (which will only be possible with some other changes coming up in another PR) and add some tests for the deprecated behavior.~~

### DIFF
--- a/doc/deprecated.rst
+++ b/doc/deprecated.rst
@@ -13,6 +13,13 @@ Inheritance
   ``Twig_Error_Syntax`` exception. It does not work anyway, so most projects
   won't need to do anything to upgrade.
 
+Tags
+----
+
+* Using the ``spaceless`` tag at the root level of a child template is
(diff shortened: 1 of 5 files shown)


doc/deprecated.rst
```diff
@@ -5,6 +5,13 @@ This document lists deprecated features in Twig 2.x. Deprecated features are
 kept for backward compatibility and removed in the next major release (a
 feature that was deprecated in Twig 2.x is removed in Twig 3.0).
 
+Tags
+----
+
+* Using the ``spaceless`` tag at the root level of a child template is depreated
```
stof: typo: `deprecated`

fabpot:
Needs #2687 first

fabpot:
This one is ready now.

nicolas-grekas (approved):

PR #2686, opened 2018-05-18
Move legacy tests that were not executed anymore in 2.x

fabpot:
While working on some new deprecations, I noticed that some fixtures were in a directory that was not used anymore in 2.x. I've moved the relevant one back to the main Fixtures directory and remove the other one, which is not relevant anymore.


stof:
shouldn't `.legacy` be removed from the file name ?

fabpot:
Indeed, renamed.

Issue #2393, opened 2017-02-13
conditionally define a block in twig 2


matthbull:
hi all,

**twig 2.1**<br>
**symfony3**

Ive asked this on [SO](http://stackoverflow.com/questions/42186007/conditionally-define-a-block-in-twig-2?noredirect=1#comment71535829_42186007), but the only answer I have sort of confirms my suspicions, but I'd just like to check with you guys if this is actually the case.

I want to conditionally define a block in a twig template, as in the example below:

** called template **
```
{% extends 'base.html.twig' %}
{% set something = true %}
{% if not something %}
    {% block left_sidebar %}
        <h2>sidebar rendered</h2>
    {% endblock %}
{% endif %}
```

** base.html.twig **
```
{% if block('left_sidebar') is defined %}
   {{ block('left_sidebar')|raw }}
{% else %}
    no block
{% endif %}
```

I have a use case where I dont actually want to define the block if its not needed (dynamic regions in the base template).

The problem is, the block always renders, regardless of a true/false result of the conditional check.  But this only seems to happen if the template is extending another.  Is this because of compile order? Or am I missing something.

Heres a [fiddle](https://twigfiddle.com/k5elut) to illustrate.

thanks in advance.

Matt

fabpot:
Think of a block as the equivalent of a method in a class. A block cannot be dynamic.

matthbull:
Interstingly though, if you look at the compiled output from twigfiddle, its resulted in this.

```
    protected function doDisplay(array $context, array $blocks = array())
    {
        // line 3
        $context["something"] = null;
        // line 5
        if ((($context["something"] ?? null) != null)) {
        }
        // line 1
        $this->parent->display($context, array_merge($this->blocks, $blocks));
    }
```

note that // line 1 has compiled outside of the line 5 if statement?


Ive also noticed that SonataAdminBundle uses the same syntax, so maybe theres a general lack of understanding in the community?

fabpot:
I think I will change Twig to throw an exception in such circumstances as it cannot work and people keep having this issue. The fact that it "fails" silently does not help. Any link to an example on SonataAdminBuindle?

javiereguiluz:
Given that in the base template you don't make anything with the dynamic block, you can do the following:

1. Define a `main` block in `base.html.twig` to hold the main contents of the page, including the optional left sidebar:

```twig
{% block main %}
{% endif %}
```

2. In the child templates, define the `left_sidebar` when needed:

```twig
[...]

Issue #1685, opened 2015-05-06
Compile Error when block tag inside another tag without whitespace


philer:
I tried removing as much whitespace as possible from some templates (via regex) and found some weird behaviour which I believe to be a bug.

As far as I can tell, if a template extends another one and defines a block inside another tag (e.g. an if), and there is no whitespace present, it'll throw an error.

**Example:**
`foo.twig` may be completely empty, it doesn't seem to matter
`foobar.twig` looks like this:
```twig
{% extends 'foo.twig' %}

{% if true %}
{% block foobar 'asdf' %}
{% endif %}
```
Note that there is no whitespace between the `if`, `block` and `endif` tags except for those two newlines. If I add *any* whitespace in between those tags, e.g. a space character or another newline, this error does *not* occur. This is why I call it a bug.
Also, if I remove those two newlines so I get everything in one line, I still get the error.
Error: **`An exception has been thrown during the compilation of a template ("Node "1" does not exist for Node "Twig_Node".") in "path/to/foobar.twig". `**

Using a different tag yields a slightly different error, for example `spaceless`:
```twig
{% extends 'foo.twig' %}

{% spaceless %}
{% block foobar 'asdf' %}
{% endspaceless %}
```
Error: **`An exception has been thrown during the compilation of a template ("Node "body" does not exist for Node "Twig_Node_Spaceless".") in "path/to/foobar.twig".`**

I haven't tried any other tags since very few make sense in this context (probably only `if`).

Tested on Twig v1.18.1 in a Laravel v4.2.17 project via @rcrowe's TwigBridge v0.6.1
Let me know if you need anything else.

reFX-Mike:
Verified: http://twigfiddle.com/6wydib

fabpot:
You cannot us a block tag inside an if statement or any other tag for that matter.

reFX-Mike:
But when having whites space around it, it works just fine! So why this artificial limitation and why does it only occur if there is no white space?

fabpot:
There is no artificial limitation, the code you reference is not valid; a block cannot be wrapped in another tag. We should probably throw an exception in this case though.

philer:
But it works just fine, why would you want to prohibit it?
Here's something I've been using, which works just as you'd expect (except for the problem described above):
```Twig
{% if user.staffMember %}
    {% block init_script %}
        {{ parent() }}
        js.doSomething();
    {% endblock %}
{% endif %}
```

SpacePossum:
[...]

Issue #1639, opened 2015-02-27
[DX] Throw exception when using conditionals outside of blocks in a template extending another


SpacePossum:
When using a setup like:

```twig
{# main.twig #}
{% set test = false %}

<h1>Hello</h1>

{% block dialog %}
    dialog base
{% endblock %}
```
```twig
{# test.twig #}

{% extends 'main.twig' %}

{% if false and test %}

    {% block dialog %}
    
        dialog override
        
        {% if test == true %}
            test is true
        {% else %}
            test is false
        {% endif %}
        
    {% endblock %}

{% endif %}
```

The output is:

```twig


<h1>Hello</h1>

    dialog override

        test is false
```

This is confusing. When printing anything in the extending template outside of the `block` a syntax exception is thrown like: `A template that extends another one cannot have a body in "test.twig" at line 19.`. I think an exception should be thrown as well when using conditionals outside of `blocks`.
Fiddle @ http://twigfiddle.com/818v58

ninsuo:
Hello,

In some cases, that's useful to put conditions outside blocks, for example, to import macros conditionally:

```jinja
{% if isMobile %}
    {% import 'mobileMacros.twig' as macros %}
{% else %}
    {% import 'desktopMacros.twig' as macros %}
{% endif %}
```

This will compile as:

```php
if ((isset($context["isMobile"]) ? $context["isMobile"] : null)) {
    // line 5
    $context["macros"] = $this->env->loadTemplate("mobileMacros.twig");
} else {
    // line 7
    $context["macros"] = $this->env->loadTemplate("desktopMacros.twig");
}
```

Then, you can use your `macros` in all blocks without importing macros in each blocks.

SpacePossum:
Hello @ninsuo,

Thank you for making TwigFiddle! It is really a great tool and helped me a lot :+1:
And thank you for you reply. Your example shows a valid case against my suggestion to always throw an exception. I didn't consider all options and jumped to conlusions,
Looking at the compiled source you added I wonder why a similar setup doesn't work for `blocks` as it does for `macros`. I'm puzzled about the way and when conditionals are checked in Twig. For example this doesn't run as well:

```twig
{% set test = false %}
{% if test %}
    {{ 'test'|myDebugFunction }}
{% else %}
    no test
{% endif %}
```

I'll try if I can find more and get back to this issue

ninsuo:
[...]

Issue #988, opened 2013-02-04
if statement outside block always evaluating to true


MisterGlass:
I had a page where I only needed extra JS/CSS files under certain circumstances, so I tried to wrap them in an if like so:

```
{% if extras %}
    {# these get output when they shouldnt #}
    {% block stylesheets %}
        <link rel="stylesheet" href="/css/extra.css" type="text/css">
    {% endblock %}
    {% block javascripts %}
        <script type="text/javascript" src="http://ajax.googleapis.com/ajax/libs/jquery/1.4.2
    {% endblock %}
{% endif %}
```

When I did this, the blocks were always output (instead of only when extras=true).

Someone on IRC said that conditionals outside of blocks are invalid for TWIG, so I have adjusted my template, but I am filing this ticket because the always true behavior is incorrect. The condition should either work properly or trigger a TWIG error.

PR #2687, opened 2018-05-18
Deprecate the possibility to define a block in a non-capturing block from a child template

### LINKED ISSUE #351: {% if true%}{% block ... %}{%endif%}
Hey,

i posted this issue on the mailingliste, but I didn't get any response. I am not sure wether I understand all ideas behind twig but I think this is a bug:

I tried the following and it doesn't work the way I wanted it to work:
{% if type == 0 %}
    {% block menuHandbooks %}currentmenu{% endblock %}
{% endif %}
{% if type == 1 %}
    {% block menuDocuments %}currentmenu{% endblock %}
{% endif %}
{% if type == 2 %}
    {% block menuBetProgram %}currentmenu{% endblock %}
{% endif %}

Is it intended, that all conditions are true? I had a template which was inherited {% extends '::admin.html.twig' %}
and with the above example, the template gets his own body.

The idea is to mark the current page in the menu.

I rebuild it this way and now it works:
{% block menuHandbooks %}
    {% if type == 0 %}currentmenu{% endif %}
{% endblock %}

{% block menuDocuments %}
    {% if type == 1 %}currentmenu{% endif %}
{% endblock %}

{% block menuBetProgram %}
{% if type == 2 %}currentmenu{% endif %}
{% endblock %}

Best regards
Fabian

P.S. Part from the admin.html.twig:
<li class="menu"><a class="{% block menuNews %}{% endblock %}" href="{{path("news_admin")}}">Nachrichten</a></li>
<li class="menu"><a class="{% block menuDocuments %}{% endblock %}" href="{{path("document_admin")}}">Dokumente</a></li>
<li class="menu"><a class="{% block menuHandbooks %}{% endblock %}" href="{{path("handbook_admin")}}">Handbücher</a></li>
<li class="menu"><a class="{% block menuVideos %}{% endblock %}" href="{{path("videos_admin")}}">Lernvideos</a></li>

fabpot:
fixes #351, #541, #703, #988, #1639, #1685, #2393

One recurring issue (see the probably non-exhaustive list of issues references above) is a misunderstanding of how blocks work in a child template.

For instance, there is no way to define a `block` conditionally:

```twig
{% extends "layout" %}

{% if ... %}
    {% block content %}
        ...
    {% endblock %}
{% endif %}
```

And wrapping a `block` definition with a `spaceless` tag does not work:

```twig
{% extends "layout" %}

{% spaceless %}
    {% block content %}
        ...
    {% endblock %}
{% endspaceless %}
```

The issue is that the above templates compile just fine, but not in a way people expect. Basically, the condition would be empty.

Interestingly enough, wrapping a `block` definition might be useful/work as expected when used with "capturing" tags like `set`:

```twig
{% extends "layout" %}

{% set content %}
    {% block content %}
        ...
    {% endblock %}
{% endset %}
```

Not sure if that's useful, but here, the `block` tag will serve as the blog definition AND it will also be displayed in place. I would not recommend such usages.

This pull request deprecates the cases where the compiled template would never reflect the developer intent, but keeps the "working" scenarii as is.

I've been trying to fix that issue for **years**. Very happy to have found a simple solution without breaking anything that worked fine.

### DIFF
--- a/lib/Twig/Parser.php
+++ b/lib/Twig/Parser.php
@@ -313,7 +313,7 @@ class Twig_Parser
         return $this->stream->getCurrent();
     }
 
-    private function filterBodyNodes(Twig_Node $node)
+    private function filterBodyNodes(Twig_Node $node, $nested = false)
     {
         // check that the body does not contain non-empty output nodes
         if (
@@ -328,17 +328,30 @@ class Twig_Parser
             throw new Twig_Error_Syntax('A template that extends another one cannot include content outside Twig blocks. Did you forget to put the content inside a {% block %} tag?', $node->getTemplateLine(), $this->stream->getSourceContext());
         }
 
-        // bypass nodes that will "capture" the output
+        // bypass nodes that "capture" the output
         if ($node instanceof Twig_NodeCaptureInterface) {
+            // a "block" tag in such a node will serve as a block definition AND be displayed in place as well
             return $node;
+        }
+
+        // "block" tags that are not captured (see above) are only used for defining
+        // the content of the block. In such a case, nesting it does not work as
+        // expected as the definition is not part of the default template code flow.
+        if ($nested && $node instanceof Twig_Node_BlockReference) {
+            //throw new Twig_Error_Syntax('A block definition cannot be nested under non-capturing nodes.', $node->getTemplateLine(), $this->stream->getSourceContext());
+            @trigger_error(sprintf('Nesting a block definition under a non-capturing node in "%s" at line %d is deprecated since version 2.5.0 and will become a syntax error in 3.0.', $this->stream->getSourceContext()->getName(), $node->getTemplateLine()), E_USER_DEPRECATED);
+            return;
         }
 
         if ($node instanceof Twig_NodeOutputInterface) {
             return;
         }
 
+        // here, $nested means "being at the root level of a child template"
+        // we need to discard the wrapping "Twig_Node" for the "body" node
+        $nested = $nested || get_class($node) !== 'Twig_Node';
         foreach ($node as $k => $n) {
-            if (null !== $n && null === $this->filterBodyNodes($n)) {
+            if (null !== $n && null === $this->filterBodyNodes($n, $nested)) {
                 $node->removeNode($k);
             }
         }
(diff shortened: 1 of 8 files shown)


lib/Twig/Parser.php
```diff
@@ -328,17 +328,30 @@ private function filterBodyNodes(Twig_Node $node)
...
+            // a "block" tag in such a node will serve as a block definition AND be displayed in place as well
             return $node;
         }
 
+        // "block" tags that are not captured (see above) are only used for defining
+        // the content of the block. In such a case, nesting it does not work as
+        // expected as the definition is not part of the default template code flow.
+        if ($nested && $node instanceof Twig_Node_BlockReference) {
+            //throw new Twig_Error_Syntax('A block definition cannot be nested under non-capturing nodes.', $node->getTemplateLine(), $this->stream->getSourceContext());
+            @trigger_error('Nesting a block definition under a non-capturing node is deprecated since version 2.5.0 and will become a syntax error in 3.0.', E_USER_DEPRECATED);
```
stof: should it mention the template name, to know where this can be fixed ? the 3.x exception will have the location, but this message does not

lib/Twig/Test/IntegrationTestCase.php
```diff
@@ -176,7 +178,14 @@ protected function doIntegrationTest($file, $message, $condition, $templates, $e
             $p->setAccessible(true);
             $p->setValue($twig, '__TwigTemplate_'.hash('sha256', uniqid(mt_rand(), true), false).'_');
 
+            $deprecations = array();
             try {
+                set_error_handler(function ($type, $msg) use (&$deprecations) {
+                    if (E_USER_DEPRECATED === $type) {
+                        $deprecations[] = $msg;
+                    }
```
stof: the original error handler should be called for any other error, to make the rendering fail. Otherwise, this is equivalent to silencing all non-deprecation errors in tests (defeating the goal of the test)

lib/Twig/Parser.php
fabpot: done

lib/Twig/Test/IntegrationTestCase.php
fabpot: fixed

stof (approved):

PR #2688, opened 2018-05-22
Implement multiple inheritance

tyurderi:
This feature brings template inheritance into a new level. I needed it for personal reasons and tried to implement it into the core. Maybe it gets accepted :)

With this feature you can extend a parent template with one or more child templates. 

For example, the normal behaviour:
```php
// index.twig
{% block title %} Hello World {% endblock %}

// child.twig
{% extends 'index.twig' %}
{% block title %} new title {% endblock %}

// index.php
$twig->render('child.twig');
```

This is fine but if you extend your app with a plugin-system with the possibility that plugins can extend templates like they want it is not suitable for.

The new behaviour (while the old behaviour is still supported):
```php
// index.twig
{% block title %} Hello World {% endblock %}

// child.twig
{% block title %} new title {% endblock %}

// index.php
$twig->extendsTemplate('index.twig', 'child.twig'); // called somewhere before the next line
$twig->render('index.twig');
```

The output will be the same. But in the second example you can extend the parent with at least one child template - as often as you like.

This is a working proof of concept and I am not sure if I am missing any important things since this is the first time I have modified the twig core. Maybe it is also a wrong approach to solve this problem (due performance lacks or similar).

Feel free to discuss this feature. :)

### DIFF
--- a/lib/Twig/Environment.php
+++ b/lib/Twig/Environment.php
@@ -43,6 +43,7 @@ class Twig_Environment
     private $runtimes = array();
     private $optionsHash;
     private $loading = array();
+    private $extends = array();
 
     /**
      * Constructor.
@@ -347,6 +348,16 @@ class Twig_Environment
      */
     public function loadTemplate($name, $index = null)
     {
+        if (isset($this->extends[$name]) && $this->extends[$name]['lock'] === false) {
+            $template = $this->buildExtendedTemplate($name);
+        
+            $this->extends[$name]['lock'] = true;
+            $template = $this->createTemplate($template);
+            $this->extends[$name]['lock'] = false;
+        
+            return $template;
+        }
+        
         $cls = $mainCls = $this->getTemplateClass($name);
         if (null !== $index) {
             $cls .= '_'.$index;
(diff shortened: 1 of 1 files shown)


stof:
As far as I can tell, this does *not* implement multiple inheritance (you can only have one parent when rendering). It implements dynamic inheritance in a weird way (and incompatible with caching AFAICT) while Twig **already** supports dynamic inheritance in a clean way.

stof:
hmm, no, that's not dynamic inheritance either. That's about generating a child template which *uses* the templates you name "children" (which are not child templates at all, which is why I got confused). I don't think that's a desirable feature.

From a performance PoV and an architecture PoV, this looks like a nightmare: `index.twig` changes source (and so must not reuse the cache) depending on some *state* in the Twig environment (which can never be reset AFAIK), which requires a special locking system to be able to make it mean its original meaning somewhere.

If you really need such stateful feature, I suggest you to first try to experiment with it on top of Twig instead (`createTemplate` is a public API)

tyurderi:
@stof How is dynamic inheritance done yet?

Oh and you are totally right about the nightmare :) I have just modified the code so it works - without matter if it is nice or anything else.

fabpot:
@tyurderi I'm going to close it as we already have many ways to "inherits" templates (via `extends` -- even dynamically, but also via horizontal reuse and the `use` tag).

PR #2689, opened 2018-05-22
Add option to inline static values

pierredup:
Sometimes projects use filters or functions with static values, which can cause a lot additional function calls in the compiled template. If these static values are inlined in the compiled template, it can improve performance during render.

One example might be the `title` filter. If you have `{{ 'hello world'|title }}` in your template, it generates the following code in the compiled template:

```php
echo twig_escape_filter($this->env, twig_title_string_filter($this->env, "hello world"), "html", null, true);
```
where the call to `twig_title_string_filter` might be unnecessary, and the transformed value can be inlined in the template, giving the following result:

```php
echo twig_escape_filter($this->env, "Hello World", "html", null, true);
```

Another example is Symfony's `|trans` filter or `asset()` function, which generates the following code in the template respectively: 

```php
echo twig_escape_filter($this->env, $this->extensions['Symfony\Bridge\Twig\Extension\TranslationExtension']->trans("pages.home.read_more", array(), "labels"), "html", null, true);

// ... 

echo twig_escape_filter($this->env, $this->extensions['Symfony\Bridge\Twig\Extension\AssetExtension']->getAssetUrl("assets/images/Homepage.jpg"), "html", null, true);
```
Inlined, those values would look like

```php
echo twig_escape_filter($this->env, "Read More", "html", null, true);

// ... 

echo twig_escape_filter($this->env, "/assets/images/Homepage.jpg", "html", null, true);
```

(The Symfony examples are just to illustrate the functionality and not an ideal scenario as there are a lot of other factors involved, like config settings etc. which would influence the result)

On the homepage of an app with roughly 32 translations and 10 images, I saw an improvement of 17% in load time (very rough benchmark, other factors might also impact the test).

Inlining a value can be done by using the new `'inline' => true` option when defining a filter or function (E.G `new Twig_Filter('camel_case', 'my_camelcase_function', array('inline' => true))`).

If any dynamic value is encountered (E.G variables, expressions etc), the default template is compiled as it would normally be done and inlining is skipped

**TODO:**
- [ ] Add unit tests
- [ ] Enable inline mode for more core functions/filters
- [ ] Support inline mode for more expression types
- [ ] docs

### DIFF
--- a/lib/Twig/Node/Expression/Call.php
+++ b/lib/Twig/Node/Expression/Call.php
@@ -17,34 +17,47 @@ abstract class Twig_Node_Expression_Call extends Twig_Node_Expression
         $callable = $this->getAttribute('callable');
 
         $closingParenthesis = false;
-        if (is_string($callable) && false === strpos($callable, '::')) {
-            $compiler->raw($callable);
-        } else {
-            list($r, $callable) = $this->reflectCallable($callable);
-            if ($r instanceof ReflectionMethod && is_string($callable[0])) {
-                if ($r->isStatic()) {
-                    $compiler->raw(sprintf('%s::%s', $callable[0], $callable[1]));
-                } else {
-                    $compiler->raw(sprintf('$this->env->getRuntime(\'%s\')->%s', $callable[0], $callable[1]));
-                }
-            } elseif ($r instanceof ReflectionMethod && $callable[0] instanceof Twig_ExtensionInterface) {
-                // For BC/FC with namespaced aliases
-                $class = (new ReflectionClass(get_class($callable[0])))->name;
-                if (!$compiler->getEnvironment()->hasExtension($class)) {
-                    throw new Twig_Error_Runtime(sprintf('The "%s" extension is not enabled.', $class));
-                }
+        $compileInline = false;
 
-                $compiler->raw(sprintf('$this->extensions[\'%s\']->%s', ltrim($class, '\\'), $callable[1]));
-            } else {
-                $closingParenthesis = true;
-                $compiler->raw(sprintf('call_user_func_array($this->env->get%s(\'%s\')->getCallable(), array', ucfirst($this->getAttribute('type')), $this->getAttribute('name')));
+        if ($this->hasAttribute('inline') && $this->getAttribute('inline')) {
+            try {
+                $this->compileInline($compiler, $callable);
+                $compileInline = true;
+            } catch (RuntimeException $e) {
+                // Continue standard source compilation
             }
         }
 
-        $this->compileArguments($compiler);
+        if (!$compileInline) {
+            if (is_string($callable) && false === strpos($callable, '::')) {
+                $compiler->raw($callable);
+            } else {
+                list($r, $callable) = $this->reflectCallable($callable);
+                if ($r instanceof ReflectionMethod && is_string($callable[0])) {
+                    if ($r->isStatic()) {
+                        $compiler->raw(sprintf('%s::%s', $callable[0], $callable[1]));
+                    } else {
+                        $compiler->raw(sprintf('$this->env->getRuntime(\'%s\')->%s', $callable[0], $callable[1]));
+                    }
+                } elseif ($r instanceof ReflectionMethod && $callable[0] instanceof Twig_ExtensionInterface) {
+                    // For BC/FC with namespaced aliases
+                    $class = (new ReflectionClass(get_class($callable[0])))->name;
+                    if (!$compiler->getEnvironment()->hasExtension($class)) {
+                        throw new Twig_Error_Runtime(sprintf('The "%s" extension is not enabled.', $class));
+                    }
 
-        if ($closingParenthesis) {
-            $compiler->raw(')');
+                    $compiler->raw(sprintf('$this->extensions[\'%s\']->%s', ltrim($class, '\\'), $callable[1]));
+                } else {
+                    $closingParenthesis = true;
+                    $compiler->raw(sprintf('call_user_func_array($this->env->get%s(\'%s\')->getCallable(), array', ucfirst($this->getAttribute('type')), $this->getAttribute('name')));
+                }
+            }
+
+            $this->compileArguments($compiler);
+
+            if ($closingParenthesis) {
+                $compiler->raw(')');
+            }
         }
     }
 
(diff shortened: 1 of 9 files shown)


nicolas-grekas:
Let's try :)

Thinking from Symfony side, this means calling services (twig extensions) at warmup time. While this might be fine, it might also break in e.g. environments that don't have access to the DB then.

Second issue: this may wrongly inline context sensitive results. E.g. the "trans" example relies on a default locale. What if this changes at runtime?

This may mean inlining should be opt-in, either per extension or per filter (taking args into account, e.g. "locale provided or not").

pierredup:
> This may mean inlining should be opt-in, either per extension or per filter

It currently is opt-in per filter/function using the `['inline' => true]` option.

I wasn't actually planning on adding this to the `trans` filter in Symfony, the examples were just to give more information and context about what this patch can do. And I was curious about the performance, so I did the benchmark against that :)

Each filter and function will need to be evaluated if it can be inlined. Ideally, a developer would add this feature to their own extensions if they know it can be inlined (E.G in one project, we have an `icon()` function that just renders the html for a FontAwesome icon, and that html can be inlined as there is no external services or dependencies)

lib/Twig/Node/Expression/Call.php
```diff
@@ -17,34 +17,47 @@ protected function compileCallable(Twig_Compiler $compiler)
...
-                if (!$compiler->getEnvironment()->hasExtension($class)) {
-                    throw new Twig_Error_Runtime(sprintf('The "%s" extension is not enabled.', $class));
-                }
+        $compileInline = false;
+
+        if ($this->hasAttribute('inline') && $this->getAttribute('inline')) {
+            try {
+                $this->compileInline($compiler, $callable);
+                $compileInline = true;
+            } catch (RuntimeException $e) {
```
stof, line 26: if you need a bailout exception, it should be a specific one IMO

lib/Twig/Node/Expression/Call.php
```diff
@@ -17,34 +17,47 @@ protected function compileCallable(Twig_Compiler $compiler)
...
+                // Continue standard source compilation
+            }
+        }
 
-                $compiler->raw(sprintf('$this->extensions[\'%s\']->%s', ltrim($class, '\\'), $callable[1]));
+        if (!$compileInline) {
```
stof, line 31: instead of indenting all this logic in a if block to cancel all the normal compilation when it was compiled inline, you should rather return earlier after compiling inline.

lib/Twig/Node/Expression/Call.php
```diff
@@ -287,6 +300,93 @@ private function reflectCallable($callable)
 
         return $this->reflector = array($r, $callable);
     }
+
+    public function compileInline(Twig_Compiler $compiler, $callable)
```
stof, line 304: should be protected, not public

lib/Twig/Node/Expression/Call.php
```diff
@@ -287,6 +300,93 @@ private function reflectCallable($callable)
...
+    {
+        $args = array();
+
+        if ($this->hasAttribute('needs_environment') && $this->getAttribute('needs_environment')) {
+            $args[] = $compiler->getEnvironment();
+        }
+
+        if ($this->hasAttribute('needs_context') && $this->getAttribute('needs_context')) {
+            $args[] = array(); // @TODO: Is it possible to get the context from here?
```
stof, line 313: no it is not (and so it should forbid inlining)

lib/Twig/Node/Expression/Call.php
```diff
@@ -287,6 +300,93 @@ private function reflectCallable($callable)
...
+        }
+
+        if ($this->hasAttribute('arguments')) {
+            foreach ($this->getAttribute('arguments') as $argument) {
+                $args[] = sprintf('"%s"', addcslashes($argument, "\0\t\"\$\\"));
```
stof, line 318: that looks wrong to me

lib/Twig/Node/Expression/Call.php
```diff
@@ -287,6 +300,93 @@ private function reflectCallable($callable)
...
+        $result = $callable(...$args);
+
+        if (is_string($result)) {
+            $compiler->string($result);
+        } else {
+            $compiler->raw(var_export($result, true));
+        }
+    }
+
+    private function compileInlineArgument(Twig_Compiler $compiler, Twig_Node $node)
```
stof, line 345: this method is not compiling. It is evaluating.

lib/Twig/Node/Expression/Call.php
```diff
@@ -287,6 +300,93 @@ private function reflectCallable($callable)
...
+    {
+        if (!$node instanceof Twig_Node_Expression_Call && !$node instanceof Twig_Node_Expression_Constant && !$node instanceof Twig_Node_Expression_Array) {
+            throw new RuntimeException('Unsupported dynamic value: ' . $node);
+        }
+
+        if ($node instanceof Twig_Node_Expression_Filter) {
+            $name = $node->getNode('filter')->getAttribute('value');
+            $filter = $compiler->getEnvironment()->getFilter($name);
+
+            return $node->compileInline($compiler, $filter->getCallable());
```
stof, line 355: this is broken, as `compileInline` does not return anything

stof (changes requested):
Inlining any filter or function needing the environment is very fragile. Most of these cases are about accessing some config of the environment (often the charset), which can currently be changed without forcing a recompilation of templates (it is a runtime config).
Evaluating these functions are compile-time would require turning these into a compile-time config (and so invalidating the cache if you change the config)

stof:
> Thinking from Symfony side, this means calling services (twig extensions) at warmup time. While this might be fine, it might also break in e.g. environments that don't have access to the DB then.

Instantiating extensions already need to work at warmup time. Then, if an extension is not sure that the function runtime is safe at warmup time, it would be its responsibility to not opt-in for inlining the function (anything dealing with a DB cannot be inlined anyway as you don't want to clear your template cache each time your DB entry gets updated)

lib/Twig/Node/Expression/Call.php
pierredup: This just mimics the `$compiler->string()` function. It's required to escape the returned values (escaping quotes is especially necessary here)

stof:
I'm wondering whether this should actually be done by the node compiler. This would make the logic very complex if you want to support inlining propagation (i.e. inlining a call might then allow inlining the next one as the argument is not dynamic anymore). Doing such logic in the compiler itself would make it a maintenance nightmare, as the inlined compilation would have to work in a totally different way.

I think this should rather be implemented as a node visitor replacing some nodes in the AST with inlined version.
But note that this requires being careful regarding auto-escaping. Constant nodes are considered safe by the auto-escaper. But the result of an unsafe operation cannot be safe. So this optimization must run after the autoescaper (so that all auto-escaping is already applied and the inlining logic already has to take care about the escaping calls).

lib/Twig/Node/Expression/Call.php
stof: but that's totally wrong. You are evaluating code here. So you want a string (which you already have), not a string containing the source code for a PHP string (you won't run this through `eval`)

stof: I'm quite sure you haven't tried this code at all.

fabpot:
I have not had a look at the code of this pull request yet, but two big warnings:

 * I've tried several times and gave up as what we can do is quite limited (see all the comments from other commenters on limitations);

 * This indeed should be done as a node visitor.

lib/Twig/Node/Expression/Call.php
pierredup: Thanks, this explains why I couldn't get it to work on a some of the core filters, since I was getting a lot of `null` values :)

stof:
Well, as long as the charset is a runtime config, almost nothing can be inlined (most "basic" string manipulation filters depend on the charset).

If we want to make this logic safe regarding caching, we should forbid any function needing the environment to be inlined, as any config accessed from the environment is a runtime one. If we want to convert the charset to a compile-time setting in the future, we should add a way to inject the charset on its own at compile-time instead of injecting the whole environment.

pierredup:
> Inlining any filter or function needing the environment is very fragile

A use case for this is a function that renders another template and returns the html. In some cases that html can be inlined (although that might not always be the case). But I understand the limitation with the charset as well

> I think this should rather be implemented as a node visitor replacing some nodes in the AST with inlined version.

I haven't worked with node visitors in a while but will give this a shot

stof:
> A use case for this is a function that renders another template and returns the html. In some cases that html can be inlined (although that might not always be the case). But I understand the limitation with the charset as well

that would be insane regarding caching, as the cache invalidation for one template now has to be aware that any change in the other template must invalidate the cache too. This will not work (your Twig filter will not be able to alter the caching behavior of the template in which it will be used).

pierredup:
That's a good point, I didn't consider cache invalidation for templates, as I'm used to always doing a full cache refresh during deployment.
I'll then skip inlining for anything that depends on the environment or context

lib/Twig/Node/Expression/Call.php
pierredup: I did try the code, I wouldn't submit a PR for something that I haven't tested.

The initial version of my changes caused errors in the compiled template if any string is returned with quotes. Since the raw value was added to the compiled template, you would end up with `echo twig_escape_filter($this->env, one "two" three, "html", null, true);` instead of `echo twig_escape_filter($this->env, "one \"two\" three", "html", null, true);`. So this was necessary.
But after the refactor I ended up using `$compiler->string` (instead of `$compiler->raw`) which now handles the escaping of the values. So this part is not needed anymore, but it was necessary with the initial patch before I switched from `->raw()` to `->string()`

stof: and now, it is breaking things

stof:
@pierredup the `auto_reload` feature of Twig is something that we can absolutely not break. It is part of the great development experience we want for Symfony (in Symfony, you never need to clear the cache manually during development. the debug mode always checks for cache freshness)

pierredup:
Inlining should then be disabled during development since you want to see any changes you make to extensions without clearing the cache.

stof:
@pierredup and then, you might face issues in prod in case the inlined variant of the compilation is not 100% equivalent to the non-inlined variant. And it means it would require a way to opt-in for the inline at the config-level of Twig too, to make the user choose whether they want to inline or no.
I would rather limit the inlining to cases where it is compatible with the automatic reloading.

nicolas-grekas:
> It currently is opt-in per filter/function using the ['inline' => true] option.

For cases where it might also be arguments-dependent, allow this option to be a closure that would take the filter's arguments to decide, and return true/false based on that?

> Instantiating extensions already need to work at warmup time

they are instantiated, but not called, so that their (lazy) runtime might still be sleeping.

> Inlining should then be disabled during development since you want to see any changes you make to extensions without clearing the cache

at some point, ppl did opt-in for inlining, so I don't think it's an issue. Yes, cache auto clear will not work. Fine to me, that's the tradeoff. We will have to document it for sure.

stof:
> at some point, ppl did opt-in for inlining, so I don't think it's an issue. Yes, cache auto clear will not work. Fine to me, that's the tradeoff. We will have to document it for sure.

no, the extension author opts in for inlining. This does not mean that the end-user is OK to loose their development experience.

apfelbox:
Quick comment from the side: inlining as static value would be really really great for inlining (static) assets like SVGs. 👍

fabpot:
Reading the comments and the state of the PR, I think we are far from something that might be considered. Let's close for now.

PR #2690, opened 2018-05-24
Update random.rst

nzubkov:
added a new way to get a random number from the range.


stof:
Well, that's the same than the first example, as you are also passing an array as argument. I don't think we need to provide examples for all possible ways to get an array to pass it in the argument.

SpacePossum:
it is a not so efficient way for a min. max. as it creates an array with an element for each value in the range, couldn't Twig support `random(min, max)` ?

fabpot:
I agree with @stof. Let's close it.

apfelbox:
@SpacePossum you can always fake it with `min + random(max - min)`. Would still be convenient to add it directly, though.

Issue #2599, opened 2017-12-07
Deprecate templates


kunicmarko20:
Hi,

I was doing a PR [here](https://github.com/sonata-project/SonataAdminBundle/pull/4833) as we had to deprecate some templates and I had to provide an extension for this so I was wondering if a feature like deprecating templates could be added directly into twig maybe?

fabpot:
Not sure I understand what "deprecating a template means". Can you tell us a bit more?

greg0ire:
The context is moving a template from a bundle to another. Sonata doctrine ORM to Sonata admin in our case. Except it cannot be a move, because that would be a BC-break, so we copy templates to the admin bundle, and want a way to tell people they should no longer rely on the old ones.

fabpot:
Ok, understood. That's not a feature that should be provided by Twig directly. We don't have any of the infrastructure needed to make it happen anyway, so that's out of scope.

yceruto:
Oh, I had not seen this issue in the past and I think https://github.com/twigphp/Twig/pull/2696 is precisely what this is about.

kunicmarko20:
At the end we created our own https://github.com/sonata-project/SonataCoreBundle/blob/3.x/src/Twig/Node/DeprecatedTemplateNode.php

kunicmarko20:
Would love to deprecate it and move to something more official 😄

PR #2696, opened 2018-06-02
[RFC] Added new "deprecated" tag

yceruto:
```twig
{% deprecated 'The message...' %}
```

This new tag would allow us to define a deprecation warning anywhere within a template, useful for Frameworks, bundles, etc. where breaking the BC is frequently when templates are involved.

It will be easy to guarantee a smooth migration path when we want e.g. rename/remove a template or block, as well as displaying an accurate depreciation message.

**Deprecating a whole template `{# base.twig #}`**
```twig
{% deprecated 'The "' ~ _self ~ '" template is deprecated, use "layout.twig" instead' %}
 
{% extends 'layout.twig' %}
```
If we're extending from this template `{% extends 'base.twig' %}` then:
```
The "base.twig" template is deprecated, use "layout.twig" instead ("base.twig" at line 1).
```
**Deprecating a block {# greeting/blocks.twig #}**
```twig
{% block hey %}
    {% deprecated 'The "hey" block is deprecated, use "greet" block instead' %}

    {{ block('greet') }}
{% endblock %}

{% block greet %}
    Hey you!
{% endblock %}
```
If we're using this block `{{ block('hey') }}` then:
```
The "hey" block is deprecated, use "greet" block instead ("greeting/blocks.twig" at line 2).
```
also other examples come from my mind like deprecating macros, and any other extension point.

### DIFF
--- a/doc/tags/deprecated.rst
+++ b/doc/tags/deprecated.rst
@@ -1,0 +1,30 @@
+``deprecated``
+==============
+
+.. versionadded:: 1.36 and 2.6
+    The ``deprecated`` tag was added in Twig 1.36 and 2.6.
+
+Twig generates a deprecation notice (via a call to the ``trigger_error()``
+PHP function) where the ``deprecated`` tag is used in a template:
+
+.. code-block:: jinja
+
+    {# base.twig #}
+    {% deprecated 'The "base.twig" template is deprecated, use "layout.twig" instead.' %}
+    {% extends 'layout.twig' %}
+
+Also you can deprecate a block in the following way:
+
+.. code-block:: jinja
+
+    {% block hey %}
+        {% deprecated 'The "hey" block is deprecated, use "greet" instead.' %}
+        {{ block('greet') }}
+    {% endblock %}
+
+    {% block greet %}
+        Hey you!
+    {% endblock %}
+
+Note that by default, the deprecation notices are silenced and never displayed nor logged.
+See :ref:`deprecation-notices` to learn how to handle them.
--- a/lib/Twig/Node/Deprecated.php
+++ b/lib/Twig/Node/Deprecated.php
@@ -1,0 +1,49 @@
+<?php
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+/**
+ * Represents a deprecated node.
+ *
+ * @author Yonel Ceruto <yonelceruto@gmail.com>
+ */
+class Twig_Node_Deprecated extends Twig_Node
+{
+    public function __construct(Twig_Node_Expression $expr, $lineno, $tag = null)
+    {
+        parent::__construct(array('expr' => $expr), array(), $lineno, $tag);
+    }
+
+    public function compile(Twig_Compiler $compiler)
+    {
+        $compiler->addDebugInfo($this);
+
+        $expr = $this->getNode('expr');
+
+        if ($expr instanceof Twig_Node_Expression_Constant) {
+            $compiler->write('@trigger_error(')
+                ->subcompile($expr);
+        } else {
+            $varName = $compiler->getVarName();
+            $compiler->write(sprintf('$%s = ', $varName))
+                ->subcompile($expr)
+                ->raw(";\n")
+                ->write(sprintf('@trigger_error($%s', $varName));
+        }
+
+        $compiler
+            ->raw('.')
+            ->string(sprintf(' ("%s" at line %d).', $this->getTemplateName(), $this->getTemplateLine()))
+            ->raw(", E_USER_DEPRECATED);\n")
+        ;
+    }
+}
+
+class_alias('Twig_Node_Deprecated', 'Twig\Node\DeprecatedNode', false);
--- a/test/Twig/Tests/Node/DeprecatedTest.php
+++ b/test/Twig/Tests/Node/DeprecatedTest.php
@@ -1,0 +1,71 @@
+<?php
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+class Twig_Tests_Node_DeprecatedTest extends Twig_Test_NodeTestCase
+{
+    public function testConstructor()
+    {
+        $expr = new Twig_Node_Expression_Constant('foo', 1);
+        $node = new Twig_Node_Deprecated($expr, 1);
+
+        $this->assertEquals($expr, $node->getNode('expr'));
+    }
+
+    public function getTests()
+    {
+        $tests = array();
+
+        $expr = new Twig_Node_Expression_Constant('This section is deprecated', 1);
+        $node = new Twig_Node_Deprecated($expr, 1, 'deprecated');
+        $node->setTemplateName('foo.twig');
+
+        $tests[] = array($node, <<<EOF
+// line 1
+@trigger_error("This section is deprecated"." (\"foo.twig\" at line 1).", E_USER_DEPRECATED);
+EOF
+        );
+
+        $t = new Twig_Node(array(
+            new Twig_Node_Expression_Constant(true, 1),
+            new Twig_Node_Deprecated($expr, 2, 'deprecated'),
+        ), array(), 1);
+        $node = new Twig_Node_If($t, null, 1);
+        $node->setTemplateName('foo.twig');
+
+        $tests[] = array($node, <<<EOF
+// line 1
+if (true) {
+    // line 2
+    @trigger_error("This section is deprecated"." (\"foo.twig\" at line 2).", E_USER_DEPRECATED);
+}
+EOF
+        );
+
+        $environment = new Twig_Environment($this->getMockBuilder('Twig_LoaderInterface')->getMock());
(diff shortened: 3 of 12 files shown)


evelync23:
A simple way to deprecate things in templates, thank you.

doc/tags/deprecated.rst
```diff
@@ -0,0 +1,30 @@
+``deprecated``
+==============
+
+.. versionadded:: 1.x
+    The ``deprecated`` tag was added in Twig 1.x.
```
stof: It should say `1.36 and 2.5`. `Added in 1.x` is worthless (and does not mention when it was added in the 1.x branch)

yceruto: This was a draft, thanks.

stof:
should the location of the deprecation (template name and original line) be included automatically in the message ? Otherwise, it might be harder to locate where they come from (the PHP stacktrace will go inside the compiled template).

yceruto:
@stof added template name and template lineno to the message automatically. I like it, thanks.

lib/Twig/Node/Deprecated.php
```diff
@@ -0,0 +1,36 @@
...
+        parent::__construct(array('expr' => $expr), array(), $lineno, $tag);
+    }
+
+    public function compile(Twig_Compiler $compiler)
+    {
+        $compiler
+            ->addDebugInfo($this)
+            ->write('@trigger_error(')
+            ->subcompile($this->getNode('expr'))
+            ->write(sprintf('.\' ("%s" at line %d).\'', $this->getTemplateName(), $this->getTemplateLine()))
```
stof: you should use `->string` instead to write a string. Otherwise, you need to handle escaping of special chars in the template name to ensure it stays valid (AFAIK, quotes are allowed in template names if your loader accepts them)

stof: and you should be using `->raw` to add some continuation to the line without indentation

test/Twig/Tests/Node/DeprecatedTest.php
```diff
@@ -0,0 +1,30 @@
...
+        $this->assertEquals($expr, $node->getNode('expr'));
+    }
+
+    public function getTests()
+    {
+        $expr = new Twig_Node_Expression_Constant('foo', 1);
+        $node = new Twig_Node_Deprecated($expr, 1, 'deprecated');
+        $node->setTemplateName('foo.twig');
+
+        return array(array($node, "// line 1\n@trigger_error(\"foo\".' (\"foo.twig\" at line 1).', E_USER_DEPRECATED);"));
```
stof: we should have a test compiling the node with some indentation set, to be able to see issues related to use `raw` vs `write`

lib/Twig/Node/Deprecated.php
```diff
@@ -0,0 +1,36 @@
...
+    {
+        parent::__construct(array('expr' => $expr), array(), $lineno, $tag);
+    }
+
+    public function compile(Twig_Compiler $compiler)
+    {
+        $compiler
+            ->addDebugInfo($this)
+            ->write('@trigger_error(')
+            ->subcompile($this->getNode('expr'))
```
stof: this should subcompile in raw mode to put it on the same line

yceruto: This should subcompile in raw mode by default, isn't it?

stof: indeed, sorry

yceruto:
@stof comments addressed.

lib/Twig/Node/Deprecated.php
```diff
@@ -0,0 +1,37 @@
...
+            ->raw(".' (")
+            ->string($this->getTemplateName())
```
stof: that's broken. You should not use `->string` to put the double quotes around the template name inside the message. You should use it to render the PHP string (i.e. ` ("foo.twig" at line 2).`). Otherwise, that's a hacky usage of the API, and could break at any time (for instance if the compiler switches to using single-quoted strings when possible)

yceruto: Fixed, thanks again.

stof:
this looks fine to me.

I still wonder about 1 thing: if the message is dynamic, any error in building it gets silenced, which might hide mistakes from developers. Here is an idea to improve this:

```php

$expr = $this->getNode('expr');

if ($expr instanceof Twig_Node_Expression_Constant) {
    $compiler->write('@trigger_error(')
        ->subcompile($expr);
} else {
    $varName = $compiler->getVarName();
    $compiler->write(sprintf('$%s = ', $varName)
        ->subcompile($expr)
        ->raw(";\n")
        ->write(sprintf('@trigger_error($%s', $varName));
}

$compiler
    ->raw('.')
// ...
``` 

this way, a dynamic message gets evaluated in a non-silenced statement and used in the next statement (and a constant message avoids the extra variable, so that the concatenation can be resolved by OPCache at compile-time for such case to optimize the code).

what do you think about that @fabpot ?

yceruto:
@stof good catch! I made the changes and I've added a test to cover non-constant expressions.

stof:
This works fine to deprecate a template. However, it does not work fine to deprecate a block, which is much harder to handle (that's similar to deprecating a protected method in PHP. Warning in the parent method won't catch cases where a child class overrides the method and expects it to keep working). Putting this inside a block will warn for cases where other templates call your block, but not for cases where they override it (there might be ways to trigger such deprecation though, but I haven't tried yet).

yceruto:
Right, I'm aware of that limitation but I think that we shouldn't do something to trigger the deprecation of the parent block if it was overridden altogether, just like in the PHP scenario.

For that case, maybe a new argument in `block` tag could work:
```twig
{% block name deprecated 'message' %}...{% endblock %}
```
Thus, we could look at the parent block and trigger the warning, but that would be out of scope for this PR, so I'd prefer to improve it later :)

This is ready for you? Thank you for all feedbacks :)

stof:
> Thus, we could look at the parent block and trigger the warning, but that would be out of scope for this PR, so I'd prefer to improve it later :)

and that is not possible due to the `{% block <name> <expression> %}` syntax for blocks.
I agree that this is out of scope of this PR (it might be a matter of combining this tag with some other features btw, as this PR does not add the possibility to deprecate a specific stuff, but to trigger deprecations where you want, including inside a `{% if %}` to make it conditional)

> This is ready for you?

I think it is

yceruto:
Bonjour @fabpot, I know your position about this feature in the past https://github.com/twigphp/Twig/issues/2599#issuecomment-350092618, do you still think this is not a feature that should be provided by Twig directly?

yceruto:
Rebased.

yceruto:
Hellou? any feedback?

doc/tags/deprecated.rst
```diff
@@ -0,0 +1,30 @@
+``deprecated``
+==============
+
+.. versionadded:: 1.36 and 2.5
```
fabpot: Would be 2.6

fabpot (changes requested):
Can you also add some functional tests (see `test/Twig/Tests/Fixtures/tags/spaceless/root_level_in_child.legacy.test` for an example)?

You should also add a note in the CHANGES file.

doc/tags/deprecated.rst
yceruto: Fixed

yceruto:
@fabpot thanks for your feedback, I made the changes.

fabpot:
Thank you @yceruto.

michaelKaefer:
@yceruto Deprecating a block like in the example can lead to confusion.

If a bundle uses:
```twig
{% block hey %}
    {% deprecated 'The "hey" block is deprecated, use "greet" block instead' %}
    {{ block('greet') }}
{% endblock %}

{% block greet %}
    Hey you!
{% endblock %}
```

Then the deprecation works as expected it if an app's child template looks like:
```twig
{# Either this #}
{{ block('hey') }}

{# Or that #}
{% block hey %}
    Foobar
    {{ parent() }}
{% endblock %}
```

If we did not remove the deprecated block we get no deprecation (already mentioned above and still to be implemented in another PR):
```twig
{% block hey %}
    Foobar
{% endblock %}
```

But if the child template already removed the deprecated block we get a deprecation which is really confusing for app developers:
```twig
{% block greet %}
    Foobar
{% endblock %}
```
App developers now have to add the deprecated block again.

(Also if the app developer does not overwrite the block, it is not possible to get rid of the deprecation.)

The only solution I can think of is to remove the block deprecation example from Twig documentation.

What do you think, am I missing something?

stof:
The way to deprecate a block depends whether you expect other templates to call your block or to override it. When you expect them to override it, the place to trigger the deprecation is different, similar to how you trigger deprecations in PHP for protected methods that you expect child classes to override and that you want to deprecate

michaelKaefer:
Then it should be noted in Twig docs (https://twig.symfony.com/doc/2.x/tags/deprecated.html) that the example is only meant for blocks that are expected to be called or overridden and there is no way to deprecate a block like `page_title` in https://github.com/EasyCorp/EasyAdminBundle/blob/master/src/Resources/views/page/content.html.twig.

I will do a PR if that is correct.

stof:
the `{% deprecated %}` tag is not about deprecating a block. It is about triggering a deprecation (consider it the equivalent of the `trigger_deprecation()` function of `symfony/deprecation-contracts`). What you deprecate depends on where you use it.

michaelKaefer:
I know, but the docs are misleading. Bundle developers copy the example code for deprecating blocks like in the EasyAdmin example. I want to add a note to Twig docs to warn bundle developers.

PR #2698, opened 2018-06-05
Add max value to random function

SpacePossum:


SpacePossum:
couple of months in, any thoughts? :)

(happy to carry this PR for couple more though)

lib/Twig/Extension/Core.php
```diff
@@ -280,19 +280,36 @@ function twig_cycle($values, $position)
...
  *
  * @throws Twig_Error_Runtime when $values is an empty array (does not apply to an empty string which is returned as is)
  *
  * @return mixed A random value from the given sequence
  */
-function twig_random(Twig_Environment $env, $values = null)
+function twig_random(Twig_Environment $env, $values = null, $max = null)
 {
     if (null === $values) {
         return mt_rand();
```
fabpot: Shouldn't we take `max` into account here?

lib/Twig/Extension/Core.php
```diff
@@ -280,19 +280,36 @@ function twig_cycle($values, $position)
...
+            }
+        } else {
+            $min = (int) $values;
+            $max = (int) $max;
+        }
+
+        return function_exists('random_int')
+            ? random_int($min, $max)
+            : mt_rand($min, $max)
+        ;
```
fabpot: should be on one line.

lib/Twig/Extension/Core.php
```diff
@@ -280,19 +280,36 @@ function twig_cycle($values, $position)
...
+            } else {
+                $max = (int) $values;
+                $min = 0;
+            }
+        } else {
+            $min = (int) $values;
+            $max = (int) $max;
+        }
+
+        return function_exists('random_int')
```
fabpot: Can you add a comment about the PHP version it was added in so that we can remove the condition when we bump the min version?

lib/Twig/Extension/Core.php
```diff
@@ -280,19 +280,36 @@ function twig_cycle($values, $position)
  *
  * @param Twig_Environment                   $env
  * @param Traversable|array|int|float|string $values The values to pick a random item from
+ * @param int|float|null                     $max    Max. value used when $values is an int or float
```
fabpot: `Maximum value ...`

lib/Twig/Extension/Core.php
```diff
@@ -280,19 +280,36 @@ function twig_cycle($values, $position)
...
+            if ($values < 0) {
+                $max = 0;
+                $min = (int) $values;
+            } else {
+                $max = (int) $values;
+                $min = 0;
+            }
+        } else {
+            $min = (int) $values;
+            $max = (int) $max;
```
fabpot: Why do you cast to `int` here?

lib/Twig/Extension/Core.php
```diff
@@ -280,19 +280,36 @@ function twig_cycle($values, $position)
...
     if (null === $values) {
         return mt_rand();
     }
 
     if (is_int($values) || is_float($values)) {
-        return $values < 0 ? mt_rand($values, 0) : mt_rand(0, $values);
+        if (null === $max) {
+            if ($values < 0) {
+                $max = 0;
+                $min = (int) $values;
```
fabpot: I had a hard time figuring out why `$values` is casted to an `int` here. Can we make it more obvious? I have not tested, but I supposed it works well on all possible inputs.

fabpot (changes requested):
Looks good to me apart from some comments. Also, this should be rebased on 1.x as we are still adding all new features in the 1.x branch. Thank you.

SpacePossum:
Thanks for the feedback and all the work lately :)

For clarity about the last commit.
I've removed:
- selecting the random function to use; this can be part of a dedicated modernizing PR somewhere in the future
- the explicit casting; this can be handled when strict types gets enabled in the future in a dedicated PR

I also update the PHPDoc on request and rebased and retargeted to the 1.x branch

doc/functions/random.rst
```diff
@@ -12,14 +12,18 @@ parameter type:
 
 * a random item from a sequence;
 * a random character from a string;
-* a random integer between 0 and the integer parameter (inclusive).
+* a random integer between 0 and the integer (or float) parameter (inclusive).
```
fabpot: The `or float` part should be reverted, `mt_rand` accepts int only.

fabpot (changes requested):

fabpot (approved):

fabpot:
Thank you @SpacePossum.

SpacePossum:
thanks @fabpot : )

PR #2700, opened 2018-06-06
Report the proper location for errors compiled in templates

stof:
The `{% use %}` and `{% with %}` tags are adding some runtime checks triggering exceptions in the compiled template. This ensures that they get the proper location.
While the guessing was generally working fine for the `{% with %}` (and so this only makes the code faster), the guessing was not working for `{% use %}` due to the exception happening in the class constructor rather than on display (and so the guessing was finding the template which was triggering the load of the faulty template).


stof:
I will send a separate PR to 1.x, as it does not have `$this->source` in compiled templates AFAIK

lib/Twig/Node/Module.php
```diff
@@ -210,11 +212,13 @@ protected function compileConstructor(Twig_Compiler $compiler)
                         ->string($key)
                         ->raw("])) {\n")
                         ->indent()
-                        ->write("throw new Twig_Error_Runtime(sprintf('Block ")
+                        ->write("throw new Twig_Error_Runtime('Block ")
                         ->string($key)
                         ->raw(' is not defined in trait ')
                         ->subcompile($trait->getNode('template'))
-                        ->raw(".'));\n")
+                        ->raw(".',")
```
fabpot: missing space after the comma

lib/Twig/Node/With.php
```diff
@@ -38,7 +38,9 @@ public function compile(Twig_Compiler $compiler)
                 ->raw(";\n")
                 ->write(sprintf("if (!is_array(\$%s)) {\n", $varsName))
                 ->indent()
-                ->write("throw new Twig_Error_Runtime('Variables passed to the \"with\" tag must be a hash.');\n")
+                ->write("throw new Twig_Error_Runtime('Variables passed to the \"with\" tag must be a hash.',")
```
fabpot: missing space after the comma

lib/Twig/Node/Module.php
```diff
@@ -198,7 +198,9 @@ protected function compileConstructor(Twig_Compiler $compiler)
                     ->indent()
                     ->write("throw new Twig_Error_Runtime('Template \"'.")
                     ->subcompile($trait->getNode('template'))
-                    ->raw(".'\" cannot be used as a trait.');\n")
+                    ->raw(".'\" cannot be used as a trait.',")
```
fabpot: missing space after the comma

fabpot (changes requested):

stof:
@fabpot fixed.

fabpot:
Thank you @stof.

PR #2702, opened 2018-06-06
Fix the error handling for the optimized extension-based function calls

stof:
Triggering a Twig_Error_Runtime at compile-time breaks the contract of the Twig environment, as such exception is for errors during the rendering.
This moves back the exception to a runtime one (same behavior than before the optimization).

Another option would be to replace this with a `Twig_Error_Syntax` instead (reporting the error earlier), but that would change the exception being thrown for such case.

### DIFF
--- a/lib/Twig/Node/Expression/Call.php
+++ b/lib/Twig/Node/Expression/Call.php
@@ -31,10 +31,13 @@ abstract class Twig_Node_Expression_Call extends Twig_Node_Expression
                 // For BC/FC with namespaced aliases
                 $class = (new ReflectionClass(get_class($callable[0])))->name;
                 if (!$compiler->getEnvironment()->hasExtension($class)) {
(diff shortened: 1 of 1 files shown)


stof:
I found this while investigating whether errors had their location set properly or no.

fabpot:
Is it possible to add some tests?

stof:
Well, I don't see how it would be possible to write a test ensuring that the Twig_Error_Runtime is actually thrown at render time and not at compile time easily. A unit test is not suited for that (well, we could add a unit test ensuring that the compilation does not throw but we coud not cover the runtime part) and an integration test would be a pain, as the current infra relies on `Twig_Environment::render`, which covers both the rendering and the compilation (steps are distinct one layer below).
If you have an idea about the way to write such test, you're welcome. Maybe combining both ?

fabpot:
Thank you @stof.

PR #2710, opened 2018-06-20
Use isset before array_key_exists

jderusse:
Small performance improvment when using

```
{% set foo = foo|default("bar") %}
```
Will render the following code
```
$context["foo"] = (((isset($context["foo"]) || array_key_exists("foo", $context))) ? (_twig_default_filter((isset($context["foo"]) || array_key_exists("foo", $context) ? $context["foo"] : (function () { throw new Twig_Error_Runtime('Variable "foo" does not exist.', 1, $this->source); })()), "bar")) : ("bar"));
```

instead of 
```
$context["foo"] = ((array_key_exists("foo", $context)) ? (_twig_default_filter((isset($context["foo"]) || array_key_exists("foo", $context) ? $context["foo"] : (function () { throw new Twig_Error_Runtime('Variable "foo" does not exist.', 1, $this->source); })()), "bar")) : ("bar"));
```

### DIFF
--- a/lib/Twig/Node/Expression/Name.php
+++ b/lib/Twig/Node/Expression/Name.php
@@ -32,7 +32,12 @@ class Twig_Node_Expression_Name extends Twig_Node_Expression
             if ($this->isSpecial()) {
                 $compiler->repr(true);
             } else {
-                $compiler->raw('array_key_exists(')->repr($name)->raw(', $context)');
+                $compiler
+                    ->raw('(isset($context[')
+                    ->string($name)
+                    ->raw(']) || array_key_exists(')
+                    ->string($name)
+                    ->raw(', $context))');
             }
         } elseif ($this->isSpecial()) {
             $compiler->raw($this->specialVars[$name]);


stof:
is there actually any performance benefit when using a PHP version without XDebug (which is known to slow down functions while having no impact on `isset` as it is not a function) ?

Please provide numbers proving the performance impact of the optimization.

jderusse:
from my own benchmark (7.2 without xdebug) `isset` is 1.5x faster than `array_key_exists`
see for similar tests. https://stackoverflow.com/a/21759158

```
php > $a = ['one' => 1, 'two' => 2, 'three' => 3];
php > $numIterations = 100000000;
php > $start = microtime(true); for ($i = 0; $i < $numIterations; $i++) { array_key_exists('none', $a); }; echo microtime(true) - $start;
2.3657739162445
php > $start = microtime(true); for ($i = 0; $i < $numIterations; $i++) { isset($a['none']); }; echo microtime(true) - $start;
1.7055630683899
```

fabpot:
Can you change the branch to 1.x?

jderusse:
@fabpot rebased onto `1.x`

fabpot (approved):

fabpot:
Thank you @jderusse.

fabpot:
I merged too fast, tests were broken. @jderusse Can you work on another PR to fix the tests?

jderusse:
sorry about that. give me 10 min

PR #2713, opened 2018-06-25
Update slice.rst

eloyrubinos:
Corrected typo on the [:2] example.

### DIFF
--- a/doc/filters/slice.rst
+++ b/doc/filters/slice.rst
@@ -32,7 +32,7 @@ As syntactic sugar, you can also use the ``[]`` notation:
     {{ '12345'[1:2] }} {# will display "23" #}
 
     {# you can omit the first argument -- which is the same as 0 #}
-    {{ '12345'[:2] }} {# will display "12" #}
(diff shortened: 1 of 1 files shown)


fabpot:
Looks like the current doc is correct: https://twigfiddle.com/oy23kw

eloyrubinos:
Hum, how come {{ '12345'[1:2] }} will display "23" according to documentation (positions 1 and 2 of the string), but {{ '12345'[:2] }} displays "12" (positions 0 and 1 instead of 0 to 2).

It's a bit inconsistent that the second argument is some times inclusive and some times exclusive. That's why I thought it was a typo.

apfelbox:
@eloyrubinos the second argument is not the end index, but the length. See https://twig.symfony.com/doc/2.x/filters/slice.html

So `[:2]` will get the first 2 characters, `[1:2]` will get the second and the third character – and just to add a very clear example: `[10:2]` will get the 11th and 12th character (if the string is long enough).

eloyrubinos:
Oh thanks a lot, so silly that I read that wrong.

PR #2721, opened 2018-07-19
Don't throw error on validate or parse name if throw var is false

yceruto:
As spotted here https://github.com/symfony/symfony/issues/28001 `$loader->exists('@foo')` shouldn't throw an exception. e.g.: https://github.com/symfony/symfony/blob/9bfa971bc5662a6f90408b58a7b2453d7dae4f83/src/Symfony/Component/HttpKernel/Fragment/HIncludeFragmentRenderer.php#L145


fabpot:
LGTM. Can you add some tests?

yceruto:
Tests added.

fabpot:
Thank you @yceruto.

PR #2726, opened 2018-07-25
Fix GlobalsInterface extends for IDE

ogizanagi:


fabpot:
Thank you @ogizanagi.

PR #2727, opened 2018-07-27
Use \Traversable instead of \IteratorAggregate (length filter)

agzr:
Especially for object implementing the \Iterator interface.


nicolas-grekas:
That's on purpose, see discussion in #2600

fabpot:
Closing then.

PR #2730, opened 2018-07-31
Add support for dynamic tests

### LINKED ISSUE #2589: Dynamic Twig_Test don't seems to work
Hi,
in Twig's documentation it's said "Tests also support dynamic name feature as filters and functions." (filter can be path_* by eg)

But I don't find how to achieve this.
I tried this :

`new \Twig_SimpleTest('format_*', [$this, 'isFormat'])`

but I get this error message :

> Unknown "format_web" test. Did you mean "format_*"

I haven't found an example, so I dig in the codebase. I found a "*" replacement for Filters : 

```
foreach ($this->filters as $pattern => $filter) {
            $pattern = str_replace('\\*', '(.*?)', preg_quote($pattern, '#'), $count);
            if ($count && preg_match('#^'.$pattern.'$#', $name, $matches)) {
                array_shift($matches);
                $filter->setArguments($matches);
                return $filter;
            }
        }
```

But not for test : 

```
public function getTest($name)
    {
        if (!$this->initialized) {
            $this->initExtensions();
        }
        if (isset($this->tests[$name])) {
            return $this->tests[$name];
        }
        return false;
    }
```

So : is it the doc that need to be fixed, or the getTest($name) method? (or do I get it wrong ?)

fabpot:
fixes #2589

### DIFF
--- a/doc/advanced.rst
+++ b/doc/advanced.rst
@@ -371,9 +371,12 @@ The ``node`` sub-node will contain an expression of ``my_value``. Node-based
 tests also have access to the ``arguments`` node. This node will contain the
 various other arguments that have been provided to your test.
 
+.. versionadded:: 1.36
+    Dynamic tests support was added in Twig 1.36.
+
 If you want to pass a variable number of positional or named arguments to the
-test, set the ``is_variadic`` option to ``true``. Tests also support dynamic
-name feature as filters and functions.
+test, set the ``is_variadic`` option to ``true``. Tests support dynamic
+names (see dynamic filters and functions for the syntax).
 
 Tags
 ----
--- a/lib/Twig/Environment.php
+++ b/lib/Twig/Environment.php
@@ -16,11 +16,11 @@
  */
 class Twig_Environment
 {
-    const VERSION = '1.35.5';
-    const VERSION_ID = 13505;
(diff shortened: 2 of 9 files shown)


doc/advanced.rst
```diff
@@ -371,9 +371,12 @@ The ``node`` sub-node will contain an expression of ``my_value``. Node-based
 tests also have access to the ``arguments`` node. This node will contain the
 various other arguments that have been provided to your test.
 
+.. versionadded:: 1.36
+    Dynamic tests support was added in Twig 1.36.
```
stof, line 375: it should say `1.36 and 2.6`, as we maintain 2 branches in parallel with minor versions in both.

stof: btw, this also impact the way we produce our changelog, as we would have to make it clear that this feature is also new in 2.5 (compared to 2.5)

lib/Twig/Environment.php
```diff
@@ -1240,6 +1240,19 @@ public function getTest($name)
             return $this->tests[$name];
         }
 
+        foreach ($this->tests as $pattern => $test) {
```
stof, line 1243: should we first do a `strpos` check on the pattern, to see whether it is actually dynamic ?

Issue #2728, opened 2018-07-27
Escaping issue with shortcut block syntax


simshaun:
Twig v2.5.0

I've been using the shortcut block syntax quite a bit, but just noticed an issue where the site name is being encoded even with `|raw`. I ended up needing to use the full-form syntax.

Given `site_name = Brown's`
Calling `$subject = $template->renderBlock('subject', $context);`

- With `{% block subject 'Analytics nag - ' ~ site_name|raw %}`  
Results in `Analytics nag - Brown&#039;s`

- With `{% block subject 'Analytics nag - ' ~ (site_name|raw) %}`  
Results in `Analytics nag - Brown&#039;s`

- With `{% block subject ('Analytics nag - ' ~ site_name|raw) %}`  
Results in `Analytics nag - Brown&#039;s`

- With `{% block subject %}Analytics nag - {{ site_name|raw }}{% endblock %}`  
Results in `Analytics nag - Brown's`

---

The Twig template itself looks something like this:
```twig
{% extends 'Email/email_template.html.twig' %}

{% block subject 'Analytics nag - ' ~ site_name|raw %}
{% block title site_name %}

{% block content %}
....
{% endblock %}
```

jvasseur:
Did you try `{% block subject ('Analytics nag - ' ~ site_name)|raw %}` ? The others can't work because the concatenation operation marks the string as unsafe and thus it has to be escaped before outputing.

simshaun:
Interesting. I would never have expected the concatenation operator to affect the escaping.

simshaun:
So the docs say it "converts all operands into strings and concatenates them".

I assume it's the string casting that forces the escaping. I'm just a little puzzled as to why.

stof:
Your shortcut is not equivalent to `{% block subject %}Analytics nag - {{ site_name|raw }}{% endblock %}`. It is equivalent to `{% block subject %}{{ 'Analytics nag - ' ~ site_name|raw }}{% endblock %}`

And here, the value being output is not mark as safe (due to the precedence between the concatenation and filtering operators). Marking a string as safe before the concatenation does not make the output of the concatenation safe.

simshaun:
@stof I understand the differences now. I just don't get why concatenation marks it "not safe".

It'd be nice to somehow have some control over this, maybe with a new operator like `~~`.
At the very least, maybe I should make a docs PR that makes it clear using `~` will mark the operands "unsafe" after they're converted to strings.

[...]

PR #2731, opened 2018-07-31
Expand documentation on concatenation operator safety

simshaun:
Helps clarify issue #2728

### DIFF
--- a/doc/templates.rst
+++ b/doc/templates.rst
@@ -784,7 +784,7 @@ The following operators don't fit into any of the other categories:
 
 * ``~``: Converts all operands into strings and concatenates them. ``{{ "Hello
   " ~ name ~ "!" }}`` would return (assuming ``name`` is ``'John'``) ``Hello
-  John!``. Note that the output will be escaped due to the brackets `{{ }}`.
+  John!``. Note that the output will be escaped due to autoescaping.
   
   If you wish to have raw output, group the concatenation in parentheses before
   passing to the raw filter: ``{{ ("Hello " ~ name ~ "!")|raw }}``
--- a/doc/templates.rst
+++ b/doc/templates.rst
@@ -782,13 +782,12 @@ The following operators don't fit into any of the other categories:
 
       (1..5)|join(', ')
 
-* ``~``: Converts all operands into strings and safely concatenates them (to
-  avoid potential XSS). ``{{ "Hello " ~ name ~ "!" }}`` would return 
-  (assuming ``name`` is ``'John'``) ``Hello John!``.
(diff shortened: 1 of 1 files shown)


SpacePossum (approved):

doc/templates.rst
```diff
@@ -782,9 +782,13 @@ The following operators don't fit into any of the other categories:
 
       (1..5)|join(', ')
 
-* ``~``: Converts all operands into strings and concatenates them. ``{{ "Hello
-  " ~ name ~ "!" }}`` would return (assuming ``name`` is ``'John'``) ``Hello
-  John!``.
+* ``~``: Converts all operands into strings and safely concatenates them (to
+  avoid potential XSS). ``{{ "Hello " ~ name ~ "!" }}`` would return
```
stof: the operator is not about *safely* concatenating. It does nothing regarding safety (and so passing it to something assuming it is already safe from XSS would actually creates XSS). The auto-escaping is doing things for safety (what is important to understand is that auto-escaping does not get applied on variables, but on what gets output using `{{ ... }}`, which is why things get escape). the safety does not come from `~`, but from `{{ ... }}` (and the fact that auto-escaping is ON by default)

simshaun: oooh

doc/templates.rst
```diff
@@ -784,7 +784,10 @@ The following operators don't fit into any of the other categories:
 
 * ``~``: Converts all operands into strings and concatenates them. ``{{ "Hello
   " ~ name ~ "!" }}`` would return (assuming ``name`` is ``'John'``) ``Hello
-  John!``.
+  John!``. Note that the output will be escaped due to autoescaping.
+  
+  If you wish to have raw output, group the concatenation in parentheses before
+  passing to the raw filter: ``{{ ("Hello " ~ name ~ "!")|raw }}``
```
xabbuh: [...] passing them to [...]

doc/templates.rst
```diff
@@ -784,7 +784,10 @@ The following operators don't fit into any of the other categories:
 
 * ``~``: Converts all operands into strings and concatenates them. ``{{ "Hello
   " ~ name ~ "!" }}`` would return (assuming ``name`` is ``'John'``) ``Hello
-  John!``.
+  John!``. Note that the output will be escaped due to autoescaping.
+  
+  If you wish to have raw output, group the concatenation operands in parentheses 
+  before passing them to the raw filter: ``{{ ("Hello " ~ name ~ "!")|raw }}``
```
fabpot, line 790: I don't think that this is the right place. It's more about operator precedence, where we already have the information. I've tried to make it clearer in #2802

PR #2736, opened 2018-08-23
Add support for negative number in look method

Chrissi2812:
If the number passed to the look function was negative the ErrorHandler Throwed an Undefined Index Exception.
Now it throws the Twig_Error_Syntax Exception as expected.

### DIFF
--- a/lib/Twig/TokenStream.php
+++ b/lib/Twig/TokenStream.php
@@ -91,19 +91,22 @@ final class Twig_TokenStream
     }
 
     /**
-     * Looks at the next token.
+     * Looks at the token at given offset.
      *
-     * @param int $number
+     * @param int $offset
      *
      * @return Twig_Token
      */
-    public function look($number = 1)
+    public function look($offset = 1)
     {
-        if (!isset($this->tokens[$this->current + $number])) {
-            throw new Twig_Error_Syntax('Unexpected end of template.', $this->tokens[$this->current + $number - 1]->getLine(), $this->source);
+        if (!isset($this->tokens[$this->current + $offset])) {
+            if ($offset < 0) {
+                throw new Twig_Error_Syntax('Unexpected start of template.', $this->tokens[$this->current + $offset + 1]->getLine(), $this->source);
+            }
+            throw new Twig_Error_Syntax('Unexpected end of template.', $this->tokens[$this->current + $offset - 1]->getLine(), $this->source);
         }
 
-        return $this->tokens[$this->current + $number];
+        return $this->tokens[$this->current + $offset];
     }
 
     /**


stof:
The method says `Looks at the next token.`. So passing a negative number looks weird to me.

Thus, the message `Unexpected end of template` does not make sense in such case, as you got hit by the *start* of the template.

What is your use case for looking behind the current position ?

Chrissi2812:
> The method says Looks at the next token.. So passing a negative number looks weird to me.

But there is another method named next (for that). The description of the `look` method should be `look at the token at specified offset.`

## Example

I tried to break a for loop and used the implemantation found [here](https://stackoverflow.com/a/40949346)

But the check if we are in a loop fails.
Currently I have to catch all Exceptions and then it works.
```php
class TokenParserBreak extends \Twig_TokenParser
{
    public function parse(\Twig_Token $token)
    {
        $stream->expect(\Twig_Token::BLOCK_END_TYPE);

        // Trick to check if we are currently in a loop.
        $currentForLoop = 0;

        for ($i = 1; true; $i++) {
            try {
                // if we look before the beginning of the stream
                // the stream will throw a \Twig_Error_Syntax
                $token = $stream->look(-$i);
            } catch (\Twig_Error_Syntax $e) {
                break;
            } catch (\Exception $e) {
                // The error handler of Twig is causing an undefined index issue
                // https://github.com/twigphp/Twig/pull/2736
                break;
            }

            if ($token->test(\Twig_Token::NAME_TYPE, 'for')) {
                $currentForLoop++;
            } else if ($token->test(\Twig_Token::NAME_TYPE, 'endfor')) {
                $currentForLoop--;
            }
        }


        if ($currentForLoop < 1) {
            throw new \Twig_Error_Syntax(
                'Break tag is only allowed in \'for\' loops.',
                $stream->getCurrent()->getLine(),
                $stream->getSourceContext()->getName()
            );
        }

        return new TwigNodeBreak();
    }

    public function getTag()
    {
        return 'break';
    }
}
```

stof:
No, `next` is about *moving* the cursor to the next token. `look` is about looking at a next token *without moving the cursor*.

Chrissi2812:
Thanks for clarifying that.

I just think it doesn't makes sense to disallow using `look` at negative offsets because there are use cases for it. A new method just for a negative `look` seems unnecessary in my opinion. If you rather like a new method for a negative `look` i can change my PR accordingly.

fabpot:
@Chrissi2812 As this is the first time someone asks for this feature, I'm not sure there are a lot of use cases. In your specific case, using a `NodeVisitor` to detect wrong usages of `break` seems like a possible solution. Let's close for now and reevaluate if more people have the same need.

Even if adding this would be simple enough, I'm reluctant as it relies on the current implementation details.

PR #2745, opened 2018-09-10
Add example for block without endblock

alexander-schranz:
a block is not required to have a endblock tag. not sure how to describe this correctly.

### DIFF
--- a/doc/functions/block.rst
+++ b/doc/functions/block.rst
@@ -11,6 +11,14 @@ times, use the ``block`` function:
     <h1>{{ block('title') }}</h1>
 
     {% block body %}{% endblock %}
+    
+The block can also be used without an endblock when give the content of the block directly by string or variable:
+
+.. code-block:: jinja
+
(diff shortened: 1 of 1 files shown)


doc/functions/block.rst
```diff
@@ -11,6 +11,14 @@ times, use the ``block`` function:
     <h1>{{ block('title') }}</h1>
 
     {% block body %}{% endblock %}
+    
+The block can also be used without an endblock when give the content of the block directly by string or variable:
+
+.. code-block:: jinja
+
+    {% block title 'title' %}
```
yceruto, line 19: This document is about `block()` as a function. This new documentation likely fits better with `{% block %}` tags section https://github.com/twigphp/Twig/blob/1.x/doc/tags/block.rst

alexander-schranz:
@yceruto thanks seems to be documented here correctly: https://twig.symfony.com/doc/1.x/tags/extends.html#block-shortcuts

Issue #2744, opened 2018-09-10
Accessing an ArrayObject with a null value fails


X-Coder264:
Code to reproduce the problem:

```php
return $this->render('test/index.html.twig', [
     'foo' => new \ArrayObject(['bar' => null]),
]);

{% if foo.bar is defined %}
    {{ foo.bar }}
{% endif %}
```

After running this you get

`E_NOTICE: Undefined property`

which is not the expected behaviour as I specifically test if it's defined before outputting it.

linaori:
Are you sure this is the code being executed?

```php
php > $a = new ArrayObject(['foo' => null]);
php > var_dump(isset($a['foo']) || array_key_exists('foo', $a));
bool(true)
php > var_dump($a['foo']);
NULL
```
Which is the code as I think is compiled from here: https://github.com/twigphp/Twig/blob/84b2eb541026dbd39590ba80269ae7676b1bfc46/lib/Twig/Node/Expression/Name.php#L35-L40

yceruto:
The warning appears when the expression is being evaluated as `$foo->bar`, because [`$object instanceof ArrayAccess && isset($object[$arrayItem])`](https://github.com/twigphp/Twig/blob/caf2caa8f4d24f16e9e01d1e8074d488be0fbbe3/lib/Twig/Template.php#L516) returns `false` in this case, hence the expression is treated as object property instead of array access. https://3v4l.org/qtss2

I think this bug can be solved by using `$object->offsetExists('item')` instead of `isset()`.

yceruto:
The error appears when the expression is being evaluated as `$foo->bar`, because [`$object instanceof ArrayAccess && isset($object[$arrayItem])`](https://github.com/twigphp/Twig/blob/caf2caa8f4d24f16e9e01d1e8074d488be0fbbe3/lib/Twig/Template.php#L516) returns `false` in this case, hence the expression is treated as object property instead of array access. https://3v4l.org/qtss2

I think this bug can be solved by using `$object->offsetExists($arrayItem)` instead of `isset()`.

linaori:
The problem is that it _is_ an object, checking the interface/type would cause additional checks and thus a performance loss. Would it be viable to switch the array_key_exists and isset checks?

yceruto:
@iltar see the linked condition (1.x):
https://github.com/twigphp/Twig/blob/caf2caa8f4d24f16e9e01d1e8074d488be0fbbe3/lib/Twig/Template.php#L515-L517

[...]

PR #2749, opened 2018-09-19
Fixed ArrayObject access with a null value

yceruto:
fixes https://github.com/twigphp/Twig/issues/2744

### DIFF
--- a/lib/Twig/Template.php
+++ b/lib/Twig/Template.php
@@ -508,9 +508,7 @@ abstract class Twig_Template implements Twig_TemplateInterface
         if (self::METHOD_CALL !== $type) {
             $arrayItem = is_bool($item) || is_float($item) ? (int) $item : $item;
 
-            if ((is_array($object) && (isset($object[$arrayItem]) || array_key_exists($arrayItem, $object)))
-                || ($object instanceof ArrayAccess && isset($object[$arrayItem]))
-            ) {
+            if ((is_array($object) || $object instanceof ArrayAccess) && (isset($object[$arrayItem]) || array_key_exists($arrayItem, $object))) {
                 if ($isDefinedTest) {
                     return true;
                 }
--- a/test/Twig/Tests/TemplateTest.php
+++ b/test/Twig/Tests/TemplateTest.php
@@ -315,6 +315,7 @@ class Twig_Tests_TemplateTest extends \PHPUnit\Framework\TestCase
         ];
 
         $objectArray = new Twig_TemplateArrayAccessObject();
+        $arrayObject = new ArrayObject($array);
         $stdObject = (object) $array;
         $magicPropertyObject = new Twig_TemplateMagicPropertyObject();
         $propertyObject = new Twig_TemplatePropertyObject();
@@ -346,6 +347,7 @@ class Twig_Tests_TemplateTest extends \PHPUnit\Framework\TestCase
             // array(object, type of fetch)
             [$array,               $arrayType],
             [$objectArray,         $arrayType],
+            [$arrayObject,         $anyType],
             [$stdObject,           $anyType],
             [$magicPropertyObject, $anyType],
             [$methodObject,        $methodType],


stof:
`array_key_exists` on an object does not use their ArrayAccess interface, but is an alias of `property_exists`. So that code does not do the right thing

stof:
btw, when testing things for ArrayAccess, don't do it *only* with ArrayObject, as that one is a special case (ArrayObject has some magic allowing to mix array access and property access, which might be why things are working in your particular case)

yceruto:
> array_key_exists on an object does not use their ArrayAccess interface, but is an alias of property_exists. So that code does not do the right thing

It depends how your `ArrayObject` is defined, right? This solution should cover three cases https://3v4l.org/kOVRn, https://3v4l.org/16e5v and https://3v4l.org/S4rs6:
```php
$object = new ArrayObject(['foo' => null]);
// or
$object = new ArrayObject();
$object['foo'] = null;
// or
$object = new ArrayObject(['foo' => null], ArrayObject::ARRAY_AS_PROPS);
```

> btw, when testing things for ArrayAccess, don't do it only with ArrayObject, as that one is a special case (ArrayObject has some magic allowing to mix array access and property access, which might be why things are working in your particular case)

There is already a test for ArrayAccess here:
https://github.com/twigphp/Twig/blob/5d166a4297ab89da37aebc592aa049dff98862d5/test/Twig/Tests/TemplateTest.php#L317

yceruto:
However, I found that by using [the `ArrayObject` flags](http://php.net/manual/en/class.arrayobject.php#arrayobject.constants.flags) or assigning values as properties doesn't work either https://3v4l.org/vAsiv. Same issue here:
https://github.com/twigphp/Twig/blob/5d166a4297ab89da37aebc592aa049dff98862d5/lib/Twig/Template.php#L580

yceruto:
However, I found that by initializing `$object->foo = null` doesn't work either https://3v4l.org/vAsiv. So same issue here:
https://github.com/twigphp/Twig/blob/5d166a4297ab89da37aebc592aa049dff98862d5/lib/Twig/Template.php#L580

nikic:
Just as a heads up, it is likely that the ability to use `array_key_exists` on objects in general and `ArrayObject` in particular will go away in the future, so please do not rely on this.

If you would like to special-case `ArrayObject` in particular you can call the `offsetExists()` method (which will treat null values as existing). However, this is not true of `ArrayAccess` objects in general.

Tobion (approved):
Seems to be the correct way to fix #2744
That this solution might not be future-proof according to @nikic is a different problem as that would also require more changes from what I can see, like in https://github.com/twigphp/Twig/blob/5d166a4297ab89da37aebc592aa049dff98862d5/lib/Twig/Template.php#L580
This can be dealt with when the time comes.

yceruto:
Resolved conflicts.

fabpot:
Thank you @yceruto.

fabpot:
@yceruto That breaks some tests on the 2.x branch. Can you have a look please?

yceruto:
It should be fixed in https://github.com/twigphp/Twig/pull/2809

stof:
does `array_key_exists` work for any ArrayAccess implementation ? Or only for ArrayObject ?

nikic:
@stof `array_key_exists` does not work with ArrayAccess objects. `ArrayObject` uses a special overload to be usable with `array_key_exists` (and this is orthogonal to its ArrayAccess implementation -- e.g. a child class overriding offsetExists will not be able to influence the behavior). This functionality is retained for backwards compatibility reasons (I believe ZF1 used this is a pretty central location) and I expect that we'll drop it in PHP 8.

stof:
then this PR is wrong.

nikic:
To be more precise `array_key_exists` "works" on all objects, but it's going to check the existence of properties. So if there's an ArrayAccess object that has a public `->foo` but no `['foo']` then I think this code will misbehave, because `array_key_exists` will return true, but `['foo']` will access a non-existent offset. Might be worthwhile to add a test for such a case.

PR #2734, opened 2018-08-16
Fix multi-byte UTF-8 in escape('html_attr')

svalgaard:
As far as I can see `|escape('html_attr')` corrupts some UTF-8 input/emoticons. "Normal" UTF-8 characters, e.g., æøå, works fine.

Sample input:
```
{% set P = '🥔' %}
P|length........: {{ P|length }}
P...............: {{ P }}
P|e('html').....: {{ P|escape('html') }}
P|e('html_attr'): {{ P|escape('html_attr') }}
&amp;#x1F954;...: &#x1F954;
````

Output from Twig 2.5.0:
```
P|length........: 1
P...............: 🥔
P|e('html').....: 🥔
P|e('html_attr'): &#xDD54;
&amp;#x1F954;...: &#x1F954;
```
The correct entity for potato is `&#x1F954;`.
`&#xDD54;` is a  � (Unicode replacement character).

Looking at `Twig/Extension/Core.php`, it seems that the following PR fixes this.

PR #2750, opened 2018-09-20
Fix multi-byte UFT-8 in escape('html_attr')

nicolas-grekas:
Alternative to #2734

### DIFF
--- a/lib/Twig/Extension/Core.php
+++ b/lib/Twig/Extension/Core.php
@@ -902,6 +902,10 @@ function twig_escape_filter(Twig_Environment $env, $string, $strategy = 'html', 
         }
     }
 
+    if ('' === $string) {
+        return '';
+    }
+
     if (null === $charset) {
         $charset = $env->getCharset();
     }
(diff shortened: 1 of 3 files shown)


lib/Twig/Extension/Core.php
```diff
@@ -1046,21 +1037,9 @@ function twig_escape_filter(Twig_Environment $env, $string, $strategy = 'html',
...
-                static $entityMap = array(
-                    34 => 'quot', /* quotation mark */
-                    38 => 'amp',  /* ampersand */
-                    60 => 'lt',   /* less-than sign */
-                    62 => 'gt',   /* greater-than sign */
-                );
 
                 $chr = $matches[0];
-                $ord = ord($chr);
+                $ord = \ord($chr);
```
fabpot: no need as we are not namespaced.

lib/Twig/Extension/Core.php
```diff
@@ -1074,23 +1053,32 @@ function twig_escape_filter(Twig_Environment $env, $string, $strategy = 'html',
                  * Check if the current character to escape has a name entity we should
                  * replace it with while grabbing the hex value of the character.
                  */
-                if (1 == strlen($chr)) {
-                    $hex = strtoupper(substr('00'.bin2hex($chr), -2));
-                } else {
-                    $chr = twig_convert_encoding($chr, 'UTF-16BE', 'UTF-8');
-                    $hex = strtoupper(substr('0000'.bin2hex($chr), -4));
-                }
+                if (1 === \strlen($chr)) {
```
fabpot: same here

fabpot (changes requested):

lib/Twig/Extension/Core.php
nicolas-grekas: fixed

fabpot:
Thank you @nicolas-grekas.

PR #2751, opened 2018-09-20
Fix multi-byte UFT-8 in escape('html_attr')

nicolas-grekas:
same as #2750 but on  1.x


fabpot:
Thank you @nicolas-grekas.

PR #2754, opened 2018-09-27
Improved the code of an example

javiereguiluz:
The original code contains an unneeded *"pop culture"* reference. For those unaware of it, it contains two strong words ("kill" and "bastard").

The concern about this example was firstly raised in the `#diversity` channel of [Symfony Slack](https://symfony.com/support).


fabpot:
Thank you @javiereguiluz.

PR #2757, opened 2018-09-28
Enhances join filter for natural language join (i.e., "A, B and C")

LavaSlider:
Simply adds a second optional parameter to the join filter that will be placed between the last two entries when generating the string. Since the filter only took one argument previously, this should be 100% backward compatible. Allows simple "first, second and third" string from an array. Compatible with any length input array.

### DIFF
--- a/lib/Twig/Extension/Core.php
+++ b/lib/Twig/Extension/Core.php
@@ -717,9 +717,12 @@ function twig_last(Twig_Environment $env, $item)
 /**
  * Joins the values to a string.
  *
- * The separator between elements is an empty string per default, you can define it with the optional parameter.
+ * The separators between elements are empty strings per default, you can define them with the optional parameters.
  *
  * <pre>
+ *  {{ [1, 2, 3]|join(', ', ' and ') }}
+ *  {# returns 1, 2 and 3 #}
+ *
  *  {{ [1, 2, 3]|join('|') }}
  *  {# returns 1|2|3 #}
  *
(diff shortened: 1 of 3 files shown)


lib/Twig/Extension/Core.php
```diff
@@ -643,16 +646,25 @@ function twig_last(Twig_Environment $env, $item)
...
  */
-function twig_join_filter($value, $glue = '')
+function twig_join_filter($value, $glue = '', $and = null)
 {
     if ($value instanceof Traversable) {
         $value = iterator_to_array($value, false);
     }
 
-    return implode($glue, (array) $value);
+    if( empty( (array) $value ) || is_null( $and ) || $and === $glue ) {
```
stof: please perform the array casting only once

stof: The null check should use `null === $and` rather than `is_null()` according to the coding standards

lib/Twig/Extension/Core.php
```diff
@@ -643,16 +646,25 @@ function twig_last(Twig_Environment $env, $item)
  *
  * @param array  $value An array
  * @param string $glue  The separator
+ * @param string $and   The separator for the last pair
```
stof: this argument has `string|null` as type, not `string`

stof (changes requested):

fabpot:
Thank you @LavaSlider.

PR #2760, opened 2018-10-05
Make sure twig_include returns a string

thewilkybarkid:
Currently returns `null` if `$ignoreMissing` is `true` and it's not found, but the return type is `string`.


fabpot:
Thank you @thewilkybarkid.

PR #2767, opened 2018-10-25
Add warning to defined documentation that it does not check for null

benolot:
I think this is a fair warning to add, as twigs behavior is inconsistent with PHP's behavior


fabpot:
Not sure it is needed. The first sentence describes the behavior quite well already.

PR #2770, opened 2018-10-31
Fix ini_get() for boolean values

deguif:


lib/Twig/Extension/Debug.php
```diff
@@ -22,7 +22,7 @@ public function getFunctions()
             && (false === ini_get('xdebug.overload_var_dump') || ini_get('xdebug.overload_var_dump'))
```
fabpot, line 22: Why not here?

fabpot (commented):
If this is only to allow people to set `html_errors` via a var env, I'm not sure this is worth it.

fabpot:
Let's close as I don't think we need this change.

PR #2773, opened 2018-11-11
Update api.rst

jgil83000:
Hello,
I propose the suppression of this useless comma in the documentation (even if it does not cause error).
Best regards.


fabpot:
As you said, it's valid PHP and moreover, it's part of our coding standard.

jgil83000:
Sorry I didn't see it in your coding standard. In this case it's me who will correct my course document to add this comma i thought was a mistake. Thank you for your reply and for the information !

PR #2777, opened 2018-11-28
 Changed private attributes/methods to protected

mys3lf:
If you want to override or extend the class, you are currently forced to rebuild everything that's private. I would like to change this and simplify this process.

### DIFF
--- a/lib/Twig/Environment.php
+++ b/lib/Twig/Environment.php
@@ -23,26 +23,26 @@ class Twig_Environment
     const RELEASE_VERSION = 0;
     const EXTRA_VERSION = 'DEV';
 
-    private $charset;
-    private $loader;
-    private $debug;
-    private $autoReload;
-    private $cache;
-    private $lexer;
-    private $parser;
-    private $compiler;
-    private $baseTemplateClass;
-    private $globals = array();
-    private $resolvedGlobals;
-    private $loadedTemplates;
-    private $strictVariables;
-    private $templateClassPrefix = '__TwigTemplate_';
-    private $originalCache;
-    private $extensionSet;
-    private $runtimeLoaders = array();
-    private $runtimes = array();
-    private $optionsHash;
-    private $loading = array();
+    protected $charset;
+    protected $loader;
+    protected $debug;
+    protected $autoReload;
+    protected $cache;
+    protected $lexer;
+    protected $parser;
+    protected $compiler;
+    protected $baseTemplateClass;
+    protected $globals = array();
+    protected $resolvedGlobals;
+    protected $loadedTemplates;
+    protected $strictVariables;
+    protected $templateClassPrefix = '__TwigTemplate_';
+    protected $originalCache;
+    protected $extensionSet;
+    protected $runtimeLoaders = array();
+    protected $runtimes = array();
+    protected $optionsHash;
+    protected $loading = array();
 
     /**
      * Constructor.
@@ -953,7 +953,7 @@ class Twig_Environment
         return $this->extensionSet->getBinaryOperators();
     }
 
-    private function updateOptionsHash()
+    protected function updateOptionsHash()
     {
         $this->optionsHash = implode(':', array(
             $this->extensionSet->getSignature(),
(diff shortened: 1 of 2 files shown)


fabpot:
Thanks for proposing but we never move a private property/method without a very good use case. What do you have in mind? You can read http://fabien.potencier.org/pragmatism-over-theory-protected-vs-private.html for more information about this topic.

mys3lf:
Hey @fabpot. Thank you for your fast response. As you describe at your page, i'm a more pragmatic developer. We are using Twig within our own cms. Our developer and i are writing plugins that are extending the functionality and styling of a base theme, which itself can be extended by another theme. It's very common that the kind of inheritance is used within our and other systems (e.g. Shopware). Let us assume that we have the following structure and we are trying to extend the same block within the files:
Base Theme -> head.twig (add basic meta tags, scripts, css)
Plugin A -> head.twig (e.g. add specific meta tags)
Plugin B -> head.twig (e.g. add google analytics)
Final Theme -> head.twig (e.g. add html class attributes or add copyright information of theme)

Currently, we are using twig namespaces, but therefore we need to know all installed plugins. What we like to achieve is to use a "@parent" namespace to get the last defined file which is not used so far. To accomplish this and to avoid the "Circular reference detected" exception, we extended the Environment and FilesystemLoader classes. In our case it's possible to call "@parent" and get the next template file that wasn't used so far. Normally, we just need to override the Twig_Environment::loadTemplate and the FilesystemLoader::findTemplate methods, but due to the private attributes it's not possible. We have to define all used attributes and methods and therefore it's almost easier to copy the whole class and change the specific parts. And therefore, i wanted to change the private attributes and methods to protected.

stof:
The issue with protected properties is that it becomes very hard to make changes in Twig without breaking compatibility, because changing any property access or method call could become a BC break if someone created a child class using them or overriding them.

For instance, resolving the template name differently depending on the place it is used would require tweaking several places in Twig to avoid breaking things, and your own overrides might break if the Twig logic changes for them.

We faced such things in the past. Twig 1.x was using `protected` almost everywhere, and several refactorings had to be reverted because of unexpected BC breaks for projects extending the classes in unexpected ways.

stof:
Btw, here is an example in Twig_Environment. The current code computes the `optionsHash` at the time options are *changed*, relying on the fact that any change to something affecting the hash must go through a public API of the class, and so we can ensure the computed hash is in sync with the actual config. Making things protected breaks that invariant. So it leaves Twig with 2 choices:
- not precomputing the hash, but generating it on the fly when needed. This has a huge performance cost as Twig reads the hash much more often that it changes (generally, you configure the environment once and then it does not get reconfigured during usage)
- trusting child classes to preserve our invariant for any change they make. But this fails in multiple cases:
    - developers extending the class are unlikely to know about the invariant, and so might break it without even knowing it
    - Twig cannot ever change what's get included in the hash, as existing child classes might deal with the other property without updating the hash as it was not part of it
    - Twig cannot add another similar hash involving existing non-private properties, as existing child classes could deal with them without updating the new hash (as it was not there when they were coded). This implies that creating a separate hash is not even an option to workaround the previous case.

Not being able to refactor the codebase means that Twig would not be able to evolve anymore.

mys3lf:
We discussed this topic a lot and our team and i totally understand your point. If you make the attributes `protected ` it would e.g. be  possible to add extensions without an invalidation of the optionHash. You would pass this responsibility to the developer that overrides your classes and this would be error-prone.

Like I described before, we had to adjust two classes to achieve our use case. I don't see any problems with the FileSystemLoader. We could omit to override your class and just write an own loader. The real problem is the adjustment within the Twig_Environment. We totally understand your points and tried to refactor our code to avoid any protected attributes. Still there is one code part that could not be changed without an protected attribute. 

```
 // to be removed in 3.0
$this->extensionSet->initRuntime($this);
```

We see that you don't want to change the attribute access level. Could you/we add the following function instead: 

```
public function initExtensionSetRuntime() {
   $this->extensionSet->initRuntime($this);
}

```

This would give us the possibility to override the default behaviour without breaking BC.

fabpot:
I'm closing this PR as we won't make everything protected for the reasons mentioned by @stof. We might want to refactor the filesystem loader at some point to be able to share some logic though.

PR #2778, opened 2018-11-30
Remove mention of the Twig C extension

yesdevnull:
As per #2318 there's no additional installation methods so it should be removed from the intro page.


fabpot:
Thank you @yesdevnull.

PR #2785, opened 2018-12-13
WIP Adds Closure to ExpressionParser

markhuot:
I wanted to learn a little more about the Twig Lexer / Parser / Compiler flow so I did some hacking and built out functionality for closures in the ExpressionParser. There are no tests yet, the classes are in `lib` and I think they’d need to be stubbed out to `src`, and lot of other little things. All that said, I wanted to open this to get an idea if this could ever reasonably be merged upstream. If not, oh well, it was a great learning experience. If so, YAY! I really do think closures in Twig would open some nice possibilities for template developers and reduce some of the clunky filter chains in use today.

Background:
[Arrow functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions) are a part of the latest JavaScript specs and are gaining popularity. Within the React/JSX community they’re really gaining traction as a convienant way to massage data in your view layer without needing to add a bunch of overly specific helpers.

Examples:
Typically if you had an array of people and you wanted to filter them down to only show people that are older than 21 you’d either have to loop over the array once using the set tag and the merge filter to create a new array (`for person in person` then `if person.age > 21` you would `set olderPeople = olderPeople|merge([person])` or you’d have to loop over the people with a for loop and if suffix (`for person in people if person.age > 21`). The first approach is verbose and error prone with how the merge filter works. The second is much cleaner but non-standard in many languages. The second approach also breaks down when you want to count the total or `|batch` the results in to columns.

Solution
Arrow functions are great for array map/reduce callbacks. Take the previous example of showing an array of people who are older than 21. Imagine you have a custom reduce filter.

```
{% set olderPeople = people|filter(person => person.age > 21) %}
```

Because the PR implements the closure logic in the ExpressionParser you should be able to use it anywhere you can use an expression such as,

```
{% set callback = person => person.age > 21 %}
{% set olderPeople = people|filter(callback) %}
```

This would also make sense to map an array of people objects in to something else,

```
<p>With support from: {{ people|map(p => "#{p.firstName} #{p.lastName}")|join(', ') }}</p>
```

The intent here is not to expose the closure to the wider context. So you could not,

```
{% set fullName = p => “#{p.fName} #{p.lName}” %}
{{ fullName(person) }}
```

### DIFF
--- a/lib/Twig/ExpressionParser.php
+++ b/lib/Twig/ExpressionParser.php
@@ -74,7 +74,10 @@ class Twig_ExpressionParser
     {
         $token = $this->parser->getCurrentToken();
 
-        if ($this->isUnary($token)) {
+        if ($this->isClosure($token)) {
+            return $expr = $this->parseClosure();
+        }
+        elseif ($this->isUnary($token)) {
             $operator = $this->unaryOperators[$token->getValue()];
             $this->parser->getStream()->next();
             $expr = $this->parseExpression($operator['precedence']);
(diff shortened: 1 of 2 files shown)


khalwat:
It'd be beautiful if this were added to core

stof:
I don't think having a Twig syntax to define callables makes much sense. If you need to do a bunch of map/reduce, you should probably not do it in the template itself.

fabpot:
I had the first same reaction as @stof, but I think we can explore this topic, especially if the focus stay local and only for simple "filters".

stof:
The current implementation also seems to only allow closures as the top-level of a primary, but not inside expression AFAICT (otherwise, it would have to deal with operator precedence), which should probably be evaluated more in details to see whether it is the right thing.

Btw, such a feature would need tests covering it.

markhuot:
> I had the first same reaction as @stof, but I think we can explore this topic, especially if the focus stay local and only for simple "filters".

Yay! Just to be clear: I'm not proposing bringing the full power of PHP in to Twig. I understand (I think… 😬) the value of Twig being a subset of PHP and the inherent safety that comes with that. Still, as alluded to, having simple filters in Twig could be really handy.

> The current implementation also seems to only allow closures as the top-level of a primary, but not inside expression AFAICT (otherwise, it would have to deal with operator precedence), which should probably be evaluated more in details to see whether it is the right thing.

Yup. Which means you couldn't use a closure in an argument, like so:

```twig
{% set people = array_merge(people, person => person.name) %}
```

I'm happy to work on this, add tests, etc… what I'm curious about, though:

I don't have a _great_ way to identify a closure during the parsing pass. The current code calls `->look($i++)` (essentially) until a `=>` operator is found. Two thoughts on that,

1. Is that okay? Can we work off the `=>` operator and use a bunch of lookaheads to identify closures?
2. Is there a better way? Should the syntax be more strict for PHP than in ES6 and require parenthesis around the arguments? Curly braces around the body?

fabpot:
I think parsing the `=>` can be easy as before it, we should have an array of arguments (like with a function call), so having the parenthesis as required could simplify things a lot. For the body, I suppose that this can be any Twig expression, so we already have the logic for that.

fabpot:
see #2996 for an implementation.

markhuot:
I am _sooo_ happy. Thank you for considering this!

Issue #1695, opened 2015-05-20
PHP7 support on Twig extension


odoucet:
I saw Twig extension is now tested against PHP7 but compile failed : 

```
phpize-7.0.0
./configure --with-php-config=/usr/bin/php-config-7.0.0 --enable-twig
make
```

Leads to : 

```
make
/bin/sh /usr/src/build/php/Twig/ext/twig/libtool --mode=compile cc  -I. -I/usr/src/build/php/Twig/ext/twig -DPHP_ATOM_INC -I/usr/src/build/php/Twig/ext/twig/include -I/usr/src/build/php/Twig/ext/twig/main -I/usr/src/build/php/Twig/ext/twig -I/usr/include/php/7.0.0/php -I/usr/include/php/7.0.0/php/main -I/usr/include/php/7.0.0/php/TSRM -I/usr/include/php/7.0.0/php/Zend -I/usr/include/php/7.0.0/php/ext -I/usr/include/php/7.0.0/php/ext/date/lib  -DHAVE_CONFIG_H  -g -O2   -c /usr/src/build/php/Twig/ext/twig/twig.c -o twig.lo
mkdir .libs
 cc -I. -I/usr/src/build/php/Twig/ext/twig -DPHP_ATOM_INC -I/usr/src/build/php/Twig/ext/twig/include -I/usr/src/build/php/Twig/ext/twig/main -I/usr/src/build/php/Twig/ext/twig -I/usr/include/php/7.0.0/php -I/usr/include/php/7.0.0/php/main -I/usr/include/php/7.0.0/php/TSRM -I/usr/include/php/7.0.0/php/Zend -I/usr/include/php/7.0.0/php/ext -I/usr/include/php/7.0.0/php/ext/date/lib -DHAVE_CONFIG_H -g -O2 -c /usr/src/build/php/Twig/ext/twig/twig.c  -fPIC -DPIC -o .libs/twig.o
/usr/src/build/php/Twig/ext/twig/twig.c:23:40: error: ext/standard/php_smart_str.h: No such file or directory
/usr/src/build/php/Twig/ext/twig/twig.c: In function 'TWIG_ARRAY_KEY_EXISTS':
/usr/src/build/php/Twig/ext/twig/twig.c:100: warning: passing argument 2 of 'zend_hash_exists' from incompatible pointer type
/usr/include/php/7.0.0/php/Zend/zend_hash.h:157: note: expected 'struct zend_string *' but argument is of type 'char *'
[...]

PR #2786, opened 2018-12-16
Added hint that the php c extension is only for php5

alexander-schranz:
As @fabpot mention in the slack channel the php c twig extension is only available for php5 and is not needed when using php7 as there the performance boost is not significant enough.

Related Issue: #1695

### DIFF
--- a/doc/installation.rst
+++ b/doc/installation.rst
@@ -55,6 +55,7 @@ Installing the C extension
     improvements. Note that the extension is not a replacement for the PHP
     code; it only implements a small part of the PHP code to improve the
     performance at runtime; you must still install the regular PHP code.
+    The C extension is only compatible and useful for **PHP5**.
 
 Twig comes with a C extension that enhances the performance of the Twig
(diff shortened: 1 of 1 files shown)


fabpot:
In any case and to make it very clear, the C extension is not compatible with PHP 7 (you cannot even compile it).

alexander-schranz:
@fabpot changed the text to hopefully make it more clear.

doc/installation.rst
```diff
@@ -55,6 +55,9 @@ Installing the C extension
     improvements. Note that the extension is not a replacement for the PHP
     code; it only implements a small part of the PHP code to improve the
     performance at runtime; you must still install the regular PHP code.
+    The C extension is only compatible with **PHP5** as the performance
+    boost is not significantly enough for PHP7 there exist no compatible
```
fabpot: I think "The C extension is only compatible with **PHP5**" is enough.

fabpot:
Thank you @alexander-schranz.

PR #2787, opened 2018-12-19
Add ??? Empty Coalesce Operator

khalwat:
Empty Coalesce adds the `???` operator to Twig that will return the first thing that is defined, not null, and not empty. This is particularly useful when you're dealing with a number of fallback/default values that may or may not exist, and may or may not be empty.

The `???` Empty Coalescing operator is similar to the `??` null coalesce operator, but also ignores empty strings (`""`), 0, 0.0, null, false, and empty arrays (`[]`) as well.

Because this is an _Empty_ Coalesce Operator, it functions identically to the PHP `empty()` function in terms of [return values](https://www.php.net/manual/en/function.empty.php#refsect1-function.empty-returnvalues).

Example:

```twig
{% set bar = null %}
{% set foo = '' %}
{% set baz = [] %}
{% set foobar = woof ??? bar ??? foo ??? baz ??? 'bark' %}
{{ foobar }}
```
This will output:
```
bark
```

...because:
- `woof` is undefined
- `bar` is `null`
- `foo` is an empty string
- `baz` is an empty array

There is precedence for this operator in languages such as Swift:

https://medium.com/@JanLeMann/providing-meaningful-default-values-for-empty-or-absence-optionals-via-nil-or-empty-coalescing-379abd22ae77


stof:
What is the difference with the `|default` filter we have.

Well, one of them is that it relies on PHP's `empty`, meaning that `'0'` is empty while `|default` does not consider that. But is there any other difference ?

stof:
btw, your implementation also ignores `'0'`, `0` and `false`, in addition to what you documented as being ignored.

khalwat:
Thanks for looking @stof !

> What is the difference with the |default filter we have.

It's the same reason why the `??` null coalescing operator exist in Twig right now, despite `default` being available. It is far more convenient and much more clear what is going on to have:

```
{% set foobar = woof ??? bar ??? foo ??? baz ??? 'bark' %}
```
...using the `default` filter. It can be done, but it is quite ugly and verbose.

michaelrog:
I'm a big fan of this idea — It comes up frequently enough in my Twig templates (where I'm comparing a chain of values and I want to use/output the first non-empty one) that I've created my own stock extension for it, and it's a default add to all my projects.

I think `???` as an operator could be confusing vis-a-vis `??`. And also, the visual difference between `??` and `???` is not super-scannable. Perhaps `_?` for this.

My only reservation about this in-practice is the case of `"0"` being empty.... but, one can't really do anything about it; I think ultimately it's reasonable to be consistent with PHP's `empty()` and expect devs to learn that behavior.

It's definitely a non-standard op, but I think the utility of it is worth adding the sugar to Twig even though no similar operator exists upstream in PHP.

adrienne:
+1 to this PR - i am using @khalwat's null-coalescing operator as a plugin on a site and it is *really* nice to have in any context where your Twig data contains, for instance, a lot of empty arrays.

nicolas-grekas:
Why not use `?:` to provide this behavior?

khalwat:
> I think ??? as an operator could be confusing vis-a-vis ??. And also, the visual difference between ?? and ??? is not super-scannable. Perhaps _? for this.

`_?` is horrible to type; I think `???` is easier to type, and clearer in that it's `??` plus more. The fact that [others have independently](https://medium.com/@JanLeMann/providing-meaningful-default-values-for-empty-or-absence-optionals-via-nil-or-empty-coalescing-379abd22ae77) come up with the same `???` operator for this functionality makes me feel even better about it.

> My only reservation about this in-practice is the case of "0" being empty.... but, one can't really do anything about it; I think ultimately it's reasonable to be consistent with PHP's empty() and expect devs to learn that behavior.

It actually uses `empty()` under the hood, so yes, it'd be consistent with `empty()`'s behavior, which I think makes sense.

michaelrog:
@nicolas-grekas — The primary benefit is that `?:` will throw an error if it encounters an undefined operand. We'd really like this to behave like the null-coalescing operator, which treats undefined things as `null`.

(Also, it needs to be right-associative... I _think_ `?:` is, but I don't remember offhand.)

khalwat:
> Why not use ?: to provide this behavior?

@nicolas-grekas  what would the Twig code for this be using the `?:` operator for this construct:

`{% set foobar = woof ??? bar ??? foo ??? baz ??? 'bark' %}
`

In PHP `?:` just determines truthiness, and would throw an error if a variable was undefined, unlike `???`

nicolas-grekas:
> and would throw an error if a variable was undefined

twig is a different language, we can remove this behavior if we want to.

khalwat:
> twig is a different language, we can remove this behavior if we want to.

That's definitely true, but I'd rather define a new operator for new functionality than hoist new functionality on an operator for which there's already a defined behavior.

Many people who write in Twig also write in PHP, and it might be very confusing if `?:` operated differently in Twig than it does in PHP... at least to me.

stof:
> twig is a different language, we can remove this behavior if we want to.

that would require changing the existing operator, and that makes mistake detection harder in case you make typos in your variable names. I would vote for keeping the existing behavior of the operator.

brandonkelly:
Another vote to not modify existing operator behaviors! (And a vote for `???`)

khalwat:
Welp, looks like the `??=` operator is being added in PHP 7.4: https://twitter.com/nikita_ppv/status/1087662379037528064

It's not the same thing, but it does show some precedence for a three character operator... so I hope `???` can make it into Twig!

lib/Twig/Node/Expression/EmptyCoalesce.php
```diff
@@ -0,0 +1,44 @@
...
+            ['left' => $left, 'right' => $right],
+            ['ignore_strict_check' => true, 'is_defined_test' => false],
+            $lineno
+        );
+    }
+
+    public function compile(\Twig_Compiler $compiler)
+    {
+            $compiler
+                ->raw('((empty(')
```
fabpot: Using `empty` is a no go for me (see @stof comment about edge cases that makes `empty` far from being useful in this case).

fabpot (changes requested):

brandonkelly:
> btw, your implementation also ignores '0', 0 and false, in addition to what you documented as being ignored.

If we’re calling this the “_Empty_ Coalesce Operator”, I would expect that any value PHP considers “empty” would also be considered empty here, so `'0'` and `0` should in fact invoke the following operator argument. Safe to assume that if PHP ever adds the same operator, they would go with the same behavior as `empty()` as well.

khalwat:
Yeah that's the idea @brandonkelly -- it's called the Empty Coalesce Operator because the behavior is what we'd expect from chained/nested `empty()` calls.

jfcherng:
https://github.com/twigphp/Twig/pull/2787/files#diff-b445caeb1cc1391b4cc1966bd1b1f76cR28

After compilation, wouldn't the current implementation eventually evaluate both the left-hand side and the right-hand side twice at runtime? I mean it's possible that `func1` or `func2` in `func1(arg1) ??? func2(arg2)` be executed twice? In PHP, both `L` and `R` in `L ?? R` would be only evaluated once.

If there is a static variable (such as a static counter to log how many time the func has been executed) in `func1` or `func2`, a counterintuitive behavior would happen?

lib/Twig/Node/Expression/EmptyCoalesce.php
khalwat: I'd echo the comments from @brandonkelly below; it's called the Empty Coalesce Operator, so having it behave the way `empty()` does makes sense?

stof: but by PHP's empty rules, or by Twig's rules ? The `is empty` test is **not** using PHP's `empty` due to all the weird edge cases it has.

khalwat: Ah interesting, I didn't know that! Is there a way to call Twig's internal `empty` from the EmptyCoalesce node parser?

Give me a pointer, and I'll get it done!

brandonkelly: @stof You could make that same argument about `??`, the **null**-coalescing operator, which in addition to checking if a value is null will also check if it’s set at all, despite that not being something that `var == null` would cover. And yet `??` exists in Twig. I’d say that’s enough precedence for an “**empty**-coalescing operator” to have a slightly different definition of “empty” than an unrelated `empty` test.

khalwat: Okay, I updated it to use `twig_test_empty()` rather than PHP's `empty()` and also updating the namespacing!

Hopefully this will help get it accepted?

khalwat:
@jfcherng suggestions on how to mitigate the behavior you've mentioned?

lib/Twig/Node/Expression/EmptyCoalesce.php
khalwat: I wanted to call `twig_test_empty()` in `Twig\Extension\CoreExtension.php` but the method isn't exposed. So instead I just copied that function verbatim.

This obviously begs for refactoring that method out so that both the Twig `empty` test and the `???` empty coalesce operator use the same core method.

Not being that familiar with the codebase, though, I'm not sure how the maintainers would want to approach that.

jfcherng:
@khalwat Sorry I cannot really answer that. But [that is what I saw](https://github.com/php/php-src/pull/1795#issuecomment-387184002) when I was following the `??=` RFC. Initially, `L ??= R` was implemented as a simple syntax sugar for `L = L ?? R`, but soon problems came out.

> For example, consider the expression $a[print 'X'] ??= $b. A simple desugaring into $a[print 'X'] = $a[print 'X'] ?? $b will result in 'X' being printed twice. However, this is not how all other existing compound assignment operators behave: They will print X only once, as the LHS is only evaluated once. I assume that ??= would behave the same way.

```php
        $compiler
            ->raw('(('.self::class.'::empty(')
            ->raw(') ? null : ')
            ->raw(') ?? ('.self::class.'::empty(')
            ->subcompile($this->getNode('right'))
            ->raw(') ? null : ')
            ->subcompile($this->getNode('right'))
            ->raw('))')
        ;
```

With this, I am assuming `L ??? R` will be de-sugared into something like 
```php
(empty(L) ? null : L) ?? (empty(R) ? null : R)
```
The worst case is evaluating both L and R twice when executing the compiled template.

Maybe worth a note in the docs if this cannot be resolved?

dharkness:
How about using `??:` instead? From what I can tell, this new operator combines the effects of the existing `??` and `?:` operators in PHP.

brandonkelly:
@dharkness `?:` can’t be used in succession like `??` can though, e.g. `foo ?? bar ?? baz`

dharkness:
@brandonkelly I can only speak for PHP, but both `??` and `?:` can be used in a series.

    > $y = null;
    > echo $x ?? $y ?? 'foo';
    foo
    > echo 0 ?: '' ?: 'foo';
    foo

lib/Twig/Node/Expression/EmptyCoalesce.php
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+use Twig\Node\Expression\EmptyCoalesceExpression;
+
+@trigger_error(sprintf('Using the "Twig_Node_Expression_EmptyCoalesce" class is deprecated since Twig version 2.7, use "Twig\Node\Expression\EmptyCoalesceExpression" instead.'), E_USER_DEPRECATED);
+
+class_exists('Twig\Node\Expression\EmptyCoalesceExpression');
+
+if (\false) {
+    class Twig_Node_Expression_EmptyCoalesce extends EmptyCoalesceExpression
```
stof: introducing the legacy class and deprecating it directly does not make sense.

lib/Twig/Extension/Core.php
```diff
@@ -1,14 +1,1726 @@
...
+use Twig\TokenParser\SetTokenParser;
+use Twig\TokenParser\SpacelessTokenParser;
+use Twig\TokenParser\UseTokenParser;
+use Twig\TokenParser\WithTokenParser;
+use Twig\TwigFilter;
+use Twig\TwigFunction;
+use Twig\TwigTest;
 
-@trigger_error(sprintf('Using the "Twig_Extension_Core" class is deprecated since Twig version 2.7, use "Twig\Extension\CoreExtension" instead.'), E_USER_DEPRECATED);
+final class CoreExtension extends AbstractExtension
```
stof, line 73: this change is wrong

lib/Twig/Node/Expression/EmptyCoalesce.php
khalwat: I'd be happy to remove this; would you like me to do that?

lib/Twig/Extension/Core.php
khalwat: Correct, the error message is wrong (a copy and paste error) but if we want this removed as per your earlier comment, then it probably doesn't matter. It'd be fixed by just removing the whole deprecation nonsense.

lib/Twig/Node/Expression/EmptyCoalesce.php
khalwat: I pushed a change to remove the pointless `Twig_Node_Expression_EmptyCoalesce` class entirely, which should resolve this.

lib/Twig/Extension/Core.php
khalwat: I pushed a change to remove the pointless `Twig_Node_Expression_EmptyCoalesce` class entirely, which should resolve this.

stof: this is not fixed. You are still making a huge mess in the CoreExtension file, reverting the legacy file to a full class.

khalwat: I'm using this to merge into: https://github.com/twigphp/Twig/blob/2.x/src/Extension/CoreExtension.php

Should I be using it from a different branch?

khalwat: @stof I'd love to get this addressed, but I could use some guidance.

The only reason there is a change to `lib/Twig/Extension/Core.php` is to include the `???` operator. To fix this so that it's not creating a "huge mess" I just need to know what branch/source I should be diff'ing with to add this line?

stof: you should not be making any change in `lib/Twig/Extension/Core.php`, which is a legacy BC file (except you are reverting this to not be a BC file anymore). The change should be done only in `src/Extension/CoreExtension.php`

stof: your PR is reverting #2863 on this file, which is what the huge mess is.

khalwat: I see. That change to make `lib/Twig/Extension/Core.php` wasn't made until after the PR I made, thus my confusion about what's going on here.

I will revert the now legacy `lib/Twig/Extension/Core.php` and make my changes in `src/Extension/CoreExtension.php`

GromNaN:
Your example is easy to do with operators `??` and `?:`.

```twig
{% set bar = null %}
{% set foo = '' %}
{% set baz = [] %}
{% set foobar = woof ?? null ?: bar ?: foo ?: baz ?: 'bark' %}
{{ foobar }}
```

khalwat:
@GromNaN that presupposes that you _know_ the state of the variables. The point is that any of them could be undefined, null, or empty, and you don't know ahead of time.

GromNaN:
I did't imply this feature was not useful, just trying to help using the existing features.
So `<variable> ??? <default>` would be a shortcut for `(<variable> ?? null) ?: <default>`.

If the existence of any variable is unknown, the example can be:
```
{% set foobar = (woof ?? null) ?: (bar ?? null) ?: (foo ?? null) ?: (baz ?? null) ?: 'bark' %}
```

brandonkelly:
@GromNaN I don’t think anyone here is unaware of existing syntax options. You could have made the same point about `??` … that `<variable> ?? <default>` it is the same thing as doing `(<variable> is defined and <variable> is not same as(null) ? <variable> : <default>`. The point is to simplify, to make templates more readable & maintainable.

GromNaN:
Excepted that with `??` the `<variable>` have to be repeated several times. Which is very verbose when `<variable>` contains a complex expression.

acalvino4:
What's the status of this? Thought I'd bump it up since it seems it got abandoned without any particular reason.

Did the team decide this isn't desirable? the comment voting from the community seems to suggest otherwise.
Is it that the PR needs more work? as far as I can tell there isn't any specific request of what it needs.

So I guess the question is
1. Are the maintainers open to this?
2. If not, why, given that the community wants it and every objection so far has been answered?
3. If so, is there something that needs to be done on the PR to get it merged?

khalwat:
Currently, the PR is borked. I'd be happy to attempt to redo it if there is interest from the maintainers.

maxstrebel:
I want to add my +1 on that. And thanks for the work on that.

`{% set theme = entry is defined and entry.theme != "" ? entry.theme : defaults.theme != "" ? defaults.theme : "dark" %}`
vs.
`{% set theme = entry.theme ??? defaults.theme ??? "dark" %}`

The readability is just a whole other level.

jdreesen:
Wouldn't `??:` be the better choice, since it's a combination of the operators `??` and `?:`?

maxstrebel:
> Wouldn't `??:` be the better choice, since it's a combination of the operators `??` and `?:`?

I guess you're right. 

I stumbled upon this via a plugin that allows to use `???` and it felt _natural_ to me. But I actually do not care too much as long as I can use it 😄

VincentLanglet:
Just my two cents, but the first thing I thought about is `???` was "?:` does the same thing until I read
> In PHP ?: just determines truthiness, and would throw an error if a variable was undefined

`??:` seems to be more understandable then.
$a ??: $b = ($a ?? null) ?: $b

It could even be a RFC proposed to PHP first, in order to have the same symbol for both twig and php.

khalwat:
@VincentLanglet would love to see an RFC for it in PHP, too.

The reason I prefer `???` is simply that it's easier to type, and it's been used in other languages before, to do the same thing.

VincentLanglet:
> The reason I prefer `???` is simply that it's easier to type, and it's been used in other languages before, to do the same thing.

Do you have examples of other languages with `???` ?

khalwat:
@VincentLanglet link is in the first post in the thread; someone added it to Swift using `???`

ryanleichty:
Chiming in to see if there's still a possibility to add this to Twig! We use this in our Craft CMS project and love it.

This plus [optional chaining](https://github.com/twigphp/Twig/issues/3260) added to Twig would be amazing 🔥

ryanleichty:
Ok, I’m still a big fan of the empty coalesce operator, but I would want it to more closely align with how the null coalescing operator traditionally works. As I discovered and reported in this [issue](https://github.com/nystudio107/craft-emptycoalesce/issues/40), `[] ??? []` coerces the righthand operand to `null` rather than returning the empty array value.

**Null coalesce**

- `null ?? null` returns `null`
- `null ?? ''` returns `''`
- `null ?? []` returns `[]`

**Empty coalesce**

- `null ??? null` returns `null`
- `null ??? ''` returns `null`
- `null ??? []` returns `null`

PR #2796, opened 2019-01-08
Switch to short array notation

fabpot:


lib/Twig/Autoloader.php
```diff
@@ -30,9 +30,9 @@ public static function register($prepend = false)
         @trigger_error('Using Twig_Autoloader is deprecated since version 1.21. Use Composer instead.', E_USER_DEPRECATED);
 
         if (PHP_VERSION_ID < 50300) {
```
stof, line 32: this is dead code

lib/Twig/Autoloader.php
```diff
@@ -47,7 +47,7 @@ public static function autoload($class)
             return;
         }
 
-        if (is_file($file = dirname(__FILE__).'/../'.str_replace(array('_', "\0"), array('/', ''), $class).'.php')) {
+        if (is_file($file = dirname(__FILE__).'/../'.str_replace(['_', "\0"], ['/', ''], $class).'.php')) {
```
stof, line 50: can be `__DIR__` now

PR #2798, opened 2019-01-09
Added support for `css_class` function

### LINKED ISSUE #2797: [RFC] Added support for classes function
The [`classes`](https://github.com/JedWatson/classnames) function in javascript is quite useful.
Do you think it could be in twig ?
I would port it if you want

lyrixx:
fixes #2797


lib/Twig/Extension/Core.php
```diff
@@ -1634,4 +1635,31 @@ function twig_get_attribute(Twig_Environment $env, Twig_Source $source, $object,
...
+{
+    $classes = [];
+
+    foreach ($args as $i => $arg) {
+        if (is_string($arg)) {
+            $classes[] = $arg;
+        } elseif (is_array($arg)) {
+            foreach ($arg as $class => $condition) {
+                if (!is_string($class)) {
+                    throw new Twig_Error_Runtime(sprintf('Key #%d of argument #%d should be a string. Got "%s".', $class, $i, gettype($class)));
```
yceruto: For `2.x` branch use `RuntimeError` instead (Same below)

yceruto (commented):
Shouldn't this be implemented for `1.x` branch first?

doc/functions/classes.rst
```diff
@@ -0,0 +1,12 @@
+``classes``
+============
```
yceruto: extra `=`

lib/Twig/Extension/Core.php
yceruto: Nevermind, this class doesn't use the `2.x` namespace anyway.

stof: Saying `Key #2` makes me think it is the 3rd item in the array (as we count from 0). But `array(2 => true)` would defeat that. I think you should remove the `#` here.

lib/Twig/Extension/Core.php
```diff
@@ -1634,4 +1635,31 @@ function twig_get_attribute(Twig_Environment $env, Twig_Source $source, $object,
     return $ret;
 }
 
+function twig_classes(...$args): string
+{
+    $classes = [];
+
+    foreach ($args as $i => $arg) {
+        if (is_string($arg)) {
```
stof: should we allow a `Twig_Markup` object here, by casting it ?

lyrixx: I guess yes :) I will update the PR

lyrixx:
What should be the base branch?

stof:
I think it should be 1.x. AFAIK, 1.x still gets minor version releases.

fabpot:
IIUC, it's purely related to HTML. If that's the case, I don't think it belongs to the core.

fabpot:
Might be more appropriate in Twig-extensions.

fabpot:
At least, the name should be changed to something like `css_class(es)` as suggested by @javiereguiluz in the linked issue.

lib/Twig/Extension/Core.php
lyrixx: @stof I added a test for that. I did not had to update the code :+1:

lib/Twig/Extension/Core.php
lyrixx: Thanks Fixed

lyrixx: Fixed

doc/functions/classes.rst
lyrixx: Thanks Fixed

lyrixx (commented):
I addressed your comments. Thanks

carlcs:
For inspiration, in Craft CMS we've got a `attr()` function ([see docs](https://github.com/craftcms/cms/blob/3.2/docs/dev/functions.md#attr)).

```twig
{% set myAttributes = {
    class: ['one', 'two'],
    disabled: true,
    readonly: false,
    data: {
        baz: 'Escape this "',
        qux: {
            some: ['data', '"quoted"']
        }
    },
    style: {
        'background-color': 'red',
        'font-size': '20px'
    },
} %}

<div {{ attr(myAttributes) }}></div>
```

For conditional classes you can use ternary notation in the array.

```twig
{% set myAttributes = {
    class: [
        'one',
        loop.last ? 'two',
    ],
} %}

<div {{ attr(myAttributes) }}></div>
```

If `class` is the only attribute you want to output.

```twig
{% set class = [
    'one',
    loop.last ? 'two',
] %}

<div {{ attr({ class: class }) }}></div>
```

Another great use case where I would love to see shorthand notation for object props become a thing in Twig. Maybe one day… https://github.com/twigphp/Twig/pull/3002

jvasseur:
AFAI class are not part of CSS but of HTML and then other specs that build on top of HTML can use them (CSS class selectors and Javascript `getElementsByClassName`).

According to this I don't think we should use CSS in the name of the function.

fabpot:
I've rebased to current 2.x and made some minor changes. The biggest one being the move from the Core extension to the new HTML extension and the renaming to `html_classes`.

fabpot:
Thank you @lyrixx.

lyrixx:
Thanks Fabien :)

PR #2813, opened 2019-01-14
fix key exists check for non ArrayObject objects

xabbuh:
fixes #2810

PR #2815, opened 2019-01-14
Revert "minor #2809 Fixed 2.x tests (yceruto)"

xabbuh:
This reverts commit 59c26ae211d849fdcacafa6e0f7f7b31d0f7f513, reversing
changes made to 3fd85beda4f39f800f4eb85a0a27a26298624157.

### DIFF
--- a/test/Twig/Tests/TemplateTest.php
+++ b/test/Twig/Tests/TemplateTest.php
@@ -606,7 +606,9 @@ class Twig_TemplatePropertyObjectAndArrayAccess extends Twig_TemplatePropertyObj
         'zero' => 0,
         'null' => null,
         'bar' => true,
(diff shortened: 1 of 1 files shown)


xabbuh:
see https://github.com/twigphp/Twig/pull/2813#pullrequestreview-192151553

yceruto (approved):

stof:
Should these tests be backported to 1.x ?
AFAICT, this test failures were actually detecting the regression (they failed because the public properties were not accessed properly). And so having them in 1.x would have prevented the bug already.

xabbuh:
as far as I can see they are now "just" duplicates of the test added in #2813

fabpot:
Thank you @xabbuh.

PR #2825, opened 2019-02-04
remove duplicated directory separator

xabbuh:
The $rootPath property is already initialized with a trailing directory
separator in the class constructor.


xabbuh:
see https://github.com/twigphp/Twig/commit/a343c92805c3a9b0f982b19914dc301bb6d37f3a#r32179193

fabpot:
Thank you @xabbuh.

PR #2834, opened 2019-02-13
Clarify the difference between the include function and tag

### LINKED ISSUE #2829: Documentation should call out difference between include tag and include() function
The `include()` function was added in f4f88a51e275d380663b04bd371b215ef70bb7f0 (10 months after #605 was closed as wontfix), with a line in the changelog that says that it has identical functionality to the `include` tag. #1899 suggested that the `include` tag be deprecated in favor of the `include()` function, which was declined, in part because of the suggestion that the output of the `include()` function could be assigned to a variable. However, the response to #1095 suggests that that is not the case.

So at this point, I am unclear whether there is any difference between the `include` tag and the `include()` function, and the documentation does not even indicate that they both exist for generally the same purpose.

At the very least, I think the following pages should be updated to mention each other:

https://twig.symfony.com/doc/2.x/tags/include.html
https://twig.symfony.com/doc/2.x/functions/include.html

fabpot:
should close #2829

### DIFF
--- a/doc/advanced.rst
+++ b/doc/advanced.rst
@@ -244,8 +244,9 @@ last argument to the filter call as an array::
         // ...
     }, ['is_variadic' => true]);
 
-Be warned that named arguments passed to a variadic filter cannot be checked
-for validity as they will automatically end up in the option array.
+Be warned that :ref:`named arguments <named-arguments>` passed to a variadic
+filter cannot be checked for validity as they will automatically end up in the
+option array.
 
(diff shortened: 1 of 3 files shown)


GPHemsley:
Is it documented anywhere else that "a tag should not display anything"? Where does that leave `embed`?

stof:
embed defines some structure of the template, so it cannot be implemented in another way.

Maybe the clarification sentence is not the best out to describe that.

PR #2842, opened 2019-02-15
Tweak PHP-CS-Fixer config

fabpot:


.php_cs.dist
```diff
@@ -11,6 +11,7 @@ return PhpCsFixer\Config::create()
         'heredoc_to_nowdoc' => false,
         'ordered_imports' => true,
         'phpdoc_types_order' => ['null_adjustment' => 'always_last', 'sort_algorithm' => 'none'],
+        'native_function_invocation' => ['include' => ['@compiler_optimized'], 'scope' => 'all'],
```
stof, line 14: isn't it already part of `@Symfony:risky` in the latest versions of PHP-CS-Fixer ?

SpacePossum: the `'scope'` default for the Symfony set is `'namespaced'` and not `'all'`
(https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.14.2/src/RuleSet.php#L189)

PR #2849, opened 2019-02-17
Fix undefined class Profile in templates

chalasr:
Fixes https://travis-ci.org/symfony/symfony/jobs/494506116.
~~The calling line is https://github.com/twigphp/Twig/blob/2.x/lib/Twig/Profiler/Node/EnterProfile.php#L34, let me know if we should just rollback to `new \Twig\Profiler\Profile()` there.~~ just reverted the failing line to its original state (using FQCN)

### DIFF
--- a/lib/Twig/Profiler/Node/EnterProfile.php
+++ b/lib/Twig/Profiler/Node/EnterProfile.php
@@ -11,7 +11,6 @@
 
 use Twig\Compiler;
 use Twig\Node\Node;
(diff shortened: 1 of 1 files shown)


xabbuh:
The `use` statement is not needed, is it?

chalasr:
@xabbuh use statement removed

fabpot:
Thank you @chalasr.

PR #2850, opened 2019-02-22
Adding psr-4 support without aliases 

melbahja:
Aliases removed and psr-0 style replaced with psr-4

BUT i have 6 test failures!

you can clone : https://github.com/melbahja/Twig and run simple-phpunit

Thanks


stof:
Removing non-namespaced classes is a BC break. This cannot be merged into 2.x.

melbahja:
@stof i agree with you maybe in new branch 3.x

fabpot:
I'm working on the migration. Stay tuned :)

PR #2854, opened 2019-02-27
Add the email file extension to the escaping strategy guesser

fabpot:


stof:
so, this is for things like `foo.email.twig` ? But this assumes that such a template is a plain-text email, not an HTML email.

fabpot:
The idea is more something like this:

```twig
{% block subject %}
{% endblock %}

{% block text %}
{% endblock %}

{% block html %}
    {% autoescape 'html' %}
    {% endautoescape %}
{% endblock %}
```

stof:
I have similar templates in the mailing system of Incenteev. But I decided to actually use `{% autoescape false %}` in the `subject` and `text` blocks rather than having developers re-enable HTML escaping in `html`.
Forgetting to disable escaping in the subject and body text creates visual bugs (extra HTML escaping) while forgetting to enable escaping in the HTML body creates security issues (XSS).

And this gets even worse: if you have any `{% embed %}` inside your `html` block, you also have to remember to reenable HTML escaping for all of them, because the anonymous inline template is not actually a child node of your `{% autoescape 'html' %}` node. It is a node attached to the root of the template (the `Twig_Module` node). Only the rendering of that template is a child node of your `{% autoescape 'html' %}` node.

PR #2855, opened 2019-02-27
Encourage using prepared statements

ntzm:
Although the previous code isn't vulnerable, it could be if copied and used to store user input.


stof:
The previous code is actually vulnerable if the template source contains some single quote.

SpacePossum (approved):

fabpot:
Thank you @ntzm.

PR #2857, opened 2019-03-03
added Twig\Twig as the prefered entrypoint for Environment

### LINKED ISSUE #2694: [RFC] Rename (or alias) Environment -> Twig
It always a bit weird to write the following code:

```php
    public function __construct(Environment $twig)
    {
        $this->twig = $twig;
    }
```

I would prefer to type-hint with Twig instead of Environment. 

WDYT ?

### LINKED ISSUE #2856: Alias Twig_Environment as Twig_Twig & Twig\Twig
Twig is a concept, but its main entry point `Environment` (im using the namespaced classes) looks 100% arbitrary.

We fix it using `Environment $twig`, but ideally this be `Twig $twig` (or `TwigEnvironment` if you like) IMHO.

Im here to see if putting `Environment as TwigEnvironment` everywhere can be avoided by doing it in core :) if not  i guess we'll settle with  `Environment $twig` for convenience :cry: 

It could also be an alternative alias to avoid deprecation hassle.

WDYT?

fabpot:
closes twigphp/Twig#2856
closes twigphp/Twig#2694


src/Environment.php
```diff
@@ -5,6 +5,8 @@
 class_exists('Twig_Environment');
 
 if (\false) {
+    @trigger_error(sprintf('Using the "Twig\Environment" class is deprecated since Twig version 1.38, use "Twig\Twig" instead.'), E_USER_DEPRECATED);
```
ogizanagi: This deprecation will never be triggered. It shouldn't be enclosed in the above condition, right?

src/Environment.php
```diff
@@ -5,6 +5,8 @@
...
+
     class Environment extends \Twig_Environment
```
ogizanagi: What about adding `@deprecated`, so IDEs can highlight this class is deprecated?

fabpot:
I was working on making it work and I realized that the word "environment" is used in many other parts of Twig, like in the `needs_environment` option (for filters, functions, ...). So, renaming `Twig\Environment` to `Twig\Twig` is probably not a good idea as it would imply renaming `needs_environment` to something else as well.

The other option could be to rename `Twig\Environment` to `Twig\TwigEnvironment`, but now that I can see the work involved in doing so in 1.x and 2.x (including docs changes), I've decided to give up. Not worth it. Sorry.

ro0NL:
honestly i think `Twig\Twig` still represents _the_ "environment", i dont see the correlation between the classname and the option name here.

Within twig itself, the `Twig` instance implies the environment.

fabpot:
Think of someone who never saw the legacy `Environment` class.
Anyway, as I said, there are other reasons, migration path, deprecation notices, documentation updates, that's a huge amount of work. I realized that when working on this PR.

ericmorand:
Plus, Twig is the language _and_ the namespace. Environment is the compilation and rendering environment. I don't see any confusion here. Having the word Twig being used to represent the environment would lead to even more confusion: Twig, what is it? The language? The namespace? The compilation environment?

PR #2864, opened 2019-03-05
Deprecate PSR-0 class names

fabpot:


stof:
this won't work if the old name is used *after* the namespaced name, as the class_alias will already have been defined.

fabpot:
@stof I know

stof:
however, you should add the `@deprecated` on this "fake" child classes, so that IDEs know it is deprecated.

fabpot:
Just noticed that version is wrong in the deprecated message, upcoming PR.

fabpot:
@stof see #2865

PR #2869, opened 2019-03-06
Move some more PSR0 class names to namespaced ones

fabpot:


src/Node/ModuleNode.php
```diff
@@ -33,7 +33,7 @@ class ModuleNode extends Node
     public function __construct(\Twig_NodeInterface $body, AbstractExpression $parent = null, \Twig_NodeInterface $blocks, \Twig_NodeInterface $macros, \Twig_NodeInterface $traits, $embeddedTemplates, $name, $source = '')
```
stof, line 33: shouldn't the typehints be updated too ?

fabpot: which ones? Twig_NodeInterface is deprecated, so we don't have a namespaced equivalent.

stof: ah indeed, I forgot that.

PR #2870, opened 2019-03-06
document support for defining default argument values in the macro tag

olets:
The `macro` tag documentation reads

> Macros differ from native PHP functions in a few ways:
> - Default argument values are defined by using the default filter in the macro body

and provides the example

```twig
{% macro input(name, value, type, size) %}
    <input type="{{ type|default('text') }}" name="{{ name }}" value="{{ value|e }}" size="{{ size|default(20) }}" />
{% endmacro %}
```

The implication is that native PHP function-like default argument value definition is not supported.

v1.12 added support for

```twig
{% macro input1(name, value, type = 'text', size = 20) %}
    <input type="{{ type }}" name="{{ name }}" value="{{ value|e }}" size="{{ size }}" />
{% endmacro %}
```

This updates the documentation to cover both ways of defining default argument values.

### DIFF
--- a/doc/tags/macro.rst
+++ b/doc/tags/macro.rst
@@ -13,10 +13,16 @@ Here is a small example of a macro that renders a form element:
         <input type="{{ type|default('text') }}" name="{{ name }}" value="{{ value|e }}" size="{{ size|default(20) }}" />
     {% endmacro %}
 
-Macros differ from native PHP functions in a few ways:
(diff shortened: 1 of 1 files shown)


stof:
should be merged in the 1.x branch

olets:
Thank @stof I wondered if that might be the case. Updated

stof (approved):

fabpot:
Thank you @olets.

PR #2871, opened 2019-03-06
Fix some issues

fabpot:


stof:
isn't it a BC break to change it to `-2` instead of `false` for disabling the line ?

fabpot:
That's  purely internal code.

PR #2872, opened 2019-03-07
Add spaceless filter

fabpot:


stof:
I suggest adding a note in the doc about the performance impact of this. Spaceless happens at runtime, and calling it a lot can hurt performance.
In the past, Symfony optimized the form rendering a lot by using whitespace control instead of having lots of spaceless tags (note that this is not a general replacement, but it was a solution in this case as removed spaces in the Symfony form rendering were all coming from the template source code, and so could be stripped at compile-time instead).

src/TokenParser/SpacelessTokenParser.php
```diff
@@ -30,13 +33,23 @@ class SpacelessTokenParser extends AbstractTokenParser
...
+        $ref = new BlockReferenceExpression(new ConstantExpression($name, $token->getLine()), null, $token->getLine(), $this->getTag());
+
+        $filter = $this->parser->getExpressionParser()->parseFilterExpressionRaw($ref, $this->getTag());
         $this->parser->getStream()->expect(Token::BLOCK_END_TYPE);
+
         $body = $this->parser->subparse([$this, 'decideSpacelessEnd'], true);
         $this->parser->getStream()->expect(Token::BLOCK_END_TYPE);
 
-        return new SpacelessNode($body, $lineno, $this->getTag());
+        $block = new BlockNode($name, $body, $token->getLine());
```
stof, line 49: Using a block node here is a BC break: a block creates its own isolated block, while `{% spaceless %}` is not a scoping element.

fabpot: This part has been reverted since the merge.

fabpot:
@stof note added in 68bbf5eb

emodric:
@fabpot What's the reason for deprecating the tag? Removing code duplication or something else?

stof:
Removing duplication, and also advocating for the right usage of extension points. Modifying a string to remove spaces is what a filter should be doing, not a tag.

emodric:
Thanks @stof !

PR #2874, opened 2019-03-10
Remove runtime deprecation that prevent smooth update

nicolas-grekas:
This deprecation is already triggered when a class that implements `InitRuntimeInterface` is loaded.
This runtime deprecation prevents creating a continuous upgrade path.


fabpot:
Thank you @nicolas-grekas.

PR #2875, opened 2019-03-10
Fix deprecation notes

nicolas-grekas:
`ExpressionParser` still uses `since version [...]` because the message is not only about Twig: any package providing extensions can trigger it.


fabpot:
Thank you @nicolas-grekas.

PR #2877, opened 2019-03-11
Tweak sandbox generated code

fabpot:


src/NodeVisitor/SandboxNodeVisitor.php
```diff
@@ -77,7 +77,7 @@ protected function doLeaveNode(Node $node, Environment $env)
         if ($node instanceof ModuleNode) {
             $this->inAModule = false;
 
-            $node->setNode('display_start', new Node([new CheckSecurityNode($this->filters, $this->tags, $this->functions), $node->getNode('display_start')]));
+            $node->setNode('constructor_end', new Node([new CheckSecurityNode($this->filters, $this->tags, $this->functions), $node->getNode('display_start')]));
```
fabpot, line 80: Throwing sandbox exceptions earlier is better, no need to wait for the display.

PR #2879, opened 2019-03-11
Try to avoid guessing template info on errors

fabpot:
Right now, for syntax errors triggered in a Node, we guess the template line and source, but we then try to set the information in the Parser class. This PR avoids the guessing as it's slow.

### DIFF
--- a/src/Template.php
+++ b/src/Template.php
@@ -221,8 +221,7 @@ abstract class Template implements \Twig_TemplateInterface
 
                 // this is mostly useful for \Twig\Error\LoaderError exceptions
                 // see \Twig\Error\LoaderError
-                if (-2 === $e->getTemplateLine()) {
-                    $e->setTemplateLine(-1);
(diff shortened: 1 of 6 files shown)


src/Template.php
```diff
@@ -221,8 +221,7 @@ public function displayBlock($name, array $context, array $blocks = [], $useBloc
 
                 // this is mostly useful for \Twig\Error\LoaderError exceptions
                 // see \Twig\Error\LoaderError
-                if (-2 === $e->getTemplateLine()) {
-                    $e->setTemplateLine(-1);
+                if (-1 === $e->getTemplateLine()) {
```
stof, line 224: couldn't we avoid this line guessing in most cases by passing the appropriate location to the `displayBlock` method when compiling a BlockReferenceNode instead ?

fabpot: sure, and that's more or less what we are already doing in `Template::loadTemplate()`.

PR #2884, opened 2019-03-12
Fix "batch filter clobbers array keys when fill parameter is used "

### LINKED ISSUE #2568: batch filter clobbers array keys when fill parameter is used
batch filter clobbers array keys when fill parameter is used

On the last array chunk if there are fewer keys than the chuck size when the batch filter fills in the missing value it re-numbers the array keys.

fabpot:
closes #2568

### DIFF
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -1638,23 +1638,20 @@ function twig_constant_is_defined($constant, $object = null)
  *
  * @return array
  */
-function twig_array_batch($items, $size, $fill = null)
+function twig_array_batch($items, $size, $fill = null, $preserveKeys = true)
 {
     if ($items instanceof \Traversable) {
-        $items = iterator_to_array($items, false);
+        $items = iterator_to_array($items, $preserveKeys);
     }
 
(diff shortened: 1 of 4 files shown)


src/Extension/CoreExtension.php
```diff
@@ -1638,23 +1638,22 @@ function twig_constant_is_defined($constant, $object = null)
...
 
-    if (null !== $fill && !empty($result)) {
+    if (null !== $fill && $result) {
         $last = \count($result) - 1;
         if ($fillCount = $size - \count($result[$last])) {
-            $result[$last] = array_merge(
-                $result[$last],
-                array_fill(0, $fillCount, $fill)
-            );
+            foreach (array_fill(0, $fillCount, $fill) as $v) {
```
stof: do we actually need to use `array_fill` to generate an array here ? This is a loop from `0` to `$fillCount`, putting `$fill` in the array.

fabpot: yep, forgot to change that.

fabpot: done

PR #2887, opened 2019-03-12
Force loading class aliases used in typehints

### LINKED ISSUE #2886: Declaration of Twig_Extensions_TokenParser_Trans::parse() must be compatible with Twig\TokenParser\TokenParserInterface::parse(Twig\Token $token)
Hello,

It seems that there is an error inside the latest release of Twig 1.x (1.38.0), after I updated I have this error upon each cache clear on Symfony.

`Declaration of Twig_Extensions_TokenParser_Trans::parse() must be compatible with Twig\TokenParser\TokenParserInterface::parse(Twig\Token $token) in [...]/vendor/twig/extensions/lib/Twig/Extensions/TokenParser/Trans.php on line 86`

Is there any breaking change? I haven't seen any in the changelog...

Thanks!

stof:
Fixes #2886

### DIFF
--- a/src/NodeVisitor/NodeVisitorInterface.php
+++ b/src/NodeVisitor/NodeVisitorInterface.php
@@ -13,6 +13,9 @@ namespace Twig\NodeVisitor;
 
 use Twig\Environment;
 
+// Ensure that the aliased name is loaded to keep BC for classes implementing the typehint with the old aliased name.
+class_exists('Twig\Environment');
+
 /**
  * Interface for node visitor classes.
  *
--- a/src/Extension/ExtensionInterface.php
+++ b/src/Extension/ExtensionInterface.php
@@ -18,6 +18,9 @@ use Twig\TwigFilter;
 use Twig\TwigFunction;
 use Twig\TwigTest;
 
+// Ensure that the aliased name is loaded to keep BC for classes implementing the typehint with the old aliased name.
+class_exists('Twig\Environment');
+
 /**
  * Interface implemented by extension classes.
  *
--- a/src/Node/Node.php
+++ b/src/Node/Node.php
@@ -14,6 +14,9 @@ namespace Twig\Node;
 
 use Twig\Compiler;
 
+// Ensure that the aliased name is loaded to keep BC for classes implementing the typehint with the old aliased name.
+class_exists('Twig\Compiler');
+
 /**
  * Represents a node in the AST.
  *
--- a/src/TokenParser/TokenParserInterface.php
+++ b/src/TokenParser/TokenParserInterface.php
@@ -15,6 +15,10 @@ use Twig\Error\SyntaxError;
 use Twig\Parser;
 use Twig\Token;
 
+// Ensure that the aliased name is loaded to keep BC for classes implementing the typehint with the old aliased name.
+class_exists('Twig\Token');
+class_exists('Twig\Parser');
+
 /**
  * Interface implemented by token parsers.
  *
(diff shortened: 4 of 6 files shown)


stof:
I haven't added tests for this, because it is hard to reproduce in a testsuite. The error depends on the order in which classes are loaded. If the typehint class is already loaded (and so its alias is already defined), the error does not happen (which is exactly how this gets solved).

src/NodeVisitor/NodeVisitorInterface.php
```diff
@@ -13,6 +13,9 @@
 
 use Twig\Environment;
 
+// Ensure that the aliased name is loaded to keep BC for classes implementing the typehint with the old aliased name.
+class_exists('Twig\Environment');
```
nicwortel, line 17: Is it enough to add this to the `NodeVisitorInterface`, or should it also be added to implementations such as `AbstractNodeVisitor`? I'm asking because I found this PR because we were getting:

> Compile Error: Declaration of Translation\Bundle\Twig\Visitor\NormalizingNodeVisitor::doEnterNode(Twig_Node $node, Twig_Environment $env) must be compatible with Twig\NodeVisitor\AbstractNodeVisitor::doEnterNode(Twig\Node\Node $node, Twig\Environment $env)

Will the fact that the `AbstractNodeVisitor` implements the `NodeVisitorInterface` be enough to trigger the autoloading?

Kocal: Also see https://github.com/php-translation/symfony-bundle/issues/288, we have the same issue there

kiler129:
To avoid problems after this path is merged library authors which rely on Twig and use non-NS class references basically need to include:

```
    "conflict": {
        "twig/twig": "2.7.0 || 1.38.0"
    },
```

This will prevent installing `2.7.0` and breaking builds :D

*Edit: add 1.38.0 conflict*

src/NodeVisitor/NodeVisitorInterface.php
nicwortel: @Kocal that is indeed the bundle we are using, too :smile: I probably should have included the ~upstream~downstream issue / repository

stof: @nicwortel yes, it will be enough, as AbstractNodeVisitor implementing the interface will force the interface to be loaded.

stof:
note that 1.38 is also affected.

wadjeroudi:
@kiler129 The 2.7.1 release should be the good one, why exclude it ?

Gummibeer:
@kiler129 same for `1.38.0` :/

kiler129:
@wadjeroudi @Gummibeer misread the changelog - I was under impression `2.7.0` introduced the break while `2.7.1` was a security fix while both arrived in `2.7.0`. Corrected the comment.

nicolas-grekas (approved):

fabpot:
Thank you @stof.

kiler129:
@fabpot Can you maybe tag this as `2.7.1`?

fabpot:
@kiler129 Working as fast as possible. I asked people to test the code before release, tweeted about it, but apparently nobody cared back then.

fabpot:
1.38.1 and 2.7.1 released.

kiler129:
@fabpot @stof Thank you! You're doing an amazing job for open source :)

goba:
Thanks @stof and @fabpot!

Natshah:
Thank you

PR #2893, opened 2019-03-12
Add TemplateWrapper::getTemplateName()

### LINKED ISSUE #2892: Is 1.38.1 broken?
Was the release of Twig 1.38.1 correct? It seems like a _huge_ delta for a minor release, and some things broke for us. 

One of the things that's breaking is that this doesn't work anymore: 

```php
$template = $twig->resolveTemplate($template);
$template->getTemplateName();
```

Would give something like this on 1.37.1: `"@bolt/dashboard/dashboard.twig"`

On 1.38.1 it gives: 

Schermafbeelding 2019-03-12 om 18 05 48

Same for `$var = (string) $template;`. On 1.37.1 it works, on 1.38.1 it gives: 

Schermafbeelding 2019-03-12 om 18 06 50

It's hard to pinpoint where it broke, because of the huge delta of the update. I assume it's because of this: 

> That's why, starting from the next 2.x version, all non- namespaced classes will be deprecated and > they will be removed in Twig 3.x, which is expected to be released during 2019.

Was this intentional for the 1.x patch?

fabpot:
closes #2892

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,6 @@
 * 1.38.2 (2019-XX-XX)
 
- * n/a
+ * added TemplateWrapper::getTemplateName()
 
 * 1.38.1 (2019-03-12)
 
--- a/src/TemplateWrapper.php
+++ b/src/TemplateWrapper.php
@@ -134,6 +134,14 @@ final class TemplateWrapper
     {
         return $this->template->getSourceContext();
     }
+
+    /**
+     * @return string
+     */
+    public function getTemplatename()
+    {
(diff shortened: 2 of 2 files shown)


bobdenotter:
Thanks, for us this solves the issues with 1.38.1: 

Screenshot 2019-03-12 at 19 34 03

bobdenotter (approved):

fabpot:
@bobdenotter new releases for 1.x and 2.x are out.

bobdenotter:
@fabpot Much appreciated! 💟

PR #2894, opened 2019-03-12
Rewrite the implementation of the set tag

fabpot:
The current implementation of the `set` tag relies on `ob_start/ob_get_clean`, which is not a good practice anymore. Like for the `filter` tag, it's better to create a method that wraps the output. That way, we don't need to rely on the fact that the output is generated via `echo` calls. In Twig 3.x, we won't have `echo` calls anymore, but `yield` calls, so that would break.

As a bonus, it makes the code easier.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,6 @@
 * 1.38.3 (2019-XX-XX)
 
- * n/a
+ * rewrote the implementation of the set tag
 
 * 1.38.2 (2019-03-12)
 
--- a/src/Node/SetNode.php
+++ b/src/Node/SetNode.php
@@ -25,19 +25,8 @@ class SetNode extends Node implements NodeCaptureInterface
     {
         parent::__construct(['names' => $names, 'values' => $values], ['capture' => $capture, 'safe' => false], $lineno, $tag);
 
-        /*
-         * Optimizes the node when capture is used for a large block of text.
-         *
-         * {% set foo %}foo{% endset %} is compiled to $context['foo'] = new Twig\Markup("foo");
-         */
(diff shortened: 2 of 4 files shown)


stof:
this makes it a scoping tag now

stof:
and as you compile that as a generated block, this can also cause issues if you have such a `set` tag in both the parent and child templates, as the parent template would not use the block generated in the child.
If you really want to compile that as a function (with some reference-based magic to share the context scope), it should not be a block, to avoid the inheritance issue.

fabpot:
I don't think we have an inheritance issue as the method name is random.

fabpot:
I've added some more tests in 8c18e7d0631ef8876df88257245dadb24c7d95d4 to check that the behavior stay the same. Any case I missed?

stof:
no, the method name is not random: `sprintf('__internal_%s', hash('sha256', __METHOD__.$this->stream->getSourceContext()->getCode().$this->varNameSalt++))`

But it indeed uses the source code in the hash, which I forgot, and so won't reuse the same name name in the parent and the child (unless they have the same source code, which would be weird).

stof:
Well, we are missing a test where both templates are defining their own `foo` variable and using it immediately, to show that the parent does not use the definition of the variable in the child by mistake.

And we would also need tests covering the case of mutating another variable inside the body of the `{% set %}` tag and showing that it is indeed mutated after the tag, to ensure we don't introduce scoping in the refactoring.
Currently, the only scoping elements are `with`, `block`, `macro` and `embed`, as well as `for` partially (it does not isolate the scope, but cleans it from *new* variables defined during the loop).

PR #2899, opened 2019-03-14
Fixed spaceless filter

tifabien:
As now the `spaceless` tag is deprecated, I tried to replace it by the new `spaceless` filter instead but it broke my application. Actually there is a difference between the tag and the filter. Spaceless tag trim also whistespace (or line feed).

I propose to do the same thing on the new filter.


fabpot:
Thank you @tifabien.

PR #2901, opened 2019-03-15
Allow Traversable objects to be used in the with tag

thewilkybarkid:
Currently `with` does an `in_array()` check, this allows `Traversable`~/`JsonSerializable`~ objects to be used as they can also contain a hash.


stof:
I don't think supporting JsonSerializable here makes sense. We don't support it in other places dealing with arrays or iterators in Twig (the `merge` tag for instance).

fabpot:
I agree with @stof

fabpot (approved):
Should be rebased on 1.x (and a note in the CHANGELOG could be added as well)

fabpot:
Thank you @thewilkybarkid.

PR #2904, opened 2019-03-16
Add some type hints

fabpot:


src/ExtensionSet.php
```diff
@@ -107,39 +93,29 @@ public function getExtension($class)
...
     {
         foreach ($extensions as $extension) {
             $this->addExtension($extension);
         }
     }
 
-    /**
-     * Returns all registered extensions.
-     *
-     * @return array An array of extensions
```
stof, line 125: I suggest keeping a phpdoc, but with a more precise type `@return ExtensionInterface[]`

src/ExtensionSet.php
```diff
@@ -107,39 +93,29 @@ public function getExtension($class)
         return $this->extensions[$class];
     }
 
-    /**
-     * Registers an array of extensions.
-     *
-     * @param array $extensions An array of extensions
```
stof, line 113: I suggest keeping a phpdoc, but with a more precise type `@param ExtensionInterface[] $extensions`

src/ExtensionSet.php
```diff
@@ -191,13 +170,9 @@ public function getFunctions()
     }
 
     /**
-     * Get a function by name.
-     *
-     * @param string $name function name
-     *
      * @return TwigFunction|false
```
stof, line 176: it would probably make sense to change that to `TwigFunction|null` in Twig 3 (or maybe even in current versions in case this is internal code)

src/Source.php
```diff
@@ -27,14 +27,14 @@ final class Source
      * @param string $name The template logical name
      * @param string $path The filesystem path of the template if any
      */
-    public function __construct($code, $name, $path = '')
+    public function __construct(string $code, $name, string $path = '')
```
stof: why not adding the type on the name ?

src/ExtensionSet.php
fabpot: done

src/ExtensionSet.php
fabpot: That's internal code, but I prefer to change it in 3.0.

src/Source.php
fabpot: test did not pass and I hadn't time to have a look, fixed now.

src/ExtensionSet.php
fabpot: done

stof (approved):

PR #2905, opened 2019-03-19
add .gitattributes file 

rvitaliy:
- remove useless files for production from dist
- configure git to automatic line endings normalized


fabpot:
Thanks for the PR, I'm going to close it as this was rejected in the past.

Issue #2898, opened 2019-03-13
 Twig the version 1.38.2, breaks the site in drupal


slzno:
Twig the version 1.38.2, breaks the site in drupal completely.

![image](https://user-images.githubusercontent.com/44006503/54308283-55975180-4593-11e9-81a3-b5b0074056fc.png)

Kocal:
Maybe an error message will help, can you provide one?

Boegie:
I'm experiencing this as well, and I think I've tracked it down to the embed tag.
Templates without it work fine.

The breaking commit seems to be this one: 4eeaf76

The Drupal Error message in its watchdog log is 
```
  | Twig\Error\RuntimeError:  Failed to load Twig template  "[name-of-template].twig",  index "10993019": cache might be corrupted.  __TwigTemplate_6c0bcbf998a75ffe406ad7eb908165951d42a7d9ce59c8eedb0f7b3123e9fbc4___10993019  in Twig\Environment->loadClass() (l
-- | --
```

hctom:
Exactly the same here... also tracked it down to the `embed` tag. If this is replaced with an `include` tag, the template is loaded again (for full template path and for namespaced template paths).
For the `embed` tag neither full template paths nor namespaced template paths work and throw the following exception:

```
The website encountered an unexpected error. Please try again later.

Twig\Error\RuntimeError</em>: Failed to load Twig template "path/to/user.html.twig", index "1128131273": cache might be corrupted. in Twig\Environment->loadClass() (line 14 of path/to/user.html.twig).
```

The corresponding template throwing the error has the following code:

```
{% extends '@my_namespace/partial/content.twig' %}

{% block content_header %}
  {{- title_suffix -}}

  {%- embed '@my_namespace/partial/misc/page-title.twig' %}
    {% block page_title_primary %}
      {{- user.getAccountName() -}}
    {% endblock page_title_primary %}
  {% endembed -%}

  {{- title_prefix -}}
{% endblock content_header %}

{% block content_content %}
  {%- if content %}
    {{- content -}}
  {% endif -%}
{% endblock %}
```

timmillwood:
Which version of Drupal are you using?

It might be worth reporting issues in https://www.drupal.org/project/drupal/issues/3039408

hctom:
I use Drupal 8.6.11 and I just created a new issue over there (while also adding a note to the issue linked by @timmillwood): https://www.drupal.org/project/drupal/issues/3040210

[...]

PR #2906, opened 2019-03-19
Fix typo in Environment->getTags()

jproctor:
I found this typo (already fixed in the 2.x branch) while looking for a problem that presented like #2898 but turned out to be unrelated (see [drupal.org #3040269](https://www.drupal.org/project/drupal/issues/3040269)).

### DIFF
--- a/src/Environment.php
+++ b/src/Environment.php
@@ -1079,7 +1079,7 @@ class Environment
     public function getTags()
     {
         $tags = [];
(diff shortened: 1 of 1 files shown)


fabpot:
This patch looks wrong to me for 1.x where `getTokenParsers()` returns an instance of `Twig_TokenParserBroker`.

fabpot:
And the tests are indeed broken by  the patch.

PR #2909, opened 2019-03-21
Allow Traversable objects to be used in the with keyword of the include and embed tags

thewilkybarkid:
Same as #2901 but for `include` and `embed`.


fabpot:
Thank you @thewilkybarkid.

PR #2910, opened 2019-03-22
fix(with): comment some strings

Kocal:
I don't know if it was intentional or not, but I find the following screenshot weird:
Sélection_027

I think it's better with a comment instead:
Sélection_028

What do you think?


fabpot:
Yep, that's much better

fabpot:
Thank you @Kocal.

PR #2916, opened 2019-03-25
fix major version in Environment

xabbuh:


xabbuh:
This should make the Symfony test suite pass again as we will no longer choose the wrong execution path for the compiler.

fabpot:
Thank you @xabbuh.

PR #2918, opened 2019-03-25
Add spaceless filter to the docs index

mbabker:


fabpot:
Good catch, thanks @mbabker.

Issue #2922, opened 2019-03-29
Length filter not working with iterators or traversables


dmolineus:
The build in length filter works with objects implementing `ArrayIterator` but not with objects implmenting `Iterator` or `Traversable`.

https://github.com/twigphp/Twig/blob/7597a742b7a5d3d11f4568b0da64e0d94e46efe3/src/Extension/CoreExtension.php#L1243-L1245

Is there any reason for this limitation? I saw that the `loop.length` property was removed on purpose but when using `|length` filter is a different case. As `iterator_count` is used here, I propose also support  `Iterator` and `Traversable`.

stof:
hmm, limiting it to IteratorAggregate is probably a mistake here, indeed.

xabbuh:
IIRC the reasoning was that `iterator_count()` would not play well with iterators that are not rewindable.

PR #2923, opened 2019-04-01
Add Traversable support of length filter

SpacePossum:
closes https://github.com/twigphp/Twig/issues/2922

### DIFF
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -1301,20 +1301,16 @@ if (\function_exists('mb_get_info')) {
             return mb_strlen($thing, $env->getCharset());
         }
 
-        if ($thing instanceof \SimpleXMLElement) {
+        if ($thing instanceof \Countable || \is_array($thing) || $thing instanceof \SimpleXMLElement) {
             return \count($thing);
         }
 
-        if (\is_object($thing) && method_exists($thing, '__toString') && !$thing instanceof \Countable) {
-            return mb_strlen((string) $thing, $env->getCharset());
-        }
-
-        if ($thing instanceof \Countable || \is_array($thing)) {
-            return \count($thing);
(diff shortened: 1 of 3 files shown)


src/Extension/CoreExtension.php
```diff
@@ -1298,22 +1298,18 @@ function twig_length_filter(Environment $env, $thing)
         }
 
         if (is_scalar($thing)) {
-            return mb_strlen($thing, $env->getCharset());
+            return mb_strlen((string) $thing, $env->getCharset());
```
fabpot: I think this can be reverted.

src/Extension/CoreExtension.php
```diff
@@ -1298,22 +1298,18 @@ function twig_length_filter(Environment $env, $thing)
...
+        if (\is_object($thing) && method_exists($thing, '__toString')) {
             return mb_strlen((string) $thing, $env->getCharset());
         }
 
-        if ($thing instanceof \Countable || \is_array($thing)) {
-            return \count($thing);
-        }
-
-        if ($thing instanceof \IteratorAggregate) {
+        if ($thing instanceof \Traversable) {
```
fabpot, line 1308: I would do that before using `__toString` (also to be BC).

fabpot (changes requested):

src/Extension/CoreExtension.php
SpacePossum: good catch, done and I've added update the test to catch this in the future : )

src/Extension/CoreExtension.php
SpacePossum: :+1:  done

fabpot (approved):

fabpot:
Thank you @SpacePossum.

PR #2924, opened 2019-04-01
Fixes twigphp/Twig#1423

petski:
This pull request fixes twigphp/Twig#1423

Patches more or less copied from twigphp/Twig#1569

PR #2925, opened 2019-04-03
Whitespace improvements

### LINKED ISSUE #1005: Optional whitespace control for well-formed result (not a duplicate)
I know, there were a lot of issues about whitespace control, but I think that I've found a solution that might satisfy everyone. So, the rules for special whitespace removal are:
1. It is optional, might be enabled easily.
2. It removes only whitespace (no newlines) **before** control structures.
3. Only if this is just a single control structure (or maybe comment also) on line.
4. And only if there is nothing after it.

For example:

```
{% if good %}
        <hr />
{% endif %}
```

Would produce exactly the same result as

```
    {% if good %}
        <hr />
    {% endif %}
```

That is,

```
        <hr />
```

Isn't it logical? That is, lines with single control structure are ignored, no matter if there are spaces in the beginning or there aren't.

The problem is, that currently, in some cases there are just no way in Twig to achieve this without breaking readability of templates. Let's say you have something like this (a bit simplified code from real project, "set menu" added for easy reproduction):

```
{% set menu = {'home': {'url': '/', 'title': 'Home'},'more': {'title': 'More', 'items': [{'url': '/', 'title': 'foo'}, {'url': '/', 'title': 'bar'}]}, 'about': {'url': '/', 'title': 'About'}} %}
<nav>
    <ul>
    {% for id, item in menu %}
      {% if item.items %}
        <li class="dropdown">
            <a href="#">{{- item.title }}</a>
            <ul>
            {% for subitem in item.items %}
                <li><a href="{{ subitem.url }}">{{ subitem.title }}</a></li>
            {% endfor %}
            </ul>
        </li>
      {% else %}
        <li><a href="{{ item.url }}">{{ item.title }}</a></li>
      {% endif %}
    {% endfor %}
    </ul>
</nav>
```

You'd like to produce something beautiful, like this:

```
<nav>
    <ul>
        <li><a href="/">Home</a></li>
        <li class="dropdown">
            <a href="#">More</a>
            <ul>
                <li><a href="/">foo</a></li>
                <li><a href="/">bar</a></li>
            </ul>
        </li>
        <li><a href="/">About</a></li>
    </ul>
</nav>
```

But there is just **no simple way** to do this with Twig. You're not able to achieve this with spaceless or with '-'. Or you end up in putting '-' in weirdest places without any evident logic, or mess up everything with spaceless. That's not how template engine should let you produce well-formed result.

However, you could:
1. Manually remove spaces before all control structures. And voila! you've got the result you wished, but have broken template readability.
2. Move control structures opening tags to the beginning of the lines, increasing spaces within them correspondingly. This might work, but it breaks auto-indent in all editors.
3. Comment out spaces before tags.

I've written simple [Twig Preprocessor loader](https://gist.github.com/TiGR/5002699), so that I could easily strip all whitespace before control structures and comments and it all worked out just perfectly. As result Twig just ignores lines having only one control structure - no matter how many spaces are there before it (or there are none).

So, my proposal is to add this into twig somehow as an option.

### LINKED ISSUE #1423: Whitespace control - trim spaces, but not newlines?

fabpot:
This pull request introduces a new whitespace control modifier in addition to the `-` one. The new `~` modifier consume whitespace excluding newlines.

closes #2924, closes #1005, closes #1423, closes #1569, and many already closed ones.

### DIFF
--- a/doc/templates.rst
+++ b/doc/templates.rst
@@ -844,25 +844,28 @@ Whitespace Control
 .. versionadded:: 1.1
     Tag level whitespace control was added in Twig 1.1.
 
-The first newline after a template tag is removed automatically (like in PHP.)
+.. versionadded:: 1.39
+    Tag level Line whitespace control was added in Twig 1.39.
+
+The first newline after a template tag is removed automatically (like in PHP).
 Whitespace is not further modified by the template engine, so each whitespace
 (spaces, tabs, newlines etc.) is returned unchanged.
 
-Use the ``spaceless`` tag to remove whitespace *between HTML tags*:
+You can also control whitespace on a per tag level. By using the whitespace
+control modifiers on your tags, you can trim leading and or trailing whitespace.
 
-.. code-block:: jinja
+Twig supports two modifiers:
 
-    {% spaceless %}
-        <div>
-            <strong>foo bar</strong>
-        </div>
-    {% endspaceless %}
+* *Whitespace trimming* via the ``-`` modifier: Removes all whitespace
+  (including newlines);
 
-    {# output will be <div><strong>foo bar</strong></div> #}
+* *Line whitespace trimming* via the ``~`` modifier: Removes all whitespace
+  (excluding newlines). Using this modifier on the right disables the default
+  removal of the first newline inherited from PHP.
 
-In addition to the spaceless tag you can also control whitespace on a per tag
-level. By using the whitespace control modifier on your tags, you can trim
-leading and or trailing whitespace:
+The modifiers can be used on either side of the tags like in ``{%-`` or ``-%}``
+and they consume all whitespace for that side of the tag. It is possible to use
+the modifiers on one side of a tag or on both sides:
 
 .. code-block:: jinja
 
@@ -871,20 +874,34 @@ leading and or trailing whitespace:
     {%- if true -%}
         {{- value -}}
     {%- endif -%}
-
     {# output 'no spaces' #}
 
-The above sample shows the default whitespace control modifier, and how you can
-use it to remove whitespace around tags. Trimming space will consume all whitespace
-for that side of the tag.  It is possible to use whitespace trimming on one side
-of a tag:
+    <li>
+        {{ value }}    </li>
+    {# outputs '<li>\n    no spaces    </li>' #}
 
-.. code-block:: jinja
-
-    {% set value = 'no spaces' %}
-    <li>    {{- value }}    </li>
-
+    <li>
+        {{- value }}    </li>
     {# outputs '<li>no spaces    </li>' #}
+
+    <li>
+        {{~ value }}    </li>
+    {# outputs '<li>\nno spaces    </li>' #}
+
+.. tip::
+
+    In addition to the whitespace modifiers, Twig also has a ``spaceless`` filter
+    that removes whitespace **between HTML tags**:
+
+    .. code-block:: jinja
+
+        {% filter spaceless %}
+            <div>
+                <strong>foo bar</strong>
+            </div>
+        {% endfilter %}
+
+        {# output will be <div><strong>foo bar</strong></div> #}
 
 Extensions
 ----------
--- a/src/Lexer.php
+++ b/src/Lexer.php
@@ -62,6 +62,8 @@ class Lexer implements \Twig_LexerInterface
             'tag_block' => ['{%', '%}'],
             'tag_variable' => ['{{', '}}'],
             'whitespace_trim' => '-',
+            'whitespace_line_trim' => '~',
+            'whitespace_line_chars' => ' \t\0\x0B',
             'interpolation' => ['#{', '}'],
         ], $options);
 
@@ -71,6 +73,8 @@ class Lexer implements \Twig_LexerInterface
                 \s*
                 (?:'.
                     preg_quote($this->options['whitespace_trim'].$this->options['tag_variable'][1]).'\s*'. // -}}\s*
+                    '|'.
+                    preg_quote($this->options['whitespace_line_trim'].$this->options['tag_variable'][1]).'['.$this->options['whitespace_line_chars'].']*'. // ~}}[ \t\0\x0B]*
                     '|'.
                     preg_quote($this->options['tag_variable'][1]). // }}
                 ')
@@ -82,21 +86,26 @@ class Lexer implements \Twig_LexerInterface
                 (?:'.
                     preg_quote($this->options['whitespace_trim'].$this->options['tag_block'][1]).'\s*\n?'. // -%}\s*\n?
                     '|'.
+                    preg_quote($this->options['whitespace_line_trim'].$this->options['tag_block'][1]).'['.$this->options['whitespace_line_chars'].']*'. // ~%}[ \t\0\x0B]*
+                    '|'.
                     preg_quote($this->options['tag_block'][1]).'\n?'. // %}\n?
                 ')
             }Ax',
 
             // {% endverbatim %}
-            'lex_raw_data' => '{
-                ('.
-                    preg_quote($this->options['tag_block'][0].$this->options['whitespace_trim']). // {%-
+            'lex_raw_data' => '{'.
+                preg_quote($this->options['tag_block'][0]). // {%
+                '('.
+                    $this->options['whitespace_trim']. // -
                     '|'.
-                    preg_quote($this->options['tag_block'][0]). // {%
-                ')\s*'.
+                    $this->options['whitespace_line_trim']. // ~
+                ')?\s*'.
                 '(?:end%s)'. // endraw or endverbatim
                 '\s*'.
                 '(?:'.
                     preg_quote($this->options['whitespace_trim'].$this->options['tag_block'][1]).'\s*'. // -%}
+                    '|'.
+                    preg_quote($this->options['whitespace_line_trim'].$this->options['tag_block'][1]).'['.$this->options['whitespace_line_chars'].']*'. // ~%}[ \t\0\x0B]*
                     '|'.
                     preg_quote($this->options['tag_block'][1]). // %}
                 ')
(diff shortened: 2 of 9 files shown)


fabpot:
Note that the new modifier works on both sides (unlike the previous attempt of others PRs on the same matter).

fabpot:
If you want to review the PR, I suggest you to read each commit, which is much easier.

src/Lexer.php
```diff
@@ -66,16 +66,81 @@ public function __construct(Environment $env, array $options = [])
...
 
         $this->regexes = [
-            'lex_var' => '/\s*(?:'.preg_quote($this->options['whitespace_trim'].$this->options['tag_variable'][1], '/').'\s*|'.preg_quote($this->options['tag_variable'][1], '/').')/A',
-            'lex_block' => '/\s*(?:'.preg_quote($this->options['whitespace_trim'].$this->options['tag_block'][1], '/').'\s*|'.preg_quote($this->options['tag_block'][1], '/').')\n?/A',
-            'lex_raw_data' => '/('.preg_quote($this->options['tag_block'][0].$this->options['whitespace_trim'], '/').'|'.preg_quote($this->options['tag_block'][0], '/').')\s*(?:end%s)\s*(?:'.preg_quote($this->options['whitespace_trim'].$this->options['tag_block'][1], '/').'\s*|\s*'.preg_quote($this->options['tag_block'][1], '/').')/s',
+            // }}
+            'lex_var' => '{
+                \s* # any number of white space before the ending of a var
+                (?:'.
+                    preg_quote($this->options['whitespace_trim'].$this->options['tag_variable'][1]).'\s*'. # -}}\s*
```
stof: I suggest using `//` rather than `#` to write a PHP comment.

doc/templates.rst
```diff
@@ -871,20 +874,30 @@ leading and or trailing whitespace:
...
-for that side of the tag.  It is possible to use whitespace trimming on one side
-of a tag:
+    <li>
+        {{- value }}    </li>
+    {# outputs '<li>no spaces    </li>' #}
 
-.. code-block:: jinja
+    <li>
+        {{~ value }}    </li>
+    {# outputs '<li>\nno spaces    </li>' #}
```
stof, line 889: I suggest adding the same example without whitespace control at all to see the difference.

doc/templates.rst
```diff
@@ -844,25 +844,28 @@ Whitespace Control
...
-        <div>
-            <strong>foo bar</strong>
-        </div>
-    {% endspaceless %}
+* *Whitespace trimming* via the ``-`` modifier: Removes all whitespace
+  (including newlines);
 
-    {# output will be <div><strong>foo bar</strong></div> #}
+* *Line whitespace trimming* via the ``~`` modifier: Removes all whitespace
+  (excluding newlines). Using this modifier disables the default removal of the
```
stof: Shouldn't the sentence be `Using this modifier on the right disables ...` ? I don't think using it on the left disables the default removal (as that concerns what happens at the end)

src/Lexer.php
```diff
@@ -240,8 +255,17 @@ protected function lexData()
...
+        // trim?
         if (isset($this->positions[2][$this->position][0])) {
-            $text = rtrim($text);
+            if ($this->options['whitespace_trim'] === $this->positions[2][$this->position][0]) {
+                // whitespace_trim detected ({%-, {{- or {#-)
+                $text = rtrim($text);
+            } else {
+                // whitespace_line_trim detected ({%~, {{~ or {#~)
+                // don't trim \r and \n
+                $text = rtrim($text, " \t\0\x0B");
```
stof, line 267: shouldn't this use the `whitespace_line_chars` option ?

src/Lexer.php
```diff
@@ -378,8 +402,16 @@ protected function lexRawData($tag)
...
-            $text = rtrim($text);
+        // trim?
+        if (isset($match[1][0])) {
+            if ($this->options['whitespace_trim'] === $match[1][0]) {
+                // whitespace_trim detected ({%-, {{- or {#-)
+                $text = rtrim($text);
+            } else {
+                // whitespace_line_trim detected ({%~, {{~ or {#~)
+                // don't trim \r and \n
+                $text = rtrim($text, " \t\0\x0B");
```
stof, line 413: shouldn't this use the `whitespace_line_chars` option ?

stof:
To optimize the lexing, I think we should be able to make lots of quantifiers possessive in these regexes (`\s*` becoming `\s*+`). I think it would work fine for the default syntax.
But this change might be risky for cases where the syntax is customized (they might be inventing crazy delimiters). I'm not even sure we should keep these options in 3.x (TwigBundle never exposed them in Symfony projects as that would break the ecosystem. Changing the syntax creates a new templating language).

javiereguiluz (approved):
👍 

I'm not a big fan of using `~` because it's too similar to `-` (it will be difficult to differentiate them depending on the font family and font size used in the editor) but I can't think of anything better.

Jinja uses `-` and `+`, but their use case is different: http://jinja.pocoo.org/docs/2.10/templates/#whitespace-control

doc/templates.rst
fabpot: correct

src/Lexer.php
fabpot: done

fabpot:
I thought about using quantifiers possessive, but did not do it for BC.

src/Lexer.php
fabpot: no, `whitespace_line_chars` are quoted

fabpot:
@javiereguiluz I would argue that using similar characters is a good thing as they are doing almost the same thing :)

doc/templates.rst
fabpot: added

src/Lexer.php
```diff
@@ -175,8 +255,17 @@ protected function lexData()
...
+        // trim?
         if (isset($this->positions[2][$this->position][0])) {
-            $text = rtrim($text);
+            if ($this->options['whitespace_trim'] === $this->positions[2][$this->position][0]) {
+                // whitespace_trim detected ({%-, {{- or {#-)
+                $text = rtrim($text);
+            } else {
+                // whitespace_line_trim detected ({%~, {{~ or {#~)
+                // don't trim \r and \n
+                $text = rtrim($text, " \t\0\x0B");
```
brunoais, line 267: I think this should consume one `\r\n` if they are immediately behind the opening brackets. That allows 
```
{{~
```
not to be an NOOP and do something that can be useful.
E.g.
```
abcd

{{~
```
Equivalent to:
```
abcd
{{
```

src/Lexer.php
```diff
@@ -313,8 +402,16 @@ protected function lexRawData($tag)
...
-            $text = rtrim($text);
+        // trim?
+        if (isset($match[1][0])) {
+            if ($this->options['whitespace_trim'] === $match[1][0]) {
+                // whitespace_trim detected ({%-, {{- or {#-)
+                $text = rtrim($text);
+            } else {
+                // whitespace_line_trim detected ({%~, {{~ or {#~)
+                // don't trim \r and \n
+                $text = rtrim($text, " \t\0\x0B");
```
brunoais, line 413: I think this should consume one `\r\n` if they are immediately behind the opening brackets. That allows 
```
{{~
```
not to be an NOOP and do something that can be useful.
E.g.
```
abcd

{{~
```
Equivalent to:
```
abcd
{{
```

src/Lexer.php
```diff
@@ -62,20 +62,100 @@ public function __construct(Environment $env, array $options = [])
...
-            'lex_var' => '/\s*'.preg_quote($this->options['whitespace_trim'].$this->options['tag_variable'][1], '/').'\s*|\s*'.preg_quote($this->options['tag_variable'][1], '/').'/A',
-            'lex_block' => '/\s*(?:'.preg_quote($this->options['whitespace_trim'].$this->options['tag_block'][1], '/').'\s*|\s*'.preg_quote($this->options['tag_block'][1], '/').')\n?/A',
-            'lex_raw_data' => '/('.preg_quote($this->options['tag_block'][0].$this->options['whitespace_trim'], '/').'|'.preg_quote($this->options['tag_block'][0], '/').')\s*(?:end%s)\s*(?:'.preg_quote($this->options['whitespace_trim'].$this->options['tag_block'][1], '/').'\s*|\s*'.preg_quote($this->options['tag_block'][1], '/').')/s',
+            // }}
+            'lex_var' => '{
+                \s*
+                (?:'.
+                    preg_quote($this->options['whitespace_trim'].$this->options['tag_variable'][1]).'\s*'. // -}}\s*
+                    '|'.
+                    preg_quote($this->options['whitespace_line_trim'].$this->options['tag_variable'][1]).'['.$this->options['whitespace_line_chars'].']*'. // ~}}[ \t\0\x0B]*
```
brunoais, line 77: I think this should consume one `\r\n` if they are immediately after the closing brackets. That allows 
```
~}}
```
not to be an NOOP and do something that can be useful.
E.g.
```
~}}

abcd
```
Equivalent to:
```
}}
abcd
```

brunoais (commented):
Solution for the operator to do something like in my proposal:
https://github.com/twigphp/Twig/compare/2.x...brunoais:linespaceControl

src/Lexer.php
fabpot: I don't think so. `~` does/should not consume newlines to make the modifier super clear.

brunoais: I understand both parts of the argument and, tbh, I don't know which one is better.

1. Make it always functional by making it consume 1 "character" and then only line whitespace.
2. Make it do only 1 very specific thing.

PR #2926, opened 2019-04-06
Fix "include" with "ignore missing" when an error loading occurs in the included template

### LINKED ISSUE #2524: {% include ... ignore missing %} skips Twig_Error_Loader exceptions thrown inside the included template
Lets assume the following `base.twig`:

    {% include 'child.twig' ignore missing %}

And a child with the following:

    {% include '_' %}

The exception inside `child.twig` is skipped, and twig simply carries like if the file didn't exist.
You can see this on https://twigfiddle.com/9b14cx

-----------------------------------------------

Funny side effect, if a child has a "bad" `{% extends %}`, it starts executing the next file. (Only happens with `{% extends %}` or `{% use %}` and not with `{% include %}`).

Take as example the following `base.twig`:

    {% include ['bad.twig', 'good.twig'] ignore missing %}

And the `bad.twig` file has the following:

    {% extends 'die.twig' %}

And the `good.twig` has the following:

    good!

This will display the following:

    good!

You can try it on https://twigfiddle.com/swkjpu

fabpot:
closes #2524

### DIFF
--- a/test/Twig/Tests/EnvironmentTest.php
+++ b/test/Twig/Tests/EnvironmentTest.php
@@ -498,33 +498,6 @@ EOF
         $this->assertEquals('foo', $twig->render('func_string_named_args'));
     }
 
-    /**
-     * @expectedException \Twig\Error\RuntimeError
-     * @expectedExceptionMessage Circular reference detected for Twig template "base.html.twig", path: base.html.twig -> base.html.twig in "base.html.twig" at line 1
-     */
-    public function testFailLoadTemplateOnCircularReference()
-    {
-        $twig = new Environment(new ArrayLoader([
-            'base.html.twig' => '{% extends "base.html.twig" %}',
-        ]));
-
-        $twig->load('base.html.twig');
-    }
-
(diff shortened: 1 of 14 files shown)


test/Twig/Tests/EnvironmentTest.php
```diff
@@ -498,33 +498,6 @@ public function testAddRuntimeLoader()
         $this->assertEquals('foo', $twig->render('func_string_named_args'));
     }
 
-    /**
-     * @expectedException \Twig\Error\RuntimeError
-     * @expectedExceptionMessage Circular reference detected for Twig template "base.html.twig", path: base.html.twig -> base.html.twig in "base.html.twig" at line 1
```
stof, line 503: so, no such protection against circular references anymore ?

fabpot: No,  as we don't load templates recursively, so not needed anymore. And when the parent is dynamic, we didn't have such protection anyway (same for includes).

PR #2927, opened 2019-04-07
Prepare for the new serialization mechanism

fancyweb:
#eufossa

related to https://github.com/symfony/symfony/pull/30965

So we are future proof and it will prevent deprecations for https://github.com/symfony/symfony/issues/30304.

Almost future proof actually because the void return type cannot be added in 7.0. I guess we can see the minimum required PHP version here once 7.4 is out and see how we can handle compatibility 😕


fabpot:
Can you rebase on 1.x?

fancyweb:
@fabpot Done

fabpot:
Thank you @fancyweb.

PR #2932, opened 2019-04-10
Tweak twig_to_array

fabpot:
I've change `twig_to_array()` to **not** cast to an array. That should not be a problem as we almost always call `twig_test_iterable()` before calling `twig_to_array()` and for the cases where it was not done, the PHP code would not work anyway.

follow-up to #2901 and #2909, /cc @thewilkybarkid

### DIFF
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -514,7 +514,9 @@ function twig_replace_filter($str, $from, $to = null)
         @trigger_error('Using "replace" with character by character replacement is deprecated since version 1.22 and will be removed in Twig 2.0', E_USER_DEPRECATED);
 
         return strtr($str, $from, $to);
-    } elseif (!twig_test_iterable($from)) {
+    }
+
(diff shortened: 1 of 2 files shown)


src/Extension/CoreExtension.php
```diff
@@ -739,9 +741,13 @@ function twig_last(Environment $env, $item)
  */
 function twig_join_filter($value, $glue = '', $and = null)
 {
+    if (!twig_test_iterable($value)) {
+        $value = (array) $value;
```
fabpot, line 745: I don't like that, but this was in the original code before the introduction of `twig_to_array()`, so I kept it here.

thewilkybarkid:
Is it worth calling `twig_test_iterable()` from `twig_to_array()`?

fabpot:
I don't think so as the function is marked as internal and the error message is different depending on the caller.

thewilkybarkid (approved):

PR #2934, opened 2019-04-10
Make escaping on ternary expressions more fine-grained

### LINKED ISSUE #449: The ternary operator breaks the auto-escaping
The auto-escaping is broken when using a ternary operator like this:

``` jinja
{{ var is defined ? var|nl2br : "something else" }}
```

The `nl2br` filter is safe but the value will be escaped again here, thus leading to a double escaping.

### LINKED ISSUE #539: Raw output not possible in some ternary statements
According to the [Twig docs](http://twig.sensiolabs.org/doc/api.html#escaper-extension):

``` jinja
{{ foo ? "Twig<br />" : "<br />Twig" }} {# won't be escaped #}

{% set text = "Twig<br />" %}
{{ foo ? text : "<br />Twig" }} {# will be escaped #}

{% set text = "Twig<br />" %}
{{ foo ? text|raw : "<br />Twig" }} {# won't be escaped #}

{% set text = "Twig<br />" %}
{{ foo ? text|escape : "<br />Twig" }} {# the result of the expression won't be escaped #}
```

However, I'm seeing:

``` jinja
{% set foo = '<h1>Twig</h1>' %}

{% if true %}{{ foo|raw }}{% else %}{{ foo }}{% endif %} {# won't be escaped #}
{{ true ? foo|raw : foo }} {# will be escaped #}
{{ true ? foo|raw : 'string-literal' }} {# won't be escaped #}
```

Even though the ternary's else condition is never reached, it seems to affect the true condition's processing. This case is absent from the documentation example, which always used string literals in the else condition.

### LINKED ISSUE #1174: The ternary operator doesn't respect safe filters, functions  ...
If I specify a different safe strategies for members of the ternary operator, escaping is wrong.

``` jinja
{# char = "&" #}
{{ 1 ? char|raw : char }}
```

And bad result is:

```
&amp; // expected &
```

### LINKED ISSUE #1320: Should html safe methods work in ternary if-conditions?
Imagine i have the html-safe twig function `icon()` ( i.e. [the one  provided by BCBoostrapBundle](https://github.com/braincrafted/bootstrap-bundle/blob/develop/Twig/BootstrapIconExtension.php#L48) ). 

Then i would expect the following to work out of the box:

```
{{ article.description ?: icon('minus') }}     -> output is being escaped :(
```

It looks like you currently have to use this (ugly) workaround:

```
{{ ( article.description ?: icon('minus') )|raw }}    -> works but ugly :(
```

Am i missing something (i.e. some `autoescape` setting ) or should we probably fix this ?

### LINKED ISSUE #1739: Twig "raw" filter ignored in ternary operator
If you apply filter in ternary operator, like this:
`{{ ( entity.description ) ? entity.description|raw : 'entity.no_decription'|trans }}`
the "raw" filter will be ignored. If you just write as usual:
`{{ entity.description|raw }}`
all works well. Probably not a problem?

### LINKED ISSUE #2146: Static expressions don't work with variables on other side of ternary
Here's an example. I'll try to debug, but I guess the expression needs to be evaluated separately.
http://twigfiddle.com/xo63vc/3

Here's the generated code for that.

```
  escaped: {{ name }}
  static: {{ '&' }}
  true: {{ true ? '&' : '&' }}
  false: {{ false ? '&' : '&' }}
  true with variable false: {{ true ? '&' : name }}
  false with variable true: {{ false ? name : '&' }}

  true with variable false: {{ true ? '&'|raw : name }}
  false with variable true: {{ false ? name : '&'|raw }}
```

```
    {
        // line 1
        echo "  escaped: ";
        echo twig_escape_filter($this->env, (isset($context["name"]) ? $context["name"] : null), "html", null, true);
        echo "
  static: ";
        // line 2
        echo "&";
        echo "
  true: ";
        // line 3
        echo ((true) ? ("&") : ("&"));
        echo "
  false: ";
        // line 4
        echo ((false) ? ("&") : ("&"));
        echo "
  true with variable false: ";
        // line 5
        echo twig_escape_filter($this->env, ((true) ? ("&") : ((isset($context["name"]) ? $context["name"] : null))), "html", null, true);
        echo "
  false with variable true: ";
        // line 6
        echo twig_escape_filter($this->env, ((false) ? ((isset($context["name"]) ? $context["name"] : null)) : ("&")), "html", null, true);
    }
```

fabpot:
closes #2146 
closes #449
closes #539
closes #1174
closes #1320
closes #1739
closes #1333

### DIFF
--- a/src/Node/Expression/InlinePrint.php
+++ b/src/Node/Expression/InlinePrint.php
@@ -1,0 +1,35 @@
+<?php
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Twig\Node\Expression;
+
+use Twig\Compiler;
+use Twig\Node\Node;
+
+/**
+ * @internal
+ */
+final class InlinePrint extends AbstractExpression
+{
+    public function __construct(Node $node, $lineno)
+    {
+        parent::__construct(['node' => $node], [], $lineno);
+    }
+
+    public function compile(Compiler $compiler)
+    {
+        $compiler
+            ->raw('print (')
+            ->subcompile($this->getNode('node'))
+            ->raw(')')
+        ;
+    }
+}
--- a/src/NodeVisitor/EscaperNodeVisitor.php
+++ b/src/NodeVisitor/EscaperNodeVisitor.php
@@ -15,8 +15,11 @@ use Twig\Environment;
 use Twig\Node\AutoEscapeNode;
 use Twig\Node\BlockNode;
 use Twig\Node\BlockReferenceNode;
+use Twig\Node\DoNode;
+use Twig\Node\Expression\ConditionalExpression;
 use Twig\Node\Expression\ConstantExpression;
 use Twig\Node\Expression\FilterExpression;
+use Twig\Node\Expression\InlinePrint;
 use Twig\Node\ImportNode;
 use Twig\Node\ModuleNode;
 use Twig\Node\Node;
@@ -69,8 +72,13 @@ class EscaperNodeVisitor extends AbstractNodeVisitor
             $this->blocks = [];
         } elseif ($node instanceof FilterExpression) {
             return $this->preEscapeFilterNode($node, $env);
-        } elseif ($node instanceof PrintNode) {
-            return $this->escapePrintNode($node, $env, $this->needEscaping($env));
+        } elseif ($node instanceof PrintNode && false !== $type = $this->needEscaping($env)) {
+            $expression = $node->getNode('expr');
+            if ($expression instanceof ConditionalExpression && $this->shouldUnwrapConditional($expression, $env, $type)) {
+                return new DoNode($this->unwrapConditional($expression, $env, $type), $expression->getTemplateLine());
+            }
+
+            return $this->escapePrintNode($node, $env, $type);
         }
 
         if ($node instanceof AutoEscapeNode || $node instanceof BlockNode) {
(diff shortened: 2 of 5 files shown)


fabpot:
@stof @nicolas-grekas  Could you review this PR please? It's an important one as it involves escaping and I don't want to introduce a security issue.

nicolas-grekas (approved):

test/Twig/Tests/Fixtures/tags/autoescape/literal.test
```diff
@@ -11,19 +11,36 @@
 
 3. Conditional expression with a variable
 {{ true ? "<br />" : someVar }}
+{{ false ? "<br />" : someVar }}
+{{ true ? someVar : "<br />" }}
+{{ false ? someVar: "<br />" }}
```
mykiwi:
```suggestion
{{ false ? someVar : "<br />" }}
```

src/Node/Expression/InlinePrint.php
```diff
@@ -0,0 +1,35 @@
...
+{
+    public function __construct(Node $node, $lineno)
+    {
+        parent::__construct(['node' => $node], [], $lineno);
+    }
+
+    public function compile(Compiler $compiler)
+    {
+        $compiler
+            ->raw('print (')
```
stof, line 30: aren't we using `echo` in compiled templates ?

src/NodeVisitor/EscaperNodeVisitor.php
```diff
@@ -82,6 +90,41 @@ protected function doLeaveNode(Node $node, Environment $env)
...
+    {
+        $expression = $node->getNode('node');
+
+        if ($this->isSafeFor($type, $expression, $env)) {
+            return $node;
+        }
+
+        $class = \get_class($node);
+
+        return new $class(
```
stof: why making this dynamic ? We now that the class is `InlinePrint` print here. It cannot be a child class as it is final.

src/NodeVisitor/EscaperNodeVisitor.php
```diff
@@ -82,6 +90,41 @@ protected function doLeaveNode(Node $node, Environment $env)
...
     }
 
+    private function unwrapConditional(ConditionalExpression $expression, Environment $env)
+    {
+        // convert "echo a ? b : c" to "a ? echo b : echo c" recursively
+        $expr2 = $expression->getNode('expr2');
+        if ($expr2 instanceof ConditionalExpression) {
+            $expr2 = $this->unwrapConditional($expr2, $env);
+        } else {
+            $expr2 = $this->escapeInlinePrintNode(new InlinePrint($expr2, $expr2->getTemplateLine(), $expr2->getNodeTag()), $env, $this->needEscaping($env));
```
stof: `InlinePrint` does not take a node tag as third argument. It only has 2 (this applies to all instantiations)

stof (changes requested):

stof:
The existing auto-escaping rule (requiring both sides to be safe for the ternary to be safe) still applies when the ternary is nested in a more complex expressions (`(var ?? 'something')|nl2br` for instance), right ?

test/Twig/Tests/Fixtures/tags/autoescape/literal.test
fabpot: fixed

src/Node/Expression/InlinePrint.php
fabpot: because echo does not work here :)

src/NodeVisitor/EscaperNodeVisitor.php
fabpot: indeed, just a copy/paste, fixed.

src/NodeVisitor/EscaperNodeVisitor.php
fabpot: removed

fabpot:
@stof that's correct.

stof:
Then the doc needs to make this more clear IMO. The note you added may make people think this would not be the case anymore.

src/Node/Expression/InlinePrint.php
stof: hmm, indeed. echo can only start a statement (as does `throw`), while print can be used in expressions. I forgot that they are not just aliases (I haven't used print since at least 7 years :smile:).

src/NodeVisitor/EscaperNodeVisitor.php
```diff
@@ -82,6 +90,39 @@ protected function doLeaveNode(Node $node, Environment $env)
...
     }
 
+    private function unwrapConditional(ConditionalExpression $expression, Environment $env)
+    {
+        // convert "echo a ? b : c" to "a ? echo b : echo c" recursively
+        $expr2 = $expression->getNode('expr2');
+        if ($expr2 instanceof ConditionalExpression) {
+            $expr2 = $this->unwrapConditional($expr2, $env);
+        } else {
+            $expr2 = $this->escapeInlinePrintNode(new InlinePrint($expr2, $expr2->getTemplateLine()), $env, $this->needEscaping($env));
```
stof: why not passing the escaping type from the caller, as you already have it on line 75 (and you would then have it in this method for recursive calls), instead of re-checking it each time ?

stof:
We have an optimization in NullCoalesceExpression to use `??` when we have a NameExpression as `expr2`. This change will kill that in print nodes as any NullCoalesceExpression would be turned into a normal ConditionalExpression (with the right test though, but not the optimization).

Should we unwrap things and wrap then in an InlinePrintNode all the time, or only when the safety rules would be different for both branches ? If rules are the same in both branches, applying the escaping as done today does not hurt, but let the ConditionalExpression optimize itself when possible.

src/NodeVisitor/EscaperNodeVisitor.php
fabpot: done

fabpot:
@stof we now only unwrap when the safety is not the same on the two branches.

stof (approved):

Issue #1918, opened 2015-11-12
Twig reports incorrect file / line number if exception is thrown during the rendering of an include 


ken-spencer:
If I throw an exception in a twig extension function run from an include, the reported file and line number are for the file that called the extension not the included file.  This can make debugging fairly tricky as we are looking for a needle in a haystack. 

Twig appears to be detecting the template by looping through  a debug_backtrace. This seems to result in the last template in the trace getting returned when I believe the first would be desired. 

If I make the following change in Error.php it appears to solve the problem: 

```
@@ -201,6 +201,7 @@
                 if (null === $this->filename || ($this->filename == $trace['object']->getTemplateName() && !$isEmbedContainer)) {
                     $template = $trace['object'];
                     $templateClass = get_class($trace['object']);
+                    break;
                 }
             }
         }
```

stof:
Can you provide a reproducing testcase (and tell which version of Twig you are using) ?

ken-spencer:
Sure. I am testing in * * 1.23.1, but this problem has existed for at least a year. 

## index.php

```
<?php


require __DIR__ . '/vendor/twig/twig/lib/Twig/Autoloader.php';
Twig_Autoloader::register();


$loader = new Twig_Loader_Filesystem(__DIR__);
$twig = new Twig_Environment($loader, [
    'cache' => __DIR__ . '/cache',
    'debug' => true,
    'auto_reload' => true,
    'strict_variables' => true,
]);

class Error extends Twig_Extension
{

    public function getFunctions()
    {
        return [
            'trigger_error' => new \Twig_Function_Method($this, 'triggerError'),
        ];                
    }

    function triggerError()
    {
        $error = new \Twig_Error_Runtime("This is an error");

        $str =  '<div>Error is in: ' . $error->getTemplateFile() . 'on line: ' . $error->getTemplateLine(). '</div>';
        $str .= '<div>Expecting:  Error is in: content.html.twigon line: 3 </div>';


        return $str;
    }

    public function getName()
    {
        return 'error';
    }

}

$twig->addExtension(new Error);
        

$template = $twig->loadTemplate('index.html.twig');
echo  $template->render([]);
```

## index.html.twig

```
<!DOCTYPE html>
<html>
    <head>
        <title>Test of error</title>
    </head>
    <body>
    <h1> get error </h1>
    {% include "content.html.twig" %}
    </body>
</html>
```

## content.html.twig

```
<div class="content">

{{ trigger_error() |raw }}

[...]

PR #2935, opened 2019-04-12
Add missing hints in some runtime exceptions

fabpot:


src/Node/Expression/Test/DefinedTest.php
```diff
@@ -47,7 +47,7 @@ public function __construct(\Twig_NodeInterface $node, $name, \Twig_NodeInterfac
         } elseif ($node instanceof ConstantExpression || $node instanceof ArrayExpression) {
             $node = new ConstantExpression(true, $node->getTemplateLine());
         } else {
-            throw new SyntaxError('The "defined" test only works with simple variables.', $this->getTemplateLine(), null, null, false);
+            throw new SyntaxError('The "defined" test only works with simple variables.', $this->getTemplateLine(), $this->getSourceContext());
```
stof: wouldn't this break because they are not set yet here (as we are in the constructor) ? this should use `$lineno` for the line btw.

src/Node/Node.php
```diff
@@ -235,6 +237,19 @@ public function getTemplateName()
         return $this->name;
     }
 
+    public function setSourceContext(Source $source)
+    {
+        $this->sourceContext = $source;
+        foreach ($this->nodes as $node) {
```
stof, line 243: In 1.x, you need to account for `null` and for custom implementations of the interface (which would not have the method) here. Both are deprecated, but this code should not break the BC layer.

fabpot: correct

src/Error/Error.php
```diff
@@ -79,13 +79,7 @@ public function __construct($message, $lineno = -1, $source = null, \Exception $
 
         $this->lineno = $lineno;
         $this->filename = $name;
-
-        if ($autoGuess && (-1 === $lineno || null === $name || null === $this->sourcePath)) {
-            $this->guessTemplateInfo();
-        }
```
fabpot, line 85: That's one of the main goal of this PR: getting rid of guessing in the constructor.

src/Node/Expression/Test/DefinedTest.php
fabpot: removed for now

fabpot:
Removing auto-guessing is great for performance, but also because it fixes #1918 (tests added to prove it)

src/Error/Error.php
```diff
@@ -79,13 +79,7 @@ public function __construct($message, $lineno = -1, $source = null, \Exception $
 
         $this->lineno = $lineno;
         $this->filename = $name;
-
-        if ($autoGuess && (-1 === $lineno || null === $name || null === $this->sourcePath)) {
-            $this->guessTemplateInfo();
```
stof, line 84: is `guessTemplateInfo` still used anywhere ?

stof: OK, I saw we still have some guessing involved for the wrapping of exceptions.

stof (approved):

stof:
hmm, actually, Node still breaks in case of a child not implement NodeInterface directly, as it would not have the new `setSourceContext` method.

PR #2936, opened 2019-04-12
Fix compat with NodeInterface

fabpot:
/cc @stof


stof (approved):

src/Node/Node.php
```diff
@@ -236,7 +236,7 @@ public function setSourceContext(Source $source)
     {
         $this->sourceContext = $source;
         foreach ($this->nodes as $node) {
-            if (null !== $node) {
+            if (null !== $node && $node instanceof Node) {
```
stof: you could even remove the null check here actually.

PR #2937, opened 2019-04-12
Deprecate Node::setTemplateName()

fabpot:


src/Node/Node.php
```diff
@@ -168,17 +168,25 @@ public function getIterator()
...
+     * @deprecated since 2.8 (to be removed in 3.0)
+     */
+    public function setTemplateName($name/*, $triggerDeprecation = true */)
     {
+        $triggerDeprecation = 2 > \func_num_args() || \func_get_arg(1);
+        if ($triggerDeprecation) {
+            @trigger_error('The '.__METHOD__.' method is deprecated since version 2.8 and will be removed in 3.0. Use setSourceContext() instead.', E_USER_DEPRECATED);
+        }
+
         $this->name = $name;
```
stof, line 181: is this property still used ?

PR #2939, opened 2019-04-16
Fix various small bugs

fabpot:


src/Loader/FilesystemLoader.php
```diff
@@ -170,7 +170,7 @@ public function exists($name)
         }
 
         try {
-            return false !== $this->findTemplate($name, false);
+            return false !== $this->findTemplate($name);
```
stof: this change looks wrong to me. We have this as a hidden argument.

Tobion: Is this intentional? There is logic explicitly for this: `$throw = \func_num_args() > 1 ? func_get_arg(1) : true;`

fabpot: indeed :)

fabpot: reverting

fabpot:
green now!

PR #2940, opened 2019-04-16
Deprecate passing a 4th and 5th arguments to the Sandbox exception classes

fabpot:


fabpot:
Passing a string is deprecated (a Source instance must be passed) anyway and this is done automatically, so no use case to pass something explicitly (it will be overriden anyway).

PR #2941, opened 2019-04-16
Deprecate the 3rd argument of Sandbox error classes

fabpot:


stof:
Shouldn't we also pass the lineno to `twig_get_attribute` to set the proper line in exceptions ?

fabpot:
@stof yes, that would be great... but I'm doing it step by step :)

PR #2949, opened 2019-04-21
Fix Lexer when using custom options containing the # char

### LINKED ISSUE #2948: Readability refactoring on Lexer's regexes introduced a breaking change (PHP <7.3)
Hi, :hand: 

Today I've encountered an issue when trying to use [manalize](https://github.com/manala/manalize):

```
PHP Warning:  preg_match_all(): Compilation failed: missing ) at offset 51 in /home/kocal/.composer/vendor/twig/twig/src/Lexer.php on line 169
```

They configured Twig Lexer to use `['[#', '#]']` for tag comments and `['{#', '#}']` for tag variables from 2+ years and it worked fine, but it suddenly broke.

The corresponding line is:
https://github.com/twigphp/Twig/blob/a90d98458708ad7b0a9fc75179e6afb7b552d02d/src/Lexer.php#L169
I've checked how the regex `lex_tokens_start` is built, and I noticed that it has been modified in 7e3ec0fbeee78a008394025efc8df58c3f7f37b3 (#2925).

I'm also sure that can be related to `preg_quote` changes on 7.3, where character `#` is now quoted ([changelog](https://www.php.net/manual/en/function.preg-quote.php#refsect1-function.preg-quote-changelog))

I've opened https://github.com/twigphp/Twig/pull/2947 that add a test that fails on 7.0 - 7.2, but pass on 7.3.

Also, I've created https://3v4l.org/oDcFS that build the regex `lex_tokens_start` (with the previous code and the actual code) on different PHP versions. As you can see, the previous code works well on PHP 7.1 - 7.3, and the actual code only works on PHP 7.3.

What should we do to fix this? Revert regex delimiters/flags? Pass `/` to  `preg_quote` again? :thinking: 

Thanks! :)

fabpot:
closes #2948

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,6 @@
 * 1.39.2 (2019-XX-XX)
 
- * n/a
+ * fixed Lexer when using custom options containing the # char
 
 * 1.39.1 (2019-04-16)
 
--- a/src/Lexer.php
+++ b/src/Lexer.php
@@ -67,16 +67,17 @@ class Lexer implements \Twig_LexerInterface
             'interpolation' => ['#{', '}'],
         ], $options);
 
+        // when PHP 7.3 is the min version, we will be able to remove the '#' part in preg_quote as it's part of the default
         $this->regexes = [
             // }}
             'lex_var' => '{
                 \s*
                 (?:'.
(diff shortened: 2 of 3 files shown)


Kocal:
Thank you! :)

PR #2950, opened 2019-04-22
Add a note about escaping

### LINKED ISSUE #2623: Autoescape does not escape javascript in a href

fabpot:
closes #2623, closes #2661

### DIFF
--- a/doc/filters/escape.rst
+++ b/doc/filters/escape.rst
@@ -8,7 +8,7 @@
 .. versionadded:: 1.14.0
     The ability to define custom escapers was added in Twig 1.14.0.
 
-The ``escape`` filter escapes a string for safe insertion into the final
+The ``escape`` filter escapes a string for safe insertion into the final HTML
 output. It supports different escaping strategies depending on the template
 context.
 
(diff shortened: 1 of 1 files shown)


doc/filters/escape.rst
```diff
@@ -8,7 +8,7 @@
 .. versionadded:: 1.14.0
     The ability to define custom escapers was added in Twig 1.14.0.
 
-The ``escape`` filter escapes a string for safe insertion into the final
+The ``escape`` filter escapes a string for safe insertion into the final HTML
```
xabbuh, line 11: Using HTML here seems misleading to me. It may give the feeling that `escape` is only useful if your template produces HTML.

fabpot: I've tried something else in #2980. Does that work better for you?

PR #2954, opened 2019-04-23
"in" operator strictness

### LINKED ISSUE #340: Casting Issue / Threequals request
Today, I wanted to omit one item from a group.  For some reason though, the item I wanted to omit **and** the first item in the array were being omitted.  My code looked like this:

```
{% for name, item in somearray %}
  {% if (name != 'somestring') %}
    {# do the stuff for the ones not omitted #}
  {% endif %}
{% endfor %}
```

Well, I found that the issue was a cause of the fact that the following resolved to true:

```
{% if (0 == 'somestring') %}
{# and this also #}
{% if ('somestring' == 0) %}
```

This was being converted to PHP, which was also resolving to true, so the problem **isnt** in Twig.  Hooray!  Now this is definitely a "gotcha", so perhaps twig could implement casting if the string comes first?  Or also twig could implement a `===` (threequals) operator.  

I solved it by casting the `name` above to a string like so:

```
{% if (name ~ '' != 'somestring') %}
```

Whatever the case, I daresay template designers will be befuddled by the above issue.  I'm not sure there's an easy solution here, but if there IS, I'd glady submit a pull request.

Thoughts anyone?

### LINKED ISSUE #341: Strict comparison
It would be great to add strict comparison in Twig to be able to avoid such cases where the test returns true:

``` jinja

{% if 0 == 'some-string' %}

```

this is due to PHP casting the string to an int to compare it with 0 instead of casting the integer to a string. The only way to avoid this is to use strict comparison but it is not available in Twig.

### LINKED ISSUE #2824: Bug in containment operator?
it seems that 
`{% if 'text' in [0] %}bug{% else %}no bug{% endif %}`
and 
`{% if 'text' in {0: 0} %}bug{% else %}no bug{% endif %}`
return 'bug', whereas
`{% if 'text' in ['0'] %}bug{% else %}no bug{% endif %}`
and 
`{% if 'text' in {0: '0'} %}bug{% else %}no bug{% endif %}`
return 'no bug'

Shouldn't `'text' in [0] ` evaluate as false?

fabpot:
The PHP non-strict comparison operator has some counter-intuitive behaviors, which we inherit in Twig. For instance, `{{ 'text' in [0] }}` returns `true`.

PHP is probably going to fix it in version 8: https://wiki.php.net/rfc/string_to_number_comparison

I propose to implement the same semantics in Twig. We have two options here: do the changes in 1.x as a bug fix or implement it in 3.0. WDYT?

closes #2824, closes #341, closes #340

### DIFF
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -939,29 +939,89 @@ function twig_in_filter($value, $compare)
         $value = (string) $value;
     }
 
-    if (\is_array($compare)) {
-        return \in_array($value, $compare, \is_object($value) || \is_resource($value));
-    } elseif (\is_string($compare) && (\is_string($value) || \is_int($value) || \is_float($value))) {
-        return '' === $value || false !== strpos($compare, (string) $value);
-    } elseif ($compare instanceof \Traversable) {
-        if (\is_object($value) || \is_resource($value)) {
-            foreach ($compare as $item) {
-                if ($item === $value) {
-                    return true;
-                }
-            }
-        } else {
-            foreach ($compare as $item) {
-                if ($item == $value) {
-                    return true;
-                }
-            }
+    if (\is_string($compare)) {
+        if (\is_string($value) || \is_int($value) || \is_float($value)) {
+            return '' === $value || false !== strpos($compare, (string) $value);
         }
 
         return false;
     }
 
+    if (\is_object($value) || \is_resource($value)) {
+        return in_array($value, $compare, true);
+    }
+
+    if (!\is_array($compare) && !$compare instanceof \Traversable) {
+        return false;
+    }
+
+    foreach ($compare as $item) {
+        if (twig_compare($value, $item)) {
+            return true;
+        }
+    }
+
     return false;
+}
+
+/**
+ * Compares two values using a more strict version of the PHP non-strict comparison operator.
+ *
+ * @see https://wiki.php.net/rfc/string_to_number_comparison
+ * @see https://wiki.php.net/rfc/trailing_whitespace_numerics
+ *
+ * @internal
+ */
+function twig_compare($a, $b)
+{
+    // int <=> string
+    if (is_int($a) && is_string($b)) {
+        $b = trim($b);
+        if (!is_numeric($b)) {
+            return $b === (string) $a;
+        }
+        if ((int) $b == $b) {
+            return (int) $b === $a;
+        } else {
+            return (float) $b === (float) $a;
+        }
+    }
+    if (is_string($a) && is_int($b)) {
+        $a = trim($a);
+        if (!is_numeric($a)) {
+            return $a === (string) $b;
+        }
+        if ((int) $a == $a) {
+            return (int) $a === $b;
+        } else {
+            return (float) $a === (float) $b;
+        }
+    }
+
+    // float <=> string
+    if (is_float($a) && is_string($b)) {
+        if ($a === NAN) {
+            return true;
+        }
+        if (!is_numeric($b)) {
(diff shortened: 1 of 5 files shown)


fabpot:
The current PR only makes the change for the `in` operator, but we could even change the `==` operator in Twig to use the new `twig_compare()` function.

fabpot:
I've added a second commit showing how we could replace the `==` operator with the new `twig_compare` function (easy!).

src/Extension/CoreExtension.php
```diff
@@ -939,31 +939,100 @@ function twig_in_filter($value, $compare)
...
-            }
+    if (\is_string($compare)) {
+        if (\is_string($value) || \is_int($value) || \is_float($value)) {
+            return '' === $value || false !== strpos($compare, (string) $value);
         }
 
         return false;
     }
 
+    if (\is_object($value) || \is_resource($value)) {
```
stof, line 950: isn't this missing a check for `\is_array($compare)` ?

src/Extension/CoreExtension.php
```diff
@@ -939,31 +939,100 @@ function twig_in_filter($value, $compare)
...
+ * @see https://wiki.php.net/rfc/string_to_number_comparison
+ * @see https://wiki.php.net/rfc/trailing_whitespace_numerics
+ *
+ * @internal
+ */
+function twig_compare($a, $b)
+{
+    // strict?
+    if (\is_object($a) || \is_resource($a) || \is_object($b) || \is_resource($b)) {
+        if ($a instanceof \SimpleXMLElement || $b instanceof \SimpleXMLElement) {
```
stof: what about DateTime and DateTimeImmutable value objects too ?

fabpot:
This does not change the semantics of `<=>`, but we could do it as well by changing `twig_compare` to return 1, 0, -1.

src/Extension/CoreExtension.php
jvasseur: Maybe instances of `\DateTimeInterface` should be compared with `==` to, it would be deceptive if dates are equals only if they are the same instance.

stof:
@fabpot looks like you forgot to push the second commit.

src/Extension/CoreExtension.php
fabpot: the whole condition has been removed now.

fabpot:
@stof now with the 2 commits :)

stof:
also, making `==` strict for objects will hurt people using value objects, as they would not have a comparison by value anymore (and that would be a BC break)

src/Extension/CoreExtension.php
fabpot: I've removed this whole condition, an artefact of the first version I wrote.

fabpot:
@stof this was a mistake, removed now :)

stof:
What is the performance impact of the `==` change ?

stof:
Also should we wait for the vote on the PHP RFC to be sure that PHP 8 will indeed have this behavior ?

fabpot:
There is a performance impact as we add some function calls.
Regarding the RFC, I don't think we need to wait, the new way makes more sense IMHO.

stof:
is the new behavior strictly equivalent to the PHP RFC ? I.e. could we change the EqualBinary compilation to call the parent method on PHP 8+ in the future ? Or is there some difference ?

stof:
btw, this question may be a reason to wait for the RFC to be voted, to be sure that the answer would not change.

fabpot:
Closing, will reopen for 3.0.

fabpot:
see #2966

PR #2955, opened 2019-04-23
Let's talk about Twig 3.0

fabpot:


src/Environment.php
```diff
@@ -304,31 +266,27 @@ public function getTemplateClass($name, $index = null)
     /**
      * Renders a template.
      *
-     * @param string|TemplateWrapper $name    The template name
-     * @param array                  $context An array of parameters to pass to the template
```
stof, line 272: shouldn't we keep the argument description here ?

src/Environment.php
```diff
@@ -490,31 +424,19 @@ public function isTemplateFresh($name, $time)
...
-            }
-            if ($name instanceof TemplateWrapper) {
-                return $name;
-            }
-
             try {
-                return $this->loadTemplate($name);
+                return $this->load($name);
             } catch (LoaderError $e) {
-                if (1 === \count($names)) {
```
stof: is it expected to remove that case ?

src/Environment.php
```diff
@@ -715,8 +593,6 @@ public function setExtensions(array $extensions)
     }
 
     /**
-     * Returns all registered extensions.
-     *
      * @return ExtensionInterface[] An array of extensions (keys are for internal usage only and should not be relied on)
      */
     public function getExtensions()
```
stof, line 598: should it have a `array` return type ?

fabpot:
Better to review each commit separately to avoid the noise of the removal of the `lib` directory.

src/Extension/EscaperExtension.php
```diff
@@ -81,19 +81,15 @@ public function getDefaultStrategy($name)
...
 namespace {
 /**
  * Marks a variable as being safe.
  *
  * @param string $string A PHP variable
- *
- * @return string
  */
-function twig_raw_filter($string)
+function twig_raw_filter(string $string): string
```
stof, line 92: This is actually a BC break AFAICT, as it was not casting previously, which can have an impact on the actual behavior

src/ExtensionSet.php
```diff
@@ -333,10 +327,7 @@ public function getTests(): array
         return $this->tests;
     }
 
-    /**
-     * @return TwigTest|false
-     */
-    public function getTest(string $name)
+    public function getTest($name): ?TwigTest
```
nicolas-grekas: missing type hint?

src/Loader/FilesystemLoader.php
```diff
@@ -62,21 +58,13 @@ public function getPaths($namespace = self::MAIN_NAMESPACE)
...
-    public function getNamespaces()
+    public function getNamespaces(): array
     {
         return array_keys($this->paths);
     }
 
-    /**
-     * Sets the paths where templates are stored.
-     *
-     * @param string|array $paths     A path or an array of paths where to look for templates
```
stof, line 76: `@param string|array $paths` should not be removed, as there is no typehint for it.

src/Loader/FilesystemLoader.php
```diff
@@ -165,20 +143,19 @@ public function exists($name)
...
     }
 
     /**
      * Checks if the template can be found.
      *
-     * @param string $name  The template name
-     * @param bool   $throw Whether to throw an exception when an error occurs
+     * @param bool $throw Whether to throw an exception when an error occurs
      *
      * @return string|false The template name or false
```
stof, line 156: Should that one be changed to `string|null` as well (maybe adding support for `string|false|null` in 2.x to allow child classes to migrate to returning `null` earlier ?) to allow typehinting ?

src/Parser.php
```diff
@@ -361,24 +338,15 @@ private function filterBodyNodes(Node $node, bool $nested = false)
...
         }
 
-        // the "&& !$node instanceof SpacelessNode" part of the condition must be removed in 3.0
-        if ($node instanceof NodeOutputInterface && !$node instanceof SpacelessNode) {
-            return;
+        if ($node instanceof NodeOutputInterface) {
+            return null;
         }
 
         // here, $nested means "being at the root level of a child template"
```
stof, line 352: the `'Twig_Node' !== \get_class($node)` part of the next condition must be removed too

src/Token.php
```diff
@@ -37,12 +35,7 @@ final class Token
     const INTERPOLATION_START_TYPE = 10;
     const INTERPOLATION_END_TYPE = 11;
 
-    /**
-     * @param int    $type   The type of the token
-     * @param string $value  The token value
```
stof, line 42: the type of `$value` should still be documented

src/Token.php
```diff
@@ -81,39 +72,22 @@ public function test($type, $values = null)
...
-     * @return int
-     */
-    public function getType()
+    public function getType(): int
     {
         return $this->type;
     }
 
-    /**
-     * @return string
```
stof, line 101: the return type needs to be documented

src/ExtensionSet.php
fabpot: fixed

stof:
I guess that the target branch will be changed before merging, right ?

src/ExtensionSet.php
```diff
@@ -116,9 +84,14 @@ public function getSignature(): string
         return json_encode(array_keys($this->extensions));
     }
 
+    public function initialize()
+    {
+        $this->initialized = true;
```
stof, line 89: this change is wrong, because `getFunction` would not initialize the extensions to load the functions in case this method is called. This will break the usage of non-optimized callables in functions/filters/tests/etc... in case templates are loaded from the cache (compiling a template before rendering it would not call this method before the actual initialization).

I think we need to keep a separate `runtimeInitialized` in case we need to keep the current behavior of `isInitialized` becoming true after the first rendering (btw, the code preventing registering new stuff was not looking at `runtimeInitialized` before, which might be a bug).

fabpot:
@stof oops, I have not used the right branch for the PR. Let's try again.

stof:
@fabpot you could update the target branch, to keep the same PR

src/Token.php
fabpot: I've removed it as this is mixed.

src/Token.php
fabpot: I don't think so as this is mixed.

src/Environment.php
fabpot: Yes, as I've done an early return at the beginning of the method instead.

src/Token.php
stof: is it actually mixed, or `scalar|null` which is more precise ?

And then, documenting it is still better, as it tells people what this indeed is (plus, we need to update the phpdoc in older branch if it is wrong when saying `string`)

src/Parser.php
fabpot: see #2972

src/Environment.php
fabpot: I don't think so

src/Environment.php
fabpot: added

src/Extension/EscaperExtension.php
fabpot: reverted

src/Loader/FilesystemLoader.php
fabpot: indeed, reverted

src/Loader/FilesystemLoader.php
fabpot: @stof Good idea, implemented in 2.x in #2973

src/ExtensionSet.php
fabpot: see #2970

PR #2958, opened 2019-04-23
Change Environment::resolveTemplate() to always return a TemplateWrapper

fabpot:


src/Environment.php
```diff
@@ -490,31 +490,21 @@ public function isTemplateFresh($name, $time)
...
-            }
-            if ($name instanceof TemplateWrapper) {
-                return $name;
-            }
-
             try {
-                return $this->loadTemplate($name);
+                return $this->load($name);
             } catch (LoaderError $e) {
-                if (1 === \count($names)) {
```
stof, line 515: is it expected to remove this direct re-throwing in case of resolving a single template name ? This would hurt DX by replacing the original message with `Unable to find one of the following templates: "foo.html.twig".`

fabpot: We don't remove the throw as in the case of 1 template, it's thrown in the `if` before the `foreach`.

stof: hmm, indeed, I missed that.

stof (approved):

PR #2960, opened 2019-04-23
Remove the spaceless tag

fabpot:


src/Parser.php
```diff
@@ -361,23 +359,14 @@ private function filterBodyNodes(Node $node, bool $nested = false)
...
-        if (!$nested && $node instanceof SpacelessNode) {
-            @trigger_error(sprintf('Using the spaceless tag at the root level of a child template in "%s" at line %d is deprecated since Twig 2.5.0 and will become a syntax error in 3.0.', $this->stream->getSourceContext()->getName(), $node->getTemplateLine()), E_USER_DEPRECATED);
-        }
-
         // "block" tags that are not captured (see above) are only used for defining
         // the content of the block. In such a case, nesting it does not work as
         // expected as the definition is not part of the default template code flow.
-        if ($nested && ($node instanceof BlockReferenceNode || $node instanceof \Twig_Node_BlockReference)) {
-            //throw new SyntaxError('A block definition cannot be nested under non-capturing nodes.', $node->getTemplateLine(), $this->stream->getSourceContext());
-            @trigger_error(sprintf('Nesting a block definition under a non-capturing node in "%s" at line %d is deprecated since Twig 2.5.0 and will become a syntax error in 3.0.', $this->stream->getSourceContext()->getName(), $node->getTemplateLine()), E_USER_DEPRECATED);
```
stof: this change is not actually about the removal of the spaceless tag. It would be better to make it a separate commit to keep the history readable about what each commit does.

fabpot: indeed, removed from this PR.

stof (approved):

fabpot:
Use `{% apply spaceless %}...{% endapply %}` instead.

PR #2964, opened 2019-04-23
Simplify code

fabpot:


stof (approved):

src/Environment.php
```diff
@@ -376,15 +376,7 @@ public function load($name)
      *
      * @internal
      */
-    public function loadTemplate($name, $index = null)
```
yceruto, line 379: Shouldn't this be deprecated first in the previous version? It's a BC break without previous warning.

yceruto: NVM, I just saw that it's marked as `@internal` method

PR #2965, opened 2019-04-23
Change false to null to allow for type hinting

fabpot:


src/ExtensionSet.php
```diff
@@ -362,10 +356,7 @@ public function getTests(): array
         return $this->tests;
     }
 
-    /**
-     * @return TwigTest|false
-     */
-    public function getTest(string $name)
+    public function getTest($name): ?TwigTest
```
stof: why removing the argument typehint ?

stof (changes requested):

src/ExtensionSet.php
fabpot: fixed

fabpot:
@stof Is this one ok for you?

stof (approved):

PR #2966, opened 2019-04-23
Make the "in" operator and <,>,<=,>=,==,!= more strict when comparing strings and integers/floats

### LINKED ISSUE #340: Casting Issue / Threequals request

### LINKED ISSUE #341: Strict comparison

### LINKED ISSUE #2824: Bug in containment operator?

fabpot:
The PHP non-strict comparison operator has some counter-intuitive behaviors, which we inherit in Twig. For instance, `{{ 'text' in [0] }}` returns `true`.

PHP is probably going to fix it in version 8: https://wiki.php.net/rfc/string_to_number_comparison

I propose to implement the same semantics in Twig for 3.0.

closes #2824, closes #341, closes #340

### DIFF
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -907,29 +907,99 @@ function twig_in_filter($value, $compare)
         $value = (string) $value;
     }
 
-    if (\is_array($compare)) {
-        return \in_array($value, $compare, \is_object($value) || \is_resource($value));
-    } elseif (\is_string($compare) && (\is_string($value) || \is_int($value) || \is_float($value))) {
-        return '' === $value || false !== strpos($compare, (string) $value);
-    } elseif ($compare instanceof \Traversable) {
-        if (\is_object($value) || \is_resource($value)) {
-            foreach ($compare as $item) {
-                if ($item === $value) {
-                    return true;
-                }
-            }
-        } else {
-            foreach ($compare as $item) {
-                if ($item == $value) {
-                    return true;
-                }
-            }
+    if (\is_string($compare)) {
+        if (\is_string($value) || \is_int($value) || \is_float($value)) {
+            return '' === $value || false !== strpos($compare, (string) $value);
         }
 
         return false;
     }
 
+    if (!is_iterable($compare)) {
+        return false;
+    }
+
+    if (\is_object($value) || \is_resource($value)) {
+        if (!is_array($compare)) {
+            foreach ($compare as $item) {
+                if ($item === $value) {
+                    return true;
+                }
+            }
+
+            return false;
+        }
+
+        return in_array($value, $compare, true);
+    }
+
+    foreach ($compare as $item) {
+        if (0 === twig_compare($value, $item)) {
+            return true;
+        }
+    }
+
     return false;
+}
+
+/**
+ * Compares two values using a more strict version of the PHP non-strict comparison operator.
+ *
+ * @see https://wiki.php.net/rfc/string_to_number_comparison
+ * @see https://wiki.php.net/rfc/trailing_whitespace_numerics
+ *
+ * @internal
+ */
+function twig_compare($a, $b)
+{
+    // int <=> string
+    if (is_int($a) && is_string($b)) {
+        $b = trim($b);
+        if (!is_numeric($b)) {
+            return (string) $a <=> $b;
+        }
+        if ((int) $b == $b) {
+            return $a <=> (int) $b;
+        } else {
+            return (float) $a <=> (float) $b;
+        }
+    }
+    if (is_string($a) && is_int($b)) {
+        $a = trim($a);
+        if (!is_numeric($a)) {
+            return $a <=> (string) $b;
+        }
+        if ((int) $a == $a) {
+            return (int) $a <=> $b;
+        } else {
+            return (float) $a <=> (float) $b;
+        }
+    }
+
+    // float <=> string
+    if (is_float($a) && is_string($b)) {
+        if (is_nan($a)) {
+            return 1;
+        }
+        if (!is_numeric($b)) {
+            return (string) $a <=> $b;
+        }
+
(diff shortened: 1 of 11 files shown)


src/Extension/CoreExtension.php
```diff
@@ -911,31 +911,91 @@ function twig_in_filter($value, $compare)
...
-            }
+    if (\is_string($compare)) {
+        if (\is_string($value) || \is_int($value) || \is_float($value)) {
+            return '' === $value || false !== strpos($compare, (string) $value);
         }
 
         return false;
     }
 
+    if (\is_object($value) || \is_resource($value)) {
```
stof: this is missing the `is_array($compare)` check. Calling `in_array` on a Traversable won't work fine AFAICT.

src/Extension/CoreExtension.php
```diff
@@ -911,31 +911,91 @@ function twig_in_filter($value, $compare)
...
+        return in_array($value, $compare, true);
+    }
+
+    if (!\is_array($compare) && !$compare instanceof \Traversable) {
```
stof: this could be `!is_iterable($compare)` as you target PHP 7.2+

src/Extension/CoreExtension.php
```diff
@@ -911,31 +911,91 @@ function twig_in_filter($value, $compare)
...
+            return true;
+        }
+        if (!is_numeric($a)) {
+            return $a <=> (string) $b;
+        }
+
+        return (float) $a <=> $b;
+    }
+
+    // fallback to ==
```
stof: comment looks wrong

src/Node/Expression/Binary/EqualBinary.php
```diff
@@ -15,6 +15,21 @@
 
 class EqualBinary extends AbstractBinary
 {
+    public function compile(Compiler $compiler)
+    {
+        if (version_compare(PHP_VERSION, '8.0.0', '>=')) {
```
stof: should use `PHP_VERSION_ID` for consistency

stof (approved):

src/Extension/CoreExtension.php
fabpot: done

src/Extension/CoreExtension.php
fabpot: fixed

src/Node/Expression/Binary/EqualBinary.php
fabpot: done

src/Extension/CoreExtension.php
stof: @fabpot what about this one ?

fabpot: I'm moved the `is_iterable` check above this condition, does that work for you?

stof: nope. We still don't want to pass iterators there.

stof: and tests are not catching this, because the case of `value in iterator` is not covered by the tests.

fabpot: fixed, test added

fabpot:
@nikic Could you have a look at the `twig_compare()` method implemented in this pull request and tell me if that's a correct implementation of the https://wiki.php.net/rfc/string_to_number_comparison RFC? Thank you.

src/Extension/CoreExtension.php
```diff
@@ -907,31 +907,101 @@ function twig_in_filter($value, $compare)
...
+        if ((int) $a == $a) {
+            return (int) $a <=> $b;
+        } else {
+            return (float) $a <=> (float) $b;
+        }
+    }
+
+    // float <=> string
+    if (is_float($a) && is_string($b)) {
+        if ($a === NAN) {
```
nikic: It's not possible to check for NAN in this way, `is_nan()` has to be used instead.

nikic (commented):
Apart from the NAN check, this looks right to me.

src/Extension/CoreExtension.php
fabpot: Fixed. I've also added a test to be sure that "NAN" does not compare to NAN or any other strings as per the RFC.

PR #2970, opened 2019-04-23
Remove InitRuntimeInterface

fabpot:


src/ExtensionSet.php
```diff
@@ -106,9 +84,14 @@ public function getSignature(): string
         return json_encode(array_keys($this->extensions));
     }
 
+    public function initialize()
+    {
+        $this->initialized = true;
```
stof: my comment on https://github.com/twigphp/Twig/pull/2955#discussion_r277698843 still applies.

stof (changes requested):

src/ExtensionSet.php
fabpot: I know :)

fabpot: "

this change is wrong, because getFunction would not initialize the extensions to load the functions in case this method is called. This will break the usage of non-optimized callables in functions/filters/tests/etc... in case templates are loaded from the cache (compiling a template before rendering it would not call this method before the actual initialization).

I think we need to keep a separate runtimeInitialized in case we need to keep the current behavior of isInitialized becoming true after the first rendering (btw, the code preventing registering new stuff was not looking at runtimeInitialized before, which might be a bug).
"

fabpot: @stof fixed now

PR #2971, opened 2019-04-23
Add type hints

fabpot:


src/Extension/AbstractExtension.php
```diff
@@ -13,32 +13,32 @@
...
 abstract class AbstractExtension implements ExtensionInterface
 {
-    public function getTokenParsers()
+    public function getTokenParsers(): array
     {
         return [];
     }
 
-    public function getNodeVisitors()
+    public function getNodeVisitors(): array
```
stof, line 21: this will affect all extension authors, while they could already add it in older versions.

@nicolas-grekas is there any way to make Symfony's DebugClassLoader warn about future return types, similar to how you added support for future arguments ?

fabpot: @stof https://github.com/symfony/symfony/pull/30323

PR #2973, opened 2019-04-23
Allow Twig\Loader\FilesystemLoader::findTemplate() to return "null" instead of "false" (same meaning)

fabpot:
That will allow to type hint properly in Twig 3.0.


stof:
should we also do it in 1.x, to allow migrating Symfony's loader, which supports both 1.x and 2.x ?

fabpot:
@stof done

stof (approved):

PR #2975, opened 2019-04-24
Refactor the with node to avoid leaking the parent scope to the template

stof:
This is a small BC break for anyone relying on the fact that the `_parent` variable could be used to access (and replace) the parent scope. See https://github.com/twigphp/Twig/pull/2246#discussion_r93816236

### DIFF
--- a/src/Node/WithNode.php
+++ b/src/Node/WithNode.php
@@ -34,6 +34,10 @@ class WithNode extends Node
     {
         $compiler->addDebugInfo($this);
 
+        $parentContextName = $compiler->getVarName();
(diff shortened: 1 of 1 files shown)


fabpot:
`_parent` being prefixed with `_` is internal, so no BC break IMHO here.

fabpot:
Thank you @stof.

stof:
is there anything defining this in Twig ? I don't think so. We only have 3 special variable names (and btw, we don't forbid assigning them, even though we won't read the assigned value when reading them)

PR #2977, opened 2019-04-24
Add the apply tag

### LINKED ISSUE #2929: Variable declared inside spaceless filter is unknown outside the filter
Hi,

due to the recently introduced deprecation I changed all occurrences from `{% spaceless %}` to `{% filter spaceless %}` in my bundle, just to find out that this [introduced a bug](https://github.com/kevinpapst/AdminLTEBundle/issues/56). 

Before:
```
{% spaceless %}
{% set item = 'foo' %}
{% endspaceless %}
{{ dump(item) }}
```
results in `"foo"`

After the switch:
```
{% filter spaceless %}
{% set item = 'foo' %}
{% endfilter %}
{{ dump(item) }}
```
results in an Exception `Variable "item" does not exist.`.

I searched the docs whether this is intended behavior or not, but couldn't find anything.
Looking at the compiled source I see why that happens, everything between the filter tags becomes its own renderBlock function, while before it was inside the original block function... so I guess its by design. 

But: I think the information about changed scopes should be included in here: https://twig.symfony.com/doc/2.x/deprecated.html#tags
Is that worth a PR?

fabpot:
The `apply` tag replaces the `filter` tag to be able to fix #2929 without a BC break. The two tags do the same thing except that the `apply` tag does not create a new scope.

### DIFF
--- a/doc/advanced.rst
+++ b/doc/advanced.rst
@@ -400,16 +400,21 @@ Most of the time though, a tag is not needed:
       {{ '**markdown** text'|markdown }}
 
   If you want use this filter on large amounts of text, wrap it with the
-  :doc:`filter <tags/filter>` tag:
+  :doc:`apply <tags/apply>` tag:
 
   .. code-block:: jinja
 
-      {% filter markdown %}
+      {% apply markdown %}
       Title
       =====
 
       Much better than creating a tag as you can **compose** filters.
-      {% endfilter %}
+      {% endapply %}
+
+ .. note::
+
+      The ``apply`` tag was introduced in Twig 1.40; use the ``filter`` tag with
+      previous versions.
 
 * If your tag does not output anything, but only exists because of a side
   effect, create a **function** that returns nothing and call it via the
--- a/src/NodeVisitor/OptimizerNodeVisitor.php
+++ b/src/NodeVisitor/OptimizerNodeVisitor.php
@@ -111,6 +111,7 @@ class OptimizerNodeVisitor extends AbstractNodeVisitor
                 if (!$expression && 'Twig_Node' !== \get_class($node) && $prependedNodes = array_shift($this->prependedNodes)) {
                     $nodes = [];
                     foreach (array_unique($prependedNodes) as $name) {
+                        die();
                         $nodes[] = new SetTempNode($name, $node->getTemplateLine());
                     }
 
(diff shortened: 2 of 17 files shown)


doc/advanced.rst
```diff
@@ -400,16 +400,21 @@ Most of the time though, a tag is not needed:
...
       =====
 
       Much better than creating a tag as you can **compose** filters.
-      {% endfilter %}
+      {% endapply %}
+
+ .. note::
+
+      The ``apply`` tag was introduced in Twig 1.40; use the ``filter`` tag with
+      previous versions.
```
stof, line 417: I would add a note about `but be careful that it introduces a new scope for variables`.

src/TokenParser/ApplyTokenParser.php
```diff
@@ -0,0 +1,58 @@
...
+        $ref->setAttribute('always_defined', true);
+
+        $filter = $this->parser->getExpressionParser()->parseFilterExpressionRaw($ref, $this->getTag());
+
+        $this->parser->getStream()->expect(Token::BLOCK_END_TYPE);
+        $body = $this->parser->subparse([$this, 'decideApplyEnd'], true);
+        $this->parser->getStream()->expect(Token::BLOCK_END_TYPE);
+
+        return new Node([
+            new SetNode(true, $ref, $body, $lineno, $this->getTag()),
```
stof, line 44: This will still leak a variable in the context after that node. Could we have an internal `UnsetNode` after the printing ? Or maybe even better a way to reference a local variable rather than using a `NameExpression` ? `SetNode` in capture mode already creates a locale variable first.

stof (changes requested):

doc/advanced.rst
fabpot: Not sure it's relevant here as I added a note about this in the `filter`docs and nobody was really aware before.

src/TokenParser/ApplyTokenParser.php
fabpot: I've replaced `NameExpression` by `TempNameExpression`, which is exactly what we are looking for here.

src/NodeVisitor/OptimizerNodeVisitor.php
```diff
@@ -111,6 +111,7 @@ protected function doLeaveNode(Node $node, Environment $env)
                 if (!$expression && 'Twig_Node' !== \get_class($node) && $prependedNodes = array_shift($this->prependedNodes)) {
                     $nodes = [];
                     foreach (array_unique($prependedNodes) as $name) {
+                        die();
```
fabpot, line 114: oops, fixed

PR #2981, opened 2019-04-25
Allow Twig\NodeVisitor\NodeVisitorInterface::leaveNode() to return "null" instead of "false" (same meaning)

fabpot:


fabpot:
We don't have any node visitor in Twig core using this feature (but Symfony has one).

PR #2983, opened 2019-04-26
Add Parser typehint

ruudk:
This makes it easier to work with.


fabpot:
Thank you @ruudk.

fabpot:
Thank you @ruudk.

PR #2985, opened 2019-04-27
Optimize context access on PHP 7.4

fabpot:


stof:
is `array_key_exists` optimized by the compiler on PHP 7.4 ? that's a good news

fabpot:
@stof https://github.com/php/php-src/pull/3360

stof:
yeah, that's a great news.

CHANGELOG
```diff
@@ -1,5 +1,6 @@
 * 1.40.0 (2019-XX-XX)
```
stof: 1.40.0 is already released.

fabpot: fixed

PR #2991, opened 2019-04-29
Fix regression in NodeTraverser

fabpot:


src/NodeTraverser.php
```diff
@@ -69,7 +69,9 @@ protected function traverseForVisitor(NodeVisitorInterface $visitor, \Twig_NodeI
         $node = $visitor->enterNode($node, $this->env);
 
         foreach ($node as $k => $n) {
-            if (false !== ($m = $this->traverseForVisitor($visitor, $n)) && null !== $m) {
+            if (null === $n) {
+                $node->setNode($k, $n);
```
stof: this could be `continue` instead. You don't need to set it again (as it is already set) and it will avoid a deprecation here about setting `null`

stof (commented):
A test should be added covering this to avoid accidental removal.

src/NodeTraverser.php
fabpot: Using `continue` makes the code more verbose. In 2.x, this code is not needed as `$n` can never be `null`.

stof: why is `continue` more verbose than `$node->setNode($k, $n);` ?

fabpot:
With a test now.

src/NodeTraverser.php
fabpot: Oh, got it now :)

fabpot:
and green!

stof (approved):

alexpott:
I've tested this against Drupal 8.6.x and it looks like it works nicely. Thank you for the speedy response!

PR #2993, opened 2019-04-30
Add tests to avoid regressions (via Drupal)

fabpot:


alexpott:
This looks amazing. The only concern I have is the dependency on Drupal demo content that is subject to change - unlikely but obviously you have little control over that.

Not seen the blackfire player before either - that looks very neat.

fabpot:
If the demo changes, it will break and I will update the test, so that's a no-brainer IMHO.

.travis.yml
```diff
@@ -7,37 +7,40 @@ cache:
...
+    - wget https://get.symfony.com/cli/installer -O - | bash
 
 before_script:
     - if [ "$TWIG_EXT" == "yes" ]; then sh -c "cd ext/twig && phpize && ./configure --enable-twig && make && make install"; fi
     - if [ "$TWIG_EXT" == "yes" ]; then echo "extension=twig.so" >> `php --ini | grep "Loaded Configuration" | sed -e "s|.*:\s*||"`; fi
 
-script: ./vendor/bin/simple-phpunit
+script: |
+    ./vendor/bin/simple-phpunit
+    if [ "$DRUPAL" == "yes" ]; then ./drupal_test.sh; fi
```
stof: please do that with 2 separate entries for `script` (it supports arrays, like other sections), to be sure to have proper error handling from Travis here.

drupal_test.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
+
+set -x
+set -e
+
+export
+
+composer create-project --no-interaction drupal-composer/drupal-project:8.x-dev drupal
+cd drupal
+composer config repositories.fabpot vcs https://github.com/fabpot/Twig
```
stof: you should not add your own fork as a repository here. That won't work for PR builds (and will run the wrong code for branch builds in the main repo). You should instead add a `path` repository pointing to the local Twig clone instead (the one available here). This would probably require making this `drupal` folder a sibling instead of a child though

alexpott:
In that case this just look amazing :) Thanks for doing this.

.travis.yml
```diff
@@ -7,37 +7,40 @@ cache:
...
     - TWIG_EXT=no
+    - DRUPAL=no
 
 before_install:
-    # turn off XDebug
     - phpenv config-rm xdebug.ini || return 0
 
 install:
     - travis_retry composer install
+    - wget https://get.symfony.com/cli/installer -O - | bash
```
stof: why do we need that on all jobs ? `./drupal_test.sh` already does that

stof (changes requested):

.travis.yml
fabpot: done

.travis.yml
fabpot: a left-over, removed

drupal_test.sh
fabpot: Good idea, I will do that.

stof:
idea: override the `script` section entirely for the job running drupal tests, to run *only* the Drupal tests (no need to run the testsuite again). Btw, the `drupal_test.sh` script may then be split into 2 different script, to move all the installation part to `install` or `before_script` rather than `script`, for a better output.

And maybe move it to a separate stage so that it runs only if the testsuite is green ?

drupal_test.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+
+set -x
+set -e
+
+cd /tmp
+composer create-project --no-interaction drupal-composer/drupal-project:8.x-dev drupal
+cd drupal
+composer config repositories.pr path $HOME
```
stof: is `$HOME` the actual location of the cloned repo ? I don't think so (and this would not be reusable locally). I suggest reading the location with `pwd` before changing to `/tmp` instead.

drupal_test.sh
```diff
@@ -0,0 +1,51 @@
...
+        param name name
+        param pass pass
+        expect status_code() == 303
+    follow
+        expect status_code() == 200
+    click link('Structure')
+        expect status_code() == 200
+EOF
+./blackfire-player.phar run drupal-tests.bkf --endpoint=$ENDPOINT --variable name=$DRUPAL_Username --variable pass=$DRUPAL_Password
+symfony server:stop
```
stof, line 51: this wil not run if the player fails, right ?

fabpot: Right, but this is only needed locally when testing (Travis will kill it anyway).

drupal_test.sh
```diff
@@ -0,0 +1,51 @@
...
+set -x
+set -e
+
+REPO=`pwd`
+cd /tmp
+rm -rf drupal
+composer create-project --no-interaction drupal-composer/drupal-project:8.x-dev drupal
+cd drupal
+(cd vendor/twig && rm -rf twig && ln -sf $REPO twig)
+echo '$config["system.logging"]["error_level"] = "verbose";' >> web/sites/default/settings.php
```
stof, line 12: should we have some `after_failure` step in the Travis job which would display the latest logs ?

fabpot: This line allows to get detailed errors in the browser, which helps debugging errors locally.

stof: but is there a way to get the error message on Travis ?

drupal_test.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+
+set -x
+set -e
+
+REPO=`pwd`
+cd /tmp
+rm -rf drupal
```
stof: should we name the folder `drupal-twig-test` or something like that, to be less likely to remove some meaningful drupal tmp content when running this locally ?

fabpot:
This is now ready and I tested that reverting a recent fix that broke Drupal, the tests would not have passed.

drupal_test.sh
fabpot: Ok

drupal_test.sh
fabpot: Not really. My idea (for now at least) is to have a red feedback when we break something and be able to replicate the error locally for easier debugging. The current script works well on my laptop, which is enough I think for now.

drupal_test.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+
+set -x
```
stof, line 3: `set -x` should probably be removed, to make the Travis output more readable (we could still display some specific messages at some points if we need them to understand what gets output)

PR #2995, opened 2019-05-03
Fix partial output leak when a PHP fatal error occurs

### LINKED ISSUE #1962: PHP Catchable fatal errors causes partial rendered template to leak to client
Hi, I'm using twig to format HTML emails to be sent from an API endpoint implemented in Silex. I discovered that if a "PHP Catchable fatal error" is triggered during template rendering, the partially rendered template is output to the API client (!). This almost feels like a security issue, as the rendered template is not intended for the API client but for someone else entirely.

This only seems to happen with some errors, for example when specifying an {{ object }} which cannot be converted to a string. Other errors seem to just throw exceptions and php errors but do not leak the template.

Using twig/twig v1.23.3 from composer, here is a sample case:

``` php
<?php
require('vendor/autoload.php');

class SomeClass { public $value1; public $value2; }
$templates = array(
  'SECRET MAIL CONTENTS no leaks: {{0/0}}',
  'SECRET MAIL CONTENTS no leaks: {{invalid}}',
  'SECRET MAIL CONTENTS no leaks: {{invalid()}}',
  'SECRET MAIL CONTENTS no leaks: {{invalid',
  'SECRET MAIL CONTENTS no leaks: {{obj.invalid}}',
  'SECRET MAIL CONTENTS template DOES leak: {{obj}}',
  'SECRET MAIL CONTENTS no leaks: {{constant(PHP_VERSION)}}',
);
$idx = $_SERVER['argv'][1];
echo "\n\nTrying template $idx\n";
$twig = new Twig_Environment(new Twig_Loader_Array(array('index' => $templates[$idx])));
$obj = new SomeClass();
$htmlForAdmin = $twig->render('index', array('obj' => $obj));
echo "\nDone with template $idx, thanks.\n\n";
```

If you run through the possible templates, you'll see that the fifth template leaks the partial template to the client:

``` sh
for i in 1 2 3 4 5 6 7; do php -d error_reporting=0 testbug.php $i; done
```

Yields:

```
Trying template 1
Done with template 1, thanks.
Trying template 2
Trying template 3
Trying template 4
Done with template 4, thanks.
Trying template 5
SECRET MAIL CONTENTS template DOES leak: 
Trying template 6
Done with template 6, thanks.
Trying template 7
```

And with error reporting for the leaking template:

``` sh
php -d error_reporting=8191 testbug.php 5
```

yields:

```
Trying template 5
PHP Catchable fatal error:  Object of class SomeClass could not be converted to string in /private/tmp/twig_bug/vendor/twig/twig/lib/Twig/Environment.php(403) : eval()'d code on line 20
PHP Stack trace:
PHP   1. {main}() /private/tmp/twig_bug/testbug.php:0
PHP   2. Twig_Environment->render() /private/tmp/twig_bug/testbug.php:18
PHP   3. Twig_Template->render() /private/tmp/twig_bug/vendor/twig/twig/lib/Twig/Environment.php:347
PHP   4. Twig_Template->display() /private/tmp/twig_bug/vendor/twig/twig/lib/Twig/Template.php:366
PHP   5. Twig_Template->displayWithErrorHandling() /private/tmp/twig_bug/vendor/twig/twig/lib/Twig/Template.php:355
PHP   6. __TwigTemplate_6927891b70918e9a5f4eaf6d00ccd7cf0bfc98b7f781b770f99869a64797c516->doDisplay() /private/tmp/twig_bug/vendor/twig/twig/lib/Twig/Template.php:381
SECRET MAIL CONTENTS template DOES leak: 
```

(This is on PHP 5.6.17 via homebrew on OSX 10.11, by the way)

fabpot:
closes #1962 (thank you @fluff for the "trick")

### DIFF
--- a/src/Extension/DebugExtension.php
+++ b/src/Extension/DebugExtension.php
@@ -54,7 +54,7 @@ function twig_var_dump(Environment $env, $context, array $vars = [])
         return;
     }
 
-    ob_start();
+    ob_start(function () { return ''; });
 
     if (!$vars) {
         $vars = [];
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
 * 1.40.2 (2019-XX-XX)
 
+ * fixed partial output leak when a PHP fatal error occurs
  * optimized context access on PHP 7.4
 
 * 1.40.1 (2019-04-29)
--- a/src/Node/MacroNode.php
+++ b/src/Node/MacroNode.php
@@ -104,7 +104,7 @@ class MacroNode extends Node
             ->outdent()
             ->write("]);\n\n")
             ->write("\$blocks = [];\n\n")
-            ->write("ob_start();\n")
+            ->write("ob_start(function () { return ''; });\n")
             ->write("try {\n")
             ->indent()
             ->subcompile($this->getNode('body'))
--- a/src/Node/SetNode.php
+++ b/src/Node/SetNode.php
@@ -58,7 +58,7 @@ class SetNode extends Node implements NodeCaptureInterface
         } else {
             if ($this->getAttribute('capture')) {
                 $compiler
-                    ->write("ob_start();\n")
+                    ->write("ob_start(function () { return ''; });\n")
                     ->subcompile($this->getNode('values'))
                 ;
             }
--- a/src/Node/SpacelessNode.php
+++ b/src/Node/SpacelessNode.php
@@ -31,7 +31,7 @@ class SpacelessNode extends Node
     {
         $compiler
             ->addDebugInfo($this)
-            ->write("ob_start();\n")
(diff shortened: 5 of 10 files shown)


src/Extension/DebugExtension.php
```diff
@@ -54,7 +54,7 @@ function twig_var_dump(Environment $env, $context, array $vars = [])
         return;
     }
 
-    ob_start();
+    ob_start(function ($in) {return ''; });
```
nicolas-grekas: Missing space before return.

nicolas-grekas: I suppose there is no need for the argument btw.

fabpot: fixed

PR #2996, opened 2019-05-03
Add "filter", "map", and "reduce" filters

fabpot:
This PR adds support for 3 new filters: `filter`, `map`, and `reduce`. They take an arrow function as an argument (a PHP closure).

I have restricted the usage of arrow functions as much as possible as it makes no sense to support them everywhere. So, for now, they are only accepted as arguments to filters (using them as arguments to function is not supported but could be easily added if we have a use case).

The syntax is the following: `(x) => x + 3` where `(x)` is the list of arguments, `=>` starts the body, and the body is any Twig expression. Within the arrow function, the context is also available: `(x) => x + offset` works if `offset` is defined in the current context.

~~These new filters will allow us to deprecate the `if` support on the `for` tag, which does not work well with the `loop` variable:~~

```twig

{% set sizes = {xs: 34, s: 36, m: 38, l: 40, xl: 42} %}

{# before #}
{% for name, size in sizes if size < 38 %}
    {{ name }} = {{ size }}
    {% loop.last ? 'LAST' %} {# <--- works with this PR #}
{% endfor %}

{# after #}
{% for name, size in sizes|filter(size => size < 38) %}
    {{ name }} = {{ size }}
    {{ loop.last ? 'LAST' }}
{% endfor %}
```

This closes #2785

### DIFF
--- a/src/ExpressionParser.php
+++ b/src/ExpressionParser.php
@@ -14,6 +14,7 @@ namespace Twig;
 
 use Twig\Error\SyntaxError;
 use Twig\Node\Expression\ArrayExpression;
+use Twig\Node\Expression\ArrowFunctionExpression;
 use Twig\Node\Expression\AssignNameExpression;
 use Twig\Node\Expression\Binary\ConcatBinary;
 use Twig\Node\Expression\BlockReferenceExpression;
@@ -68,8 +69,12 @@ class ExpressionParser
         }
     }
 
-    public function parseExpression($precedence = 0)
+    public function parseExpression($precedence = 0, $allowArrow = false)
     {
+        if ($allowArrow && $arrow = $this->parseArrow()) {
+            return $arrow;
+        }
+
         $expr = $this->getPrimary();
         $token = $this->parser->getCurrentToken();
         while ($this->isBinary($token) && $this->binaryOperators[$token->getValue()]['precedence'] >= $precedence) {
@@ -96,6 +101,38 @@ class ExpressionParser
         }
 
         return $expr;
+    }
+
+    /**
+     * @return ArrowFunctionExpression|null
+     */
+    private function parseArrow()
+    {
+        $stream = $this->parser->getStream();
+        $token = $this->parser->getCurrentToken();
+        $line = $token->getLine();
+        if (!$token->test(Token::PUNCTUATION_TYPE, '|')) {
+            return null;
+        }
+
+        $stream->next();
+        $names = [];
+        while (true) {
+            $token = $this->parser->getCurrentToken();
+            if (!$token->test(Token::NAME_TYPE)) {
+                throw new SyntaxError(sprintf('Unexpected token "%s" of value "%s".', Token::typeToEnglish($token->getType()), $token->getValue()), $token->getLine(), $stream->getSourceContext());
+            }
+            $names[] = $token->getValue();
+            $stream->next();
+            if (!$stream->nextIf(Token::PUNCTUATION_TYPE, ',')) {
+                break;
+            }
+        }
+
+        $stream->expect(Token::PUNCTUATION_TYPE, '|');
+        $stream->expect(Token::ARROW_TYPE);
+
+        return new ArrowFunctionExpression($this->parseExpression(0), $names, $line);
     }
 
     protected function getPrimary()
@@ -499,7 +536,7 @@ class ExpressionParser
             if (!$this->parser->getStream()->test(Token::PUNCTUATION_TYPE, '(')) {
                 $arguments = new Node();
             } else {
-                $arguments = $this->parseArguments(true);
+                $arguments = $this->parseArguments(true, false, true);
             }
 
             $class = $this->getFilterNodeClass($name->getAttribute('value'), $token->getLine());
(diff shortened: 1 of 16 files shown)


hason:
The `if` on the `for` tag should be preserved as syntactic sugar and internally should be converted to the `filter`.

hason:
What about to pass both value and key as arguments to `filter` arrow function (see [`ARRAY_FILTER_USE_BOTH`](https://www.php.net/manual/en/function.array-filter.php)):

```twig
{% for name, size in sizes|filter(|size[, name]| => name != '...') %}
```

src/Token.php
```diff
@@ -200,6 +204,8 @@ public static function typeToEnglish($type)
                 return 'begin of string interpolation';
             case self::INTERPOLATION_END_TYPE:
                 return 'end of string interpolation';
+            case self::ARROW_TYPE:
+                return 'arrow type';
```
kevin-verschaeve: should be just `arrow` no ?

src/Node/Expression/ArrowFunctionExpression.php
```diff
@@ -0,0 +1,55 @@
...
+            if ($i) {
+                $compiler->raw(', ');
+            }
+
+            $compiler->raw('$__'.$name.'__');
+        }
+        $compiler
+            ->raw(') use ($context) { ')
+        ;
+        foreach ($this->getAttribute('names') as $i => $name) {
```
kevin-verschaeve: `$i` is not used

javiereguiluz:
Don't know if it's technically possible (or desirable) but the `|x| => ...` syntax looks like Ruby to me. I fear PHP developers could find this syntax strange.

Could we use the just approved and implemented `(x) => ...` syntax of PHP arrow functions?

```twig
{% for name, size in sizes|filter((size) => size < 38) %}
    {{ name }} = {{ size }}
    {{ loop.last ? 'LAST' }}
{% endfor %}
```

fabpot:
@javiereguiluz I chose `|a| => ` for a good reason :) It makes parsing way more easier. Using `(a) =>` means that we have parsing ambiguities as `()` have already a meaning. It's far from impossible, but I'd like to avoid adding too much complexity for such a feature.

linaori:
I find `|foo|` very confusing, not because it's ruby, but because `|` means that it's a filter over something. If I see the example: `{% for name, size in sizes|filter(|size| => size < 38) %`, I have a lot of issues seeing what this is as I'm used to treat `|` as a filter over the left side. Reading this, it looks like you're trying to do: `sizes | filter( | size | => size`, which looks really weird to me

src/Token.php
fabpot: I wanted to write `arrow function` :)

src/Node/Expression/ArrowFunctionExpression.php
fabpot: removed

fabpot:
@hason I have changed the condition implementation of the `for` tag to remove the limitation (but we need to now convert the iterated var to an array).

fabpot:
I will see if we can use the more natural `(x) => x + 1` without too many changes.

jfcherng:
How about using the same syntax with PHP 7.4's [array function](https://wiki.php.net/rfc/arrow_functions_v2)? The RFC implementation [has been merged](https://github.com/php/php-src/pull/3941#issuecomment-488666209) into the PHP 7.4 branch already.

```php
fn(x) => x + 1
```

The preceding `fn` makes it much easier to parse.

hason:
> @hason I have changed the condition implementation of the `for` tag to remove the limitation (but we need to now convert the iterated var to an array).

It would be useful to support iterators [CallbackFilterIterator](https://www.php.net/manual/en/class.callbackfilteriterator.php).

fabpot:
@hason I'm going to remove the code for the `for` tag from this PR and create another one after merging the main feature of this PR.

fabpot:
Good news, I've changed the syntax to `(x) => x + 1`!

fabpot:
Now with the possibility to omit the `()` when having only one argument: `x => x + 1`

src/ExpressionParser.php
```diff
@@ -98,6 +103,64 @@ public function parseExpression($precedence = 0)
...
+        }
+
+        // first, determine if we are parsing an arrow function by finding => (long form)
+        $i = 0;
+        if (!$stream->look($i)->test(Token::PUNCTUATION_TYPE, '(')) {
+            return null;
+        }
+        $i++;
+        while (true) {
+            // variable name
```
stof, line 130: shouldn't we validate that this is a NAME_TYPE token instead of just going over it ?

fabpot: That's a first pass to check if we can find the arrow. We will throw an exception if needed when parsing for "real".

stof:
should we allow arrow functions only in filter arguments ? What about function arguments, test arguments or custom calls ?

willrowe:
Would it be possible to pass the key as a second argument for `filter` and `map` and as a third argument for `reduce`?

```
filter((value, key) => ...)
map((value, key) => ...)
reduce((carry, value, key) => ...)
```

fabpot:
@stof I don't want to expose arrow function without a good use case. For now, we only need them for functions, so that's why I've only enabled them there.

fabpot:
@willrowe Done for `filter` and `map`, not sure it makes sense for `reduce` (I don't see a use case).

fabpot:
The feature looks "finished" to me now.

PR #2997, opened 2019-05-03
Add the possibility to use `loop.last` when there is a condition on a `for` tag

fabpot:
The `if` part of a `for` tag is now handled by a `filter` filter:

```twig
{% set sizes = {xs: 34, s: 36, m: 38, l: 40, xl: 42} %}

{# this #}
{% for name, size in sizes if size < 38 %}
    {{ name }} = {{ size }}
    {% loop.last ? 'LAST' %} {# <--- works now with this PR #}
{% endfor %}

{# is the same as #}
{% for name, size in sizes|filter(s => s < 38) %}
    {{ name }} = {{ size }}
    {{ loop.last ? 'LAST' }}
{% endfor %}
```


fabpot:
Only the last commit is relevant (based on #2996 for now).

/cc @hason

apfelbox:
I really love this!

One question though: will this also work if the condition depends on the body of the loop, like in this totally not contrived example:

```twig
{% set a = 1..10 %}
{% set c = false %}

{% for i in a if (i is even or c) %}
{{ i }}
{% set c = i > 5 %}
{% endfor %}
```

<details>
<summary>which should output</summary>
<pre>
2
4
6
7
8
9
10
</pre>
</details>

PS: In this case the condition can be inlined, but that is not the point of my question 😉 
PPS: If this isn't possible and you need to decide on what to support from now on: I would like this feature more than being able to use variables from the condition in the body.

fabpot:
@apfelbox No, that won't work as the array is computed **before** the loop, which is how we can know the **last** element.

apfelbox:
Okay, I assumed that this wouldn't be possible.

Two possible fixes would be:

1. run the body of the loop when filtering (mainly all `set` nodes) as well to the callback to produce the same results.
2. Document that the variables outside `key & value` of the loop are practically captured by value.

Although 2 is a BC break, I would just document it and do it like this PR, as it should be fine.

apfelbox:
Another issue is if the condition is using some objects that might change the internal state in the body of the loop due to side effects.

fabpot:
Avoiding issues like this one if one of the main reason I still think we need to deprecate the `if` construct of the `for` tag in favor of the new `filter` filter which makes it clear that the filter is applied before looping over the values.

stof:
Given that this change is a BC break (the case of https://github.com/twigphp/Twig/pull/2997#issuecomment-489599871 works today), I don't think we should do it.

fabpot:
Deprecation in #2998

stof:
@fabpot shouldn't this PR be closed then ? Breaking BC (in a non-obvious way) for the implementation of a feature getting deprecated looks wrong to me.

stof:
why do we have this `added a test` commit in this PR, while it adds tests for the `filter` filter ? Shouldn't it have been added in the original PR ?

fabpot:
The test has been added in the meantime in 1.x. But anyway, this PR will never be merged :)

PR #2998, opened 2019-05-09
Deprecate "if" conditions on "for" tags

fabpot:


doc/deprecated.rst
```diff
@@ -39,6 +39,9 @@ Tags
 * The ``filter`` tag is deprecated in Twig 2.9. Use the ``apply`` tag instead
   (the ``Twig\TokenParser\FilterTokenParser`` classes is also deprecated).
 
+* Adding an ``if`` condition on a ``for`` tag is deprecated in Twig 2.10. Use a
+``filter`` filter instead.
```
stof: I would say ``use a `filter` filter or a condition inside the `if` instead`` (if your condition depends on a variable updated inside the loop, you cannot migrate to a filter)

stof: same in the deprecation warning

fabpot: Done

lyrixx:
Why is the benfit of deprecating it? It's convenient. It could be seen as syntaxic sugar.
We have already many way to do the same thing with twig. I would let it FWIW

fabpot:
@lyrixx The problem is that there are many edge cases and people keep having issues with it. So, better to use something else and now with the `filter` filter, we have a replacement for all use cases.

lyrixx:
But we can move the if "node" into a filter node thanks to the AST, isn't ?

doc/deprecated.rst
```diff
@@ -39,6 +39,10 @@ Tags
 * The ``filter`` tag is deprecated in Twig 2.9. Use the ``apply`` tag instead
   (the ``Twig\TokenParser\FilterTokenParser`` classes is also deprecated).
 
+* Adding an ``if`` condition on a ``for`` tag is deprecated in Twig 2.10. Use a
+  ``filter`` filter or a condition inside the ``if`` instead (if your condition
```
apfelbox:
```suggestion
  ``filter`` filter or a condition inside the ``loop`` body instead (if your condition
```

Not sure if typo?

stof:
@lyrixx no. If your condition relies on a variable updated inside the loop, moving the condition to a filter would break it (as the filter would run with the initial value of the variable all the time). That's why #2997 (which does exactly that) is a BC break.

doc/deprecated.rst
fabpot: indeed, I've reworded the sentence to read: "or an `if` condition inside the ``for`` body instead"

apfelbox: Great! 👍 

(PS: Oh yeah, `for` oops 😅 )

hason:
I think that we should deprecate only edge case (condition relies on a variable updated inside the loop). The if tag could be syntactic sugar for filter in Twig 3.

Issue #2532, opened 2017-07-20
Shorthand for passing variables to included templates?


cviebrock:
I want to pass a subset of variables from one template to another.  Right now, I think the only way is for me to redefine them all:

```twig
{% include 'sometemplate.twig' with {
    'foo': foo,
    'bar': bar,
    'qua': qua
} only %}
```

It would be nice if there were a shorthand way of doing this, something along the lines of the way PHP's `compact` works.  I don't know what the syntax would look like but maybe:

```twig
{% include 'sometemplate.twig' with vars { foo, bar, qua } only %}
```

or (more typing):

```twig
{% include 'sometemplate.twig' with compact('foo', 'bar', 'qua') only %}
```

My goal is to be a bit more rigid in my templates in terms of the variables they use and expect.  Another option might be some kind of tag at the top of the template that defines the variables it uses (and drops all the others, or throws an error, or something configurable).  e.g.:

```twig
{% expect { foo, bar, qua } %}        // these vars must be present
{% expect { foo, bar, qua } only %}   // same, and all other vars are ignored
```

Is either or both of these options already possible?

cviebrock:
I see this has been brought up before in #502 and #1019.

I'm not trying to beat a (dead?) horse and perhaps more notations isn't the solution as @fabpot suggests.  But this has to be such a common use-case that an "official" solution would be great.

fabpot:
Introducing `compact` or `from_context` as I proposed in #1019 is not for core. We try to limit the amount of filter/tags/functions in core. I think that this addition does not belong to core.

As an aside, I like your `expect tag` :)

cviebrock:
Thanks @fabpot.  I ended up writing a `compact` function that seems to be working just fine: https://gist.github.com/cviebrock/321494afffb2ca22d8705423951d2bf1

Working on `expect` as we speak. :)

Issue #1019, opened 2013-02-28
Support {foo: foo} == {foo}


marfillaster:
I'm not sure what to call this, but I think it would be a nice syntactic sugar to have.

Imagine writing this
`{% include 'some.twig.html' with {a: a, b: b, c: c} %}`
When you can do something like this
`{% include 'some.twig.html' with {a,b,c} %}`

The use case is when you want to pass only some of the context's variables to an included template using the same variable names.

Edit

Ok so we have three options
@fabpot :
`{% include 'some.twig.html' with from_context('a', 'b', 'c') %}`
@stof :
`{% include 'some.twig.html' with [a,b,c] | compact %}`
my proposal :
`{% include 'some.twig.html' with {a, b, c} %}`

PR #3002, opened 2019-05-10
hash literal property value shorthand

sukei:
While doing some Twig templates recently. I was taking myself writing things like this:

```
{{ include('pagination.html.twig', { count: count, current_page: current_page, current_route: current_route }) }}
```
And was thinking about the ability of doing this:

```
{{ include('pagination.html.twig', { count, current_page, current_route }) }}
```

This is the purpose of this PR. This syntactic sugar comes from the JavaScript world (ES6) in the name of `Object Literal Property Value Shorthand` and It's very convenient to use.

### DIFF
--- a/src/ExpressionParser.php
+++ b/src/ExpressionParser.php
@@ -359,6 +359,7 @@ class ExpressionParser
                 }
             }
             $first = false;
+            $keyName = null;
 
             // a hash key can be:
             //
@@ -366,7 +367,9 @@ class ExpressionParser
             //  * a string -- 'a'
             //  * a name, which is equivalent to a string -- a
             //  * an expression, which must be enclosed in parentheses -- (1 + 2)
-            if (($token = $stream->nextIf(/* Token::STRING_TYPE */ 7)) || ($token = $stream->nextIf(/* Token::NAME_TYPE */ 5)) || $token = $stream->nextIf(/* Token::NUMBER_TYPE */ 6)) {
+            if (($token = $stream->nextIf(/* Token::STRING_TYPE */ 7)) || ($token = $stream->nextIf(/* Token::NAME_TYPE */ 5))) {
+                $key = new ConstantExpression($keyName = $token->getValue(), $token->getLine());
+            } elseif ($token = $stream->nextIf(/* Token::NUMBER_TYPE */ 6)) {
                 $key = new ConstantExpression($token->getValue(), $token->getLine());
             } elseif ($stream->test(/* Token::PUNCTUATION_TYPE */ 9, '(')) {
                 $key = $this->parseExpression();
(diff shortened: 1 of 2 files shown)


src/ExpressionParser.php
```diff
@@ -376,8 +377,18 @@ public function parseHashExpression()
                 throw new SyntaxError(sprintf('A hash key must be a quoted string, a number, a name, or an expression enclosed in parentheses (unexpected token "%s" of value "%s".', Token::typeToEnglish($current->getType()), $current->getValue()), $current->getLine(), $stream->getSourceContext());
             }
 
-            $stream->expect(/* Token::PUNCTUATION_TYPE */ 9, ':', 'A hash key must be followed by a colon (:)');
-            $value = $this->parseExpression();
+            // a hash can be shorten when the key and value got the same identifier ({ foo: foo, 'bar': 'bar' } === { foo, 'bar' })
+            if ($stream->nextIf(/* Token::PUNCTUATION_TYPE */ 9, ':')) {
+                $value = $this->parseExpression();
+            } else if ($value === null) {
```
fabpot: CS: `} elseif (null === $value) {` (notice the removed whitespace between `else` and `if`) and the inversion of `null` and `$value`

src/ExpressionParser.php
```diff
@@ -367,7 +368,7 @@ public function parseHashExpression()
             //  * a name, which is equivalent to a string -- a
             //  * an expression, which must be enclosed in parentheses -- (1 + 2)
             if (($token = $stream->nextIf(/* Token::STRING_TYPE */ 7)) || ($token = $stream->nextIf(/* Token::NAME_TYPE */ 5)) || $token = $stream->nextIf(/* Token::NUMBER_TYPE */ 6)) {
-                $key = new ConstantExpression($token->getValue(), $token->getLine());
+                $key = new ConstantExpression($value = $token->getValue(), $token->getLine());
```
fabpot: Here, this the key name, so what about renaming `$value` to `$keyName`?

test/Twig/Tests/ExpressionParserTest.php
```diff
@@ -79,7 +79,6 @@ public function getFailingTestsForArray()
     {
         return [
             ['{{ [1, "a": "b"] }}'],
-            ['{{ {"a": "b", 2} }}'],
```
fabpot: This should be kept and supported as this is valid.

fabpot (changes requested):

fabpot:
I've just re-read some older issues and pull requests on the same topic (#2532, #502, #1019) to name a few). Similar notations were rejected in the past as it makes hashes looks like an array (`{ 'foo', 'bar' }` vs ``[ 'foo', 'bar' ]``), and that might confusing. Thanks for proposing.

sukei:
Too bad :(. That's was a really convenient syntax (essentially for frontend developers/designers who are used to it thanks to ES6).

carlcs:
https://symfony.com/blog/twig-adds-filter-map-and-reduce-features

I’ve just became aware of the newly added filter, map and reduce filters, and seeing how the syntax is borrowed from ES6 I was again reminded of this PR.

@fabpot can you please give this another thought. Seeing the amount of issues, votes and PRs leaves me with the impression that this feature has the potential to make many people happy.

sukei:
I have the same feeling.

PR #3006, opened 2019-05-15
update docs for for loop with conditional to reference filter filters

shieldo:
I don't know if there needs to be the same emphasis for this case as there's no special syntax for this any more? This PR assumes the same emphasis.

### DIFF
--- a/doc/tags/for.rst
+++ b/doc/tags/for.rst
@@ -87,6 +87,13 @@ Variable              Description
 Adding a condition
 ------------------
 
+.. tip::
+
+    As of Twig 2.10, use the :doc:`filter <../filters/filter>` filter instead,
+    or an ``if`` condition inside the ``for`` body (if your condition
+    depends on a variable updated inside the loop and you are not using the
+    ``loop`` variable).
+
 Unlike in PHP, it's not possible to ``break`` or ``continue`` in a loop. You
 can however filter the sequence during iteration which allows you to skip
 items. The following example skips all the users which are not active:


doc/tags/for.rst
```diff
@@ -85,13 +85,14 @@ Adding a condition
...
-items. The following example skips all the users which are not active:
+can however apply a :doc:`filter<filters/filter>` filter onto the sequence
+which allows you, effectively, to skip items. The following example skips all
+the users which are not active:
 
 .. code-block:: twig
 
     <ul>
-        {% for user in users if user.active %}
+        {% for user in users|filter(v => v.active) %}
```
fabpot: I would use `user` or `u` for the variable.

fabpot:
Can you rebase on 1.x?

doc/tags/for.rst
fabpot: You should also add a note to say that `filter` is available as of Twig 1.41/2.10 (and mention the old syntax with the `if` condition).

shieldo:
@fabpot Might it be better for me to submit another PR for the 1.x branch so I can reference 1.x releases there and 2.x releases here? I'm just looking at e.g. https://twig.symfony.com/doc/2.x/filters/filter.html#filter and it seems confusing/ possibly misleading that only a 1.x release is mentioned there in the 2.x docs (as the `filter` filter wasn't available in older 2.x releases).

fabpot:
You're right

shieldo:
@fabpot OK I changed my approach here and just added a tip for the deprecation/ replacement syntax, given it maybe doesn't strictly need a syntax example, and seemed more consistent with docs elsewhere (such as https://twig.symfony.com/doc/2.x/tags/spaceless.html). If that makes sense then I can submit a separate  PR for 1.x, and add a 3.x one that uses `filter` only.

doc/tags/for.rst
```diff
@@ -84,6 +84,13 @@ Variable              Description
 Adding a condition
 ------------------
 
+.. tip::
+
+    As of Twig 2.10, use the :doc:`filter <../filters/filter>` filter instead,
+    or an ``if`` condition inside the ``for`` body (if your condition
+    depends on a variable updated inside the loop and you are not using the
+    `loop` variable).
```
fabpot: double tick

fabpot:
@shieldo Works for me.

doc/tags/for.rst
shieldo: Updated

fabpot:
Thank you @shieldo.

PR #3011, opened 2019-05-16
Fix a regression

fabpot:


src/Parser.php
```diff
@@ -334,7 +334,15 @@ public function addImportedSymbol($type, $alias, $name = null, AbstractExpressio
 
     public function getImportedSymbol($type, $alias)
     {
-        return isset($this->importedSymbols[0][$type][$alias]) ? $this->importedSymbols[0][$type][$alias] : null;
+        if (null !== $this->peekBlockStack()) {
+            foreach ($this->importedSymbols as $functions) {
```
fabpot, line 338: This code is making things work "by chance". Adding it back until we find a better solution (which will only be for 2.x anyway).

fabpot: I'm adding more test to cover more edge cases, especially for the interaction between macros and blocks (we had none).

oleg-andreyev:
@fabpot just curious what kind of regression was fixed?

I see that these changes were "reverted" in v2.11.0 (#3012 )

I know that 1.x branch is not supported anymore, but is it possible to backport changes #3012 to 1.x branch? Asking because I'm trying to upgrade a quite huge project to Symfony 3.4 with a lot of dependencies including twig and encountered this issue...

PR #3012, opened 2019-05-16
Macros imported "globally" in a template are now available in macros without re-importing them

### LINKED ISSUE #1052: Acessing globally imported macro in another macro fails with PHP fatal error

### LINKED ISSUE #1790: Macro imports with {% from %} should not rely on the context
currently, the `{% from %}` generates an internal variable in the context. But this means that using the macro is subject to context scoping, while the Twig compiler is not aware of it (leading to weird failure messages).

Here are 2 cases where it will break at runtime:
- recursive macros (yeah, my example is totally dumb. See the form panel of the Symfony profiler for a real-world use case)
- calling a block with a different context than the original one (harder to achieve, but I can provide a template triggering it if you want)

``` twig
{% from _self import recursive_macro %}

{{ recursive_macro(10) }}

{% macro recursive_macro(n) %}
    {% if n > 0 %}
        {{- recursive_macro(n - 1) -}}{# this breaks here #}
    {% endif %}
    {{ n }}
{% endmacro %}
```

currently, the failure looks like this:

```
An exception has been thrown during the rendering of a template 
("Notice: Undefined index: __internal_3fdeda406e109fab86e32799388d7e009376041033cede9a9e555c402985ed3b") 
in @WebProfiler/Collector/form.html.twig at line 437
```

In case the notice does not throw an exception, you then get a fatal error just after that because of calling `getrecursive_macro` on `null` (see https://github.com/twigphp/Twig/pull/1309 trying to fix it, but in a way which may degrade performance because it wraps all macro calls in runtime checks).

We should either detect such mistakes at compile time, relying on the knowledge that macros are not inheriting context, or we should change the way macro calls are compiled to avoid relying on the context (which would be possible in 2.0 thanks to the fact we could break BC on the fact that macros can be called through property access as reported in https://github.com/twigphp/Twig/issues/1217)

### LINKED ISSUE #2414: Import _self macros with same name in child template generates undefined method call
[Basic reproduce code](https://twigfiddle.com/2x862p)

I'm porting my Twig templates to 2.x and updating them to reflect that macros defined in the same file must be imported now. However it seems that although macros are now only [template scope](https://github.com/twigphp/Twig/issues/2336) the names are global scope. If I change one of the files to use `selfMacro` instead of `_selfMacro` it works.

Is this expected behavior and if so what is the correct way of doing this?

### LINKED ISSUE #2442: Macros in embed tags
Macros imported in a template are not accessible in `embed` tags of the same template.

With following templates:

**foo.html.twig**
```twig
{% import 'bar.html.twig' as bar_macros %}

{% embed 'baz.html.twig' %}
    {% block baz %}
        {{ bar_macros.bar() }}
    {% endblock %}
{% endembed %}
```

**bar.html.twig**
```twig
{% macro bar() %}
    BAR
{% endmacro %}
```

**baz.html.twig**
```twig
{% block baz %}{% endblock %}
```

trying to render foo.html.twig will throw an exception: `Twig_Error_Runtime: Accessing Twig_Template attributes is forbidden`.

The only workaround I found is to import the macros inside the `embed` tag. Explicitly passing `bar_macros` using `with` keyword does not help. I have this issue with Twig 2.3.0, is it the expected behavior?

fabpot:
closes #1790, closes #1052, closes #2414, closes #2442

### DIFF
--- a/src/Node/ImportNode.php
+++ b/src/Node/ImportNode.php
@@ -22,19 +22,27 @@ use Twig\Node\Expression\NameExpression;
  */
 class ImportNode extends Node
 {
-    public function __construct(AbstractExpression $expr, AbstractExpression $var, int $lineno, string $tag = null)
+    public function __construct(AbstractExpression $expr, AbstractExpression $var, int $lineno, string $tag = null, bool $global = true)
     {
-        parent::__construct(['expr' => $expr, 'var' => $var], [], $lineno, $tag);
+        parent::__construct(['expr' => $expr, 'var' => $var], ['global' => $global], $lineno, $tag);
     }
 
     public function compile(Compiler $compiler)
     {
         $compiler
             ->addDebugInfo($this)
-            ->write('')
-            ->subcompile($this->getNode('var'))
-            ->raw(' = ')
+            ->write('$macros[')
+            ->repr($this->getNode('var')->getAttribute('name'))
+            ->raw('] = ')
         ;
+        
+        if ($this->getAttribute('global')) {
+            $compiler
+                ->raw('$this->macros[')
+                ->repr($this->getNode('var')->getAttribute('name'))
+                ->raw('] = ')
+            ;
+        }
 
         if ($this->getNode('expr') instanceof NameExpression && '_self' === $this->getNode('expr')->getAttribute('name')) {
             $compiler->raw('$this');
--- a/src/TokenParser/FromTokenParser.php
+++ b/src/TokenParser/FromTokenParser.php
@@ -47,7 +47,7 @@ final class FromTokenParser extends AbstractTokenParser
         $stream->expect(/* Token::BLOCK_END_TYPE */ 3);
 
         $var = new AssignNameExpression($this->parser->getVarName(), $token->getLine());
-        $node = new ImportNode($macro, $var, $token->getLine(), $this->getTag());
+        $node = new ImportNode($macro, $var, $token->getLine(), $this->getTag(), $this->parser->isMainScope());
 
         foreach ($targets as $name => $alias) {
             $this->parser->addImportedSymbol('function', $alias, 'macro_'.$name, $var);
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,8 @@
 * 2.11.0 (2019-XX-XX)
 
- * n/a
+ * fixed macros "import" when using the same name in the parent and child templates
+ * fixed recursive macros
+ * macros imported "globally" in a template are now available in macros without re-importing them
 
 * 2.10.0 (2019-05-14)
 
--- a/src/Node/BlockNode.php
+++ b/src/Node/BlockNode.php
@@ -32,6 +32,7 @@ class BlockNode extends Node
             ->addDebugInfo($this)
             ->write(sprintf("public function block_%s(\$context, array \$blocks = [])\n", $this->getAttribute('name')), "{\n")
             ->indent()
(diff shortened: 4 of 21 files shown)


fabpot:
I wanted to make these changes on 1.x and I was almost there, but as some behaviors have been deprecated in 1.x, that's very complex, so I gave up.

src/TokenParser/FromTokenParser.php
```diff
@@ -47,7 +47,7 @@ public function parse(Token $token)
         $stream->expect(/* Token::BLOCK_END_TYPE */ 3);
 
         $var = new AssignNameExpression($this->parser->getVarName(), $token->getLine());
-        $node = new ImportNode($macro, $var, $token->getLine(), $this->getTag());
+        $node = new ImportNode($macro, $var, $token->getLine(), $this->getTag(), $this->parser->isMainScope());
```
stof, line 50: how does `parser->isMainScope()`  behave when parsing an `embed` ?

src/Node/ImportNode.php
```diff
@@ -22,18 +22,28 @@
...
+            ;
+        } else {
+            $compiler->write('$');
+        }
         $compiler
-            ->addDebugInfo($this)
-            ->write('')
-            ->subcompile($this->getNode('var'))
-            ->raw(' = ')
+            ->raw('macros[')
```
stof: why not making the code look like that instead:

```php
$compiler
    ->addDebugInfo($this)
    ->write('$macros[')
    ->repr($this->getNode('var')->getAttribute('name'))
    ->raw('] = ')
;

if ($this->getAttribute('global')) {
    $compiler
        ->raw('$this->macros[')
        ->repr($this->getNode('var')->getAttribute('name'))
        ->raw('] = ')
    ;
}
```

That would be more readable that this magic with `$` or `$this->` before that code, as we would be adding the global part after the local one instead of repurposing the local one into the global one after having duplicated the local one.

The generated code would be the same.

stof (changes requested):

src/Node/ImportNode.php
fabpot: done

src/TokenParser/FromTokenParser.php
fabpot: When using an `embed` tag, the parser is reset to an initial stage, so, it behaves like it would for a new template, which is what we expect (see tests).

stof (approved):

Issue #3015, opened 2019-05-17
ext-ds Set not supported by the new `filter` filter


keksa:
We use `ext-ds` and since the `if` condition on `for` was deprecated in Twig 2.10.0, we moved to `filter` tag with arrow functions.

When trying to use the filter on `Ds\Set`, it crashes with:
> CallbackFilterIterator::__construct() must implement interface Iterator, instance of Ds\Set given

This is the (simplified) template code:
```twig
{% for error in errors|filter(error => constant('ERRORS', item)[error] is defined) %}
    {{ constant('ERRORS', item)[error]|trans }}
{% endfor %}
```
The `errors` variable is instance of `Ds\Set`.

Since this was working before, I think that either `Ds\Set` should be supported or even anything implementing `\Traversable`.

enumag:
It's not just a matter of Ds\Set but of any internal PHP structure that implements Traversable. See this note from PHP [docs](https://www.php.net/manual/en/class.traversable.php):

```
Internal (built-in) classes that implement this interface can be used in a foreach construct and do not need to implement IteratorAggregate or Iterator. 
```

fabpot:
You can also use:

```twig
{% for error in errors %}
    {% if constant('ERRORS', item)[error] is defined %}
        {{ constant('ERRORS', item)[error]|trans }}
    {% endif %}
{% endfor %}
```

keksa:
@fabpot I know there are workarounds, either the one you posted, or calling 'Set::toArray()', but that's not the point of my issue.
IMO there is no reason, why `filter` should not support `Traversable` interface, since that is, what is effectively supported by twig's `for` loop.

fabpot:
@keksa Do you want to work on a pull request to add support for each Traversable objects?

keksa:
@fabpot I can look into it, if you'll accept such PR.

fabpot:
fixed by #3017

PR #3016, opened 2019-05-17
filter filter - support Traversable

keksa:
solves #3015


fabpot:
Thank you for the pull request. As the bug also exists on 1.x, I've submitted a pull request for that version (#3017), which is a bit different as we need to support PHP 5.5 on 1.x. I've also written a test that only relies on internal PHP classes (`\SimpleXMLElement` is traversable but does not implement `\Iterator`).

PR #3017, opened 2019-05-18
Fix the "filter" filter when the argument is \Traversable but does not implement \Iterator

### LINKED ISSUE #3015: ext-ds Set not supported by the new `filter` filter

fabpot:
close #3015, closes #3016

### DIFF
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -1686,21 +1686,30 @@ function twig_array_batch($items, $size, $fill = null, $preserveKeys = true)
     return $result;
 }
 
-function twig_array_filter($array, $arrow)
-{
-    if (\is_array($array)) {
-        if (\PHP_VERSION_ID >= 50600) {
-            return array_filter($array, $arrow, \ARRAY_FILTER_USE_BOTH);
+if (\PHP_VERSION_ID < 50500) {
+    function twig_array_filter($array, $arrow)
+    {
+        if (\is_array($array)) {
+            if (\PHP_VERSION_ID >= 50600) {
+                return array_filter($array, $arrow, \ARRAY_FILTER_USE_BOTH);
+            }
+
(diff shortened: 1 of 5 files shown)


src/Extension/CoreExtension.php
```diff
@@ -1686,21 +1686,30 @@ function twig_array_batch($items, $size, $fill = null, $preserveKeys = true)
...
-    while ($array instanceof \IteratorAggregate) {
-        $array = $array->getIterator();
-    }
+        // some internal PHP classes are \Traversable but do not implement \Iterator
+        if ($array instanceof \Iterator) {
+            return new \CallbackFilterIterator($array, $arrow);
+        }
 
-    return new \CallbackFilterIterator($array, $arrow);
+        throw new RuntimeException('The "filter" filter argument only supports objects implementing \Iterator; upgrade to PHP 5.5+.');
```
enumag, line 1709: I think you could use [IteratorIterator](https://www.php.net/manual/en/class.iteratoriterator.php) to change Traversable into Iterator on PHP 5.4. Also you should be able to get rid of the `while ($array instanceof \IteratorAggregate) {` that way.

fabpot: You're right, much better, see #3021

PR #3019, opened 2019-05-18
Local macros auto-import

### LINKED ISSUE #861: Consider auto-importing macros in the same file

fabpot:
closes #1758, closes #861

This implements auto-import for macros defined in a template. To avoid BC breaks, I've decided to do the auto-import as the equivalent of `{% import _self as _self %}`, so local macros are loaded under the `_self` pseudo-variable:

```twig
{{ _self.hello("Fabien") }}

{% macro hello(name) -%}
    Hello {{ name }}
{% endmacro %}
```

~~Note that this is done via a new extension that you need to explicitly enable for now. We can remove the need for the extension, just not sure if we want it. WDYT?~~

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
 * 2.11.0 (2019-XX-XX)
 
+ * macros are now auto-imported in the template they are defined (under the ``_self`` variable)
  * added support for macros on "is defined" tests
  * fixed macros "import" when using the same name in the parent and child templates
  * fixed recursive macros
--- a/doc/tags/macro.rst
+++ b/doc/tags/macro.rst
@@ -74,7 +74,20 @@ via the ``from`` tag:
 
 .. tip::
 
-    To import macros from the current file, use the special ``_self`` variable:
+    When macro usages and definitions are in the same template, you don't need to
+    import the macros as they are automatically available under the special
+    ``_self`` variable:
+
+    .. code-block:: twig
+
+        <p>{{ _self.input('password', '', 'password') }}</p>
+
+        {% macro input(name, value, type = "text", size = 20) %}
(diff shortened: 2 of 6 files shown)


fabpot:
The second commit removes the extension so that auto-import is always on. ~~We need to decide which one we want.~~

ruudk:
Wouldn't it be nicer to execute the macro without the `_self.` prefix?

fabpot:
@ruudk That would be a huge BC break. Using `_self` makes us BC with current templates.

lyrixx:
I like this idea a lot!

javiereguiluz:
Very nice! This is something that most of us have needed at some point. Thanks for this feature.

Evoke-PHP:
@ruudk 

> Wouldn't it be nicer to execute the macro without the `_self.` prefix?

I tried the following which seems to work for single file usage:

```
{% macro input(name, value, type = "text", size = 20) %}
    <input type="{{ type }}" name="{{ name }}" value="{{ value|e }}" size="{{ size }}" />
{% endmacro %}
{% macro textarea(name, value, rows = 10, cols = 40) %}
    <textarea name="{{ name }}" rows="{{ rows }}" cols="{{ cols }}">{{ value|e }}</textarea>
{% endmacro %}
{% from _self import input as input_field, textarea %}

{# Usage in the same template #}
{{ input_field('foo', 'abc', 'def', 10) }}
{{ textarea('bar', 'baz') }}
```
By using the: ` {%from _self import <import list> %}` we seem to be able to import them into the current namespace.

[Importing Macros documentation](https://twig.symfony.com/doc/3.x/tags/macro.html)

PR #3023, opened 2019-05-19
Update advanced.rst

palmiak:
I added the example on how to use a newly added function. I think it may be usefull, as some might think that

`{{ function_name }}` which isn't as we need to add `()`.


fabpot:
People reading this advanced docs have probably used Twig before or read the main documentation where this is already explained. I prefer to not repeat the same information. Thank you for proposing.

PR #3024, opened 2019-05-20
doc(filter): add example of `filter` without `for` use

Kocal:
First, thanks for #2996! :)

This PR add an example of `filter` usage without using a `for` loop, to prevend some misinterpretation to people.

At first read, I thought `filter` was only usable with `for` loop, but nope. In fact it iterates on the result produced by `filter`.


fabpot:
Thank you @Kocal.

PR #3025, opened 2019-05-20
Add the possibility to register classes/interface as being safe

### LINKED ISSUE #2548: Mark types as "safe"
Today when a value object is inherently safe to output it must extend {{Twig\Markup}} and therefore introduce a hard dependency to Twig. It would be appreciated if one could reverse that relationship and mark a type as safe. Something like:

```
$env = new \Twig\Environment(…, ['object_safelist' => ['js' => [ObjectThatIsSafeForJs::class], 'html' => [Object1ThatIsSafeForOutputInHtml::class, Object2ThatIsSafe::class]]]);
```

`twig_escape_filter()` would use that list to look up if the passed object is considered safe for the current escaping and output the result of `__toString()` as is.

fabpot:
closes #2548

To avoid a too big performance impact on the escaper, we aggressively cache the safe classes, which means that changing the. configuration at runtime is not possible (and having different ones on 2 Twig instances is not possible either, this is really *globally* configured).

### DIFF
--- a/src/Extension/EscaperExtension.php
+++ b/src/Extension/EscaperExtension.php
@@ -14,11 +14,17 @@ use Twig\FileExtensionEscapingStrategy;
 use Twig\NodeVisitor\EscaperNodeVisitor;
 use Twig\TokenParser\AutoEscapeTokenParser;
 use Twig\TwigFilter;
+use Twig\Environment;
+use Twig\Error\RuntimeError;
+use Twig\Extension\CoreExtension;
+use Twig\Markup;
 
 final class EscaperExtension extends AbstractExtension
 {
     private $defaultStrategy;
     private $escapers = [];
+    private $safeClasses = [];
+    private $safeLookup = [];
 
     /**
      * @param string|false|callable $defaultStrategy An escaping strategy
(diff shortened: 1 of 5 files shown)


lstrojny:
Looks promising! If you want to avoid having quasi singleton semantics in `EscaperExtension` and `CoreExtension` the static cache variables could be indexed by `spl_object_hash()` in order to allow two Twig instances with different extension configurations in the same system

stof:
@lstrojny that would not work, as the cached compiled template would have to be invalidated when that change, so using the spl object hash would force to always invalidate the cache for each HTTP request.

src/Extension/EscaperExtension.php
```diff
@@ -100,6 +104,41 @@ public function getName()
...
+    {
+        self::$safeClassesFrozen = true;
+
+        return $this->safeClasses;
+    }
+
+    public function setSafeClasses(array $safeClasses = []): void
+    {
+        if (self::$safeClassesFrozen) {
+            throw new \LogicException('Unable to set the safe classes for escaping as twig_escape_filter() has already been initialized.');
```
stof: wouldn't this cause issues when running tests of a Symfony project where the kernel could be booted lots of time in the same process ? The first boot will instantiate Twig and freeze safe classes at some point. The next unit test will boot a kernel, which will then create another Twig instance trying to set the same safe classes, but this will break here. Forcing to use process isolation for the whole testsuite as soon as the project uses safe classes is a bad idea.
And the same will happen for phppm users btw, but they cannot use process isolation (as that would defeat the whole point).

stof (commented):
btw, changing safe classes does not invalidate the cache AFAICT.

stof:
@fabpot maybe `twig_escape_filter` should become a method of the extension, to be able to store some private state for the instance.

stof:
> btw, changing safe classes does not invalidate the cache AFAICT.

nvm, this is runtime escaping, not compile-time one.

fabpot:
Before going further here, I propose to move everything related to the escaping implementation to EscaperExtension. See #3026. One consequence of these changes will be that this feature will only be available as of Twig 2.x (as in Twig 1.x, we cannot be sure that both CoreExtension and EscaperExtension are available).

lstrojny:
> @lstrojny that would not work, as the cached compiled template would have to be invalidated when that change, so using the spl object hash would force to always invalidate the cache for each HTTP request.

Not sure I get that: what’s the relationship of this topic and compiled templates, aren’t all escaping checks runtime only?

src/Extension/EscaperExtension.php
fabpot: Fixed

fabpot:
I think it's ready for another round of reviews.

mpdude:
What is the suggested/easiest way to configure this in Symfony?

PR #3026, opened 2019-05-21
Deprecate CoreExtension::setEscaper() and CoreExtension::getEscaper() in favor of the same methods on EscaperExtension

fabpot:
This is some preliminary work to ease #3025. Everything related to escaping is now part of the `EscaperExtension` instead of `CoreExtension`. This PR is submitted on 2.x because both extensions are always available in 2.x (which is not the case on 1.x).

### DIFF
--- a/src/Extension/EscaperExtension.php
+++ b/src/Extension/EscaperExtension.php
@@ -18,6 +18,7 @@ use Twig\TwigFilter;
 final class EscaperExtension extends AbstractExtension
 {
     private $defaultStrategy;
+    private $escapers = [];
 
     /**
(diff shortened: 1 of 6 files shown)


src/Extension/EscaperExtension.php
```diff
@@ -97,4 +129,252 @@ function twig_raw_filter($string)
...
+
+        case 'url':
+            return rawurlencode($string);
+
+        default:
+            static $escapers;
+
+            if (null === $escapers) {
+                // merge the ones set on CoreExtension for BC (to be removed in 3.0)
+                $escapers = array_merge(
```
fabpot, line 349: This is where we deal with BC.

PR #3027, opened 2019-05-21
Add information for when `filter` filter was added in 2.x branch

YetiCGN:
The current "versionadded" info can be confusing to developers, if they assume 2.9 > 1.41 and don't know that 1.x and 2.x branches are still developed simultaneously.


fabpot:
Thank you @YetiCGN.

PR #3031, opened 2019-05-27
Add missing template name and line no to some deprecation messages

yceruto:
before:
twig-before

after:
twig-after

### DIFF
--- a/src/TokenParser/FilterTokenParser.php
+++ b/src/TokenParser/FilterTokenParser.php
@@ -30,21 +30,24 @@ final class FilterTokenParser extends AbstractTokenParser
 {
     public function parse(Token $token)
     {
-        @trigger_error('The "filter" tag is deprecated since Twig 2.9, use the "apply" tag instead.', E_USER_DEPRECATED);
+        $stream = $this->parser->getStream();
+        $lineno = $token->getLine();
+
+        @trigger_error(sprintf('The "filter" tag in "%s" at line %d is deprecated since Twig 2.9, use the "apply" tag instead.', $stream->getSourceContext()->getName(), $lineno), E_USER_DEPRECATED);
 
         $name = $this->parser->getVarName();
-        $ref = new BlockReferenceExpression(new ConstantExpression($name, $token->getLine()), null, $token->getLine(), $this->getTag());
+        $ref = new BlockReferenceExpression(new ConstantExpression($name, $lineno), null, $lineno, $this->getTag());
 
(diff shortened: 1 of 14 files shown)


stof:
Wasn't it done on purpose in the past to avoid breaking the grouping @nicolas-grekas ?

nicolas-grekas:
Yes, that's true. The goal is to limit the number of repetitive messages.
But I guess there is not a huge list of different file+line, so this should be fine?

stof:
Well, every time this deprecation is triggered, the file+line will be different (at least if you write your templates in a readable way rather than putting them in a single very long line).

yceruto:
I'm fine with adding the template name only if we want to keep the group per template

yceruto:
Note that the real problem here is that neither "Show context" nor "Show trace" show the affected template for those deprecation notices, so it's very hard to find which template contains the error.

fabpot:
Not related, but I've just released a new version of SwiftmailerBundle to get rid of these deprecation notices.

fabpot:
I think we need to add this information as Twig templates are complied to PHP, so without doing it, it's quite complex to find the usage (at least if you don't know how to use grep :)). So 👍 on my side.

nicolas-grekas (approved):

src/TokenParser/FilterTokenParser.php
```diff
@@ -30,21 +30,24 @@ final class FilterTokenParser extends AbstractTokenParser
 {
     public function parse(Token $token)
     {
-        @trigger_error('The "filter" tag is deprecated since Twig 2.9, use the "apply" tag instead.', E_USER_DEPRECATED);
+        $stream = $this->parser->getStream();
+        $lineno = $token->getLine();
+
+        @trigger_error(sprintf('The "filter" tag is deprecated since Twig 2.9, use the "apply" tag instead. ("%s" at line %d).', $stream->getSourceContext()->getName(), $lineno), E_USER_DEPRECATED);
```
fabpot: I would use `tag instead in "%s" at line %d.` to be more similar to what we're doing with exceptions.

fabpot: Or perhaps like you did elsewhere here: `The "filter" tag in "%s" at line %d is deprecated ...`

yceruto: done.

fabpot:
Thank you @yceruto.

PR #3035, opened 2019-05-28
Use v5.0 of the phpunit bridge

nicolas-grekas:
Not critical but still better IMHO :)


fabpot:
Thank you @nicolas-grekas.

PR #3038, opened 2019-05-30
Remove unneeded requirement

fabpot:
mbstring is not used anywhere anymore in Twig 2.x.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,6 @@
     ],
     "require": {
         "php": "^7.0",
-        "symfony/polyfill-mbstring": "^1.3",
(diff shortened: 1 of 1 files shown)


stof:
aren't we using it for the `length` filter and other things like that ?

SpacePossum:
it is still on 2.x https://github.com/twigphp/Twig/blob/2.x/src/Extension/CoreExtension.php#L684 ?

fabpot:
As you can see, I closed the PR.

SpacePossum:
sorry about that! :+1:

Issue #3042, opened 2019-06-03
Accessing macros from extended parent templates


nea:
Hi there

With the change https://github.com/twigphp/Twig/pull/3012 the `macro` import context structure has changed.

Now, I am facing the issue of having a parent with a macro, which I would like to access from an extending template. Before, I was using `{% import _self as me %}` to access it. This is not working anymore. Also with the `{% from %}` syntax it is not working.

What would be the new prefered way to access a parent macro from an extending child template? Is it even possible anymore?

Thanks

fabpot:
Can you share a template that does not work anymore? Via https://twigfiddle.com/?

nea:
Hey @fabpot 

Of course: https://twigfiddle.com/ts0vl6


Thanks

fabpot:
Your example does not seem to work in any Twig version, right?

nea:
Hey @fabpot 

It worked until this morning, until I updated Twig to 2.11. It worked before, last week with latest Symfony and Twig worked fine.


Cheers

fabpot:
@nea Have a look at the fiddle that you have provided, it does not. work with any versions. So, your code is probably different.

nea:
Hi @fabpot 

I am sorry but it worked for ages. In the given twigfiddle here I forgot the brackets at the end of the macro. But it worked before.

The exact template is:
```
{% extends '@PecMtuDvp/BackendEvent/PhysicalTestPlanning_changeset.html.twig' %}
{% trans_default_domain "PecMtuDvpBundle" %}
{% import _self as me %}

{% block changeset_field_block %}
	{% if key in ['progress'] %}
		{% set baseTranslationPath = 'pec_mtu_dvp.backendevent.physical_test_planning_progress' %}
		{% set baseTranslationDomain = 'PecMtuDvpBundle' %}
		{{ me.renderChange(key, entity.oldProgress|default(0), item[1], baseTranslationPath, baseTranslationDomain, []) }}
	{% elseif key in ['oldProgress'] %}
	{% else %}
		{{ parent() }}
	{% endif %}
{% endblock %}
```

and down the `extends` line there is a template with
```
{%- macro renderChange(key, oldValue, newValue, baseTranslationPath, baseTranslationDomain, defaultEvents) -%}
	{% import _self as me %}
	<li>
		<strong>
			{{ me.getKeyTranslation(key, baseTranslationPath, baseTranslationDomain, defaultEvents) }}
		</strong>:
[...]

PR #3046, opened 2019-06-03
Fixed: Endless loop if more than one parent without the macro exists

nea:
Hi @fabpot 

I made a quick fix for the parent macro method call fix you did for me in #3042. Your code works in case the direct parent has the macro. But if there is a longer hierarchy and the looked for macro is in a parent template +1 it runs into an endless loop because the `->getParent()` method is always called on `$template` but never the newer `$parent`.

Best


fabpot:
oops, good catch!

fabpot:
Thank you @nea.

stof:
a test should be added covering this.

PR #3047, opened 2019-06-03
Update the documentation for newly added features

ascirpus:
The `map`, `filter` and `reduce` functions were added in Twig 1.41 and 2.10, but the documentation for `map` and `reduce` did not contain the information in which 2.x version these functions were added. 
This minor PR fixes this.


fabpot:
Thank you @nexsja.

PR #3050, opened 2019-06-04
Add a doc note

### LINKED ISSUE #3045: BC break in 2.11 when importing macros inside an embed
Imports are no longer in scope inside nested blocks within embeds.
Example: https://twigfiddle.com/fhfwr0
The block "inner" causes it to fail. The example works fine before Twig 2.11.

fabpot:
closes #3045

### DIFF
--- a/doc/tags/macro.rst
+++ b/doc/tags/macro.rst
@@ -138,6 +138,12 @@ When calling ``import`` or ``from`` from a ``macro`` tag, the imported macros
 are only defined in the current macro and they override macros defined at the
 template level with the same names.
 
+.. note::
+
+    Before Twig 2.11, it was possible to use macros imported in a block in a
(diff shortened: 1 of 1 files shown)


doc/tags/macro.rst
```diff
@@ -138,6 +138,12 @@ When calling ``import`` or ``from`` from a ``macro`` tag, the imported macros
 are only defined in the current macro and they override macros defined at the
 template level with the same names.
 
+.. note::
+
+    Before Twig 2.11, it was possible to use macros imported in a block in a
+    "sub-block". When upgrading to 2.11, you need to either move the import in
+    the global scope or reimport the macros explicitely in the "sub-blocks".
```
xabbuh: explicitly

fabpot: fixed

PR #3052, opened 2019-06-04
Remove docs about overloading

fabpot:


fabpot:
see https://github.com/symfony/symfony/issues/31841

PR #3053, opened 2019-06-05
Deprecate overloading filters/functions/tags/operators

fabpot:
see discussion on https://github.com/symfony/symfony/issues/31841

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,6 @@
 * 2.12.0 (2019-XX-XX)
 
- * n/a
+ * deprecated overloading filters/functions/tags/operators
 
 * 2.11.1 (2019-06-04)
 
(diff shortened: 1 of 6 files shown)


nicolas-grekas:
I'm wondering if this isn't too opinionated. It's not needed technically yet it forces users to deal with conflicts while they didn't have to before. Yes you can go past the limitation, but that might be difficult sometimes.

fabpot:
It's not a question of being opinionated. What happens if a filter is overloaded twice? If the behavior is different, it should be another filter. If there is a bug, it should be fixed instead of overloaded. I don't see any valid use cases for overloading TBH.

stof:
this affects Symfony overloading the `dump` function to make it use the better VarDumper, because it cannot stop Twig Environment from registering the DebugExtension.

PR #3059, opened 2019-06-06
Don't clear PHP buffer when an error occurs with debug=true

### LINKED ISSUE #3058: Blank screen when die() in an AbstractExtension
I have a package that wraps [larapack/dd](https://github.com/larapack/dd) in a Twig Extension and I'm having a problem on projects that are using Twig 2.11 (issue on my [repo](https://github.com/studioespresso/craft3-dumper/issues/10))

Larapack has a 'dump & die' method (`dd`) and using that results in a blank screen, no matter where in the template the extension gets called.

Further debugging reveals that it is not related to larapack/dd of my extension specifically, any extension with a `die()` or an `exit()` in it results in a blank screen. 

For reference, this is built as a  plugin for [craftcms/cms](http://github.com/craftcms/cms). 

This worked on 2.81, but yesterday's Craft update bumped Twig from 2.8.1 to 2.11, which is why the issue shows up now. 

Is this an intentional change and if so how can I make this work again? Thanks!

lyrixx:
fixes #3058

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,6 @@
 * 1.42.2 (2019-XX-XX)
 
- * n/a
+ * Display partial output (PHP buffer) when an error occurs in debug mode
 
 * 1.42.1 (2019-06-04)
 
--- a/src/Extension/DebugExtension.php
+++ b/src/Extension/DebugExtension.php
@@ -54,7 +54,7 @@ function twig_var_dump(Environment $env, $context, array $vars = [])
         return;
     }
 
-    ob_start(function () { return ''; });
+    ob_start();
 
     if (!$vars) {
         $vars = [];
--- a/src/Node/MacroNode.php
+++ b/src/Node/MacroNode.php
@@ -104,7 +104,13 @@ class MacroNode extends Node
             ->outdent()
             ->write("]);\n\n")
             ->write("\$blocks = [];\n\n")
-            ->write("ob_start(function () { return ''; });\n")
+        ;
+        if ($compiler->getEnvironment()->isDebug()) {
+            $compiler->write("ob_start();\n");
+        } else {
+            $compiler->write("ob_start(function () { return ''; });\n");
+        }
+        $compiler
             ->write("try {\n")
             ->indent()
             ->subcompile($this->getNode('body'))
--- a/test/Twig/Tests/ErrorTest.php
+++ b/test/Twig/Tests/ErrorTest.php
@@ -207,6 +207,20 @@ EOHTML
             ],
         ];
     }
+
+    public function testTwigLeakOutputInDebugMode()
+    {
+        $output = exec(sprintf('%s %s debug', \PHP_BINARY, __DIR__.'/Fixtures/errors/leak-output.php'));
+
+        $this->assertSame('Hello OOPS', $output);
+    }
+
+    public function testDoesNotTwigLeakOutput()
+    {
+        $output = exec(sprintf('%s %s', \PHP_BINARY, __DIR__.'/Fixtures/errors/leak-output.php'));
+
+        $this->assertSame('', $output);
+    }
 }
 
 class Twig_Tests_ErrorTest_Foo
--- a/test/Twig/Tests/Fixtures/errors/leak-output.php
+++ b/test/Twig/Tests/Fixtures/errors/leak-output.php
@@ -1,0 +1,31 @@
+<?php
+
+require __DIR__.'/../../../../../vendor/autoload.php';
+
+use Twig\Environment;
+use Twig\Extension\AbstractExtension;
+use Twig\Loader\ArrayLoader;
+use Twig\TwigFilter;
+
+class BrokenExtension extends AbstractExtension
+{
+    public function getFilters()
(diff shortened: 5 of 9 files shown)


src/Extension/DebugExtension.php
```diff
@@ -54,7 +54,7 @@ function twig_var_dump(Environment $env, $context, array $vars = [])
         return;
     }
 
-    ob_start(function () { return ''; });
+    ob_start();
```
lyrixx, line 57: In that case, we are already in debug mode

src/Node/MacroNode.php
```diff
@@ -104,7 +104,13 @@ public function compile(Compiler $compiler)
             ->outdent()
             ->write("]);\n\n")
             ->write("\$blocks = [];\n\n")
-            ->write("ob_start(function () { return ''; });\n")
+        ;
+        if ($compiler->getEnvironment()->isDebug()) {
```
stof, line 108: Do we take the debug mode into account in the cache key ? If no, we would need to make that check at runtime rather than at compile time.

lyrixx: I don't recall ... :/

stof: checked it. The Environment puts it in its option hash, so that's fine.

fabpot:
@lyrixx Can you add a test to check that we don't leak output on non-debug mode and that we have the output on debug mode? Also, can you add a note in the CHANGELOG?

lyrixx:
@fabpot Done. That was not easy because PHPUnit buffer output + has an error handler. So the simplest thing to do was to use an external script

test/Twig/Tests/ErrorTest.php
```diff
@@ -216,3 +238,20 @@ public function bar()
         throw new \Exception('Runtime error...');
     }
 }
+
+
+class BrokenExtension extends \Twig\Extension\AbstractExtension
```
stof: this one looks unused

lyrixx: Oups, thanks.

stof (approved):

fabpot:
@lyrixx Tests do not pass for 5.4 :)

lyrixx:
Woho, PHP 5.4. Thanks for report. I fixed that in a very clean way :joy:

fabpot:
Yay for PHP 5.4... but some other PHP versions are now broken :joy:

CHANGELOG
```diff
@@ -1,6 +1,6 @@
 * 1.42.2 (2019-XX-XX)
 
- * n/a
+ * Display partial output (PHP buffer) when an eeor occurs in debug mode
```
fabpot: typo: error

composer.json
```diff
@@ -30,6 +30,7 @@
     "require-dev": {
         "symfony/phpunit-bridge": "^3.4.19|^4.1.8|^5.0",
         "symfony/debug": "^2.7",
+        "symfony/process": "^2.7|3.4|^4.0|^5.0",
```
fabpot: typo: `^3.4`

test/Twig/Tests/Fixtures/errors/leak-output.php
```diff
@@ -0,0 +1,29 @@
...
+    public function getFilters()
+    {
+        return [
+            new TwigFilter('broken', [$this, 'broken']),
+        ];
+    }
+
+    public function broken()
+    {
+        die('OUPS');
```
fabpot: I would prefer the English version `OOPS` 😂

test/Twig/Tests/Fixtures/errors/leak-output.php
```diff
@@ -0,0 +1,29 @@
...
+    }
+}
+
+$loader = new Twig_Loader_Array([
```
fabpot: You need to use the namespaced classes (same below).

fabpot (changes requested):

lyrixx:
I stop using symfony/process for using plain `exec`. Simpler to manage! And now it's all green :green_heart: 
I also addressed your comments.
Thanks.

fabpot:
Thank you @lyrixx.

test/Twig/Tests/ErrorTest.php
```diff
@@ -207,6 +207,20 @@ public function getErroredTemplates()
             ],
         ];
     }
+
+    public function testTwigLeakOutputInDebugMode()
+    {
+        $output = exec(sprintf('%s %s debug', \PHP_BINARY, __DIR__.'/Fixtures/errors/leak-output.php'));
```
stof, line 213: you need to escape the argument for the path (`__DIR__` might have spaces in it)

lyrixx: Good catch. https://github.com/twigphp/Twig/pull/3062

lyrixx:
@fabpot Could you do a release with this patch? This is quite annoying.  Thanks

fabpot:
@lyrixx New releases out just for you :)

lyrixx:
Thanks :)

PR #3061, opened 2019-06-10
Remove duplicate phpdoc

tvdijen:


fabpot:
Good catch, thanks @tvdijen.

PR #3070, opened 2019-06-24
[doc] "map" filter: fix args order (value, key)

guilliamxavier:


fabpot:
Good catch, thanks @guilliamxavier.

Issue #3072, opened 2019-06-24
Probably wrong behaviour in split-function


adriansuter:
Hi there

If the delimiter is the string `"0"`, then the `split` function might not work as expected.
```twig
{% set foo = "hello0world"|split('0') %}
```
We get an array of chars, but shouldn't `foo` be an array of two items only?

## Proposition

Maybe we should change the line
https://github.com/twigphp/Twig/blob/fb5a1f43e653ba4af08333e8ba3f6330b1017d34/src/Extension/CoreExtension.php#L781
into
```php
    if (strlen($delimiter)) {
```

What do you mean?

stof:
indeed, this looks wrong.

fabpot:
That's why `empty` should **never** be used in PHP.

@adriansuter Can you submit a PR to fix this bug (adding your example as a test case)? Thank you.

adriansuter:
Of course.

PR #3075, opened 2019-06-24
[1.x] Patch split filter

adriansuter:
This PR addresses #3072 vor Twig 1.x.


fabpot:
Thank you @adriansuter.

PR #3077, opened 2019-06-26
Replace dirname(__FILE__) with usage of __DIR__ in test fixtures

SenseException:


fabpot:
Thank you @SenseException.

PR #3086, opened 2019-07-10
Update include.rst

Chrysweel:
Fix in function include to set a filter

### DIFF
--- a/doc/tags/include.rst
+++ b/doc/tags/include.rst
@@ -36,7 +36,7 @@ of that file:
               {% include 'template.html' %}
           {% endfilter %}
           {# vs #}
-          {% include('template.html')|upper %}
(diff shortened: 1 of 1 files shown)


lyrixx (approved):
:+1: Good catch. But this goes on 1.x

Chrysweel:
And I think that if it is recommended to use the function include instead of the tag, I could change all the examples to use the function. But I have not done it because maybe you do not want to change.

Chrysweel:
Yes @lyrixx but this documentation is about 2.x the documentation about 1.x is in https://twig.symfony.com/doc/1.x/tags/include.html and has not this error in include.rst

fabpot:
Good catch, thanks @Chrysweel.

PR #3087, opened 2019-07-11
Extend opcache check if it is enabled

vosco88:
When trying to deploy October on shared hosting (www.websupport.sk) I received following error when accessing the frontend pages - ErrorException:> Zend OPcache API is restricted by "restrict_api" configuration directive.
Backend was working normally. 
I propose the following change to check not only if the opcache_invalidate exists but also if opcache is enabled.
The change was also needed in octobercms files, but to correspond I am also proposing to add it directly to twig.


fabpot:
Thank you @vosco88.

PR #3094, opened 2019-07-24
Add extra argument to filter column

DarkBee:
Added optional argument to `twig_array_column`, `$index_key`, to reflect proper arguments for underlaying `PHP` function `column`


fabpot:
Thank you @DarkBee.

PR #3102, opened 2019-08-07
Add the data_uri filter

### LINKED ISSUE #54: add a base64 encoding filter
this can be convenient for inlining images. the filter should accept a variable containing a binary string, a stream resource and maybe also just a filename.

fabpot:
Implemented as a popular request here: closes https://github.com/twigphp/Twig-extensions/issues/54


composer.json
```diff
@@ -24,13 +24,14 @@
         }
     ],
     "require": {
-        "php": "^7.0",
+        "php": "^7.1.3",
```
nicolas-grekas, line 27: We should be careful here: Symfony 3.4 would e.g. bump to `"twig/twig": "^1.43|^2.12"`, people on php 7.0 would be forced to downgrade to Twig 1 in order to continue receiving Symfony upgrades.

derrabus: I have one client who will be locked on Twig 2.11 now. I don't want to complain becuase nobody really should be on php 7.0 anymore these days. (Also, I urge them to upgrade php everytime we talk)

Anyway, if that change is kept, we should at least document the bump in the ChangeLog.

nicolas-grekas: see  #3157

PR #3103, opened 2019-08-07
Update PhpUnitBridge and fix deprecations

jderusse:
This PR:
- update dev dependency to `phpunit-bridge` to `^4.4@dev` in order to get polyfill methods
- fix deprecated tests
- update cs-fixer rule

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -4,6 +4,8 @@ return PhpCsFixer\Config::create()
     ->setRules([
         '@Symfony' => true,
         '@Symfony:risky' => true,
+        '@PHPUnit75Migration:risky' => true,
+        'php_unit_dedicate_assert' => ['target' => '5.6'],
         'array_syntax' => ['syntax' => 'short'],
         'php_unit_fqcn_annotation' => true,
         'no_unreachable_default_argument_value' => false,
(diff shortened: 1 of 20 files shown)


jderusse:
Hmmm test on 5.4 are red because:

> symfony/phpunit-bridge 5.0.x-dev requires php >=5.5.9 -> your PHP version (5.4.45) does not satisfy that requirement

We should either : 
- close this PR
- bump version of twig 1.x to 5.5
- stop testing twig 1.x with PHP 5.4
- downgrade version of PHPUnitBridge to 5.4

any advice @fabpot

fabpot:
Let's bump the minimum PHP version to 5.5.

jderusse:
Finally green... Travis were not really stable. I had to `push -f` 3 times to get it pass.
RFR

fabpot:
Thank you @jderusse.

fabpot:
@jderusse I've created a followup here: #3110. Do we have to change some configuration as the min version of PHP is different? Or perhaps in 3.x?

jderusse:
@fabpot `.php_cs.dist` is already configured to the highest value. 
you can remove `SYMFONY_PHPUNIT_REMOVE_RETURN_TYPEHINT` from `.travis.yml` but need to make `setUp` methods compatible with PHPUnit 8 signature `setUp: void`

PR #3105, opened 2019-08-07
Handle opcache_invalidate() being disabled on shared hosting platforms

RhvsW:
This fix makes twig gracefully fail on opcache API being disabled on shared hosting platforms for security reasons.
Tested with Grav CMS.

### DIFF
--- a/src/Cache/FilesystemCache.php
+++ b/src/Cache/FilesystemCache.php
@@ -68,7 +68,7 @@ class FilesystemCache implements CacheInterface
             if (self::FORCE_BYTECODE_INVALIDATION == ($this->options & self::FORCE_BYTECODE_INVALIDATION)) {
                 // Compile cached file into bytecode cache
                 if (\function_exists('opcache_invalidate') && filter_var(ini_get('opcache.enable'), FILTER_VALIDATE_BOOLEAN)) {
-                    opcache_invalidate($key, true);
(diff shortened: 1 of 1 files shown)


src/Cache/FilesystemCache.php
```diff
@@ -68,7 +68,7 @@ public function write($key, $content)
             if (self::FORCE_BYTECODE_INVALIDATION == ($this->options & self::FORCE_BYTECODE_INVALIDATION)) {
                 // Compile cached file into bytecode cache
                 if (\function_exists('opcache_invalidate') && filter_var(ini_get('opcache.enable'), FILTER_VALIDATE_BOOLEAN)) {
-                    opcache_invalidate($key, true);
+                    @opcache_invalidate($key, true);
```
fabpot, line 71: So, you're saying that the method is available, that's opcache is enabled, but some hosters disable the function somehow?

RhvsW: The method will work but it will just return that it is disabled for security reasons instead of it's expected use, using the opcache API you can enumerate all files on the host even those that are not owned. Apnscp does this by default.

`E_WARNING: Zend OPcache API is restricted by "restrict_api" configuration directive
`

fabpot:
Thank you @RhvsW.

PR #3107, opened 2019-08-07
Upgrade minimal version of php

jderusse:
Following this comment https://github.com/twigphp/Twig/pull/3103#issuecomment-519282356
Bump the minimal required version of PHP to 5.5

### DIFF
--- a/doc/intro.rst
+++ b/doc/intro.rst
@@ -26,7 +26,7 @@ Slim, Yii, Laravel, and Codeigniter — just to name a few.
 Prerequisites
 -------------
 
-Twig needs at least **PHP 5.4.0** to run.
+Twig needs at least **PHP 5.5.0** to run.
 
 Installation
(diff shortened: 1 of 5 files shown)


doc/intro.rst
```diff
@@ -26,7 +26,7 @@ Slim, Yii, Laravel, and Codeigniter — just to name a few.
 Prerequisites
 -------------
 
-Twig needs at least **PHP 5.4.0** to run.
+Twig needs at least **PHP 5.5.0** to run.
```
jderusse, line 29: hmm not sure about that. 5.4 can still be used with `version <= 1.42.2`

fabpot: That's what we did before when we bumped the min PHP version for 1.x, so LGTM

fabpot:
Thank you @jderusse.

PR #3108, opened 2019-08-07
Add an Intl extension

### LINKED ISSUE #167: localizedcurrency pattern feature request
This is a great library, but an issue I have is that it returns the current output
"RUB8,000.00"
While I appreciate this a lot, I'd like to add a space between the currency and value, after some digging, I was able to do that with modifiying the twig_localized_currency_filter function like the following

```
function twig_localized_currency_filter($number, $currency = null, $locale = null)
{
    $formatter = twig_get_number_formatter($locale, 'currency');
    $formatter->setPattern(str_replace('¤#', '¤ #', $formatter->getPattern()));
    return $formatter->formatCurrency($number, $currency);
}
```

I thought it could be convenient to override a pattern configuration write below where I set the locale in my symfony parameters.yml
and/or add a new a parameter to the above function

### LINKED ISSUE #211: localizednumber doesn't support feminine ordinal numbers
Ordinal numbers e.g. in Italian can be feminine or masculine, but localizednumber doesn't support this distinction.

Here is example PHP code to show the difference:

    $m = new NumberFormatter('it', NumberFormatter::ORDINAL);
    $m->setTextAttribute(NumberFormatter::DEFAULT_RULESET, "%digits-ordinal-masculine");
    var_dump($m->format(10));

    $f = new NumberFormatter('it', NumberFormatter::ORDINAL);
    $f->setTextAttribute(NumberFormatter::DEFAULT_RULESET, "%digits-ordinal-feminine");
    var_dump($f->format(10));

Output:

    string(4) "10º"
    string(4) "10ª"

### LINKED ISSUE #2567: Add a better support to INTL
Hi, 

For now, Twig does not provide any intl related functions.
And it's fairly common to add i18n / l10n features to a website.

So I think it's time to integrate https://github.com/twigphp/Twig-extensions/blob/master/lib/Twig/Extensions/Extension/Intl.php into the core.

More over, even in the twig ext, there are no support for translating language / locale / country.
It's quit common to have a symfony form with a CountryType. The country type set in the model something like "FR". And if you want to render a template with this information, it's better to display "France".

So it would be very nice to have also localized_country, localized_language etc.

This might require a new dependency on symfony/intl, but I don't really think it's an issue.

What do you think?

Obviously, I could work on this feature if it's accepted.

### LINKED ISSUE #13641: Should we make twig extension functions for accessing ICU Data ?
Tell me if it would be a good idea to add a extension/twig folder, containing a IcuExtension.php with those functions : 

```
public function getFunctions()
{
    return array(
        new \Twig_SimpleFunction('languageName', array($this, 'getLanguageName')),
        new \Twig_SimpleFunction('scriptName', array($this, 'getScriptName')),
        new \Twig_SimpleFunction('countryName', array($this, 'getCountryName')),
    );
}

public function getLanguageName($language, $region = null, $displayLocale = null)
{
    return Intl::getLanguageBundle()->getLanguageName($language, $region, $displayLocale);
}

public function getScriptName($script, $language = null, $displayLocale = null)
{
    return Intl::getLanguageBundle()->getScriptName($script, $language = null, $displayLocale = null);
}

public function getCountryName($country, $displayLocale = null)
{
    return Intl::getRegionBundle()->getCountryName($country, $displayLocale = null);
}
```

So we can use them in twig like that for ex

```
{{ languageName('en') }}
{{ languageName('en', null, 'de') }}
```

If it's a good idea, I'll make a PR.

K.

fabpot:
This new Intl extension is made of 2 parts: one is about formatting dates/currencies/numbers (which is what we already have partially in twig-extensions) and the second part is about ICU data as available in Symfony 4.3+.

I've decided to put this extension into Twig instead of the Symfony bridge because of the functions that do not rely on Symfony (and I wanted everything in the same place). Also, some Twig features now depends on some external libraries, so it makes more sense now than a few years ago where Twig had no external dependencies. Last, but not least, intl is an important topic, and I think it deserves to be a Twig first class citizen.

TODO:

 - [x] add documentation
 - [ ] what do we want to do with the existing `date` and `number_format` filters?

closes https://github.com/symfony/symfony/pull/20053
closes #2567
closes https://github.com/symfony/symfony/issues/13641
closes https://github.com/twigphp/Twig-extensions/pull/134
closes https://github.com/twigphp/Twig-extensions/pull/137
closes https://github.com/twigphp/Twig-extensions/pull/183
closes https://github.com/twigphp/Twig-extensions/pull/229
closes https://github.com/twigphp/Twig-extensions/issues/167
closes https://github.com/twigphp/Twig-extensions/issues/211

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,11 +12,13 @@ before_install:
 install:
     - travis_retry composer install
     - (cd extra/html-extra && travis_retry composer install)
+    - (cd extra/intl-extra && travis_retry composer install)
     - (cd extra/markdown-extra && travis_retry composer install)
 
 script:
     - ./vendor/bin/simple-phpunit
     - (cd extra/html-extra && ./vendor/bin/simple-phpunit)
+    - (cd extra/intl-extra && ./vendor/bin/simple-phpunit)
     - (cd extra/markdown-extra && ./vendor/bin/simple-phpunit)
 
 jobs:
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,7 +1,14 @@
 * 2.12.0 (2019-XX-XX)
 
- * added the Markdown extension in the "extra" repositories
- * added the HtmlExtension extension in the "extra" repositories
+ * added Intl extension in the "extra" repositories: "country_name",
+   "currency_name", "currency_symbol", "language_name", "locale_name",
+   "timezone_name", "country_timezones", "format_currency", "format_number",
+   "format_*_number", "format_datetime", "format_date", and "format_time"
+   filters
+ * added the Markdown extension in the "extra" repositories: "markdown_to_html",
+   and "html_to_markdown" filters
+ * added the HtmlExtension extension in the "extra" repositories: "date_uri"
+   filter, and "html_classes" function
  * fixed cache when opcache is installed but disabled
  * fixed using macros in arrow functions
 
--- a/doc/filters/country_name.rst
+++ b/doc/filters/country_name.rst
@@ -1,0 +1,41 @@
+``country_name``
+================
+
+.. versionadded:: 2.12
+    The ``country_name`` filter was added in Twig 2.12.
+
+The ``country_name`` filter returns the country name given its ISO-3166
+two-letter code:
+
+.. code-block:: twig
+
+    {# France #}
+    {{ 'FR'|country_name }}
+
+By default, the filter uses the current locale. You can pass it explicitly:
+
+.. code-block:: twig
+
+    {# États-Unis #}
+    {{ 'US'|country_name('fr') }}
+
+.. note::
+
+    The ``country_name`` filter is part of the ``IntlExtension`` which is not
+    installed by default. Install it first:
+
+    .. code-block:: bash
+
+        $ composer req twig/intl-extra
+
+     Then, add it on the Twig environment::
+
+        use Twig\Extra\Intl\IntlExtension;
+
+        $twig = new \Twig\Environment(...);
+        $twig->addExtension(new IntlExtension());
+
+Arguments
+---------
+
+* ``locale``: The locale
--- a/doc/filters/country_timezones.rst
+++ b/doc/filters/country_timezones.rst
@@ -1,0 +1,29 @@
+``country_timezones``
+=====================
+
+.. versionadded:: 2.12
+    The ``country_timezones`` filter was added in Twig 2.12.
+
+The ``country_timezones`` filter returns the names of the countries associated
+with a given timezone:
+
+.. code-block:: twig
+
+    {# Europe/Paris #}
+    {{ 'FR'|country_timezones|join(', ') }}
+
+.. note::
+
+    The ``country_timezones`` filter is part of the ``IntlExtension`` which is not
+    installed by default. Install it first:
+
+    .. code-block:: bash
+
+        $ composer req twig/intl-extra
+
+     Then, add it on the Twig environment::
+
+        use Twig\Extra\Intl\IntlExtension;
+
+        $twig = new \Twig\Environment(...);
+        $twig->addExtension(new IntlExtension());
--- a/doc/filters/currency_name.rst
+++ b/doc/filters/currency_name.rst
@@ -1,0 +1,44 @@
+``currency_name``
+================
+
+.. versionadded:: 2.12
+    The ``currency_name`` filter was added in Twig 2.12.
+
+The ``currency_name`` filter returns the currency name given its three-letter
+code:
+
+.. code-block:: twig
+
+    {# Euro #}
+    {{ 'EUR'|currency_name }}
+
+    {# Japanese Yen #}
+    {{ 'JPY'|currency_name }}
+
+By default, the filter uses the current locale. You can pass it explicitly:
+
+.. code-block:: twig
+
+    {# yen japonais #}
+    {{ 'JPY'|currency_name('fr_FR') }}
+
+.. note::
+
+    The ``currency_name`` filter is part of the ``IntlExtension`` which is not
+    installed by default. Install it first:
+
+    .. code-block:: bash
+
+        $ composer req twig/intl-extra
+
+     Then, add it on the Twig environment::
+
+        use Twig\Extra\Intl\IntlExtension;
+
+        $twig = new \Twig\Environment(...);
+        $twig->addExtension(new IntlExtension());
+
+Arguments
+---------
+
+* ``locale``: The locale
--- a/doc/filters/currency_symbol.rst
+++ b/doc/filters/currency_symbol.rst
@@ -1,0 +1,44 @@
+``currency_symbol``
+===================
+
+.. versionadded:: 2.12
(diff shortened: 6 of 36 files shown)


src/Extension/IntlExtension.php
```diff
@@ -0,0 +1,393 @@
...
+        return $this->formatNumber($number, $locale, $attrs, $style, $type);
+    }
+
+    /**
+     * @param \DateTimeInterface|string|null  $date     A date or null to use the current time
+     * @param \DateTimeZone|string|false|null $timezone The target timezone, null to use the default, false to leave unchanged
+     */
+    public function formatDateTime(Environment $env, $date, string $locale = null, ?string $dateFormat = 'medium', ?string $timeFormat = 'medium', string $pattern = '', $timezone = null, string $calendar = 'gregorian'): string
+    {
+        if (!class_exists('Locale')) {
```
Tobion: I'd recommend to use `\Locale::class` for discoverability

src/Extension/IntlExtension.php
```diff
@@ -0,0 +1,393 @@
...
+            throw new RuntimeException('The "format_date" filter requires the PHP intl extension or the symfony/intl package to be installed.');
```
Tobion: I think you want to recommend this package instead https://github.com/symfony/polyfill/tree/master/src/Intl/Icu
That the polyfill is in symfony/intl is meant to be an implementation detail that might change.

src/Extension/IntlExtension.php
fabpot: done

src/Extension/IntlExtension.php
fabpot: Right,  changed

Tobion:
It would be cleaner if we created a new package so we can add a real dependency on symfony/intl and ext-intl. I know it's a little more work but it worth it in the long run IMO.

fabpot:
I want to avoid creating new packages as it makes things way more complex for everyone. I'm also thinking about merging my other extensions in core (Markdown, CSS Inliner, ...).

composer.json
```diff
@@ -31,6 +31,7 @@
     "require-dev": {
         "symfony/phpunit-bridge": "^3.4.19|^4.1.8|^5.0",
         "symfony/debug": "^2.7",
```
Tobion: looks like a very outdated version is used

fabpot: fixed

Tobion:
> I want to avoid creating new packages as it makes things way more complex for everyone.

It should be as simple as `composer require twig/intl`. The problem with optional dependencies is that they only trigger at runtime. For example you create a template with `format_currency` and test it and voila it works. A different template uses `country_name` and does not work because you don't have symfony/intl. And if you don't have tested all templates, you won't find the problem at all.

CHANGELOG
```diff
@@ -1,5 +1,7 @@
 * 2.12.0 (2019-XX-XX)
 
+ * added the "country_name", "currency_name", "currency_symbol", "language_name", "locale_name",
+   "timezone_name", "country_timezones", "format_currency", "format_number", "format_number_*", "format_date" filters
```
jderusse: missing `format_datetime` and `format_time`

src/Extension/IntlExtension.php
```diff
@@ -0,0 +1,393 @@
...
+    public function formatNumberStyle(string $style, $number, string $locale = null, array $attrs = [], string $type = 'default'): string
+    {
+        return $this->formatNumber($number, $locale, $attrs, $style, $type);
+    }
+
+    /**
+     * @param \DateTimeInterface|string|null  $date     A date or null to use the current time
+     * @param \DateTimeZone|string|false|null $timezone The target timezone, null to use the default, false to leave unchanged
+     */
+    public function formatDateTime(Environment $env, $date, string $locale = null, ?string $dateFormat = 'medium', ?string $timeFormat = 'medium', string $pattern = '', $timezone = null, string $calendar = 'gregorian'): string
```
jderusse: shouldn't be `$locale` the last argument?  (like in `localizeddate` filter)
most of the time, we configure the application to use a default locale (which is always the same on all templates) and vary the length of date per template. Repeating the first argument could be annoying:

ie. A sample of template I often meet
```diff
 <table class="list">
-  <tr><td>{{ item.createdAt|format_date(null, 'short') }}</td></tr>
-  <tr><td>{{ item.createdAt|format_date(dateFormat='short') }}</td></tr>
+  <tr><td>{{ item.createdAt|format_date('short') }}</td></tr>
 </table>
 <div class="detail">
-  <div>{{ item.createdAt|format_date(null, 'long') }}</div>
-  <div>{{ item.createdAt|format_date(dateFormat='long') }}</div>
+  <div>{{ item.createdAt|format_date('long') }}</div>
 </div>
```

CHANGELOG
fabpot: fixed

src/Extension/IntlExtension.php
fabpot: That makes sense, changed

fabpot:
@Tobion I propose to create an `extra/` directory where I'm going to store all extensions that are not "core" or depend on some external libraries. I will then subtree-split the extra directory (and exclude it from the main Composer package). That way, I can keep everything in the same place and fix your concern.

Tobion:
Sounds good.

extra/intl-extension/composer.json
```diff
@@ -0,0 +1,39 @@
...
+    "license": "MIT",
+    "authors": [
+        {
+            "name": "Fabien Potencier",
+            "email": "fabien@symfony.com",
+            "homepage": "http://fabien.potencier.org",
+            "role": "Lead Developer"
+        }
+    ],
+    "require": {
```
Tobion: don't forget symfony/intl and ext-intl here ;)

fabpot: I've moved symfony/intl from require-dev to require, but I don't want to add ext-intl as one can use the polyfill.

.travis.yml
```diff
@@ -10,8 +10,11 @@ before_install:
 
 install:
     - travis_retry composer install
+    - (cd extra/intl-extension && travis_retry composer install)
 
-script: ./vendor/bin/simple-phpunit
+script:
+    - ./vendor/bin/simple-phpunit
+    - (cd extra/intl-extension && ./vendor/bin/simple-phpunit)
```
jderusse: what's about `./vendor/bin/simple-phpunit ./extra` in order to let simple_phpunit run future extensions in parallel?

fabpot: That does not work.

Trainmaster:
Imho there shouldn't be the dependency on `symfony/intl` (as already said in https://github.com/symfony/symfony/issues/13641#issuecomment-278628872).

stof:
The `extra` folder should be excluded from the download of the main package through `.gitattributes`, to avoid shipping the class twice in projects (which will confuse IDEs)

fabpot:
@stof That's the case :) I've done some followup PRs/commits.

PR #3110, opened 2019-08-08
Fix PHPUnit deprecations

fabpot:
#3103 followup for 2.x

### DIFF
--- a/tests/Cache/FilesystemTest.php
+++ b/tests/Cache/FilesystemTest.php
@@ -81,7 +81,7 @@ class FilesystemTest extends TestCase
 
     public function testWriteFailMkdir()
     {
-        $this->expectException('\RuntimeException');
+        $this->expectException(\RuntimeException::class);
         $this->expectExceptionMessage('Unable to create the cache directory');
 
         if (\defined('PHP_WINDOWS_VERSION_BUILD')) {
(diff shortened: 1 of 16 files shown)


tests/TemplateTest.php
```diff
@@ -386,7 +383,7 @@ public function getGetAttributeTests()
 
     public function testGetIsMethods()
     {
-        $this->expectException('\Twig\Error\RuntimeError');
```
fabpot, line 389: Was a bad merge

tests/Cache/FilesystemTest.php
```diff
@@ -81,7 +81,7 @@ public function testWrite()
 
     public function testWriteFailMkdir()
     {
-        $this->expectException('\RuntimeException');
+        $this->expectException(\RuntimeException::class);
```
jderusse, line 84: Good catch, I had to stringify classes because of PHP 5.4. 
But then we removed it and I forgot to revert it

jderusse (approved):

jderusse:
@fabpot next step is to add `: void` return type on `setup` and `teardown` methods and remove the travis env variable `SYMFONY_PHPUNIT_REMOVE_RETURN_TYPEHINT`

But I'm not sure it's something to do in this PR

fabpot:
@jderusse I've now merged this PR. Can you do the next steps in another PR?

jderusse:
I'm on it

PR #3111, opened 2019-08-08
Fix compatibility with PHPUnit 8

jderusse:
adds `: void` return typehint on methods `setup` and `teardown`


fabpot:
Thank you @jderusse.

PR #3113, opened 2019-08-08
move the HTML extension to an external Composer package

fabpot:


extra/html-extension/composer.json
```diff
@@ -0,0 +1,39 @@
...
+            "role": "Lead Developer"
+        }
+    ],
+    "require": {
+        "php": "^7.1.3",
+        "twig/twig": "^2.4|^3.0"
+    },
+    "require-dev": {
+        "symfony/mime": "^4.3",
+        "symfony/phpunit-bridge": "^4.4@dev"
```
Tobion, line 22: I think this should be moved to "require" or?

fabpot: The mime component is "only" used for 1 filter (not the other) and only if you don't pass the content-type. I hesitated on this one. Might be "better" to always require it.

Tobion: I would try to get rid of optional dependencies.

Tobion: And installing it when it's not needed does seem better than not installing it, when it is needed.

Tobion:
I don't know if you planned this already: With the extra optional extensions, it would be cool to enable extensions automatically when they are installed. So similar to symfony framework that enables stuff based on the existence. So we could have an extension in core that automatically installes other extra extensions when they exist.

fabpot:
That's already the case. Have a look at the `extra/twig-extra-bundle` directory. It automatically register the extensions when they are installed and even better, they hint you when trying to use a filter/function that does not exist but is part of an extra extension.

PR #3117, opened 2019-08-10
Fix PHPDoc of twig_get_attribute()

szepeviktor:
@fabpot Repel the attack of the French keyboard

props. @phpstan


fabpot:
Good catch, thanks @szepeviktor.

szepeviktor:
You're welcome.

Actually `phpstan analyze -l max src/` could tell you way more.

szepeviktor:
Although there is a need to write a phpstan extension for `getExtension()` as it always returns with a different type.

PR #3118, opened 2019-08-10
Add suggestions for extra extensions

fabpot:


extra/twig-extra-bundle/TwigExtraBundle.php
```diff
@@ -11,8 +11,16 @@
...
 
+use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\HttpKernel\Bundle\Bundle;
+use Twig\Extra\TwigExtraBundle\DependencyInjection\Compiler\MissingExtensionSuggestorPass;
 
 class TwigExtraBundle extends Bundle
 {
+    public function build(ContainerBuilder $container)
+    {
+        parent::build($container);
```
yceruto: the parent doesn't have any operation, can it be removed?

fabpot: I think we are doing the same in Symfony, just to be on the safe side if one day we add something in the parent?

yceruto:
in TwigExtraExtension -> `new FileLocator(\dirname(__DIR__, 2).'/Resources/config')`

yceruto:
or `new FileLocator(\dirname(__DIR__).'/Resources/config')` if you want to move it into `src/`

extra/twig-extra-bundle/src/TwigExtraBundle.php
```diff
@@ -11,10 +11,19 @@
...
+    public function build(ContainerBuilder $container)
+    {
+        parent::build($container);
+
+        $container->addCompilerPass(new MissingExtensionSuggestorPass());
+    }
+
     public function getPath()
     {
         return \dirname(__DIR__);
```
yceruto: if `src/Resources/config`, this should be removed, relying on the default (`src/`).

PR #3119, opened 2019-08-10
Revamping bundle structure to not mix source code with config files

yceruto:
Still compatible with any Symfony version.


fabpot:
Thank you @yceruto.

yceruto:
or `Resources/config` at the root if need some path convention.

fabpot:
`Resources/` should be under the `src/` directory. I will move it.

PR #3125, opened 2019-08-12
Fix intl-extra docs and links

pierredup:


fabpot:
Thank you @pierredup.

PR #3146, opened 2019-09-06
Minor fixes in the docs of some intl filters

javiereguiluz:
I'll do another PR to fix https://github.com/twigphp/intl-extra/README.md

### DIFF
--- a/doc/filters/country_timezones.rst
+++ b/doc/filters/country_timezones.rst
@@ -4,8 +4,8 @@
 .. versionadded:: 2.12
     The ``country_timezones`` filter was added in Twig 2.12.
 
-The ``country_timezones`` filter returns the names of the countries associated
-with a given timezone:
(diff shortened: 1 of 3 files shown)


fabpot:
Thank you @javiereguiluz.

doc/filters/country_timezones.rst
```diff
@@ -4,8 +4,8 @@
 .. versionadded:: 2.12
     The ``country_timezones`` filter was added in Twig 2.12.
 
-The ``country_timezones`` filter returns the names of the countries associated
-with a given timezone:
+The ``country_timezones`` filter returns the names of the timezones associated
+with a given country code:
```
stof, line 8: should this actually be a filter ? It looks like a usecase for a function to me. It does not transform the value, but loads something different.

fabpot: I agree, see #3149

PR #3147, opened 2019-09-06
Minor fix in a filter description

javiereguiluz:


fabpot:
Thank you @javiereguiluz.

PR #3152, opened 2019-09-17
Introduce limit to string replace

tommyvdv:
usage example:

```twig
{{ "A number of words"|replace({" ": "</br>"}, 1) }}
{{ "A number of words"|replace({" ": "</br>"}, 2) }}
```

```html
A<br/>number of words
A<br/>number<br/>of words
```

```estimation
A
number of words
A
number
of words
```

### DIFF
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -543,13 +543,22 @@ function twig_date_converter(Environment $env, $date = null, $timezone = null)
  *
  * @return string
  */
-function twig_replace_filter($str, $from)
+function twig_replace_filter($str, $from, $limit = null)
 {
     if (!twig_test_iterable($from)) {
         throw new RuntimeError(sprintf('The "replace" filter expects an array or "Traversable" as replace values, got "%s".', \is_object($from) ? \get_class($from) : \gettype($from)));
     }
 
-    return strtr($str, twig_to_array($from));
+    $pairs = twig_to_array($from);
+    if (is_int($limit)) {
+        array_walk($from, function ($substitution, $needle) use (&$str, $limit) {
+            $needle = preg_quote($needle);
+            $str = preg_replace("/{$needle}/", "{$substitution}", $str, $limit);
+        });
+        return $str;
+    }
+
+    return strtr($str, $pairs);
 }
 
 /**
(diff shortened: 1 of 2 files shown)


OskarStark:
Do you have a real world example where this could be helpful?

src/Extension/CoreExtension.php
```diff
@@ -543,13 +543,22 @@ function twig_date_converter(Environment $env, $date = null, $timezone = null)
...
  * @return string
  */
-function twig_replace_filter($str, $from)
+function twig_replace_filter($str, $from, $limit = null)
 {
     if (!twig_test_iterable($from)) {
         throw new RuntimeError(sprintf('The "replace" filter expects an array or "Traversable" as replace values, got "%s".', \is_object($from) ? \get_class($from) : \gettype($from)));
     }
 
-    return strtr($str, twig_to_array($from));
```
GromNaN, line 552: For performance the function `strtr` should be used when `$limit === null`.
`array_walk` is know to be relatively slow. https://medium.com/@steveclifton_12558/array-walk-performance-test-b8cfb11511a8

src/Extension/CoreExtension.php
```diff
@@ -543,13 +543,22 @@ function twig_date_converter(Environment $env, $date = null, $timezone = null)
...
+    $pairs = twig_to_array($from);
+    if (is_int($limit)) {
+        array_walk($from, function ($substitution, $needle) use (&$str, $limit) {
+            $needle = preg_quote($needle);
+            $str = preg_replace("/{$needle}/", "{$substitution}", $str, $limit);
```
stof, line 556: some chars are special in substitutions of `preg_replace` too. So this might not be enough to implement it.

And this behavior is not equivalent to the case without limit. `strtr` guarantees that it won't perform a replacement in a part already replaced, and it starts with the longer strings to lookup: https://www.php.net/strtr
Your implemented applying a limit does not have these behavior.

stof (changes requested):

stof:
what is the use case for such feature ? AFAIK, none of the string replacement functions from PHP have it either (`preg_replace` has it for the regex-based replacement).

fabpot:
Closing as there is no feedback.

PR #2799, opened 2019-01-10
Create .gitattributes

reedy:

PR #3153, opened 2019-09-19
Update .gitattributes to remove tests from "dist"

barryvdh:
This was proposed many times before, but considering https://github.com/symfony/symfony/pull/33579 is now merged, I propose this again. I refer to that issue as for the reasoning, as it applies here also.

I did not add the `docs` directory as there was much discussion about that before (discussion in  #942)

Related: #942 #1776 #1981 #2292 #2799 2905

### DIFF
--- a/extra/cssinliner-extra/.gitattributes
+++ b/extra/cssinliner-extra/.gitattributes
@@ -1,3 +1,2 @@
-/extra/** export-ignore
 /tests export-ignore
 /phpunit.xml.dist export-ignore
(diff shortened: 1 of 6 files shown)


stof (approved):

stof:
The same should be done for subtree in the `extra` folder.

barryvdh:
What do you mean? `extra` was already excluded, right?

stof:
Each folder inside `extra` is a subtree split. These subtree splits also need the `.gitattributes` file excluding their tests.

barryvdh:
@stof done

fabpot:
Thank you @barryvdh.

PR #3155, opened 2019-09-20
Optimize block('foo') ?? 'bar'

### LINKED ISSUE #3154: Redundant renderBlock() call when using block() function with ?? default
The following template line:

```twig
{{ block('foo') ?? 'bar' }}
```

will be compiled to this PHP code: (formatted for clarity)

```php
echo 
    (
        $this->hasBlock("foo", $context, $blocks) &&
        !(null === $this->renderBlock("foo", $context, $blocks))
    )
    ? $this->renderBlock("foo", $context, $blocks)
    : "bar";
```

Notice how `renderBlock()` is called twice there, when once should have been enough.

```php
echo
    (
        $this->hasBlock("foo", $context, $blocks)
            ? $this->renderBlock("foo", $context, $blocks)
            : null
    )
    ?? "bar";
```

If it’s too much trouble to avoid the redundant `renderBlock()` call, perhaps at least `renderBlock()` could internally memoize its results, in case it is called more than once.

fabpot:
closes #3154

### DIFF
--- a/src/Node/Expression/NullCoalesceExpression.php
+++ b/src/Node/Expression/NullCoalesceExpression.php
@@ -22,11 +22,15 @@ class NullCoalesceExpression extends ConditionalExpression
 {
     public function __construct(\Twig_NodeInterface $left, \Twig_NodeInterface $right, $lineno)
     {
-        $test = new AndBinary(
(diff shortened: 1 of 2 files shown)


stof:
See my comment on the issue. This could be made much simpler.

PR #3156, opened 2019-09-20
[3.0] remove the most blocking return types

nicolas-grekas:
These return types make it needlessly hard to move to Twig 3.
This should make Symfony 4.4 compatible with Twig 3, and should help make the CI of Symfony 5 green.
> `Declaration of Twig\Extra\CssInliner\CssInlinerExtension::getFilters() must be compatible with Twig\Extension\AbstractExtension::getFilters(): array in /home/travis/build/symfony/symfony/vendor/twig/cssinliner-extra/src/CssInlinerExtension.php`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
         }
     ],
     "require": {
-        "php": "^7.1.3",
+        "php": "^7.2.9",
         "symfony/polyfill-mbstring": "^1.3",
         "symfony/polyfill-ctype": "^1.8"
     },
--- a/src/Extension/AbstractExtension.php
+++ b/src/Extension/AbstractExtension.php
@@ -13,32 +13,32 @@ namespace Twig\Extension;
 
 abstract class AbstractExtension implements ExtensionInterface
 {
-    public function getTokenParsers(): array
+    public function getTokenParsers()
     {
         return [];
     }
 
-    public function getNodeVisitors(): array
+    public function getNodeVisitors()
     {
         return [];
     }
 
-    public function getFilters(): array
+    public function getFilters()
     {
         return [];
     }
 
-    public function getTests(): array
+    public function getTests()
     {
         return [];
     }
 
-    public function getFunctions(): array
+    public function getFunctions()
     {
         return [];
     }
 
-    public function getOperators(): array
+    public function getOperators()
     {
         return [];
     }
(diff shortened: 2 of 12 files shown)


xabbuh (approved):

composer.json
```diff
@@ -24,7 +24,7 @@
         }
     ],
     "require": {
-        "php": "^7.1.3",
+        "php": "^7.2.0",
```
fabpot: Would it make sense to use 7.2.9 like Symfony 5.0?

nicolas-grekas: sure, updated.

fabpot (approved):

fabpot:
Thank you @nicolas-grekas.

stof:
Why are they making it hard ? Couldn't Symfony 4.4 have return types on these methods ?

nicolas-grekas:
That'd be a BC break for Symfony.

stof:
Well, then we will need a 4.0 version of Twig to get return types on extension points just because we want to support Twig 3 and SF 4.4 and our Twig extensions were not final yet ?

nicolas-grekas:
Because the ecosystem deserves it. There are extensions into the wild and adding return types is asking everyone to work or die. We can and should do friendlier. What this PR provides.

smnandre:
Sorry to digg that far... should we add those return types back now.. if we want to add them in Twig 4.0 ?

fabpot:
I suppose so @smnandre 
Do you want to work on this?

stof:
We kept this in phpdoc, so the Symfony DebugClassLoader triggers deprecations since years if you don't add them in your child classes.
the only remaining work is to add them back in the 4.x branch

smnandre:
> I suppose so @smnandre Do you want to work on this?

With pleasure! 

>  the only remaining work is to add them back in the 4.x branch

Ok so i "revert" this commit on the 4.0 branch..  or is there more to do / some tricks along the way ? :)

fabpot:
> > I suppose so @smnandre Do you want to work on this?
> 
> With pleasure!
> 
> > the only remaining work is to add them back in the 4.x branch
> 
> Ok so i "revert" this commit on the 4.0 branch.. or is there more to do / some tricks along the way ? :)

It should be that simple… let’s see if that's the case in practice 🧐

PR #3157, opened 2019-09-20
Twig 2 is compatible with PHP 7.0

nicolas-grekas:
7.1 was needed for #3102, but this is not the case anymore since #3113

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,28 +6,34 @@ cache:
         - extra/*/vendor
         - $HOME/.composer/cache/files
 
+env:
+    global:
+        - SYMFONY_PHPUNIT_REMOVE_RETURN_TYPEHINT=1
+        - SYMFONY_PHPUNIT_DISABLE_RESULT_CACHE=1
+
 before_install:
     - phpenv config-rm xdebug.ini || return 0
 
 install:
     - travis_retry composer install
-    - (cd extra/cssinliner-extra && travis_retry composer install)
-    - (cd extra/html-extra && travis_retry composer install)
-    - (cd extra/inky-extra && travis_retry composer install)
-    - (cd extra/intl-extra && travis_retry composer install)
-    - (cd extra/markdown-extra && travis_retry composer install)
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -6,28 +6,34 @@ cache:
         - extra/*/vendor
         - $HOME/.composer/cache/files
 
+env:
+    global:
+        - SYMFONY_PHPUNIT_REMOVE_RETURN_TYPEHINT=1
+        - SYMFONY_PHPUNIT_DISABLE_RESULT_CACHE=1
```
nicolas-grekas, line 12: Except for this line which can be kept, this commit should be basic reverted when merging in 3.x.

nicolas-grekas (commented):
Now green :)

.travis.yml
```diff
@@ -6,28 +6,34 @@ cache:
...
-    - (cd extra/cssinliner-extra && travis_retry composer install)
-    - (cd extra/html-extra && travis_retry composer install)
-    - (cd extra/inky-extra && travis_retry composer install)
-    - (cd extra/intl-extra && travis_retry composer install)
-    - (cd extra/markdown-extra && travis_retry composer install)
+    - ([[ $TRAVIS_PHP_VERSION = 7.0 ]] || cd extra/cssinliner-extra && travis_retry composer install)
+    - ([[ $TRAVIS_PHP_VERSION = 7.0 ]] || cd extra/html-extra && travis_retry composer install)
+    - ([[ $TRAVIS_PHP_VERSION = 7.0 ]] || cd extra/inky-extra && travis_retry composer install)
+    - ([[ $TRAVIS_PHP_VERSION = 7.0 ]] || cd extra/intl-extra && travis_retry composer install)
+    - ([[ $TRAVIS_PHP_VERSION = 7.0 ]] || cd extra/markdown-extra && travis_retry composer install)
```
stof, line 23: it might be more readable to override the `install` and `script` sections in the 7.0 job instead.

fabpot:
Thank you @nicolas-grekas.

PR #3161, opened 2019-09-24
fix syntax

andreybolonin:


stof:
this change is wrong. `=>` is the syntax for arrow functions.

PR #3170, opened 2019-10-13
doc: number_format add math calculation hint

gimler:
Add hint about number_format and math calculation


fabpot:
Thank you @gimler.

PR #3171, opened 2019-10-13
Change array_key_exists to property_exists for object check

ghost:
PHP7.4 throws an error using array_key_exists 
changed code to use correct function: property_exists

### DIFF
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -1403,7 +1403,7 @@ function twig_get_attribute(Environment $env, Source $source, $object, $item, ar
 
     // object property
     if (/* Template::METHOD_CALL */ 'method' !== $type) {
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
There is nothing to do here: `array_key_exists` is always called with an array, see the `(array)` cast.

fabpot:
Let's close then.

floverdevel:
I have upgraded to PHP 7.4 today and now i have this error :
```
PHPUnit\Framework\Error\Deprecated: array_key_exists(): Using array_key_exists() on objects is deprecated. Use isset() or property_exists() instead
/path/to/project/vendor/twig/twig/src/Extension/CoreExtension.php:1396
```

nicolas-grekas:
Please upgrade your version of twig.

floverdevel:
Ok.
I thought I updated it, I will double-check tonight (or Monday morning) 😊

PR #3178, opened 2019-10-16
Add twig/string-extra extension

nicolas-grekas:
Last step before deprecating https://github.com/twigphp/Twig-extensions entirely.

See https://github.com/twigphp/Twig/blob/2.x/doc/filters/u.rst for doc.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,6 @@
 * 2.12.1 (2019-XX-XX)
 
- * n/a
+ * added the String extension in the "extra" repositories: "u" filter
 
 * 2.12.0 (2019-10-05)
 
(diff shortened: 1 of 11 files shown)


extra/string-extra/README.md
```diff
@@ -0,0 +1,10 @@
+String Extension
+================
+
+This package is a Twig extension that provides integration with the Symfony String component.
+
+It provides a single [`u`][1] filter that wraps a text in a `UnicodeString` object to give access to [methods of the class](https://symfony.com/doc/current/components/string.html).
+
+`{{ 'Symfony String + Twig = <3'|u.wordwrap(5) }}`
```
fabpot: Can we add some more relevant examples?

fabpot: This should be moved to the docs anyway.

fabpot:
Thank you @nicolas-grekas.

CHANGELOG
```diff
@@ -1,14 +1,14 @@
 * 2.12.1 (2019-XX-XX)
 
- * n/a
+ * added the String extension in the "extra" repositories: "u" filter
```
lyrixx, line 3: It should have been better to add it in 2.13.0

PR #3182, opened 2019-10-22
Add support for exponential numbers

fabpot:


CHANGELOG
```diff
@@ -1,6 +1,6 @@
 * 1.42.4 (2019-XX-XX)
```
stof, line 1: should be changed to 1.43 as this is a new feature.

tests/Fixtures/expressions/exponential_numbers.test
```diff
@@ -0,0 +1,8 @@
+--TEST--
+Twig manages exponentiel numbers correctly
```
apfelbox, line 2: Small typo:

```suggestion
Twig manages exponential numbers correctly
```

fabpot: Good catch

PR #3184, opened 2019-10-24
feat: interface renderer

Grafikart:
I think it would be nice to have a simple interface that could be used for testing (so we won't have to mock Environment).


fabpot:
You should not mock the environment, just the loader.

PR #3186, opened 2019-11-01
Fix twig compare

dpinheiro:
On twig 2.x the comparison between two variables was being done using the php operators (<, >, <=..), although on 3.x the function `twig_compare` was introduced and it had some issues

As an example
`'200' > 399` was returning `false` on twig 2.x, but on 3.x it was returning `true`.

Digging into the code I found out the comparison binaries (`GreaterBinary`, `LessBinary`, etc) were doing the wrong comparison, and the function `twig_compare` was also returning an wrong value if both variables had the same type

Now it's following the convention `-1` if first var is less, 0 if they are equal, or 1 if first var is greater

This is my first PR, so please comment if anything is wrong


stof:
For the CI setup, it will be fixed by #3187

stof:
this can now be rebased to make the CI run

dpinheiro:
Rebase done

xabbuh (approved):

nicolas-grekas (approved):

fabpot:
Thank you @dpinheiro.

PR #3187, opened 2019-11-04
Fix the Travis setup

stof:
This looks like a bad merge from 2.x to 3.x


fabpot:
Thank you @stof.

stof:
hmm, actually, this is not enough. It seems like only the first one works.

PR #3188, opened 2019-11-04
Fix the Travis config again

stof:
The parenthesis around the statements are isolating the directory changes in a sub-command. So the right fix was not to remove the closing parenthesis, but to re-add the opening one.

This replaces #3187, which was merged even though Travis was still broken.


stof:
OK, now this is working.

fabpot:
Thank you @stof.

PR #3200, opened 2019-11-18
Remove mentions of the obsolete twig-extensions repository

fabpot:


stof:
Actually, new `*-extra` repos are only available for Twig 2.x+ so this should not have been merged in the 1.x branch.

fabpot:
The repo is abandoned, so it's not maintained anymore even for 1.x.

PR #3204, opened 2019-11-27
Add globals option in constructor and set globals by array with new method

jeremyNouveau:
I added the globals option in the constructor for set $globals when we create a new environment instance or if we declare a new service and we want the globals option to be passed as arguments.

I also added the addGlobals method which calls the addGlobal method if we want to add an array of globals directly.

### DIFF
--- a/src/Environment.php
+++ b/src/Environment.php
@@ -118,7 +118,7 @@ class Environment
         $this->strictVariables = (bool) $options['strict_variables'];
         $this->setCache($options['cache']);
         $this->extensionSet = new ExtensionSet();
-        $this->globals = $options['globals'];
+        $this->globals = is_array($options['globals']) ? $options['globals'] : [] ;
 
         $this->addExtension(new CoreExtension());
         $this->addExtension(new EscaperExtension($options['autoescape']));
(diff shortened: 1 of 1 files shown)


src/Environment.php
```diff
@@ -115,6 +118,7 @@ public function __construct(LoaderInterface $loader, $options = [])
         $this->strictVariables = (bool) $options['strict_variables'];
         $this->setCache($options['cache']);
         $this->extensionSet = new ExtensionSet();
+        $this->globals = $options['globals'];
```
stof: this does not ensure anymore that `$this->globals` is an array, as it accepts any incoming values.

jeremyNouveau: And if i replace :
`$this->globals = $options['globals'];`
by : 
`$this->globals = (is_array($options['globals']))? $options['globals'] : [] ;`

fabpot:
Thank you for your contributions. For a bit of background, I've tried in the recent years to remove some methods from the the main Environment class. Globals is also something I don't want to promote too much. That's probably why there is currently no way to pass an array of globals in the constructor. So, that's why I won't accept this PR.

PR #3205, opened 2019-11-27
Updated spaceless deprecation to point to `apply` tag, as the `filter` tag is now deprecated too

snebes:
Title says it all.  No reason to chase our tail if we find a random `spaceless` tag yet.

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -19,6 +19,7 @@
   <php>
     <ini name="error_reporting" value="-1" />
     <ini name="xdebug.overload_var_dump" value="0" />
+    <server name="SYMFONY_PHPUNIT_VERSION" value="7.5" />
   </php>
 
   <listeners>
--- a/src/TokenParser/SpacelessTokenParser.php
+++ b/src/TokenParser/SpacelessTokenParser.php
@@ -24,7 +24,7 @@ use Twig\Token;
  *   {% endspaceless %}
  *   {# output will be <div><strong>foo</strong></div> #}
  *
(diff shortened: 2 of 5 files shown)


phpunit.xml.dist
```diff
@@ -19,6 +19,7 @@
   <php>
     <ini name="error_reporting" value="-1" />
     <ini name="xdebug.overload_var_dump" value="0" />
+    <server name="SYMFONY_PHPUNIT_VERSION" value="7.5" />
```
snebes, line 22: I don't know if this is the convention outside of Symfony, but without this, phpunit 8.x was being pulled down.

src/TokenParser/SpacelessTokenParser.php
```diff
@@ -24,7 +24,7 @@
  *   {% endspaceless %}
  *   {# output will be <div><strong>foo</strong></div> #}
  *
- * @deprecated since Twig 2.7, to be removed in 3.0 (use the spaceless filter instead)
+ * @deprecated since Twig 2.7, to be removed in 3.0 (use the "apply" tag instead)
```
fabpot: Here, it did not say to use the filter tag, but the spaceless filter. Maybe we should be more precise.
What about `use the "spaceless" filter with the "apply" tag instead`?

snebes: I'm open to any verbiage change.  I'll push that up.

stof (approved):

fabpot:
Thank you @snebes.

PR #3206, opened 2019-11-28
Fix tests

fabpot:


.travis.yml
```diff
@@ -19,6 +19,7 @@ before_install:
     - phpenv config-rm xdebug.ini || return 0
 
 install:
+    - if [[ $TRAVIS_PHP_VERSION == 5* ]]; then composer remove symfony/error-handler --dev; fi
```
stof: this looks weird to me, given that it is not in the list of dependencies.

PR #3207, opened 2019-11-28
added pre render events

ghost:
As a developer I would like to hook context into the Environment based on a listener which is fired before rendering the content. I would like this as an option, so we keep backward compatibility.

The event which I would like to see fired is:
`twig.pre_render:{template-name}`
The template name will be the name from the render paramaters. It should only fire the event. And not implement the listener. Those should be build in te application.

Recommended would be to extend the TwigEvents class, for example:
```
namespace App\Event;

use Twig\Event\TwigEvents as BaseEvent;

class TwigEvents Extends BaseEvent
{

    /**
     * Called directly before the render the foo index.
     * 
     * Listeners have the opportunity to add or change the context for the environment.
     * 
     * @Event("Twig\Event\PreRenderEvent")
     */
    const PRE_RENDER_FOO_INDEX = self::PRE_RENDER . 'foo/index.html.twig';
}
```
 And make the listeners in the application.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,9 +29,9 @@
         "symfony/polyfill-ctype": "^1.8"
     },
     "require-dev": {
-        "symfony/phpunit-bridge": "^4.4@dev|^5.0",
-        "symfony/debug": "^3.4|^4.2|^5.0",
-        "psr/container": "^1.0"
+        "symfony/phpunit-bridge": "^4.4|^5.0",
+        "psr/container": "^1.0",
+        "symfony/event-dispatcher": "^4.4|^5.0"
     },
     "autoload": {
         "psr-4" : {
--- a/drupal_test.sh
+++ b/drupal_test.sh
@@ -18,7 +18,6 @@ source output
 wget https://get.symfony.com/cli/installer -O - | bash
 export PATH="$HOME/.symfony/bin:$PATH"
 symfony server:start -d --no-tls
-ENDPOINT=`symfony var:export SYMFONY_DEFAULT_ROUTE_URL`
 
 curl -OLsS https://get.blackfire.io/blackfire-player.phar
 chmod +x blackfire-player.phar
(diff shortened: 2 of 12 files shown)


stof:
I would vote against that, and even more when the event name is dynamic (you won't be able to write a generic listener using these events).

I think this use case of altering the context passed to template should rather be solved at a level wrapping Twig than by dispatching such even in Twig itself (I never saw such use case in 9 years of using Twig).

stof:
thus, depending on the way you call the rendering logic, the event is not even guaranteed to be dispatched in your implementation, making it even less useful.

stof:
> As a developer I would like to hook context into the Environment based on a listener which is fired before rendering the content. I would like this as an option, so we keep backward compatibility.

Keeping backward compatibility with what ?

ghost:
@stof backward compatibility in the sentence, that no one should need to add the event dispatcher. So, they can just use it the way they use it now.

Indeed optional, once you added it, it will be used before every template which is rendered.
The point of the option is to make an event before rendering the template. In this way, I can add context with the listeners. It is like composing additional context for the template. It is not a functionality for generic context. 

The concept it self is new for Twig, but very well known in Blade. And really useful for applications where we do not use Doctrine and still want to assign models or an array of models to a view or other parameters. For example static data in a form. And not want the same code multiple times in the controller or.....

ghost:
@stof  I see my mistake btw, I changed the code, so the event will be fired before every template rendering. Now it will also work for the includes in the template. I think that will change a lot. Maybe you want to check it out again ;)

xabbuh:
This looks like a feature to me that creates quite some overhead when rendering templates. Since I also never saw a use case in practice for that and given the performance implications I would rather not add this feature.

ghost:
@xabbuh if you are using twig now as it is and leave the event dispatcher to null, no change is being made, so no additional load what so ever. 

A use case for this scenario can be found in: https://github.com/symfony/symfony-docs/pull/12731/files
The case for the the partials and one place to manage your code and simple unit tests are very handy. Another way to do the same is by calling a controller in your template. Those cases are equal in load. 
The positive side on that is, that is much easier to test.

fabpot:
As you say, this is already possible by using a controller. Or wrapping Twig. As nobody ever asked for such a "hook" before, I would rather not introduce a new way of doing things. Thanks for proposing but that's not something I will add to Twig core.

stof:
> Another way to do the same is by calling a controller in your template. Those cases are equal in load.

that's not equal in load. If you call a controller (or a custom Twig function registered by an extension), you have no impact on the rendering of other templates. If you inject an event dispatcher in Twig and dispatch an event for each rendered template, you are adding overhead for **all** rendered templates (even the one not having any listener registered)

PR #3208, opened 2019-11-28
Pre render events for version 2.x

ghost:
As a developer I would like to hook context into the Environment based on a listener which is fired before rendering the content. I would like this as an option, so we keep backward compatibility.

The event which I would like to see fired is:
`twig.pre_render:{template-name}`
The template name will be the name from the render paramaters. It should only fire the event. And not implement the listener. Those should be build in te application.

Recommended would be to extend the TwigEvents class, for example:
```
namespace App\Event;

use Twig\Event\TwigEvents as BaseEvent;

class TwigEvents Extends BaseEvent
{

    /**
     * Called directly before the render the foo index.
     * 
     * Listeners have the opportunity to add or change the context for the environment.
     * 
     * @Event("Twig\Event\PreRenderEvent")
     */
    const PRE_RENDER_FOO_INDEX = self::PRE_RENDER . 'foo/index.html.twig';
}
```
 And make the listeners in the application.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
     "require-dev": {
         "symfony/phpunit-bridge": "^4.4|^5.0",
         "psr/container": "^1.0",
-        "symfony/event-dispatcher": "^4.4|^5.0"
+        "symfony/event-dispatcher": "^3.4"
     },
     "autoload": {
         "psr-0" : {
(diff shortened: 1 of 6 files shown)


ghost:
for version 2, there is compatibility with php 7.0. There for we can not use the event dispatcher 4.4 or 5, both require an higher php. Adding an version of an old event dispatcher also breaks. Therefor closed.

PR #3212, opened 2019-12-05
Update api.rst

abdounikarim:
Hello,

I just add a link to the Extending Twig page.

Cheers 😉

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -336,7 +336,7 @@ This section describes the features added by the built-in extensions.
 
 .. tip::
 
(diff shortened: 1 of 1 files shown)


stof:
@fabpot can we use doc references in the Twig doc, as done in symfony-docs, to avoid hardcoding the URL of the page (which would require editing the URL in each branch to have the right version) ? Or do we need to handle internal links of the Twig doc as if they were external links ?

fabpot:
We need to use a doc link instead as done in 335dcad4b00c6ff1aa7da6409fc6d5ab5d3d9d46.

PR #3213, opened 2019-12-05
Updated HtmlExtension namespace in html_classes

mheesters:
Updated HtmlExtension namespace in html_classes to point to correct path


stof:
There is a use statement in that code-block, so the code is already valid.

mheesters:
> There is a use statement in that code-block, so the code is already valid.

Missed that one.. Sorry for the trouble! 🙏

PR #3214, opened 2019-12-08
[RFC] Avoid `strto(low|upp)er` due to special case in the `tr_TR` locale

mpdude:
This is a weird issue that I do not really know how to best address.

When the `tr_TR` Turkish locale is active, `strtolower('I') !== 'i'` and `strtoupper('i') !== 'I'`. In Turkish, there are two special "i" characters, namely the "lowercase i without dot" and the "uppercase I with dot". See http://www.i18nguy.com/unicode/turkish-i18n.html for a really neat writeup.

Because `strtolower()` is used when scanning classes for attributes and methods, template markup like `{{ object.id }}` does no longer match a `function getId()` when the `tr_TR` locale is active.

On the one hand, I think it's a bug if a working Twig template suddenly fails due to a locale switch. On the other hand, I don't know how people how use `tr_TR` on a daily basis handle this... changing Twig behaviour in this regard might be a BC issue for them.

For now, I'll just try to modify the Travis setup and a test method to show that the issue exists.

What you you think how we should approach this?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,13 +2,17 @@ language: php
 
 dist: trusty
 
-sudo: true
+sudo: false
 
 cache:
     directories:
         - vendor
         - $HOME/.composer/cache/files
 
+addons:
+    apt:
+        packages:
+            - locales
 
 env:
     global:
@@ -22,7 +26,6 @@ install:
     - travis_retry composer install
 
 before_script:
-    - sudo locale-gen tr_TR.UTF-8
     - locale -a
     - if [ "$TWIG_EXT" == "yes" ]; then sh -c "cd ext/twig && phpize && ./configure --enable-twig && make && make install"; fi
     - if [ "$TWIG_EXT" == "yes" ]; then echo "extension=twig.so" >> `php --ini | grep "Loaded Configuration" | sed -e "s|.*:\s*||"`; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,8 @@ install:
     - travis_retry composer install
 
 before_script:
-    - sudo locale-gen tr_TR.UTF-8 && locale -a | grep -q tr_TR.utf8
+    - sudo locale-gen tr_TR.UTF-8
+    - locale -a
     - if [ "$TWIG_EXT" == "yes" ]; then sh -c "cd ext/twig && phpize && ./configure --enable-twig && make && make install"; fi
     - if [ "$TWIG_EXT" == "yes" ]; then echo "extension=twig.so" >> `php --ini | grep "Loaded Configuration" | sed -e "s|.*:\s*||"`; fi
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,17 +2,10 @@ language: php
 
 dist: trusty
 
-sudo: false
-
 cache:
     directories:
         - vendor
         - $HOME/.composer/cache/files
-
-addons:
-    apt:
-        packages:
-            - locales
 
 env:
     global:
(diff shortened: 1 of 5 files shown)


mpdude:
What steps do I need to make the `tr_TR` locale available on Travis?

fabpot:
That's a very well known problem... I even did a talk about this topic a few years ago.
Short answer: this is not supported and we should not try to support it.

mpdude:
For the record, this is how PHP addresses the issue to make case-insensitive access to methods, fields, classes etc. possible: https://github.com/php/php-src/commit/582514d4c7b216dbdc7a8429962cf3e5776206f0. They have a special "to_lower" mapping that just maps ASCII `A-Z` to `a-z`, no other changes.

@fabpot Is this your last word? I understand we're possibly opening a can of worms here. 

We could also try to add a mapping or treat `i` as a special case and see how far get can get...

fabpot:
That's my last word :) Trust me, we don't want to go down this path.

mpdude:
😢😭

Do you have any suggestions how to work around this? I mean, I need the locale to be able to use Turkish translations, display localized month names and stuff. Need to find a way to sell my customer that we cannot support this particular locale 😱

mpdude:
Test fail in the Twig C extension

nicolas-grekas:
Don't use the system locale for doing localization. There are other ways, e.g. the Symfony Intl + Translation component don't rely on it.

stof:
@mpdude to be able to display localized month names, you can use the IntlDateFormatter, which is configured based on `Locale::setDefault`, not on `setlocale`.

mpdude:
Thank you @nicolas-grekas and @stof for the suggestions. Unfortunately, it's a legacy code base and I will need to see if we can change this at all.

This is so bitter... On of the methods that fail is `getI18NDate(...)` because it has the capital `I` in it.

What if we'd just create additional aliases in the table that Twig uses to map identifiers to methods? 

I guess I am at phase 3 of the [Kübler-Ross model](https://en.wikipedia.org/wiki/K%C3%BCbler-Ross_model).

OskarStark:
Maybe I am getting it a bit wrong but could the new string component help here somehow? 🧐

apfelbox:
Not sure, but wouldn't it be enough (although not very elegant) to just use the full method name (so instead `{{ object.id }}` just `{{ object.getId }} // {{ object.getId() }}`?
That could be at least a mitigation in legacy code bases until the code can be refactored to avoid using system locales.

mpdude:
@apfelbox In fact that's what we're doing right now while assessing the situation. But it's a constant source of mishaps because everybody is so used to writing just `{{ object.id }}` and it affects all methods with an `i` or `I` in them (and errors only surface when running the app in`tr_TR`).

Is it really necessary that Twig uses `strtolower()`/`strtoupper()` when it deals with method names? Case insensitivity rules in the Zend Engine only map `[a-z]` to `[A-Z]` and vice-versa. They stopped using the internal `strtolower/strtoupper` implementations a few years ago to avoid this issue.

stof:
doing the strtolower in userland rather than using the C implementation provided by PHP will likely have a performance impact.

nicolas-grekas:
Actually `strtr($s, 'ABCDEFGHIJKLMNOPQRSTUVWXYZ', 'abcdefghijklmnopqrstuvwxyz')` might be a little bit faster, provided it's written inline and not wrapped in a helper method.

mpdude:
You agree we should try that and see how far we get?

I could try to re-frame this as „consistency with case-sensitivity rules in the ZE“ and avoid mentioning tr_TR altogether.

fabpot:
"consistency with case-sensitivity rules in the ZE" sounds like a great fix :)

fabpot:
Closing in favor of #3237

mpdude:
Twig 2.12.4 including #3237 has been released this morning. We've updated and deployed that version and so far are running without issues on the `tr_TR` locale 🎉🥳🚀.

PR #3215, opened 2019-12-09
[Minor] Correct return type of twig_date_converter

mcfedr:
if a `\DateTimeImmutable` is passed to `twig_date_converter` it will return a `\DateTimeImmutable`, not a `\DateTime`.


fabpot:
Thank you @mcfedr.

PR #3216, opened 2019-12-11
Add missing suggestions in Twig Extra Bundle

fabpot:


stof:
I'm wondering whether we could manage to write some kind of test which would automatically discover all filters and functions available in extra packages, and assert that they are listed in the suggestions.

PR #3224, opened 2019-12-22
Update doc with new escaping behavior on ternary statements

l-vo:
According to https://symfony.com/blog/better-white-space-control-in-twig-templates#fine-grained-escaping-on-ternary-expressions


fabpot:
Thank you @l-vo.

PR #3226, opened 2019-12-26
Allow Symfony 5

### LINKED ISSUE #3225: html-extra does not work with Symfony 5
Mime dependency is fixed to `4.3` here: https://github.com/twigphp/html-extra/blob/2.x/composer.json#L18

Guite:
This fixes #3225

### DIFF
--- a/extra/html-extra/composer.json
+++ b/extra/html-extra/composer.json
@@ -15,7 +15,7 @@
     ],
     "require": {
         "php": "^7.1.3",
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
can you please check and confirm that the other extra extensions are ok?

Guite:
They are because I could install them. Otherwise composer would have complained I think.

fabpot:
Thank you @Guite.

PR #3233, opened 2020-01-02
lower php requirement

Guite:
This PR adjusts the minimum PHP version to `7.2.5+` which is the version that is also required by Symfony 5.


fabpot:
Thank you @Guite.

PR #3237, opened 2020-01-08
Replace strtolower() with strtr() when dealing with method names

mpdude:
This is to improve consistency with the implementation of case-insensitivity for method names etc. in the Zend Engine: In fact, the ZE does not apply the internal `strtolower` implementation, but just maps ASCII `A-Z` to `a-z`.

This was done in https://github.com/php/php-src/commit/582514d4c7b216dbdc7a8429962cf3e5776206f0 to fix issues like https://bugs.php.net/bug.php?id=18556.

Additionally, `strtok` might give a tiny (measurable?) performance gain over `strtolower`, plus it is not locale-dependant.

### DIFF
--- a/src/ExpressionParser.php
+++ b/src/ExpressionParser.php
@@ -657,7 +657,7 @@ class ExpressionParser
                 $stream->expect(Token::NAME_TYPE, null, 'Only variables can be assigned to');
             }
             $value = $token->getValue();
-            if (\in_array(strtolower($value), ['true', 'false', 'none', 'null'])) {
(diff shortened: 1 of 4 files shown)


src/Util/LowercaseMethodHelper.php
```diff
@@ -0,0 +1 @@
+<?php
```
fabpot: Is this file useful?

mpdude:
⚠️ Tests are green, but in fact some of the jobs did not really execute. See https://github.com/symfony/symfony/pull/35254.

src/Util/LowercaseMethodHelper.php
mpdude: Ouch, sorry – leftover from some other approach.

nicolas-grekas (approved):

fabpot (approved):

fabpot:
Thank you @mpdude.

PR #3240, opened 2020-01-11
Fix versionadded in doc

fancyweb:
I think those versionadded blocks don't show because they have a typo.


fabpot:
Thank you @fancyweb.

PR #3243, opened 2020-01-13
Syntax error in doc

JakeFr:
This code sample in advanced.rst is not highlighted in php because the second semicolon is missing


fabpot:
Thank you @JakeFr.

PR #3245, opened 2020-01-13
Add optional $throwException set to false

ybenhssaien:
It will be great I guess to have the possibility to disable throwing an exception while displaying Country

> Example : Data source with a mix of countries codes and countries names

image


fabpot:
That would not be consistent with the other similar functions.

PR #3249, opened 2020-01-21
fix(CoreExtension twig_escape_filter) custom filter for empty string

paulmhh:
custom filters should be able to decide by themselves if they want to process
empty strings or not ... for example:

->getExtension('core')->setEscaper('json', function($twigEnv, $string, $charset) {
    return json_encode($string);
});

template:
[{{value1}},{{value2}}]
=> empty strings need to be run through json_encode() too

this was working up to "v1.33.2", upgrading to "v1.42.4" for php7.4 compatibility introduced this issue

### DIFF
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -1031,10 +1031,6 @@ function twig_escape_filter(Environment $env, $string, $strategy = 'html', $char
         }
     }
 
-    if ('' === $string) {
-        return '';
-    }
-
     if (null === $charset) {
         $charset = $env->getCharset();
     }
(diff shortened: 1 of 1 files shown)


guilliamxavier:
FTR, the change of behavior in your case was probably introduced as part of #2751

Site note: I'm not sure if raw `json_encode` (without stripping the external double-quotes) is a suitable "escaper" (especially in regards of https://twig.symfony.com/doc/1.x/filters/json_encode.html and https://twig.symfony.com/doc/1.x/filters/escape.html)...
Also, instead of
`[{{ value1 }},{{ value2 }}]` with your custom `'json'` escaping, have you considered
`["{{ value1 }}","{{ value2 }}"]` with the `'js'` escaping strategy (if they're all strings)? or maybe
`[{{ value1|json_encode }},{{ value2|json_encode }}]` without escaping, or even
`{{ [value1, value2]|json_encode }}`?

paulmhh:
>  have you considered
> `[{{ value1|json_encode }},{{ value2|json_encode }}]` without escaping,

this seems like an exact equivalent to the custom filter we are using. (haven't read the twig source yet though). Thanks for the suggestion, we will try that to get back to the mainline twig lib. Only downside is, that it increases template size (of the source template and probably the compiled template too, the filter basically gets inlined)

fabpot:
I think Twig current code is correct and @guilliamxavier sounds the way to go.

PR #3255, opened 2020-02-04
Make round brackets optional for one argument tests like sameas

iquito:
Currently Twig forces round brackets when doing tests with `is` if there is an argument:

    {{ 1 is same as(1) ? 'OK' }}
    {{ 8 is divisible by(2) ? 'OK' }}

That syntax always bumped me, because without the brackets it would be easier to understand and write, especially for people who do not know Twig internals:

    {{ 1 is same as 1 ? 'OK' }}
    {{ 8 is divisible by 2 ? 'OK' }}

Twig only has the `same as` and `divisible by` tests which need arguments (in the base library), and for both you usually only need one argument, so it could make the template syntax more concise. [Jinja already supports this syntax](https://jinja.palletsprojects.com/en/2.11.x/templates/#tests) for one argument tests, so it would be one less difference between Jinja and Twig for people who use both.

The changes in this pull request should be fully backwards-compatible - instead of Twig throwing a SyntaxError it now accepts one argument when the round brackets are missing. I added an option for TwigTest to signal that there is one mandatory argument for the test, otherwise the ExpressionParser does not know it should look for that mandatory argument, and this feature is therefore available to anybody writing their own TwigTests, which could help extensions to Twig to simplify their syntax too.

I also adjusted the unit tests for `sameas` and `divisibleby` in hopefully a sensible way.


fabpot:
Thank you @iquito.

PR #3257, opened 2020-02-05
IntlExtension > Handle MissingResourceException

ruudk:
When you want to get the language name of a language you can type:
```twig
{{ inputLanguage | language_name }}
```

But if `inputLanguage` is not a valid language, a `MissingResourceException` is thrown, blowing up the whole page.

Since in Twig, there is no way to catch this exception, it becomes extremely hard to work with these helpers if you have no control over the input. There is also no way to first check if the language exists before invoking the filter.

### DIFF
--- a/extra/intl-extra/src/IntlExtension.php
+++ b/extra/intl-extra/src/IntlExtension.php
@@ -13,6 +13,7 @@ namespace Twig\Extra\Intl;
 
 use Symfony\Component\Intl\Countries;
 use Symfony\Component\Intl\Currencies;
+use Symfony\Component\Intl\Exception\MissingResourceException;
 use Symfony\Component\Intl\Languages;
 use Symfony\Component\Intl\Locales;
 use Symfony\Component\Intl\Timezones;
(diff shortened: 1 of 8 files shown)


extra/intl-extra/src/IntlExtension.php
```diff
@@ -156,37 +157,65 @@ public function getFunctions()
...
+        }
     }
 
     public function getCountryTimezones(string $country): array
     {
-        return Timezones::forCountryCode($country);
+        try {
+            return Timezones::forCountryCode($country);
+        } catch (MissingResourceException $exception) {
+            return [];
```
ruudk, line 217: Not sure if this is the right fallback?

fabpot: I think there is no better way.

fabpot:
Thank you @ruudk.

ruudk:
@fabpot Thanks for the fast merge. Would it be possible to tag a new release?

ruudk:
This should also be merged to the 3.x branch btw.

fabpot:
Releases done :)

ruudk:
@fabpot Thanks, quick question, I just deployed this to production and got some errors:

```twig
{{ inputCountry | country_name('en') }}
```

When `inputCountry` is `null` it will give this error:
```
Argument 1 passed to Twig\Extra\Intl\IntlExtension::getCountryName() must be of the type string, null given
```

I think we should allow null as input, and return null?

```diff
-public function getCountryName(string $country, string $locale = null): string
+public function getCountryName(?string $country, string $locale = null): ?string
```

And then early return null when the input is null. **_Would you accept that? Then I will create a PR._**

fabpot:
I would always return a string, so an empty string when the input is null.

ruudk:
@fabpot Created a PR > https://github.com/twigphp/Twig/pull/3268

PR #3259, opened 2020-02-06
allow null string in StringExtension

### LINKED ISSUE #3250: truncate filter fails with null string
I found this change when upgrading from old Symfony 4 project with TextExtension to Symfony 5 (with extensions in Twig core and String component).

In old extension, you can use `truncate` filter with a null string, you simply get an empty string.
This is not working anymore with `u.truncate`, you get the following exception:
> Argument 1 passed to Twig\Extra\String\StringExtension::createUnicodeString() must be of the type string, null given

I'm not sure if the new behavior is on purpose. In case, it's fine to me, but it should be documented somewhere

garak:
Fix #3250

### DIFF
--- a/extra/string-extra/src/StringExtension.php
+++ b/extra/string-extra/src/StringExtension.php
@@ -24,8 +24,8 @@ final class StringExtension extends AbstractExtension
         ];
     }
 
-    public function createUnicodeString(string $text): UnicodeString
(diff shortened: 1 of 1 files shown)


nicolas-grekas (approved):

fabpot:
Thank you @garak.

PR #3268, opened 2020-02-13
IntlExtension > Return empty string when input is null

ruudk:
```twig
{{ inputCountry | country_name('en') }}
```
When `inputCountry` is `null` it will give this error:
```
Argument 1 passed to Twig\Extra\Intl\IntlExtension::getCountryName() must be of the type string, null given
```

With this change, an empty string will be returned instead.

See discussion in https://github.com/twigphp/Twig/pull/3257#issuecomment-585740408


fabpot:
Thank you @ruudk.

PR #3269, opened 2020-02-14
Fix format of the CHANGELOG

nicolas-grekas:
Change to be applied to 2.x and 3.x when merging up of course.


fabpot:
Thank you @nicolas-grekas.

PR #3271, opened 2020-02-17
Environment Fix

DurandSacha:


doc/functions/dump.rst
```diff
@@ -15,9 +15,11 @@ introspecting its variables:
     ``\Twig\Extension\DebugExtension`` extension explicitly when creating your Twig
     environment::
 
-        $twig = new \Twig\Environment($loader, [
+        use Twig\Environment;
+        
+        $twig = new Environment($loader, [
             'debug' => true,
-            // ...
+            // define your environment.
```
stof, line 22: this is about specifying options passed to the environment here, not defining the environment (the whole statement is about that).

fabpot:
The whole doc is using inlined namespaces. So, changing just one of them would not make sense. Let's close as changing all of them would incur a lot of work when merging up from 1.x to 3.x

PR #3273, opened 2020-02-20
Update constant.rst

DurandSacha:


fabpot:
We are using ... elsewhere in the docs, I think they are enough.

PR #3275, opened 2020-02-21
Identity Equals - Best practices

hectorprats:
all the equals are with identity equals, except these.

It's better normalize all the comparatives operations.


fabpot:
Good catch, thanks @hectorprats.

PR #3277, opened 2020-02-21
Support object initialisation shortcut {a} == {a:a} from ES2015

GromNaN:
Object declaration in Twig is very close to the Javascript syntax and is targeting the same profil of developers. In order to provide a smooth developer experience, Twig should implement some new feature of ECMAScript.

Since ECMAScript 2015, a short syntax can be used to declare an object when the key is identical to a defined variable (see [MDN doc](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Operators/Object_initializer))

```js
// Shorthand property names (ES2015)
let a = 'foo', b = 42, c = {};
let o = {a, b, c}
```

This PR adds the same syntax to Twig.
This can be very useful for `include ... with ... only` :
```twig
{% set title = "My title" %}
{% set image = "image.jpg" %}
{{ include("template.twig", {title, image}, false) }}
# instead of 
{{ include("template.twig", {'title': title, 'image': image}, false) }}
```

### DIFF
--- a/doc/templates.rst
+++ b/doc/templates.rst
@@ -568,6 +568,10 @@ exist:
     {# keys as integer #}
     { 2: 'foo', 4: 'bar' }
 
+    {# keys get the variable name when omitted #}
+    {% set foo = 'foo' %}
+    { foo }
+
     {# keys as expressions (the expression must be enclosed into parentheses) -- as of Twig 1.5 #}
     {% set foo = 'foo' %}
     { (foo): 'foo', (1 + 1): 'bar', (foo ~ 'b'): 'baz' }
--- a/src/ExpressionParser.php
+++ b/src/ExpressionParser.php
@@ -376,7 +376,16 @@ class ExpressionParser
             //  * a string -- 'a'
             //  * a name, which is equivalent to a string -- a
             //  * an expression, which must be enclosed in parentheses -- (1 + 2)
-            if (($token = $stream->nextIf(Token::STRING_TYPE)) || ($token = $stream->nextIf(Token::NAME_TYPE)) || $token = $stream->nextIf(Token::NUMBER_TYPE)) {
+            if ($token = $stream->nextIf(Token::NAME_TYPE)) {
+                $key = new ConstantExpression($token->getValue(), $token->getLine());
+
+                // {a} is a shortcut for {a:a}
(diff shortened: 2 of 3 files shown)


lyrixx:
@GromNaN Usually we prefer to use the `{{ include }}` syntax over `{% include %}`: 
It's more semantic, `{{ }}` print something where `{% %}` should not (yes, initial implem was an error)

Anyway, I already dreamed about the ES2015 syntaxt in twig. I would be awesome to have the support everywhere

GromNaN:
Thanks for the recommendation @lyrixx. I've updated the description.

stof:
This seems to have been rejected in https://github.com/twigphp/Twig/issues/1019

GromNaN:
Thanks @stof for pointing. 
Times are changing : what looked weird [in 2013](https://github.com/twigphp/Twig/pull/1021#issuecomment-14244399) is something that [frontend developers expect today](https://github.com/twigphp/Twig/pull/3002#issuecomment-506720157).

By chance, I've written  the same code than https://github.com/twigphp/Twig/pull/1021.

apfelbox:
Still a big fan of this idea. It is just noise in the code, to duplicate every variable name (in the worst case).

And for a non-empty set of web developers, this syntax is very familiar and learned, as it is used in ES2015.

yoriiis:
Hi, is there any way this will eventually be merged or is this a dead option?

The current syntax has several duplicates when passing parameters, as in the example below.

```twig
{{ include 'template.html.twig' with {
    isCustomGrid: isCustomGrid,
    orderedPeriod: orderedPeriod,
    route: route,
    routeParams: routeParams
} only }}
```

The previous request dates from 2013 but the developers and languages have evolved a lot since.

The changes from @GromNaN make it easier to write the above example.

```twig
{{ include 'template.html.twig' with {
    isCustomGrid,
    orderedPeriod,
    route,
    routeParams
} only }}
```

fabpot:
Thank you @GromNaN.

oles:
Ohh - we're sooo close! Missing just a release - anything holding you back, @fabpot?

Thank you very much for applying it to all the versions also! I'm really looking forward to using it in CraftCMS :heart:

fabpot:
@oles I've just made fresh releases.

oles:
@fabpot AWESOME - thank you!!

PR #3281, opened 2020-02-27
Update Api.rst with refactorisation

DurandSacha:


doc/api.rst
```diff
@@ -164,7 +164,7 @@ You can add or prepend paths via the ``addPath()`` and ``prependPath()``
 methods::
 
     $loader->addPath($templateDir3);
-    $loader->prependPath($templateDir4);
+        ->prependPath($templateDir4);
```
xabbuh, line 167: This is now invalid PHP code if the previous line terminates the statement with a semicolon.

xabbuh: Furthermore, `addPath()` does not return the loader instance. So even when the semicolon on the line above is removed, this will not work as intended.

xabbuh (changes requested):

Arman-Hosseini (changes requested):
Unfortunately, your changes are not valid and will not be accepted.

fabpot:
IMHO, your changes make the code less clear in the examples.

PR #3286, opened 2020-03-04
added syntax highlighting in templates.rst

DurandSacha:


fabpot:
This change is wrong as this is needed to properly render HTML and Twig in the same template. It works well:

image

PR #3289, opened 2020-03-05
Fix logic for refreshing the cache

SeriousKen:
The logic for refreshing the cache is not working properly and always loads from the cache if AutoReload is false. `isTemplateFresh` never gets called. The logic should be:

<table>
<thead>
<tr><th>autoreload</th><th>cache fresh</th><th>action</th></tr>
</thead>
<tbody>
<tr><td>true</td><td>true</td><td>re-generate</td></tr>
<tr><td>true</td><td>false</td><td>re-generate</td></tr>
<tr><td>false</td><td>true</td><td>load from cache</td></tr>
<tr><td>false</td><td>false</td><td>re-generate</td></tr>
</tbody>
</table>

### DIFF
--- a/src/Environment.php
+++ b/src/Environment.php
@@ -338,7 +338,7 @@ class Environment
         if (!class_exists($cls, false)) {
             $key = $this->cache->generateKey($name, $mainCls);
 
-            if (!$this->isAutoReload() || $this->isTemplateFresh($name, $this->cache->getTimestamp($key))) {
+            if (!$this->isAutoReload() && $this->isTemplateFresh($name, $this->cache->getTimestamp($key))) {
                 $this->cache->load($key);
             }
 


xabbuh:
The tests do not seem to agree with the change being made.

stof:
always loading from the existing cache if `auto_reload` is `false` is precisely the intended behavior of `auto_reload`.

when `auto_reload` is `true`, Twig will automatically reload the template from source when the cache is not fresh anymore, to avoid having to clear the cache manually (that's a setting dedicated to dev environments, as calling `isTemplateFresh` has a performance impact due to doing lots of IO to check whether files have changed)

stof:
I'm closing this PR, as the code behaves as intended. Your table of expected behavior is the one being buggy.

PR #3290, opened 2020-03-06
[Add Link] link to parent() function in use.rst

DurandSacha:


fabpot:
I don't think this is need. Thank you.

PR #3308, opened 2020-04-06
Restrict callables to closures in filters

jderusse:
When using filter in sandboxed mode, this PR requires the use of closure.

### DIFF
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -194,9 +194,9 @@ class CoreExtension extends AbstractExtension
             new TwigFilter('sort', 'twig_sort_filter'),
             new TwigFilter('merge', 'twig_array_merge'),
             new TwigFilter('batch', 'twig_array_batch'),
-            new TwigFilter('filter', 'twig_array_filter'),
-            new TwigFilter('map', 'twig_array_map'),
-            new TwigFilter('reduce', 'twig_array_reduce'),
+            new TwigFilter('filter', 'twig_array_filter', ['needs_environment' => true]),
+            new TwigFilter('map', 'twig_array_map', ['needs_environment' => true]),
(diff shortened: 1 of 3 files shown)


src/Extension/CoreExtension.php
```diff
@@ -1693,8 +1693,12 @@ function twig_array_batch($items, $size, $fill = null, $preserveKeys = true)
     return $result;
 }
 
-function twig_array_filter($array, $arrow)
+function twig_array_filter(Environment $env, $array, $arrow)
 {
+    if ($env->getExtension('\Twig\Extension\SandboxExtension')->isSandboxed() && !$arrow instanceof Closure) {
```
stof: the leading `\` should be removed, to go through the fast path where the exact class name is provided (`get_class()` in PHP does not return a leading namespace separator)

nicolas-grekas: I would suggest swapping both operands.
Also, can't the extension be missing?

nicolas-grekas:
don't miss updating the changelog

src/Extension/CoreExtension.php
stof: hmm, indeed. We might not have the SandboxExtension registered.

jderusse: I added it for consistency with all other calls to `getExtension` and `hasExtension` in the project.

The performance impact is negligible, as the first instruction of getExtension is `$class = ltrim($class, '\\');`

stof: hmm, I forgot we un-conditionnally trim the char, without first checking for the key as is.

jderusse: good catch! fixed

nicolas-grekas (approved):

fabpot:
Thank you @jderusse.

PR #3311, opened 2020-04-14
Update api.rst

Yohann76:
formating improve

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -153,10 +153,7 @@ Compilation Cache
 ~~~~~~~~~~~~~~~~~
 
 All template loaders can cache the compiled templates on the filesystem for
-future reuse. It speeds up Twig a lot as templates are only compiled once; and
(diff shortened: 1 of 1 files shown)


doc/api.rst
```diff
@@ -135,7 +135,8 @@ Compilation Cache
 All template loaders can cache the compiled templates on the filesystem for
 future reuse. It speeds up Twig a lot as templates are only compiled once; and
 the performance boost is even larger if you use a PHP accelerator such as
-OPCache. See the ``cache`` and ``auto_reload`` options of ``\Twig\Environment``
+`OPCache <https://www.php.net/manual/fr/book.opcache.php/>`_. See the ``cache`` and
```
fabpot: I'm wondering if we should instead remove the part about the PHP accelerator as modern PHP versions have it enabled by default.

Yohann76: Yes, that may be a good idea

fabpot:
Thank you @Yohann76.

PR #3322, opened 2020-05-07
An error in format_number.rst. "format_number" invalid usage

JokubasR:
format_number does not accept second argument as a string, so I believe "format_currency" is the correct extension in this case


fabpot:
Fix in 60e2bda5

PR #3324, opened 2020-05-10
[3.0] Added support for floating-point numbers for modulo

### LINKED ISSUE #1353: % operator returns the integer result

connorhu:
Ported from #1440 and fixes #1353


src/Extension/CoreExtension.php
```diff
@@ -1595,4 +1595,25 @@ function twig_array_reduce($array, $arrow, $initial = null)
...
+ * @return int|float
+ */
+function twig_mod($divident, $divisor)
+{
+    if (is_int($divident) && is_int($divisor)) {
+        return $divident % $divisor;
+    }
+
+    $dividentDecimal = ($pos = strpos($divident, '.')) ? strlen($divident) - $pos - 1 : 0;
+    $divisorDecimal = ($pos = strpos($divisor, '.')) ? strlen($divisor) - $pos - 1 : 0;
```
stof, line 1614: you should not do it using string manipulation (especially given that float casting is locale-dependant before PHP 8 so this is not even working if the locale uses a different decimal separator).

fabpot:
Closing as there is no more activity.
@connorhu Feel free to reopen when you have time.

connorhu:
@fabpot I have time, but there is an unanswered question about other types.

PR #3325, opened 2020-05-11
Fix transform call to Michelf\MarkdownExtra

gerryd:
As per the documentation here:

  https://michelf.ca/projects/php-markdown/configuration/#configure

Calling defaultTransform() will use the default options, thus ignoring
the "hard_wrap" option. Calling transform() instead, which takes the
options into account.


fabpot:
Thank you @gerryd.

PR #3326, opened 2020-05-11
Throw exception in case non-Traversable data is passed to "filter"

Lustmored:
In PHP 7.4 passing for example null to |filter generates TypeError. I was unsure how to proceed from there as filter could silently fall back to empty result, but instead I went with throwing RuntimeExceptions to be consistent with |batch and |column.

I got confused in structure of tests here, so if that should have some kind of test please push me in the right direction and I'll gladly add it.


fabpot:
Thank you @Lustmored.

thijzert:
Ah, I think we ran into the same issue at around the same time. If you need a source for tests, I think the ones from my PR (#3341) should work with either.

PR #3334, opened 2020-05-19
Remove deprecated option

dbrumann:
See: https://docs.travis-ci.com/user/reference/overview/#deprecated-virtualization-environments

> If you’re trying to use sudo: false or dist: precise keys in your travis.yml, we recommend you remove them and switch to our current Xenial Linux infrastructure.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,6 @@ language: php
 
 dist: xenial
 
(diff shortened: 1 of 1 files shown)


fabpot:
Thank you @dbrumann.

PR #3337, opened 2020-05-21
Use dev-master when possible

nicolas-grekas:
Should fix the nightly job.


nicolas-grekas:
Applied when merging #3338 up.

PR #3340, opened 2020-05-22
Improve batch code example

y2keeper:
The current example uses an array of 7 items, resulting in 3 arrays of 3 items each (with fill). 

This can be misinterpreted at a glance as creating n arrays as opposed to arrays with n items.

We should therefore instead use an example array of 4 items resulting in 2 arrays of 3 items each (with fill). 

I think this would make the functionality less ambiguous.


fabpot:
Thank you @y2keeper.

PR #3341, opened 2020-05-28
In |filter(), |map(), and |reduce(), throw a RuntimeError instead of a TypeError

thijzert:
If a template accidentally performs `|filter(...)` on a variable that isn't traversable (e.g. `null`), the calling code will encouter the raw `TypeError` that results from Twig trying to construct an `IteratorIterator` on that variable. This means that context information (e.g., in which exact template the error occurred, and on which line) is discarded, which makes it very difficult to debug your templates.

This commit adds code to throw a `RuntimeError` if Twig encounters a situation where there's no valid way to filter an array.

### DIFF
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -1724,6 +1724,10 @@ function twig_array_map($array, $arrow)
 function twig_array_reduce($array, $arrow, $initial = null)
 {
     if (!\is_array($array)) {
+        if (!$array instanceof \Traversable) {
+            throw new RuntimeError(sprintf('The "reduce" filter only works with arrays or "Traversable", got "%s" as first argument.', \gettype($array)));
+        }
+
         $array = iterator_to_array($array);
     }
 
--- a/tests/ErrorTest.php
+++ b/tests/ErrorTest.php
@@ -161,6 +161,85 @@ EOHTML
         }
     }
 
+    public function testTwigArrayFilterThrowsRuntimeExceptions()
+    {
+        $loader = new ArrayLoader([
+            'filter-null.html' => <<<EOHTML
+{# Argument 1 passed to IteratorIterator::__construct() must implement interface Traversable, null given: #}
+{% for n in variable|filter(x => x > 3) %}
+    This list contains {{n}}.
+{% endfor %}
+EOHTML
+        ]);
+
+        $twig = new Environment($loader, ['debug' => true, 'cache' => false]);
+
+        $template = $twig->load('filter-null.html');
+        $out = $template->render(['variable' => [1, 2, 3, 4]]);
+        $this->assertEquals('This list contains 4.', trim($out));
+
+        try {
+            $template->render(['variable' => null]);
+
+            $this->fail();
+        } catch (RuntimeError $e) {
+            $this->assertEquals(2, $e->getTemplateLine());
+            $this->assertEquals('filter-null.html', $e->getSourceContext()->getName());
+        }
+    }
+
+    public function testTwigArrayMapThrowsRuntimeExceptions()
+    {
+        $loader = new ArrayLoader([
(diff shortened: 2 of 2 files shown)


guilliamxavier:
There's already #3326 that has been merged in `1.x` but not into `2.x`/`3.x`.

Also, what about `twig_array_map` and `twig_array_reduce`?

thijzert:
I guess I should use `|reduce()` more - it has the same problem. I've added a fix for it.

`twig_array_map` already fails with a RuntimeError in a way that's consistent with my subjective expectations, so I've added a unit test to fix the behaviour in place.

guilliamxavier:
Actually `twig_array_map(null, fn ($x) => $x)` doesn't throw an exception per se, just emits a Warning "Invalid argument supplied for foreach()", that I suppose is then converted to a RuntimeError by some error handler... But indeed not a TypeError either way ;)

Anyway `$array instanceof \Traversable || $array instanceof \Iterator` is redundant since `interface Iterator extends Traversable` (see e.g. `twig_array_column`)

thijzert:
> Actually `twig_array_map(null, fn ($x) => $x)` doesn't throw an exception per se, just emits a Warning "Invalid argument supplied for foreach()", that I suppose is then converted to a RuntimeError by some error handler.

Ah, that makes sense.

src/Extension/CoreExtension.php
```diff
@@ -1573,6 +1573,10 @@ function twig_array_filter($array, $arrow)
         return array_filter($array, $arrow, \ARRAY_FILTER_USE_BOTH);
     }
 
+    if (!$array instanceof \Traversable) {
+        throw new RuntimeError(sprintf('An array filter only works with arrays or "Traversable", got "%s" as first argument.', \gettype($array)));
```
guilliamxavier:
```suggestion
        throw new RuntimeError(sprintf('The filter filter only works with arrays or "Traversable", got "%s" as first argument.', \gettype($array)));
```
(or maybe
```suggestion
        throw new RuntimeError(sprintf('The "filter" filter only works with arrays or "Traversable", got "%s" as first argument.', \gettype($array)));
```
to mitigate the repetition)

src/Extension/CoreExtension.php
```diff
@@ -1590,6 +1594,10 @@ function twig_array_map($array, $arrow)
 function twig_array_reduce($array, $arrow, $initial = null)
 {
     if (!\is_array($array)) {
+        if (!$array instanceof \Traversable) {
+            throw new RuntimeError(sprintf('An array filter only works with arrays or "Traversable", got "%s" as first argument.', \gettype($array)));
```
guilliamxavier:
```suggestion
            throw new RuntimeError(sprintf('The reduce filter only works with arrays or "Traversable", got "%s" as first argument.', \gettype($array)));
```

fabpot:
Can you rebase on 1.x and remove the check in twig_array_filter as it is redundant?

thijzert:
Some would say "redundant," some would say "preemptive regression test," but sure, no problem.

src/Extension/CoreExtension.php
```diff
@@ -1707,6 +1707,10 @@ function twig_array_filter($array, $arrow)
         return array_filter($array, $arrow);
     }
 
+    if (!$array instanceof \Traversable) {
```
fabpot: This test must be removed, the same test is already done line 1698.

fabpot (changes requested):

fabpot:
@thijzert I made a comment to make myself clearer. The test was fine and can be re-added :)

thijzert:
Ah, yes, I see the confusion now.

src/Extension/CoreExtension.php
thijzert: Done.

fabpot:
Thank you @thijzert.

PR #3344, opened 2020-06-05
Fix the compilation of Twig comparison operators

### LINKED ISSUE #3318: Why twig negative (not) conditional works like this?
I don't understand why Twig `not` is working like this : 

```
{{ dump( false ) }}
{{ dump( not(false) ) }}

{{ dump ( date1.timestamp > date2.timestamp ) }}
{{ dump ( not(date1.timestamp > date2.timestamp) ) }}

{% set comp = date1.timestamp > date2.timestamp %}
{{ dump ( comp ) }}
{{ dump ( not(comp) ) }}
```

Result : 

```
false
true

false
false

false
true
```

https://twigfiddle.com/udjjti/2

Thank's for your help :)

stof:
The new compilation implemented in Twig 3 for PHP 7 (to match the comparison semantic of PHP 8) was missing the surrounding braces (which are part of the AsbtractBinary compilation). This caused issues when the full expression would involve other operators (like the not operator), as things would then rely on having the correct precedence in PHP (which would not be the case).

Closes #3318

### DIFF
--- a/src/Node/Expression/Binary/EqualBinary.php
+++ b/src/Node/Expression/Binary/EqualBinary.php
@@ -24,11 +24,11 @@ class EqualBinary extends AbstractBinary
         }
 
         $compiler
-            ->raw('0 === twig_compare(')
+            ->raw('(0 === twig_compare(')
             ->subcompile($this->getNode('left'))
             ->raw(', ')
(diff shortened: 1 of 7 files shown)


fabpot:
Thank you @stof.

PR #3346, opened 2020-06-08
Propose md file for changelog

ColonelMoutarde:
And remove old changelog


fabpot:
The current CHANGELOG file is already using md, no need to rename the file.

PR #3348, opened 2020-06-09
Fixes #3351

### LINKED ISSUE #3351: Sandbox is not disabled if syntax error occurs within {% sandbox %} tag
I added test for that: https://github.com/twigphp/Twig/pull/3348

Warxcell:
Sandbox mode is not disabled if syntax error occurs inside {% sandbox %} tag

Fixes #3351


src/Node/SandboxNode.php
```diff
@@ -34,12 +34,15 @@ public function compile(Compiler $compiler): void
             ->write("\$this->sandbox->enableSandbox();\n")
             ->outdent()
             ->write("}\n")
+            ->write("try {\n")
```
stof, line 37: to generate good looking code, please add the `->indent()` and `->outdent()` calls  to indent the generated code inside the `try` and `finally` blocks.

stof (changes requested):

src/Node/SandboxNode.php
Warxcell: Cool, thanks. Didn't knew that.

stof (approved):

fabpot:
Thank you @Warxcell.

PR #3349, opened 2020-06-09
Add test to verify that filter is not working when template_from_string

Warxcell:
When using {{ include(sandboxed = true) }} in  conjunction with {{ template_from_string() }} - sandbox is not active.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
 # 1.44.0 (2020-XX-XX)
 
+ * Fix sandbox support when using "include(template_from_string())"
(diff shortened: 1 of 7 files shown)


fabpot:
Thank you @Warxcell.

ruudk:
@Warxcell Could you please have a look [here](https://github.com/twigphp/Twig/issues/3411#issuecomment-716206187)? This seems to be causing lots of issues for people that don't have the Sandbox enabled.

fabpot:
I'm on it.

PR #3353, opened 2020-06-12
[3.x] Fix twig_compare w.r.t. whitespace trimming

### LINKED ISSUE #3352: [3.x] 3 failing tests related to trim in twig_compare
1 from https://github.com/twigphp/Twig/pull/3256#discussion_r438157867 + 2 more:

```twig
{{ 42 > ' foo' ? '.' : 'F' }}
{{ 42 == "42\f" ? '.' : 'F' }}
{{ 42 > "\x00\x34\x32" ? '.' : 'F' }}
```

(https://twigfiddle.com/h5x8fk)

PHP:

```php
echo 1 === twig_compare(42, ' foo') ? '.' : 'F', "\n";
echo 0 === twig_compare(42, "42\f") ? '.' : 'F', "\n";
echo 1 === twig_compare(42, "\x00\x34\x32") ? '.' : 'F', "\n";
```

More detailed: https://3v4l.org/VkVU2

I guess the fixes would be like
(1) use a new variable for the trimmed string and use it *except* in `return (string) $a <=> $b;`, and
(2) trim explicitly the chars `" \t\n\r\v\f"` (from https://github.com/php/php-src/blob/php-7.4.7/Zend/zend_operators.c#L3041), because the default misses `\f` and includes `\0`.

Viz. (just for the `int <=> string` case): https://3v4l.org/VFPbu

guilliamxavier:
Fixes #3352


fabpot (approved):

fabpot:
Thank you @guilliamxavier.

PR #3355, opened 2020-06-17
Minimum version fix

pgorod:
I just tried this on Composer

### DIFF
--- a/doc/intro.rst
+++ b/doc/intro.rst
@@ -26,7 +26,7 @@ Slim, Yii, Laravel, and Codeigniter — just to name a few.
 Prerequisites
 -------------
 
-Twig needs at least **PHP 7.2.0** to run.
+Twig 3.x needs at least **PHP 7.2.5** to run.
 
 Installation
 ------------


doc/intro.rst
```diff
@@ -26,7 +26,7 @@ Slim, Yii, Laravel, and Codeigniter — just to name a few.
 Prerequisites
 -------------
 
-Twig needs at least **PHP 7.2.0** to run.
+Twig 3.x needs at least **PHP 7.2.5** to run.
```
fabpot, line 29: 3.x should be removed as the docs are versioned already on the website.

fabpot (changes requested):

doc/intro.rst
pgorod: I know, but... Google doesn't. Have you tried searching the web for `twig 2.0 minimum PHP requirement` and also `twig 3.0 minimum PHP requirement`? It gets pretty confusing. I see `Twig needs at least PHP 7.1.3 to run.` in the **first** result of the Twig 3.0 query.
The only way I could actually see what I needed was to get the composer error. Google searches are sometimes messy. I think it's better to have the complete information right there. 
If you still don't agree I'll change it.

fabpot:
Thank you @pgorod.

PR #3359, opened 2020-06-19
[RFC] Add static code analysis

rvanlaak:
This PR adds running static code analysis via Github Actions. Let me know what you think, of feel free to close it.

* adds PHPStan lvl 5 with existing errors in baseline 
* runs CS Fixer with the already existing config
* adds Makefile to run the same image locally as Github Actions would
* move `IntegrationTestCase` and `NodeTestCase` out of `/src` to `/tests` namespace

The code changes in `/src` are related to running CS Fixer

Gently borrowed from https://github.com/guzzle/guzzle/blob/master/.github/workflows/static.yml

### DIFF
--- a/.github/workflows/static.yml
+++ b/.github/workflows/static.yml
@@ -1,0 +1,36 @@
+name: Static analysis
+
+on:
+    push:
+        branches:
+            - 3.x
+    pull_request:
+
+jobs:
+    phpstan:
+        name: PHPStan
+        runs-on: ubuntu-latest
+
+        steps:
+            - name: Checkout code
+              uses: actions/checkout@v2
+
+            - name: PHPStan
+              uses: OskarStark/phpstan-ga@0.12.28
+              env:
+                  REQUIRE_DEV: true
+              with:
+                  args: analyze --no-progress
+
+    php-cs-fixer:
+        name: PHP-CS-Fixer
+        runs-on: ubuntu-latest
+
+        steps:
+            - name: Checkout code
+              uses: actions/checkout@v2
+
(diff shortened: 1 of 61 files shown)


fabpot:
Thanks for the PR. Several thoughts: first, everything should first be merged into the 1.x branch, which is still maintained. It makes things more complex as the PHP versions we support are different. Then, the change from Test to Tests is wrong. Classes in Test means that developers can use them for their own tests. Running phpstan or any other tool to get some feedback from time to time is fine, having this crazy configuration in the repo is not. For the CS fixes, I will do them on 1.x first and carefully review them first.

rvanlaak:
Thanks for the feedback so far. 

> everything should first be merged into the 1.x branch, which is still maintained. It makes things more complex as the PHP versions we support are different

Would it be worth to continue working on this PR as I could get it mergeable? As in, is Twig ok with adopting PHPStan and CS-Fixer on its CI pipeline?

>  Running phpstan or any other tool to get some feedback from time to time is fine, having this crazy configuration in the repo is not.

Do you thereby refer to the `phpstan.neon.dist` file? Because that seems to be a minimalistic configuration.

fabpot:
To be very honest, I prefer to not add these tools in the CI. I prefer to run them from time to time and see what I want to change. Adding them to the CI means that contributors can see recommendations that they apply and that's not always needed/desirable. I'd like to keep contributors "safe" from these tools, especially here in Twig.

sstok:
For php-cs-fixer would it make sense to enable fabbot-io?

PR #3363, opened 2020-07-06
Allow trailing commas in macros + functions + filters

apfelbox:
I implemented support for trailing commas in

- function calls
- filter arguments
- macro parameters + arguments (so call + definition)

Lists + maps are already supported https://twigfiddle.com/6atiz3

Is something missing?

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,6 @@
 # 1.43.1 (2020-XX-XX)
 
- * n/a
+ * Allow trailing commas in argument lists (in calls as well as definitions)
 
 # 1.43.0 (2020-07-05)
 
--- a/src/ExpressionParser.php
+++ b/src/ExpressionParser.php
@@ -598,6 +598,11 @@ class ExpressionParser
         while (!$stream->test(Token::PUNCTUATION_TYPE, ')')) {
             if (!empty($args)) {
                 $stream->expect(Token::PUNCTUATION_TYPE, ',', 'Arguments must be separated by a comma');
+
+                // if the comma above was a trailing comma, early exit the argument parse loop
+                if ($stream->test(/* Token::PUNCTUATION_TYPE */ 9, ')')) {
+                    break;
+                }
             }
 
             if ($definition) {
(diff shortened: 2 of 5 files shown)


src/ExpressionParser.php
```diff
@@ -579,6 +579,11 @@ public function parseArguments($namedArguments = false, $definition = false, $al
         while (!$stream->test(/* Token::PUNCTUATION_TYPE */ 9, ')')) {
             if (!empty($args)) {
                 $stream->expect(/* Token::PUNCTUATION_TYPE */ 9, ',', 'Arguments must be separated by a comma');
+
+                // if the comma above was a trailing comma, early exit the argument parse loop
+                if ($stream->test(/* Token::PUNCTUATION_TYPE */ 9, ')')) {
+                    break;
+                }
```
apfelbox, line 605: The nice part of this addition here is that it prevents usage with no arguments.

So `_self.test(a,)` is allowed as macro call, `_self.test(,)` isn't however.

CHANGELOG
```diff
@@ -1,6 +1,7 @@
 # 3.0.5 (2020-XX-XX)
 
  * n/a
+ * Allow trailing commas in argument lists (in calls as well as definitions)
```
fabpot: Can you do it in 2.x instead?

fabpot: We might even add support in 1.x as the patch will probably apply cleanly.

apfelbox: Sure thing, I will reapply it on 1.x

apfelbox: Done.

apfelbox:
@fabpot rebased + update the target branch to also include it in v1.

CHANGELOG
apfelbox: > as the patch will probably apply cleanly.

It was exactly the same code change. The only change I needed to do was to import the macro specifically in the macro test, as `_self` is deprecated in v1.

fabpot:
Thank you @apfelbox.

PR #3367, opened 2020-07-06
Unify code style

apfelbox:
I noticed that the code style in `ExpressionParser` changed from v1 to v3. So this PR now fixes it, that the file in v1 looks more clean.

@fabpot this is a small addendum to #3363


fabpot:
Good catch, thanks @apfelbox.

PR #3371, opened 2020-07-15
Return the object property that is actually accessible.

ErikRoelofs:
Since there is a check for two types of property, we should also return the one we found.

Without this, cases where a property is accessed that exists in the cast of the object, but not directly, will not be available and will throw notices and end up with NULL.

We noticed this in our project because of using a DateTimeInterface's date property (which I wouldn't recommend, but still). The if check will respond `true` because if you cast a DateTimeImmutable to an array it has a  'date' key, but trying to access it directly does not actually work (at least not in PHP 7.4. It apparently worked before as the code had been running fine before we upgraded.)

My proposed change should work in all situations.

### DIFF
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -1450,7 +1450,10 @@ function twig_get_attribute(Environment $env, Source $source, $object, $item, ar
                 $env->getExtension(SandboxExtension::class)->checkPropertyAllowed($object, $item, $lineno, $source);
             }
 
-            return $object->$item;
+            if (isset($object->$item)) {
+                return $object->$item;
+            }
+            return ((array) $object)[(string) $item];
         }
     }
 


ErikRoelofs:
Hm, the nightly build has failed. Is that something my code did? It looks more like an issue with the nightly build or some other part of Twig?

src/Extension/CoreExtension.php
```diff
@@ -1450,7 +1450,10 @@ function twig_get_attribute(Environment $env, Source $source, $object, $item, ar
                 $env->getExtension(SandboxExtension::class)->checkPropertyAllowed($object, $item, $lineno, $source);
             }
 
-            return $object->$item;
+            if (isset($object->$item)) {
+                return $object->$item;
+            }
+            return ((array) $object)[(string) $item];
```
fabpot, line 1456:
```suggestion
            return (isset($object->$item) ? $object->$item : ((array) $object)[(string) $item];
```

fabpot (changes requested):
LGTM.
Can you add some tests that would fail before your changes?

PR #3379, opened 2020-07-29
Fix for creating dir in FilesystemCache (mkdir error)

puniserv:
Error:
`mkdir(): File exists`

This changes should fix this error.


stof:
The current code uses silencing on the error rather than a check before it precisely to be safe against race condition (another process might create the folder between the `is_dir` check and the `mkdir` call).
So your change is making the code more fragile, reverting changes done in the past to harden it.

If your error reporting mechanism is not respecting the error silencing operator, the issue is on your side, not in Twig.

fabpot:
Closing for the reason explained by @stof

PR #3386, opened 2020-08-12
Add slug

### LINKED ISSUE #3320: Use Slugger in Twig
Hi,
It would be great to be able to use Slugger (AsciiSlugger) in twig.
Thanks :)

seb-jean:
Hi,
 I fix #3320

### DIFF
--- a/extra/string-extra/src/StringExtension.php
+++ b/extra/string-extra/src/StringExtension.php
@@ -11,6 +11,7 @@
 
 namespace Twig\Extra\String;
 
+use Symfony\Component\String\Slugger\AsciiSlugger;
 use Symfony\Component\String\UnicodeString;
 use Twig\Extension\AbstractExtension;
 use Twig\TwigFilter;
@@ -21,11 +22,17 @@ final class StringExtension extends AbstractExtension
     {
         return [
             new TwigFilter('u', [$this, 'createUnicodeString']),
+            new TwigFilter('slug', [$this, 'createSlug']),
         ];
     }
 
     public function createUnicodeString(?string $text): UnicodeString
     {
         return new UnicodeString($text ?? '');
+    }
+
+    public function createSlug(?string $string, ?string $separator = '-', string $locale = null): string
+    {
+        return (new AsciiSlugger())->slug($string, $separator, $locale);
     }
 }
--- a/extra/string-extra/src/StringExtension.php
+++ b/extra/string-extra/src/StringExtension.php
@@ -11,6 +11,7 @@
 
 namespace Twig\Extra\String;
 
+use Symfony\Component\String\AbstractUnicodeString;
 use Symfony\Component\String\Slugger\AsciiSlugger;
 use Symfony\Component\String\UnicodeString;
 use Twig\Extension\AbstractExtension;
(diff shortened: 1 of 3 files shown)


extra/string-extra/src/StringExtension.php
```diff
@@ -21,11 +22,17 @@ public function getFilters()
...
+            new TwigFilter('slug', [$this, 'createSlug']),
         ];
     }
 
     public function createUnicodeString(?string $text): UnicodeString
     {
         return new UnicodeString($text ?? '');
     }
+
+    public function createSlug(?string $string, ?string $separator = '-', string $locale = null): string
```
fabpot: should be ` public function createSlug(string $string, string $separator = '-', string $locale = null): string` to be consistent with the `AsciiSlugger` signature.

And it returns a `AbstractUnicodeString` instance. Not sure if we should always return a string or the instance.

fabpot (changes requested):
Can you add some tests?

seb-jean:
> Can you add some tests?

Can you tell me the procedure for the tests? I saw that for filter "u", there's the file https://github.com/twigphp/Twig/blob/3.x/extra/string-extra/tests/Fixtures/string.test.

 Thanks

fabpot:
Indeed, you can do the same as for the u filter.

extra/string-extra/src/StringExtension.php
```diff
@@ -11,6 +11,8 @@
 
 namespace Twig\Extra\String;
 
+use Symfony\Component\String\AbstractUnicodeString;
+use Symfony\Component\String\Slugger\AsciiSlugger;
```
Korbeil, line 15: Since `AsciiSlugger` is an optional feature of `symfony/string`, the `Symfony\Contracts\Translation\LocaleAwareInterface` class required by it is not inside `symfony/string` dependencies, so you should add ` symfony/translation-contracts` to Twig dependencies.

seb-jean: I just added it to "extra/string-extra/composer.json."

seb-jean:
> Indeed, you can do the same as for the u filter.

I just added some tests.

extra/string-extra/src/StringExtension.php
```diff
@@ -21,11 +23,17 @@ public function getFilters()
...
+            new TwigFilter('slug', [$this, 'createSlug']),
         ];
     }
 
     public function createUnicodeString(?string $text): UnicodeString
     {
         return new UnicodeString($text ?? '');
     }
+
+    public function createSlug(string $string, string $separator = '-', string $locale = null): AbstractUnicodeString
```
stof, line 35: making the locale optional here seems weird to me. The behavior in AsciiSlugger means that it keeps the current locale set in the slugger. But as you instantiate the slugger here rather than using DI, there is no way to control the locale of the slugger.

seb-jean: What do you propose @stof ?

stof: IMO, the Slugger should be injected in the extension with DI. This allows this instance to be integrated with your locale management to benefit from it being locale aware.

fabpot:
Any news on this PR? @seb-jean Are you still interested in finishing it?

seb-jean:
I am still interested in this PR but I am blocking to develop it. I may not be experienced enough to finish it.

EmilePerron:
This feature is quite relevant for a project I'm working on, and I'd love for it to be available in the existing StringExtension instead of creating my own custom extension for it.

Is anyone still working on this?

If @seb-jean isn't available to work on this PR anymore, I'd love to hop in and finish the implementation so it can be merged.

fabpot:
Closing in favor of #3478

PR #3388, opened 2020-08-13
Fixed Twig\Template check

brandonkelly:
Looks like this was missed when the `Twig\Template` class was added, and `Twig_Template` was deprecated + aliased.

### DIFF
--- a/src/Error/Error.php
+++ b/src/Error/Error.php
@@ -198,7 +198,7 @@ class Error extends \Exception
 
         $backtrace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS | DEBUG_BACKTRACE_PROVIDE_OBJECT);
         foreach ($backtrace as $trace) {
-            if (isset($trace['object']) && $trace['object'] instanceof Template && 'Twig_Template' !== \get_class($trace['object'])) {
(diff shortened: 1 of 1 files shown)


src/Error/Error.php
```diff
@@ -198,7 +198,7 @@ private function guessTemplateInfo()
 
         $backtrace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS | DEBUG_BACKTRACE_PROVIDE_OBJECT);
         foreach ($backtrace as $trace) {
-            if (isset($trace['object']) && $trace['object'] instanceof Template && 'Twig_Template' !== \get_class($trace['object'])) {
+            if (isset($trace['object']) && $trace['object'] instanceof Template && 'Twig\Template' !== \get_class($trace['object'])) {
```
fabpot, line 201: Can't we have both depending on what the user is using?

stof: `get_class` always return the canonical class name, not the alias. So this was indeed forgotten when changing which name is the alias.

fabpot:
Thank you @brandonkelly.

src/Error/Error.php
fabpot: hmmm, the code is actually correct. We are testing that the object is an instance of Template, but not of class Twig_Template.

Issue #654, opened 2012-03-02
Break a for loop?


dnizetic:
Hi,

is it possible to break a loop?

I googled and searched it here, but it seems like this isn't supported in Twig.

So I was wandering - how come it's left out?

ernestob:
how dum is that ... a break should be included
who cares about the loop.index

#fails

```
{% set my_array = ['a', 'b', 'c', 'b'] %}
{% set counter = 0 %}
{% for element in my_array %}
    {% if element == 'b' %}
        {% set index = counter %}
        {% break %}
    {% endif %}
    {% set counter = counter + 1 %}
{% endfor %}
```

#workaround

```
{% set counter = 0 %}
{% set found = false %}
{% set my_array = ['a', 'b', 'c', 'b'] %}
{% for element in my_array %}
    {% if element == 'b' %}
        {% if not found %}
            {% set index = counter %}
            {% set found = true %}
        {% endif %}
    {% endif %}
    {% set counter = counter + 1 %}
{% endfor %}
```

Holly

ndusan:
I just had to do something similar and I think it could be interesting for the others.
If we take the case where u should stop looping after X, u can do it so:

{% set cnt = 10 }
{% for el in collection if cnt > 0 %}
{% set cnt = cnt - 1 %}

{# do something #}

{% endfor %}

Maybe this will help someone :)

ArlingtonHouse:
Breaking a loop is not business logic at all, it is display logic that belongs in twig at the view layer.   How many items in a specific array do I want to display on this page?  UPDATE: the way to accomplish this is using the Twig filter "slice" to cut down your array in the template .

Maxooo:
7 years later, the need is still here... 🙈

sarath49:
still in need and not implemented 
I see one craft3 implementation for twig, can it be extended/implemented.
https://github.com/marionnewlevant/craft-mnbreakandcontinue

sfaragy:
{% for x in allx %}
//statements
//conditions
 {% set break = true %}
{% endfor %}

bajahranks:
It's amazing that years later and still nothing has been done. Just incredible. If "break" is business logic then "for and if" statements are business logic as well. SMH.

fabpot:
What’s incredible is that nobody took the time to submit a PR in 7 years. Closing now. If someone wants to work on it, feel free too open a PR.

btinet:
Hi Folks,

you can achieve a break in a for loop like following:

```
[...]

PR #3395, opened 2020-09-10
BreakNode - introduction

SpacePossum:
Hi all!

The PR aims to close/implement https://github.com/twigphp/Twig/issues/654

### DIFF
--- a/src/Node/BreakNode.php
+++ b/src/Node/BreakNode.php
@@ -1,0 +1,42 @@
+<?php
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Twig\Node;
+
+use Twig\Compiler;
+
+final class BreakNode extends Node
+{
+    /**
+     * @var int
+     */
+    private $target;
+
+    public function __construct(int $target, int $linenumber, string $tag)
+    {
+        parent::__construct([], ['target' => $target], $linenumber, $tag);
+
+        $this->target = $target;
+    }
+
+    public function setTarget(int $target): void
+    {
+        $this->target = $target;
+    }
+
+    public function compile(Compiler $compiler): void
+    {
+        $compiler
+            ->addDebugInfo($this)
+            ->write($this->target > 1 ? sprintf("break %d;\n", $this->target) : "break;\n")
+        ;
+    }
+}
--- a/src/NodeVisitor/LoopNodeVisitor.php
+++ b/src/NodeVisitor/LoopNodeVisitor.php
@@ -48,8 +48,12 @@ final class LoopNodeVisitor implements NodeVisitorInterface
         } elseif ($node instanceof BreakNode) {
             $target = $node->getAttribute('target');
 
-            if (!\is_int($target) || $target < 1) {
-                throw new SyntaxError(sprintf('Break target must be an integer > 1, got "%s".', var_export($target, true)), $node->getTemplateLine(), $node->getSourceContext());
+            if (!\is_int($target)) {
+                throw new SyntaxError(sprintf('Break target must be an integer, got "%s".', \is_object($target) ? \get_class($target) : \gettype($target)), $node->getTemplateLine(), $node->getSourceContext());
+            }
+
+            if ($target < 1) {
+                throw new SyntaxError(sprintf('Break target must be greater than 1, got %s.', $target), $node->getTemplateLine(), $node->getSourceContext());
             }
 
             if (null !== $this->elseNode) {
@@ -61,7 +65,7 @@ final class LoopNodeVisitor implements NodeVisitorInterface
             }
 
             if ($target > $this->depth) {
-                if (\count($this->structureStack) > 0) {
+                if (0 < \count($this->structureStack)) {
                     [$structure] = array_pop($this->structureStack);
 
                     if ($structure instanceof BlockNode) {
--- a/src/NodeVisitor/LoopNodeVisitor.php
+++ b/src/NodeVisitor/LoopNodeVisitor.php
@@ -1,0 +1,109 @@
+<?php
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Twig\NodeVisitor;
+
+use Twig\Environment;
+use Twig\Error\SyntaxError;
+use Twig\Node\BlockNode;
+use Twig\Node\BreakNode;
+use Twig\Node\ForNode;
+use Twig\Node\MacroNode;
+use Twig\Node\Node;
+use Twig\Node\WithNode;
+
+final class LoopNodeVisitor implements NodeVisitorInterface
+{
+    /**
+     * @var int
+     */
+    private $depth = 0;
+
+    /**
+     * @var int[]
+     */
+    private $structureStack = [];
+
+    /**
+     * @var Node|null
+     */
+    private $elseNode;
+
+    public function enterNode(\Twig_NodeInterface $node, Environment $env): Node
+    {
+        if ($this->isStructureNode($node)) {
+            $this->structureStack[] = [$node, $this->depth, $this->elseNode];
+            $this->depth = 0;
+        } elseif ($node instanceof ForNode) {
+            $this->elseNode = $node->hasNode('else') ? $node->getNode('else') : null;
+            ++$this->depth;
+        } elseif ($node instanceof BreakNode) {
+            $target = $node->getAttribute('target');
+
+            if (!\is_int($target) || $target < 1) {
+                throw new SyntaxError(sprintf('Break target must be an integer > 1, got "%s".', var_export($target, true)), $node->getTemplateLine(), $node->getSourceContext());
+            }
+
+            if (null !== $this->elseNode) {
+                if (1 === $target) {
+                    throw new SyntaxError('Cannot break from itself in "else" of "for" .', $node->getTemplateLine(), $node->getSourceContext());
+                }
+
+                $node->setTarget(--$target);
+            }
+
+            if ($target > $this->depth) {
+                if (\count($this->structureStack) > 0) {
+                    [$structure] = array_pop($this->structureStack);
+
+                    if ($structure instanceof BlockNode) {
+                        throw new SyntaxError('Break tag target outside of loop, did you try to target outside the block?', $node->getTemplateLine(), $node->getSourceContext());
+                    }
+
+                    if ($structure instanceof MacroNode) {
+                        throw new SyntaxError('Break tag target outside of loop, did you try to target outside the macro?', $node->getTemplateLine(), $node->getSourceContext());
+                    }
+
+                    if ($structure instanceof WithNode) {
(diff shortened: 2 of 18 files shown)


src/TokenParser/BreakNodeParser.php
```diff
@@ -0,0 +1,64 @@
...
+
+namespace Twig\TokenParser;
+
+use Twig\Error\SyntaxError;
+use Twig\Node\BreakNode;
+use Twig\Token;
+
+final class BreakNodeParser extends AbstractTokenParser
+{
+    public function parse(Token $token)
```
SpacePossum: missing return type

fabpot (commented):
Si we want support breaking multiple levels?

src/Node/BreakNode.php
```diff
@@ -0,0 +1,37 @@
...
+
+use Twig\Compiler;
+
+final class BreakNode extends Node
+{
+    private int $level;
+
+    public function __construct(int $level)
+    {
+        parent::__construct([], [], 0, null);
```
stof: passing `0` as lineno is a no-go. the line of the node in the source must be given. Otherwise, it can do weird things for the error reporting.
And I would also pass a non-empty tag name.

src/Node/BreakNode.php
```diff
@@ -0,0 +1,37 @@
...
+
+        $this->level = $level;
+    }
+
+    public function compile(Compiler $compiler): void
+    {
+        $compiler->write(
```
stof: please add the call to `->addDebugInfo($this)` at the beginning of the compilation, to have better debug info in the generated code (this BreakNode is a statement-like node)

src/TokenParser/BreakNodeParser.php
```diff
@@ -0,0 +1,78 @@
...
+                if (0 === $loopTarget) {
+                    break;
+                }
+            }
+
+            ++$forward;
+        } while (true);
+
+        if (0 !== $loopTarget) {
+            throw new SyntaxError('Break tag target outside of loop.', $stream->getCurrent()->getLine(), $stream->getSourceContext());
```
stof: this looking forward to find how many loops get ended after the tag is very fragile to me. I think it would be better to implement such validation using a node visitor traversing the AST.

stof: btw, given that blocks and macros are generating separate methods in the compile code, you cannot break a loop higher in the AST from inside the block (that would be invalid PHP code). Your current validation does not deal with that (please add tests covering these cases too)

Allowing to break a loop accross a `{% with %}` tag might also be hard to implement, to ensure that the scope is still properly restore by `{% endwith %}`. So I would probably also forbid it.

And we clearly need more tests covering the compatibility of break with specific features of the `{% for %}` tag (the scoping rules of variables for instance), to ensure they still work fine when break is used (also multi-level break).

src/Node/BreakNode.php
SpacePossum: thanks, done!

src/Node/BreakNode.php
SpacePossum: thank you, I've added the line number and tag name, however I'm not 100% sure about the two `arrays` here

src/TokenParser/BreakNodeParser.php
SpacePossum: Thanks for all the details.
If you've the time please let me know if I'm on the right track with the visitor. 
Will add more tests for sure!

src/Node/BreakNode.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+declare(strict_types=1);
```
fabpot: Can be removed in all classes.

fabpot (changes requested):

src/Node/BreakNode.php
SpacePossum: thanks, force of habit, removed!

src/TokenParser/BreakNodeParser.php
SpacePossum: I've added the test

src/NodeVisitor/LoopNodeVisitor.php
```diff
@@ -0,0 +1,109 @@
...
+            $this->structureStack[] = [$node, $this->depth, $this->elseNode];
+            $this->depth = 0;
+        } elseif ($node instanceof ForNode) {
+            $this->elseNode = $node->hasNode('else') ? $node->getNode('else') : null;
+            ++$this->depth;
+        } elseif ($node instanceof BreakNode) {
+            $target = $node->getAttribute('target');
+
+            if (!\is_int($target) || $target < 1) {
+                throw new SyntaxError(sprintf('Break target must be an integer > 1, got "%s".', var_export($target, true)), $node->getTemplateLine(), $node->getSourceContext());
```
fabpot: I would avoid using `var_export()` here. What about having a different error for when this is an invalid integer and when it's not an integer? For the second case, we could use what we are using everywhere else: `\is_object($target) ? \get_class($target) : \gettype($target)`

src/NodeVisitor/LoopNodeVisitor.php
```diff
@@ -0,0 +1,109 @@
...
+            if (null !== $this->elseNode) {
+                if (1 === $target) {
+                    throw new SyntaxError('Cannot break from itself in "else" of "for" .', $node->getTemplateLine(), $node->getSourceContext());
+                }
+
+                $node->setTarget(--$target);
+            }
+
+            if ($target > $this->depth) {
+                if (\count($this->structureStack) > 0) {
```
fabpot: `if (0 < \count($this->structureStack)) {`

src/NodeVisitor/LoopNodeVisitor.php
```diff
@@ -0,0 +1,109 @@
...
+                    [$structure] = array_pop($this->structureStack);
+
+                    if ($structure instanceof BlockNode) {
+                        throw new SyntaxError('Break tag target outside of loop, did you try to target outside the block?', $node->getTemplateLine(), $node->getSourceContext());
```
fabpot, line 72: `A "break" tag cannot be used inside a "block" tag.`?

src/NodeVisitor/LoopNodeVisitor.php
```diff
@@ -0,0 +1,109 @@
...
+                    }
+
+                    if ($structure instanceof MacroNode) {
+                        throw new SyntaxError('Break tag target outside of loop, did you try to target outside the macro?', $node->getTemplateLine(), $node->getSourceContext());
```
fabpot, line 76: Same here

src/NodeVisitor/LoopNodeVisitor.php
```diff
@@ -0,0 +1,109 @@
...
+                    }
+
+                    if ($structure instanceof WithNode) {
+                        throw new SyntaxError('Break tag target outside of loop, did you try to target outside the "with"?', $node->getTemplateLine(), $node->getSourceContext());
```
fabpot, line 80: Same here

src/NodeVisitor/LoopNodeVisitor.php
```diff
@@ -0,0 +1,109 @@
...
+                    }
+                }
+
+                throw new SyntaxError('Break tag target outside of loop.', $node->getTemplateLine(), $node->getSourceContext());
```
fabpot, line 84: `A "break" tag cannot be used outside of a "loop" tag.`

src/NodeVisitor/LoopNodeVisitor.php
```diff
@@ -0,0 +1,109 @@
...
+        } elseif ($node instanceof BreakNode) {
+            $target = $node->getAttribute('target');
+
+            if (!\is_int($target) || $target < 1) {
+                throw new SyntaxError(sprintf('Break target must be an integer > 1, got "%s".', var_export($target, true)), $node->getTemplateLine(), $node->getSourceContext());
+            }
+
+            if (null !== $this->elseNode) {
+                if (1 === $target) {
+                    throw new SyntaxError('Cannot break from itself in "else" of "for" .', $node->getTemplateLine(), $node->getSourceContext());
```
fabpot, line 61: `A "break" tag cannot be used in the "else" clause of a "for" tag.`

src/NodeVisitor/LoopNodeVisitor.php
```diff
@@ -0,0 +1,109 @@
...
+    private $structureStack = [];
+
+    /**
+     * @var Node|null
+     */
+    private $elseNode;
+
+    public function enterNode(Node $node, Environment $env): Node
+    {
+        if ($this->isStructureNode($node)) {
```
fabpot, line 42: I think we need a `if ($node instanceof ModuleNode) {` that would reset `structureStack`.

fabpot (changes requested):
Can you an entry in the CHANGELOG?
Also, as 1.x is still maintained, can you target 1.x instead of 3.x?

src/NodeVisitor/LoopNodeVisitor.php
SpacePossum: run into a recursive `var_export` the other day... which is not supported.., good idea to avoid, updated, thanks!

src/NodeVisitor/LoopNodeVisitor.php
SpacePossum: :+1:  done, happy to follow any CS

src/NodeVisitor/LoopNodeVisitor.php
SpacePossum: I struggled with the copy and I agree it isn't really anything good. However, people can use the `break` in a `block`, what we detect here is an invalid target.

For example, this is fine:
```twig
{% block body %}
    {% for item in items %}
        {% break %}
    {% endfor %}
{% endblock %}
```

but here we see something like:
```twig
{% for item in items %}
    {% block body %}
        {% break %}
    {% endblock %}
{% endfor %}
```
which is not supported

src/NodeVisitor/LoopNodeVisitor.php
SpacePossum: TBH; I'm kind of out of my league here as I'm not familiar with the ModuleNode.
I see I missed the `EmbedNode`, `SandboxNode` and `AutoEscapeNode` in the list of "structure" nodes, not sure if or how this related to your suggestion though.

(maybe also `apply`, `verbatim` and `with` tags?)

stof: SandboxNode and AutoEscapeNode could be allowed without impacting the breaking of loops, if we change the compiled code to use a `try/finally` to ensure that their cleanup logic also runs if a `break;` interrupts the execution of a parent loop from inside the node: https://3v4l.org/RotPq (we would have to check how this impacts error handling). This might apply to more tags relying on a cleanup phase (WithNode for instance).
However, such need means that the introduction of `break` in Twig is not a small change. It could be considered as a BC break because it is the first time that a syntax allows to have jumps in the execution of the AST (a deep descendant node impacts the execution path of a node much higher), which is a significant language change (other features have to know how to interact with it)

Blocks are macros are hard blockers here because they are compiled as separate PHP functions.
And the `{% embed %}` is parsed separately as a ModuleNode. It is not a child in the AST (EmbedNode represents the call to render this associated anonymous template)

SpacePossum: Thanks for the details. I see no way how to handle this so I'm closing the PR :+1:

PR #3397, opened 2020-09-13
Replace implicit dependence on ext/iconv in JS escaper

JoshyPHP:
Twig's JS escaper fails on non-word characters if `ext/iconv` isn't available. That's because it uses iconv to replace characters with their hex codepoints and split characters outside the Basic Multilingual Plane into surrogate pairs. `symfony/polyfill-iconv` does not support the UTF-16BE conversion and cannot be used as a replacement.

This PR splits the surrogates using native PHP, after obtaining the codepoint via `mb_ord()` which is provided either natively or via `symfony/polyfill-mbstring`, which is already part of Twig's requirements.

### DIFF
--- a/src/Extension/EscaperExtension.php
+++ b/src/Extension/EscaperExtension.php
@@ -282,15 +282,18 @@ function twig_escape_filter(Environment $env, $string, $strategy = 'html', $char
                     return $shortMap[$char];
                 }
 
-                // \uHHHH
-                $char = twig_convert_encoding($char, 'UTF-16BE', 'UTF-8');
-                $char = strtoupper(bin2hex($char));
-
-                if (4 >= \strlen($char)) {
-                    return sprintf('\u%04s', $char);
+                $codepoint = mb_ord($char);
+                if (0x10000 > $codepoint) {
+                    return sprintf('\u%04X', $codepoint);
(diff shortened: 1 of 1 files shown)


nicolas-grekas (approved):
Looks good to me.
Can you please check if/confirm that both cases are covered by a unit test?

JoshyPHP:
You're right, I should have addressed testing. I did look into the current test suite and there already are tests covering non-ASCII characters and surrogates, that's why I didn't add any. I manually confirmed that a non-functional implementation would cause those tests to fail.

https://github.com/twigphp/Twig/blob/b48bd18dc6f967ad09af9eab521cdf0e68fb6a95/tests/Extension/EscaperTest.php#L72-L74

Now, a regression test would need to run with `ext/iconv` disabled and to be thorough the new implementation should be tested with `mbstring` disabled. Unfortunately, I don't think there's a simple way to do that in Travis. Also, unrelated parts of the test suite fail if `ext/iconv` is disabled.

In summary, there's already a couple of test cases for that portion of the code and I manually ran them with both `ext/iconv` and `mbstring` disabled, successfully.

fabpot:
Thank you @JoshyPHP.

PR #3401, opened 2020-09-25
Fix missing delimiter for preg_quote in lexer

waitxd:
It is a really edge case (PHP < 7.3 and '#' as whitespace_trim symbol), but for the sake of consistency here's a proper fix.


stof (approved):

fabpot:
Thank you @waitxd.

PR #3406, opened 2020-10-07
[Documentation] Emphase the fact that the include tag outputs data

nstcactus:


fabpot:
Thank you @nstCactus.

PR #3407, opened 2020-10-08
Improving truncate documentation

cavemon:
Adding description and example of third parameter for `truncate` method to preserve whole words.  This seems like a common use case and extremely useful addition that should be documented.


fabpot:
Thank you @cavemon.

PR #3408, opened 2020-10-13
Update u.rst

tomcyr:
Explained the third argument of the truncate filter (the last word before the cut is kept even if that generates a string longer than the desired length)


fabpot:
Closing as a duplicate of #3407

PR #3409, opened 2020-10-13
[doc] Quick access to ICU user guide

SiM07:
I usefull to access quicky to pattern documentation.

### DIFF
--- a/doc/filters/format_datetime.rst
+++ b/doc/filters/format_datetime.rst
@@ -28,7 +28,7 @@ You can tweak the output for the date part and the time part:
 
 Supported values are: ``none``, ``short``, ``medium``, ``long``, and ``full``.
 
-For greater flexiblity, you can even define your own pattern:
+For greater flexiblity, you can even define your own pattern. Possible patterns are documented in the `ICU user guide <https://unicode-org.github.io/icu/userguide/format_parse/datetime/#datetime-format-syntax>`.
 
 .. code-block:: twig
 


doc/filters/format_datetime.rst
```diff
@@ -25,7 +25,7 @@ You can tweak the output for the date part and the time part:
 
 Supported values are: ``none``, ``short``, ``medium``, ``long``, and ``full``.
 
-For greater flexiblity, you can even define your own pattern:
+For greater flexiblity, you can even define your own pattern. Possible patterns are documented in the `ICU user guide <http://userguide.icu-project.org/formatparse/datetime>`.
```
stof: Please use the new location of the ICU user guide at `https://unicode-org.github.io/icu/`

stof (changes requested):

doc/filters/format_datetime.rst
stof: the link you need is https://unicode-org.github.io/icu/userguide/format_parse/datetime/#datetime-format-syntax

stof: btw, the same link might useful for `format_date` and `format_time`

SiM07: I use the same link like as the symfony doc : https://github.com/symfony/symfony-docs/blob/707580ebc78c4e8da259c26474ac1f1562744d69/reference/forms/types/options/date_format.rst.inc

I can open PR to update links too.

When I check `format_date` and `format_time`, these pages redirect to this doc and only explain installation. We can copy all the doc page but ... seems not a good idea.

stof: I would suggest opening a PR on the symfony doc to update the link to the new ICU location, instead of using the old link in a new place. The ICU project migrated their user guide to a new location, which also means that the old page might not be updated anymore.

> these pages redirect to this doc and only explain installation

then, indeed, that's not needed. I forgot they were not documenting things on their own.

fabpot:
Thank you @SiM07.

PR #3410, opened 2020-10-21
Drop support for PHP 7.1

fabpot:


stof:
@fabpot what is the reason to change the min PHP version in Twig 1.x ?

Also, what is the ETA for end of support of 1.x ? We currently have 3 major versions being maintained (and developed with new features) in parallel

fabpot:
That's because of https://github.com/twigphp/Twig/pull/3397/files#diff-6fca80088d7f7eeccac5ad9e3f9c7b689944fad09f04368da524a227d771c951R285 (`mb_ord` is part of PHP 7.2).

Regarding versions, I don't have any ETA for the 1.x end of support. As long as it's sustainable, I don't see a need to stop support.

stof:
> That's because of https://github.com/twigphp/Twig/pull/3397/files#diff-6fca80088d7f7eeccac5ad9e3f9c7b689944fad09f04368da524a227d771c951R285 (`mb_ord` is part of PHP 7.2).

requiring `symfony/polyfill-php72` could have been enough to solve that though

fabpot:
Sure, but 7.1 is not supported anyway.

PR #3412, opened 2020-10-22
fixed FilesystemLoader directory level validation

churow:
This might be the inteded validation of directory level thru increasing and decreasing $level.


stof:
this change is wrong. `../foo/bar` would not be a negative level at the end. But it still goes outside the configured directory, which would allow accessing other files

stof:
Btw, the tests are failing precisely because of that. They are enforcing the proper validation.

PR #3413, opened 2020-10-26
Fix "include(template_from_string())" when not using a sandbox

### LINKED ISSUE #3411: Attempted to call an undefined method named "checkSecurity"
Hello, 

it's seem what something is broken when i updated in twig 1.44

<pre>Attempted to call an undefined method named "checkSecurity" of class "__TwigTemplate_ab8c5b3634ec2ea857069474b49ae5f5b2f484499df40afbe2063e28cae810c4".
500 Internal Server Error - UndefinedMethodException

Stack Trace
in /var/www/vhosts/vendor/twig/twig/src/Extension/CoreExtension.php at line 1572  -
    // if a Template instance is passed, it might have been instantiated outside of a sandbox, check security
    if ($template instanceof TemplateWrapper || $template instanceof Template) {
        $template->unwrap()->checkSecurity();
    }
    $loaded = null;
</pre>

happen with a {{ include(template_from_string("something")) }}

fabpot:
Closed #3411

### DIFF
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -1565,11 +1565,13 @@ function twig_include(Environment $env, $context, $template, $variables = [], $w
         if (!$alreadySandboxed = $sandbox->isSandboxed()) {
             $sandbox->enableSandbox();
         }
-    }
 
-    // if a Template instance is passed, it might have been instantiated outside of a sandbox, check security
-    if ($template instanceof TemplateWrapper || $template instanceof Template) {
-        $template->unwrap()->checkSecurity();
+        foreach ((\is_array($template) ? $template : [$template]) as $name) {
+            // if a Template instance is passed, it might have been instantiated outside of a sandbox, check security
+            if ($name instanceof TemplateWrapper || $name instanceof Template) {
+                $name->unwrap()->checkSecurity();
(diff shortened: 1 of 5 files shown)


src/Extension/CoreExtension.php
```diff
@@ -1565,11 +1565,11 @@ function twig_include(Environment $env, $context, $template, $variables = [], $w
...
             $sandbox->enableSandbox();
         }
-    }
 
-    // if a Template instance is passed, it might have been instantiated outside of a sandbox, check security
-    if ($template instanceof TemplateWrapper || $template instanceof Template) {
-        $template->unwrap()->checkSecurity();
+        // if a Template instance is passed, it might have been instantiated outside of a sandbox, check security
+        if ($template instanceof TemplateWrapper || $template instanceof Template) {
+            $template->unwrap()->checkSecurity();
```
stof: shouldn't this check also be done when the sandbox is already active, even if the `$sandboxed` argument is not explicitly set to `true` ?

fabpot: In that case, the check has already been done.

fabpot:
I've fixed another use case (when you pass an array of templates). /cc @stof

PR #3419, opened 2020-10-29
Enhancement: Use GithubActions instead of Travis for tests

OskarStark:
You can see the run here: https://github.com/OskarStark/Twig/pull/1

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -132,34 +132,36 @@ jobs:
             - name: "Run tests"
               working-directory: ${{ matrix.extension}}
               run: ../../vendor/bin/simple-phpunit
-
-    integration-tests:
-        needs:
-            - 'tests'
-
-        name: "Integration tests with PHP ${{ matrix.php-version }}"
-
-        runs-on: 'ubuntu-20.04'
-
-        continue-on-error: true
-
-        strategy:
-            matrix:
-                php-version:
-                    - '7.3'
-
-        steps:
-            - name: "Checkout code"
-              uses: actions/checkout@v2.3.3
-
-            - name: "Install PHP with extensions"
-              uses: shivammathur/setup-php@2.7.0
-              with:
-                  coverage: "none"
-                  extensions: "gd, pdo_sqlite"
-                  php-version: ${{ matrix.php-version }}
-                  ini-values: memory_limit=-1
-                  tools: composer:v2
-
-            - run: ./drupal_test.sh
-              shell: "bash"
+#
+#    Drupal does not support Twig 3 now!
+#
+#    integration-tests:
+#        needs:
+#            - 'tests'
+#
+#        name: "Integration tests with PHP ${{ matrix.php-version }}"
+#
+#        runs-on: 'ubuntu-20.04'
+#
+#        continue-on-error: true
+#
+#        strategy:
+#            matrix:
+#                php-version:
+#                    - '7.3'
+#
+#        steps:
+#            - name: "Checkout code"
+#              uses: actions/checkout@v2.3.3
+#
+#            - name: "Install PHP with extensions"
+#              uses: shivammathur/setup-php@2.7.0
+#              with:
+#                  coverage: "none"
+#                  extensions: "gd, pdo_sqlite"
+#                  php-version: ${{ matrix.php-version }}
+#                  ini-values: memory_limit=-1
+#                  tools: composer:v2
+#
+#            - run: ./drupal_test.sh
+#              shell: "bash"


OskarStark:
The action will run first time after merge

.github/workflows/ci.yml
```diff
@@ -0,0 +1,81 @@
...
+              working-directory: ${{ matrix.extension}}
+              run: composer config platform.php 7.4.99
+
+            - name: "Composer install: ${{ matrix.extension }}"
+              working-directory: ${{ matrix.extension}}
+              run: composer install
+
+            - name: "Run tests: ${{ matrix.extension }}"
+              working-directory: ${{ matrix.extension}}
+              run: ../vendor/bin/simple-phpunit
```
localheinz:
```suggestion
              run: ../../vendor/bin/simple-phpunit
```

❓

.github/workflows/ci.yml
```diff
@@ -0,0 +1,137 @@
...
+
+env:
+    fail-fast: true
+    SYMFONY_PHPUNIT_DISABLE_RESULT_CACHE: 1
+
+jobs:
+    tests:
+        name: "PHP ${{ matrix.php-version }}"
+
+        runs-on: 'ubuntu-latest'
```
stof: use `ubuntu-18.04` instead. The `ubuntu-latest` alias will point to `ubuntu-20.04` as of 30/11/2020, and `ubuntu-20.04` only supports PHP 7.4+ for the `setup-php` action

.github/workflows/ci.yml
```diff
@@ -0,0 +1,137 @@
...
+        continue-on-error: ${{ matrix.allow-failures }}
+
+        strategy:
+            matrix:
+                php-version:
+                    - '7.2.5'
+                    - '7.3'
+                    - '7.4'
+                    - '8.0'
+                allow-failures: [false]
```
stof: shouldn't this be `fail-fast: false` instead ?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,137 @@
+name: "CI"
+
+on:
+    pull_request:
+    push:
+        branches:
+            - '3.x'
```
stof: Given that 1.x and 2.x are still supported, I think it would be better to have the same CI setup in all branches rather than maintaining both.

fabpot: I agree. We might decide to stop the 1.x branch as I think there is no need for it anymore (we bump the min PHP deps so much that I don't see any reasons to not upgrade to 2.x).

OskarStark: Thats true, but I talked to Fabien, that we start with `3.x`, will add support for the other branches later on...

.github/workflows/ci.yml
OskarStark: I switched to ubuntu-18.04, thanks, but does this really matter, this is a dockerized way (and therefor independent on the platform itself) of setting up PHP, or am I missing something? 🤔

.github/workflows/ci.yml
fabpot: I think I said 2.x and 3.x :) I've just checked, that what I said ;)

.github/workflows/ci.yml
OskarStark: I use this variable above on `continue-on-error`, `fail-fast: true` is globally on (see line 9)

.github/workflows/ci.yml
stof: nope. the `setup-php` action is about setting up PHP in your action runner, and the supported versions depend on the platform: https://github.com/shivammathur/setup-php#cloud-osplatform-support

As such, they follow the official environment on actions, which also drops old versions in new platforms.

OskarStark:
@stof thanks for the review, can you please help me out and check the failing builds here?
https://github.com/OskarStark/Twig/pull/1

I don't know the difference between travis and my setup, I tried to install `intl` extension, but this seems to be not needed, as `symfony/intl` is required by `extra/intl-extra` 🤔 

Thanks

.github/workflows/ci.yml
OskarStark: Thank you, I wasn't aware of 👍

OskarStark:
The latest builds on Travis on `master` are [green](https://travis-ci.com/github/twigphp/Twig/jobs/410958289)

.github/workflows/ci.yml
shivammathur: @stof setup-php supports currently `PHP 5.3` to `PHP 8.1` (currently master).
If the version is pre-installed, it just switches to it, otherwise it installs the PHP version.
https://github.com/shivammathur/setup-php#tada-php-support

So after `ubuntu-latest` changes to `ubuntu-20.04` if you are using any version below PHP 7.4 on `ubuntu-latest`, `setup-php` will install it on the runner. The workflow will not fail, it will just take a bit longer for the version to be installed.

.github/workflows/ci.yml
stof: IMO, we should not have `fail-fast: true` on the matrix. If the PHP 7.2 job fails, I don't want to make it cancel the jobs for other PHP versions. It would hide results about which PHP versions are impacted by a failure.

OskarStark: Will remove it then, but do you agree on executing extension tests only if the tests for the main repo pass?

.github/workflows/ci.yml
stof: indeed, I misread it

stof:
> I don't know the difference between travis and my setup, I tried to install `intl` extension, but this seems to be not needed, as `symfony/intl` is required by `extra/intl-extra` thinking

The `symfony/intl` polyfill always formats using the `en` locale. so tests relying on other locales for their assertions will fail if you run them against the polyfill rather than the extension.
The polyfill is meant to allow writing code supporting intl and then running it without `ext-intl` if you only care about `en`. but if you do care about internationalization, you need to actual extension. Making a full polyfill in `symfony/polyfill` would require bundle the CLDR data, which was considered out of scope as `ext-intl` will provide better performance.

OskarStark:
If I understand it right, I need to install the `intl` extension to fix it, but it didn't help, lets see again...

extra/intl-extra/phpunit.xml.dist
```diff
@@ -9,6 +9,7 @@
          failOnWarning="true"
 >
     <php>
+        <env name="LANG" value="en_US.UTF-8" force="true" />
```
OskarStark: With this everyone has the same result on local machine 👍

extra/intl-extra/tests/Fixtures/format_date.test
```diff
@@ -15,11 +15,11 @@
...
 --EXPECT--
 Aug 7, 2019, 11:39:12 PM
-7 août 2019 23:39:12
+7 août 2019 à 23:39:12
 23:39
 07/08/2019
-mercredi 7 août 2019 23:39:12 UTC
-11 oclock PM, GMT
+mercredi 7 août 2019 à 23:39:12 heure d’été d’Europe centrale
+11 oclock PM, Central European Summer Time
```
OskarStark: IMHO this is just based on "newer" language packs. Maybe this is not failing on travis, because the use an older build of the distro or language pack 🤔

.github/workflows/ci.yml
```diff
@@ -0,0 +1,140 @@
...
+
+            - name: "Composer install"
+              working-directory: ${{ matrix.extension}}
+              run: composer install
+
+            - working-directory: ${{ matrix.extension}}
+              run: composer show -p
+
+            - working-directory: ${{ matrix.extension}}
+              run: printenv
```
OskarStark: * [x] remove this before merging

extra/intl-extra/tests/Fixtures/format_date.test
```diff
@@ -15,11 +15,11 @@
...
 --EXPECT--
 Aug 7, 2019, 11:39:12 PM
-7 août 2019 23:39:12
+7 août 2019 à 23:39:12
 23:39
 07/08/2019
-mercredi 7 août 2019 23:39:12 UTC
-11 oclock PM, GMT
+mercredi 7 août 2019 à 23:39:12 Temps universel coordonné
+11 oclock PM, Coordinated Universal Time
```
OskarStark: IMHO this is just based on "newer" language packs. Maybe this is not failing on travis, because the use an older build of the distro or language pack 🤔

.github/workflows/ci.yml
```diff
@@ -0,0 +1,140 @@
...
+
+            - name: "Set composer cache directory"
+              id: composer-cache
+              run: echo "::set-output name=dir::$(composer config cache-files-dir)"
+
+            - name: "Cache composer"
+              uses: actions/cache@v2.1.2
+              with:
+                  path: ${{ steps.composer-cache.outputs.dir }}
+                  key: ${{ runner.os }}-${{ matrix.php-version }}-composer-${{ hashFiles('composer.json') }}
```
jderusse: ` ${{ runner.os }}-${{ matrix.php-version }}-${{ matrix.extension }}-`

.github/workflows/ci.yml
```diff
@@ -0,0 +1,140 @@
...
+                    - '7.3'
+                    - '7.4'
+                    - '8.0'
+                extension:
+                    - 'extra/cssinliner-extra'
+                    - 'extra/html-extra'
+                    - 'extra/inky-extra'
+                    - 'extra/intl-extra'
+                    - 'extra/markdown-extra'
+                    - 'extra/string-extra'
```
jderusse: That's a lot of buids (6x4 = 24 jobs). Wouldn't it make sens to run all the extensions in the same job? (like we do for Symfony?)

CF what we do with @nyholm in async-aws (see "run test" step in https://github.com/async-aws/aws/pull/835/checks?check_run_id=1331289562)

Nyholm: But maybe we could merge this now, and improve it later?

.github/workflows/ci.yml
OskarStark: I think this does not work with a slash inside, does it? Also caching composer.json does not make sense maybe?

Referring to: https://github.com/symfony/maker-bundle/pull/719#pullrequestreview-520537815

.github/workflows/ci.yml
OskarStark: Yes but it's quite convenient to see exactly what is failing. And the duration is ok, also the workflow file is much softer and less copy paste

extra/intl-extra/tests/Fixtures/format_date.test
OskarStark: @jderusse what do you think?

fabpot:
Thank you @OskarStark.

Issue #3416, opened 2020-10-28
Feature idea: Recursive loop


kostislav:
# Recursive loops
## Why
While it's currently possible to loop through recursive tree-like structures in Twig, it requires creating a macro, which is slightly inconvenient and hard to get right without copy-pasting. This feature would add loop recursion as a built-in feature of Twig.

## Syntax
There are several ways the syntax could look, each with pros and cons. Here are two of them, I'm sure there are more:

### recurse keyword
```twig
<ul>
  {% for item in tree %}
    <li>
      {{ item.title }}
      {% if item.children is not empty %}
        <ul>
          {% recurse item.children }}
        </ul>
      {% endif %}
    </li>
  {% endfor %]
</ul>
```
Pros: Intuitive
Cons: Requires new keyword

### recurse function in the loop variable
```twig
<ul>
  {% for item in tree %}
    <li>
      {{ item.title }}
      {% if item.children is not empty %}
        <ul>
          {% loop.recurse(item.children) }}
        </ul>
      {% endif %}
    </li>
  {% endfor %]
</ul>
```
Pros: No new keywords
Cons: Seems like a mis-use of the loop variable; might make loops that are using the loop variable slightly less efficient unless the recursion can be optimized away if it's not used

## Implementation
An immediately invoked anonymous function that calls itself recursively, probably as an entry in the context array
Both of the previous examples would render as something along these lines:
```php
$context['loop_function'] = function($items) use (&$context) {
    foreach ($items as $item) {
        echo '<li>';
        echo $item['title'];
        if (array_key_exists('children', $item) && !empty($item['children'])) {
            echo '<ul>';
            $context['loop_function']($item['children']);
            echo '</ul>';
        }
        echo '</li>';
    }
};
echo '<ul>';
$context['loop_function']($context['tree']);
echo '</ul>';
```

## Who
I'll be happy to work on a PR if you think this feature makes sense.

pdureau:
Is it similar to what Jinja is doing?

```
<ul class="sitemap">
{%- for item in sitemap recursive %}
    <li><a href="{{ item.href|e }}">{{ item.title }}</a>
    {%- if item.children -%}
        <ul class="submenu">{{ loop(item.children) }}</ul>
    {%- endif %}</li>
{%- endfor %}
</ul>
```

https://jinja.palletsprojects.com/en/3.1.x/templates/#for

If yes, let's try to stick with syntax parity.

fabpot:
See #4153

PR #3421, opened 2020-10-30
Add the ability to search attributes deeply.

FabianoLothor:
The new feature was set as as default `true`, this could cause problems if someone is trying to search the `index` `"foo.bar"` in an array for example.

#3416

### DIFF
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -1354,9 +1354,9 @@ function twig_array_batch($items, $size, $fill = null, $preserveKeys = true)
 
     return $result;
 }
-
+    
 /**
- * Returns the attribute value for a given array/object.
+ * Returns the deeper attribute value for a given array/object.
  *
  * @param mixed  $object            The object or array from where to get the item
  * @param mixed  $item              The item to get from the array or object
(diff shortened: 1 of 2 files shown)


FabianoLothor:
#3416

src/Extension/CoreExtension.php
```diff
@@ -1374,6 +1374,25 @@ function twig_array_batch($items, $size, $fill = null, $preserveKeys = true)
  */
 function twig_get_attribute(Environment $env, Source $source, $object, $item, array $arguments = [], $type = /* Template::ANY_CALL */ 'any', $isDefinedTest = false, $ignoreStrictCheck = false, $sandboxed = false, int $lineno = -1)
 {
+    // deep attribute
+    if(\is_string($item)) {
+        list($item, $deepItem) = \array_merge(\explode("->", $item, 2), [null]);
```
stof, line 1379: Detecting the deep case based on a `->` delimiter in the string passed to `twig_get_attribute` is a BC break, because it forbids using that in the name of an attribute (and array keys allow *any* string so you won't be able to find a safe magic delimiter)

stof (changes requested):

stof:
If such deep lookup is wanted, it should be a separate function IMO, which takes an array of strings as argument (and so does not rely on a reserved delimiter).
And having it as a separate function has another advantage: `twig_get_attribute` is the most performance sensitive part of Twig (which is why Twig 1.x even went as far as shipping a C implementation of that function). Adding a `explode` in all such calls to detect that it is not a deep access will likely kill performances.

fabpot:
I agree with @stof. In any case, I don't think this is something we want in core. As it can be implemented in "userland", let's close here. Thank you for proposing.

PR #3425, opened 2020-11-03
Fix `odd` not working for negative numbers

### LINKED ISSUE #3424: `is odd` not working for negative numbers
When you do the `is odd` check with negative numbers it returns false. This is caused because the check in OddTest.php is: ` % 2 == 1` and php returns `-1` if you run `-1 % 2`.
```
{% for i in [-1,0,1,2] %}
{% if i is odd %}
true
{% else %}
false
{% endif %}
{% endfor %}

false
false
true
false
```

Twigfiddle: https://twigfiddle.com/t169i5

harrybournis:
Fixes https://github.com/twigphp/Twig/issues/3424

### DIFF
--- a/doc/advanced.rst
+++ b/doc/advanced.rst
@@ -359,7 +359,7 @@ This is used by many of the tests built into Twig::
             $compiler
                 ->raw('(')
                 ->subcompile($this->getNode('node'))
-                ->raw(' % 2 == 1')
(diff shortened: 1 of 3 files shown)


stof (approved):

jderusse (approved):

fabpot:
Thank you @harrybournis.

PR #3427, opened 2020-11-04
Enhancement: Use GithubActions instead of Travis for tests

OskarStark:
You can see the first run in https://github.com/OskarStark/Twig/pull/2


fabpot:
Thank you @OskarStark.

PR #3430, opened 2020-11-05
Enhancement: Introduce DOCtor-RST config + Sphinx build + GithubActions

OskarStark:


doc/templates.rst
```diff
@@ -874,7 +881,7 @@ Twig can be extended. If you want to create your own extensions, read the
 :ref:`Creating an Extension <creating_extensions>` chapter.
 
 .. _`Twig bundle`:                https://github.com/Anomareh/PHP-Twig.tmbundle
-.. _`Jinja syntax plugin`:        http://jinja.pocoo.org/docs/integration/#vim
+.. _`twig syntax plugin`:        http://twig.pocoo.org/docs/integration/#vim
```
stof: this link does not work

stof (changes requested):

doc/templates.rst
OskarStark: thanks

doc/filters/map.rst
```diff
@@ -2,7 +2,8 @@
 =======
 
 .. versionadded:: 1.41
-    The ``map`` filter was added in Twig 1.41 and 2.10.
+
+    The ``map`` filter was added in Twig 1.41.
```
stof, line 6: the removal of `and 2.10` looks wrong to me. This was done on purpose because 1.x and 2.x were developed in parallel (1.x was not in bugfix-only mode after the 2.0 release). So a feature added in 1.41 is not available in 2.0 even though the version number is higher (and having to switch version of the doc to discover that would not be done by devs which don't know about this specific case).

stof: this is not the only affected place in the diff.

fabpot: I agree with @stof, these changes should be reverted

doc/filters/nl2br.rst
```diff
@@ -2,16 +2,17 @@
...
 The ``nl2br`` filter inserts HTML line breaks before all newlines in a string:
 
-.. code-block:: twig
+.. code-block:: html+twig
 
     {{ "I like Twig.\nYou will like it too."|nl2br }}
     {# outputs
 
-        I like Twig.<br />
+        I like Twig.<br/>
```
fabpot: I would even remove the / (`<br>`), should be applied everywhere if done.

fabpot (commented):
LGTM apart from the version removals.

doc/filters/map.rst
OskarStark: To me it is fine to develop in parallel, but in this case the feature needs a versionadded directive in `2.x` and `1.x`, otherwise someone using `2.x` and browsing the `2.x` doc would not know when this method was introduced 🤔 

To fix this problem, I would remove the mention of `1.4.1` in `2.x`:
CleanShot 2020-11-05 at 11 58 36

As this was developed in parallel we have two new introduced methods, one per version

doc/filters/nl2br.rst
stof: I wouldn't do it. this must render what `|nl2br` renders.

fabpot: Right, so this should be reverted

doc/filters/map.rst
fabpot: ok, fair enough

stof: it should indeed have a version added in 2.x too mentioning 2.10. But the message in 1.x should keep mentioning both 1.41 and 2.10 IMO.

stof: btw, maybe the message in 2.x should also mention both in the details (so that people don't think that this feature is entirely unavailable in 1.x), changing only the main version (or maybe we should support passing multiple version to the directive directly)

doc/filters/nl2br.rst
OskarStark: Done

fabpot:
Thank you @OskarStark.

PR #3433, opened 2020-11-05
Fix: Typo

OskarStark:
The DOCtor-RST build will be fixed in #3434


fabpot:
Thank you @OskarStark.

PR #3437, opened 2020-11-06
Add return type for getGlobals() to documentation. The return type is defined in the GlobalsInterface

maartendekeizer:


stof (approved):

fabpot:
Good catch, thanks @maartendekeizer.

PR #3439, opened 2020-11-09
Fix include documentation - change filter tag to apply

### LINKED ISSUE #3438: Documentation referencing removed filter tag
Documentation in [doc/tags/include.rst](https://github.com/twigphp/Twig/blob/3.x/doc/tags/include.rst) still references removed `{% filter` tag.

kick-the-bucket:
Fixes #3438

### DIFF
--- a/doc/tags/include.rst
+++ b/doc/tags/include.rst
@@ -30,10 +30,10 @@ of that file:
           {# vs #}
           {% set content = include('template.html') %}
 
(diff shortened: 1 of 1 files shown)


stof:
I think this should even be merged into lower branches, where `{% filter %}` is deprecated.

kick-the-bucket:
Opened a new PR #3440 for 2.x

PR #3441, opened 2020-11-11
Remove is_int since typehint exists

ybenhssaien:
Since `$optimizers` param is always `int` no needs to check `is_int`


fabpot:
Thank you @ybenhssaien.

PR #3444, opened 2020-11-20
Update PHPDoc

JBlond:


extra/html-extra/src/HtmlExtension.php
```diff
@@ -44,6 +44,9 @@ public function getFunctions(): array
      * Length validation is not performed on purpose, validation should
      * be done before calling this filter.
      *
+     * @param string $data
+     * @param string|null $mime
+     * @param array $parameters
```
stof, line 49: this brings no value as it exactly duplicates the PHP typehints (and the same is true for most phpdoc you are adding)

JBlond: Well for most IDE this improves the code completion

stof: most IDEs (and all _good_ IDEs) are able to read the native typehints too

PR #3447, opened 2020-11-29
Adding link to RFC

ThomasLandauer:
Reason for rewording the `twig/extra-bundle` part: It was not clear what "use" means here:

> Then, use the twig/extra-bundle on Symfony projects

### DIFF
--- a/doc/filters/data_uri.rst
+++ b/doc/filters/data_uri.rst
@@ -5,8 +5,8 @@
 
     The ``data_uri`` filter was added in Twig 2.12.
 
-The ``data_uri`` filter generates a URL using the data scheme as defined in RFC
-2397:
+The ``data_uri`` filter generates a URL using the data scheme as defined in
+`RFC 2397`_:
 
 .. code-block:: html+twig
 
@@ -32,8 +32,8 @@ The ``data_uri`` filter generates a URL using the data scheme as defined in RFC
 
         $ composer require twig/html-extra
 
-    Then, use the ``twig/extra-bundle`` on Symfony projects or add the extension
-    explicitly on the Twig environment::
+    On Symfony projects, install the ``twig/extra-bundle``: ``composer require twig/extra-bundle``
+    Otherwise, add the extension explicitly on the Twig environment::
 
         use Twig\Extra\Html\HtmlExtension;
 
@@ -42,7 +42,7 @@ The ``data_uri`` filter generates a URL using the data scheme as defined in RFC
 
 .. note::
 
-    The filter does not perform any length validation on purpose (limits depends
+    The filter does not perform any length validation on purpose (limit depends
     on the usage context), validation should be done before calling this filter.
 
 Arguments
@@ -50,3 +50,5 @@ Arguments
 
 * ``mime``: The mime type
 * ``parameters``: An array of parameters
+
+.. _RFC 2397: https://tools.ietf.org/html/rfc2397


doc/filters/data_uri.rst
```diff
@@ -25,15 +25,15 @@ The ``data_uri`` filter generates a URL using the data scheme as defined in RFC
 
 .. note::
 
-    The ``data_uri`` filter is part of the ``HtmlExtension`` which is not
-    installed by default. Install it first:
+    The ``data_uri`` filter is part of the [``HtmlExtension``](https://github.com/twigphp/html-extra)
+    which is not installed by default. Install it first:
```
fabpot: I would not add links to the extra bundle Github repository as instructions are just below.

doc/filters/data_uri.rst
```diff
@@ -25,15 +25,15 @@ The ``data_uri`` filter generates a URL using the data scheme as defined in RFC
...
 
     .. code-block:: bash
 
         $ composer require twig/html-extra
 
-    Then, use the ``twig/extra-bundle`` on Symfony projects or add the extension
-    explicitly on the Twig environment::
+    On Symfony projects, install the [``twig/extra-bundle``](https://github.com/twigphp/twig-extra-bundle).
+    Otherwise, add the extension explicitly on the Twig environment::
```
fabpot, line 36: Same here

fabpot (changes requested):

doc/filters/data_uri.rst
ThomasLandauer: Well, below are instructions on how to *install* it, but no explanation what this thing actually *is* ;-)

So do you still want me to remove it? It's certainly not worth an elaborate discussion... :-)

fabpot: Yes, I know, I prefer to remove it.

doc/filters/data_uri.rst
ThomasLandauer: Well, just noticing that it's not really the same here, since there are no installation instructions for Symfony users ;-) The link would cover that. What do you suggest?

sstok: These are rst docs, the used syntax (markdown) doesn't work 👍🏼 

https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#external-links

ThomasLandauer: Thanks :-)

doc/filters/data_uri.rst
```diff
@@ -5,8 +5,8 @@
 
     The ``data_uri`` filter was added in Twig 2.12.
 
-The ``data_uri`` filter generates a URL using the data scheme as defined in RFC
-2397:
+The ``data_uri`` filter generates a URL using the data scheme as defined in
+[RFC 2397](https://tools.ietf.org/html/rfc2397:
```
fabpot: Please use ReST URLs (this one is Markdown and broken).

fabpot (changes requested):

doc/filters/data_uri.rst
```diff
@@ -32,8 +32,8 @@ The ``data_uri`` filter generates a URL using the data scheme as defined in RFC
 
         $ composer require twig/html-extra
 
-    Then, use the ``twig/extra-bundle`` on Symfony projects or add the extension
-    explicitly on the Twig environment::
+    On Symfony projects, install the `twig/extra-bundle`_.
```
fabpot: I would remove the link to Github's README and give the command here (as done above for the html-extra package)

fabpot (changes requested):

doc/filters/data_uri.rst
ThomasLandauer: Like this?
This line should probably go onto *all* extra-bundle filters. Do you want to create an include or something? If no, I'd commit it manually - is there a list of all those "extra" filters somewhere?

fabpot:
Thank you @ThomasLandauer.

Issue #3445, opened 2020-11-26
Cannot redeclare __TwigTemplate_xyz::checkSecurity if adding both SandboxNodeVisitor and SandboxExtension


mbessolov:
I encountered this issue when upgrading from v2.13.1 to 2.14.0

```php
<?php
require __DIR__ . '/vendor/autoload.php';
$twig = new \Twig\Environment(new \Twig\Loader\ArrayLoader(['template' => 'content']));

$twig->addExtension(new \Twig\Extension\SandboxExtension(new \Twig\Sandbox\SecurityPolicy()));
$twig->addNodeVisitor(new \Twig\NodeVisitor\SandboxNodeVisitor());

echo $twig->render('template');
```
produces
```
PHP Fatal error:  Cannot redeclare
__TwigTemplate_4d22a90d60e84fc10eff1f053099f13449bfe617a048aee5fcfb6c9154c218ca::checkSecurity()
in /home/michael/tmp/twig-bug/vendor/twig/twig/src/Environment.php(418) : eval()'d code on line 87
```
The compiled template looks like this (non-relevant lines stripped for brevity):

```php
class __TwigTemplate_4d22a90d60e84fc10eff1f053099f13449bfe617a048aee5fcfb6c9154c218ca extends \Twig\Template
{
    /* ... */
    public function __construct(Environment $env)
    {
        /* ... */
        $this->sandbox = $this->env->getExtension('\Twig\Extension\SandboxExtension');
        $this->checkSecurity();
        $this->sandbox = $this->env->getExtension('\Twig\Extension\SandboxExtension');
        $this->checkSecurity();
    }

    /* ... */

    public function checkSecurity()
    {
        /* ... */
    }
    
    public function checkSecurity()
    {
        /* ... */
    }
}
```

On the surface it looks like a result of #3413, but after digging a bit into it I think that #3413 simply revealed the side effects of  somewhat special treatment given to parsers and visitors in ExtensionSet initExtension - https://github.com/twigphp/Twig/blob/v2.14.1/src/ExtensionSet.php#L458

Both visitors and parsers are prone to such duplication, as they are not de-duped when initializing extensions and the staging.

stof:
you should not be adding a SandboxNodeVisitor yourselves. The SandboxExtension is already managing the SandboxNodeVisitor, so you end up with running the SandboxNodeVisitor twice.

mbessolov:
@stof the code above is just an over simplified example to reproduce the problem.
[...]

PR #3448, opened 2020-11-30
Mark built-in node visitors as @internal

fabpot:
refs #3445


stof:
@fabpot maybe we should do the same for token parsers ?

fabpot:
@stof done

stof (approved):

PR #3451, opened 2020-12-01
Add checks for existing callable at compilation

### LINKED ISSUE #3450: Twig assumes static callable if it does not exist when registering a filter
If I register a class method as a filter callable and it does not exist it is treated as being static. Afterwards if I define the callable as an instance method it is called statically. 

On php 8 this is a fatal error in 7.4 it is called nonetheless and goes through if the method does not use $this.

After the compiled template is removed everything works as expected.

[bug.zip](https://github.com/twigphp/Twig/files/5617282/bug.zip)

ghost:
Fixes #3450 

Currently Twig does not check if the callable passed to CallExpression is actually a callable.
This pull request adds does checks.

If it is a function it check if it exists. If it is an array indicating a class method, then it also checks for the presence of __call and __callStatic methods.
If the callable does not exist a \LogicException is thrown. (I don't know if this is the appropriate Exception).

I think Twig should check for the existence of callables if it can otherwise it can be quite cumbersome to debug when using caching, since it is possible a wrong path will be compiled. (This happened to me.)

Some unrelated Function/Filter tests broke because they were passing undefined callables so I fixed them. However, the intent was clear to pass an existing callable.

### DIFF
--- a/src/Node/Expression/CallExpression.php
+++ b/src/Node/Expression/CallExpression.php
@@ -27,6 +27,9 @@ abstract class CallExpression extends AbstractExpression
         $closingParenthesis = false;
         $isArray = false;
         if (\is_string($callable) && false === strpos($callable, '::')) {
+            if (!is_callable($callable)) {
+                throw new \LogicException("Function '$callable' does not exist!");
+            }
             $compiler->raw($callable);
         } else {
             list($r, $callable) = $this->reflectCallable($callable);
@@ -293,11 +296,16 @@ abstract class CallExpression extends AbstractExpression
         }
 
         if (\is_array($callable)) {
-            if (!method_exists($callable[0], $callable[1])) {
-                // __call()
+            list($class, $method) = $callable;
+            if (!method_exists($class, $method)) {
+                if (!method_exists($class, '__call') && !method_exists($class, '__callStatic')) {
+                    $class = is_object($class) ? get_class($class) : $class;
+                    throw new \LogicException("Class '$class' does not define the method '$method' or the magic methods __call or __callStatic.");
+                }
+                // __call() or __callStatic
                 return [null, []];
             }
-            $r = new \ReflectionMethod($callable[0], $callable[1]);
+            $r = new \ReflectionMethod($class, $method);
         } elseif (\is_object($callable) && !$callable instanceof \Closure) {
             $r = new \ReflectionObject($callable);
             $r = $r->getMethod('__invoke');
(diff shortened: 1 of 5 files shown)


src/Node/Expression/CallExpression.php
```diff
@@ -293,11 +296,16 @@ private function reflectCallable($callable)
         }
 
         if (\is_array($callable)) {
-            if (!method_exists($callable[0], $callable[1])) {
-                // __call()
+            list($class, $method) = $callable;
```
stof: The naming `$class` is confusing here, as this is really `$objectOrClass`

ghost: Is it ok that I destructed the callable?

src/Node/Expression/CallExpression.php
```diff
@@ -306,12 +314,18 @@ private function reflectCallable($callable)
             $class = substr($callable, 0, $pos);
             $method = substr($callable, $pos + 2);
             if (!method_exists($class, $method)) {
+                if (!method_exists($class, '__callStatic')) {
```
ghost, line 317: Should this also check for __call? We allow [FQCN, 'nonStaticMethod'] for instance methods even though that is not a legal callable. Is 'FQCN::nonStaticMethod' also legal in context of Twig?

ghost: @stof What do you think?

ghost:
@fabpot do you have an opinion on this pull request?

ghost:
Sorry for pinging again. Is there anything else I have to do to get this PR merged? Or are these checks not desired? @stof  @fabpot

ghost:
@stof @fabpot  do I have to do anything else?

nlemoine:
I'm having a similar issue here. 

```php
class SomeExtension extends AbstractExtension
{
    private array $filters = [
         'filter1',
         'filter2',
         'filter3',
         // and so on...
    ];

    public function getFilters(): array
    {
        $filters = [];
        foreach ($this->filters as $filter) {
            $filters[] = new TwigFilter(
                $filter,
                // ["SomeRuntime::$filter"], // Don't work
                // [SomeRuntime::class, $filter], // Don't work
                ['is_variadic' => true]
            );
        }

        return $filters;
    }
}

class SomeRuntime implements RuntimeExtensionInterface
{
    private Service $service;

    public function __construct(Service $service)
    {
        $this->service = $service;
    }

    public function __call($name, array $arguments = [])
    {
        $object = $this->service->create($arguments[0]);
        return $object->$name(...$arguments)
    }
}

```

Could you give this PR/issue some a second thought?

PR #3452, opened 2020-12-02
[intl-extra] Add test showing currency behaviour

franmomu:
This is just to show some weird behaviour that I have no idea how to fix it.

What I've found so far is that the problem seems to occur when the currency symbol `¤` is involved in the text attributes:

https://github.com/twigphp/Twig/blob/e3bca056f1331cf017893c2248d1d4951c165162/extra/intl-extra/src/IntlExtension.php#L411-L413

When `numberFormatterProto` is created with `en` locale these are its properties before (in red) and after (in green) going through the `IntlExtension`:
```diff
2,3c2,3
<   locale: "en"
<   pattern: "¤#,##0.0"
---
>   locale: "en_US"
>   pattern: "'¤'#,##0.0;'-¤'#,##0.0"
18c18
<     FORMAT_WIDTH: false
---
>     FORMAT_WIDTH: 0
```
In case of `fr` locale are these:

```diff
2,3c2,3
<   locale: "fr"
<   pattern: "#,##0.0 ¤"
---
>   locale: "en_US"
>   pattern: "#,##0.0 '¤';'-'#,##0.0 '¤'"
18c18
<     FORMAT_WIDTH: false
---
>     FORMAT_WIDTH: 0
```

Looks like it escapes the symbol and does not change it to the local currency.

And this is the PHPUnit output:

```
Testing Twig\Extra\Intl\Tests\IntlExtensionTest
FF                                                                  2 / 2 (100%)

Time: 121 ms, Memory: 6.00 MB

There were 2 failures:

1) Twig\Extra\Intl\Tests\IntlExtensionTest::testFormatterProtoWithCurrency with data set #0 ('en', '€12.3')
--- Expected
+++ Actual
@@ @@
-'€12.3'
+'¤12.3'

/intl-extra/tests/IntlExtensionTest.php:37

2) Twig\Extra\Intl\Tests\IntlExtensionTest::testFormatterProtoWithCurrency with data set #1 ('fr', '12,3 €')
--- Expected
+++ Actual
@@ @@
-'12,3 €'
+'12,3 ¤'

/intl-extra/tests/IntlExtensionTest.php:37

FAILURES!
Tests: 2, Assertions: 2, Failures: 2.
```

### DIFF
--- a/extra/intl-extra/src/IntlExtension.php
+++ b/extra/intl-extra/src/IntlExtension.php
@@ -409,11 +409,15 @@ final class IntlExtension extends AbstractExtension
         }
 
         foreach ($textAttrs as $name => $value) {
-            $this->numberFormatters[$hash]->setTextAttribute(self::NUMBER_TEXT_ATTRIBUTES[$name], $value);
+            if (false === strpos($value, '¤')) {
+                $this->numberFormatters[$hash]->setTextAttribute(self::NUMBER_TEXT_ATTRIBUTES[$name], $value);
+            }
         }
 
         foreach ($symbols as $name => $value) {
-            $this->numberFormatters[$hash]->setSymbol(self::NUMBER_SYMBOLS[$name], $value);
+            if (false === strpos($value, '¤')) {
+                $this->numberFormatters[$hash]->setSymbol(self::NUMBER_SYMBOLS[$name], $value);
+            }
         }
 
         return $this->numberFormatters[$hash];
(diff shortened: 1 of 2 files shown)


franmomu:
Just to confirm that is the problem is setting `¤` in text attributes and symbols ~, the weird thing is that this code worked to me (php 7.4), I guess there is some different configuration.~

fabpot:
Closing as I doubt the issue is in Twig.

PR #3459, opened 2020-12-31
Add a cache tag

fabpot:
Add a "cache" tag based on the Symfony cache component.

Replaces https://github.com/twigphp/twig-cache-extension and https://github.com/asm89/twig-cache-extension

### DIFF
--- a/doc/tags/cache.rst
+++ b/doc/tags/cache.rst
@@ -1,0 +1,112 @@
+``cache``
+=========
+
+.. versionadded:: 3.2
+
+    The ``cache`` tag was added in Twig 3.2.
+
+The ``cache`` tag tells Twig to cache a template fragment:
+
+.. code-block:: twig
+
+    {% cache "cache key" %}
+        Cached forever (depending on the cache implementation)
+    {% endcache %}
+
+If you want to expire the cache after a certain amount of time, specify an
+expiration in seconds via the ``ttl()`` modifier:
+
+.. code-block:: twig
+
+    {% cache "cache key" ttl(300) %}
+        Cached for 300 seconds
+    {% endcache %}
+
+The cache key can be any string that does not use the following reserved
+characters ``{}()/\@:``; a good practice is to embed some useful information in
+the key that allows the cache to automatically expire when it must be
+refreshed:
+
+* Give each cache a unique name and namespace it like your templates;
+
+* Embed an integer that you increment whenever the template code changes (to
+  automatically invalidate all current caches);
+
+* Embed a unique key that is updated whenever the variables used in the
+  template code changes.
+
+For instance, I would use ``{% cache "blog_post;v1;" ~ post.id ~ ";" ~
+post.updated_at %}`` to cache a blog content template fragment where
+``blog_post`` describes the template fragment, ``v1`` represents the first
+version of the template code, ``post.id`` represent the id of the blog post,
+and ``post.updated_at`` returns a timestamp that represents the time where the
+blog post was last modified.
+
+Using such a strategy for naming cache keys allows to avoid using a ``ttl``.
+It's like using a "validation" strategy instead of an "expiration" strategy as
+we do for HTTP caches.
+
+If your cache implementation supports tags, you can also tag your cache items:
+
+.. code-block:: twig
+
+    {% cache "cache key" tag('blog') %}
+        Some code
+    {% endcache %}
+
+    {% cache "cache key" tag(['cms', 'blog']) %}
+        Some code
+    {% endcache %}
+
+The ``cache`` tag creates a new "scope" for variables, meaning that the changes
+are local to the template fragment:
+
+.. code-block:: twig
+
+    {% set count = 1 %}
+
+    {% cache "cache key" tag('blog') %}
+        {# Won't affect the value of count outside of the cache tag #}
+        {% set count = 2 %}
+        Some code
+    {% endcache %}
+
+    {# Displays 1 #}
+    {{ count }}
+
+.. note::
+
+    The ``cache`` tag is part of the ``CacheExtension`` which is not installed
+    by default. Install it first:
+
+    .. code-block:: bash
+
+        $ composer require twig/cache-extra
+
+    On Symfony projects, you can automatically enable it by installing the
+    ``twig/extra-bundle``:
+
+    .. code-block:: bash
+
+        $ composer require twig/extra-bundle
+
+    Or add the extension explicitly on the Twig environment::
+
+        use Twig\Extra\Cache\CacheExtension;
+
+        $twig = new \Twig\Environment(...);
+        $twig->addExtension(new CacheExtension());
+
+    If you are not using Symfony, you must also register the extension runtime::
+
+        use Symfony\Component\Cache\Adapter\FilesystemAdapter;
+        use Twig\Extra\Cache\CacheRuntime;
+        use Twig\RuntimeLoader\RuntimeLoaderInterface;
+
+        $twig->addRuntimeLoader(new class implements RuntimeLoaderInterface {
+            public function load($class) {
+                if (CacheRuntime::class === $class) {
+                    return new CacheRuntime(new FilesystemAdapter());
+                }
+            }
+        });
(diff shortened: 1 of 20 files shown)


doc/tags/cache.rst
```diff
@@ -0,0 +1,99 @@
...
+If you want to expire the cache after a certain amount of time, specify an
+expiration in seconds via the ``ttl()`` modifier:
+
+.. code-block:: twig
+
+    {% cache "cache key" ttl(300) %}
+        Cached for 300 seconds
+    {% endcache %}
+
+The cache key can be any string that does not used the following reserved
```
nicolas-grekas: s/used/use

doc/tags/cache.rst
```diff
@@ -0,0 +1,99 @@
...
+        Some code
+    {% endcache %}
+
+    {% cache "cache key" tag(['cms', 'blog']) %}
+        Some code
+    {% endcache %}
+
+.. tip::
+
+    Twig automatically adds a ``twig`` tag to all caches which allows to clear
```
nicolas-grekas: Really needed ? Wouldn't this be equivalent to clearing the pool where all items are stored?

extra/cache-extra/composer.json
```diff
@@ -0,0 +1,40 @@
...
+        }
+    },
+    "autoload-dev": {
+        "psr-4" : {
+            "Twig\\Extra\\Cache\\Tests\\" : "tests/"
+        }
+    },
+    "extra": {
+        "branch-alias": {
+            "dev-master": "2.12-dev"
```
nicolas-grekas: Version is correct?

extra/cache-extra/src/CacheRuntime.php
```diff
@@ -0,0 +1,29 @@
...
+
+namespace Twig\Extra\Cache;
+
+use Symfony\Contracts\Cache\CacheInterface;
+
+class CacheRuntime
+{
+    private $cache;
+
+    public function __construct(CacheInterface $cache)
```
nicolas-grekas: Don't we have a TagAwareCacheInterface?

extra/cache-extra/LICENSE
```diff
@@ -0,0 +1,19 @@
+Copyright (c) 2021 Fabien Potencier
```
OskarStark, line 1: 🎆🥳

extra/cache-extra/src/CacheRuntime.php
fabpot: When using the default Symfony configuration from the extra bundle, yes. But we accept `CacheInterface` as well as using tags is optional (see the Node, we catch the exception and ignore it if we don't have tag support). Is there a better way?

doc/tags/cache.rst
fabpot: Not if you are using the default `cache.app` service, right? It would remove everything. Here, you can remove only the Twig cache.

nicolas-grekas: Right. But can't we wire a dedicated pool? Tags add some storage overhea this would save.

extra/cache-extra/src/CacheRuntime.php
nicolas-grekas: Then throw an exception when using tags with a non compatible item/pool?

fabpot: Not sure this is a good idea as if a bundle use some caches with tags, I still want people to be able to use a "standard" CacheInterface.

stof:
Should the `cache` tag introduce a new scope ? This would avoid weird behavior where a variable declaration inside the cached version would affect the following rendering on cache miss but not on cache hit (as the declaration would not be executed then).

fabpot:
@stof That's already the case as we execute the template fragment in a closure that `use` the `$context` var which is an array and not an object (test added).

doc/tags/cache.rst
```diff
@@ -0,0 +1,101 @@
...
+
+The cache key can be any string that does not use the following reserved
+characters ``{}()/\@:``; a good practice is to embed some useful information in
+the key that allows the cache to automatically expire when it must be
+refreshed:
+
+* Give each cache a unique name and namespace it like your templates;
+
+* Embed an integer that you increment whenever the template code changes (to
+  automatically invalidate all current caches);
```
nicolas-grekas, line 33: an idea for a future PR: this part could be done automatically by twig - adding a prefix to keys

extra/cache-extra/Node/CacheNode.php
```diff
@@ -0,0 +1,81 @@
...
+    {
+        $compiler
+            ->addDebugInfo($this)
+            ->write('$cached = $this->env->getRuntime(\'Twig\Extra\Cache\CacheRuntime\')->getCache()->get(')
+            ->subcompile($this->getNode('key'))
+            ->raw(", function (\Symfony\Contracts\Cache\ItemInterface \$item) use (\$context) {\n")
+            ->indent()
+            ->write("try {\n")
+            ->indent()
+            ->write("\$item->tag('twig');\n")
```
nicolas-grekas: If we remove this, we can also remove the try/catch, which then means that ppl that use `tag()` with a non-taggaware pool will get an exception, which could be nice for DX. That'd be my preference I think.

extra/twig-extra-bundle/src/Resources/config/cache.php
```diff
@@ -0,0 +1,31 @@
...
+        ->set('twig.extension.cache', CacheExtension::class)
+            ->tag('twig.extension')
+
+        ->set('twig.runtime.cache', CacheRuntime::class)
+            ->args([
+                service('twig.cache.default'),
+            ])
+            ->tag('twig.runtime')
+
+        ->alias('twig.cache.default', TagAwareCacheInterface::class)
```
nicolas-grekas: I can submit a PR to wire this as a dedicated pool after merge if you want.

extra/cache-extra/TokenParser/CacheTokenParser.php
```diff
@@ -0,0 +1,63 @@
...
+            $k = $stream->getCurrent()->getValue();
+            $stream->next();
+            $args = $expressionParser->parseArguments();
+
+            switch ($k) {
+                case 'ttl':
+                    $ttl = $args->getNode(0);
+                    break;
+                case 'tags':
+                    $tags = $args->getNode(0);
```
stof, line 46: should we throw a proper SyntaxError if we either have 0 arguments or multiple arguments ? For now, multiple arguments would be ignored silently (not reporting a mistake of writing `tags('foo', 'bar')` instead of `tags(['foo', 'bar'])`) and no arguments is likely to break with an error that's not clear about what the issue is (reporting a bug in the token parser instead)

stof:
@fabpot that's good, but it is not documented

extra/cache-extra/TokenParser/CacheTokenParser.php
fabpot: TBH, I was too lazy to do it, but as you noticed it, I will do it :)

doc/tags/cache.rst
fabpot: Not really (think of an include call in the fragment).

fabpot:
Thank you @nicolas-grekas.

extra/twig-extra-bundle/src/Resources/config/cache.php
```diff
@@ -0,0 +1,45 @@
...
+        ->set('twig.cache', TagAwareAdapter::class)
+            ->args([
+                service('.twig.cache.inner')
+            ])
+
+        ->set('.twig.cache.inner')
+            ->parent('cache.app')
+            ->tag('cache.pool', ['name' => 'twig.cache'])
+
+        ->alias(TagAwareCacheInterface::class.' $twigCache', 'twig.cache')
```
lyrixx, line 41: Where is defined `$twigCache` ?

nicolas-grekas: It's a named autowiring alias: here, it's a string, not a variable.

lyrixx: Hmm, ok, nevermind, I got it. It's for auto wiring

jkabat:
@fabpot is there any reason why is cache extension dependent on symfony/cache^5.0 ? Would be possible to support ^4.4 as well?

PR #3463, opened 2021-01-05
Twig extra bundle now requires twig 3.2

sylvia-vdv:
Updated requirements of composer.json in twig/extra/twig-extra-bundle: 

Due to the addition of `registerUndefinedTokenParserCallback ` in the file `Twig/extra/twig-extra-bundle/DependencyInjection/Compiler/MissingExtensionSuggestorPass.php`, version 3.2 of twig/twig is also required. 
With older versions of twig/twig installed, an error is thrown: 
```
 Attempted to call an undefined method named "registerUndefinedTokenParserCallback" of class "Twig\Environment".                                         
Did you mean to call e.g. "registerUndefinedFilterCallback" or "registerUndefinedFunctionCallback"?      
```


fabpot:
Thank you @sylvia-vdv.

PR #3472, opened 2021-01-13
Adding link to the SymfonyCasts Twig tutorial

weaverryan:
Hi!

We have (and have had) for a long time, a really nice Twig screencast on SymfonyCasts with interactive coding challenges so people can play with Twig. I thought a link from Twig's docs might be useful :). I was also thinking a link from https://twig.symfony.com/doc/3.x/ might make sense, but I think this isn't modifiable directly via the docs (but tell me if I'm wrong).

Here's the "look":

Screen Shot 2021-01-13 at 1 07 40 PM

Thanks!


fabpot:
Thank you @weaverryan.

PR #3473, opened 2021-01-15
Fix length filter documentation about computation

VincentLanglet:
According to https://github.com/twigphp/Twig/blob/1.x/src/Extension/CoreExtension.php#L1317


fabpot:
Thank you @VincentLanglet.

PR #3476, opened 2021-01-26
remove the C extension

### LINKED ISSUE #3474: Delete C extension code from Twig 1.*
The documentation says that "The C extension is only compatible and useful for **PHP5**." Twig 1.43.0 and above supports only PHP >= 7.1.3. Wouldn't it be a good idea to remove unmaintained code from the repository?

xabbuh:
fixes #3474


nicolas-grekas:
Not sure, looks like a bc break, isn't it ?

xabbuh:
I don't think so as the Twig extension is afaik not compatible with PHP 7.

fabpot:
I kept it for historical reasons, but it's about time to remove it I suppose.

fabpot:
Thank you @xabbuh.

PR #3478, opened 2021-01-29
Add slug filter (follow-up to PR #3386)

EmilePerron:
Hi,

This is a follow-up to PR #3386, in which @seb-jean had done most of the work.

I went ahead and implemented the remaining fixes for the pending comments/reviews, and added some documentation on the filter.

Let me know if anything else should be done before this can be merged, and I will get it done ASAP.

Cheers!

### DIFF
--- a/doc/filters/slug.rst
+++ b/doc/filters/slug.rst
@@ -1,0 +1,37 @@
+``slug``
+========
+
+The ``slug`` filter transforms a given string into another string that
+only includes safe ASCII characters. 
+
+Here is an example:
+
+.. code-block:: twig
+
+    {{ 'Wôrķšƥáçè ~~sèťtïñğš~~'|slug }}
+    Workspace-settings
+
+The default separator between words is a dash (``-``), but you can 
+define a selector of your choice by passing it as an argument:
+
+.. code-block:: twig
+
+    {{ 'Wôrķšƥáçè ~~sèťtïñğš~~'|slug('/') }}
+    Workspace/settings
+
+The slugger automatically detects the language of the original
+string, but you can also specify it explicitly using the second
+argument:
+
+.. code-block:: twig
+
+    {{ '...'|slug('-', 'ko') }}
+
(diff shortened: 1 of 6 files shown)


doc/filters/slug.rst
```diff
@@ -0,0 +1,29 @@
+``slug``
+========
+
+The ``slug`` filter transforms a given string into another string that
+only includes safe ASCII characters. It does so by using the method by
```
fabpot: I would move the second sentence at the end (as most people don't care about how it is implemented).

doc/filters/slug.rst
```diff
@@ -0,0 +1,29 @@
...
+    {{ 'Wôrķšƥáçè ~~sèťtïñğš~~'|slug('/') }}
+    Workspace/settings
+
+The slugger automatically detects the language of the original
+string, but you can also specify it explicitly using the second
+argument:
+
+.. code-block:: twig
+
+    {{ '...'|slug('-', 'ko') }}
```
fabpot: Can you document the argument names (as done with other filters)? This helps use named arguments instead of positional ones.

extra/string-extra/Tests/Fixtures/slug.test
```diff
@@ -0,0 +1,13 @@
...
+{{ 'Wôrķšƥáçè ~~sèťtïñğš~~'|slug }}
+
+{{ 'Wôrķšƥáçè ~~sèťtïñğš~~'|slug('/') }}
+
+--DATA--
+return []
+--EXPECT--
+Workspace-settings
+
+Workspace/settings
```
fabpot: missing EOL.

fabpot (changes requested):
Great! Thank you for taking over. I've done some quick comments.

EmilePerron:
@fabpot Thanks for the speedy response! I just pushed an update to resolve the issues you pointed out in your comment.

fabpot:
Thank you @EmilePerron.

PR #3479, opened 2021-01-29
Add byte string filter

norkunas:


norkunas:
will this be considered for merge at all?

GromNaN:
Hello @norkunas, can you tell little more about your use-case?

Also, tests are very welcome when adding a new feature.

norkunas:
Use case is that I render slack notifications with twig and in certain cases I know that string could be too long so I use truncation with symfony string, but for slack it needs to be truncated by bytes, not by unicode chars.

I know that tests would be welcome, but what's the point of retesting same things that are done in string component? :)

GromNaN:
The test can be  a single statement in the fixture. It will validate the integration and pave the way to add other tests if any bug was be found. 
https://github.com/twigphp/Twig/blob/3.x/extra/string-extra/Tests/Fixtures/string.test

I don't really get the difference between `UnicodeString` and `ByteString`, I think we should improve the documentation on this particular point (I will open an issue).

stof:
@GromNaN ByteString operates at the byte level for string operations (when returning or receiving offsets in the string) while UnicodeString operates at the level of unicode grapheme clusters (the String component also has a CodepointString that operates at the level of unicode codepoints).

stof:
See https://symfony.com/doc/current/components/string.html for the doc about the symfony/string component. It already explains the difference.

nicolas-grekas:
About the described use case, does Slack allow invalid utf8 sequences? Because cutting at bytes boundaries will lead to broken utf8 strings.

Can you please double check and share some links about that?

norkunas:
In 1+ years didn't log any errors about that.

norkunas:
sad ...

PR #3480, opened 2021-02-01
Template Wrapper Interface

rodnaph:
Adds an interface for `TemplateWrapper`, as the public interface for templates, to facilitate mocking.

Currently if mocking `Twig\Environment` the `load` method returns a `TemplateWrapper`, which is `final`. Creating an instance of this object requires mimicking it's internal behaviour, which is not desirable.

Have I missed a more obvious way of doing this? Or is this not intended usage?

Any pointers appreciated, thanks.

### DIFF
--- a/src/Environment.php
+++ b/src/Environment.php
@@ -266,7 +266,7 @@ class Environment
     /**
      * Renders a template.
      *
-     * @param string|TemplateWrapper $name The template name
+     * @param string|TemplateWrapperInterface $name The template name
      *
      * @throws LoaderError  When the template cannot be found
      * @throws SyntaxError  When an error occurred during compilation
@@ -280,7 +280,7 @@ class Environment
     /**
      * Displays a template.
      *
-     * @param string|TemplateWrapper $name The template name
+     * @param string|TemplateWrapperInterface $name The template name
      *
      * @throws LoaderError  When the template cannot be found
      * @throws SyntaxError  When an error occurred during compilation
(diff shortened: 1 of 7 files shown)


stof:
the issue with adding an interface is that this locks this object: we cannot add new methods in an interface outside major versions (while we can add new methods in a class).
What is your use case for mocking `Twig\Environment::load` ?

Btw, a common best practice is to avoid mocking APIs you don't own (as such test would only give you guarantee that your code is correct if your mock reflects the behavior of the original code, which you have no guarantee if you don't have tests using the actual implementation which will start failing of the original behavior changes). And your project definitely does not own the `Twig\Environment` API...

rodnaph:
Hey, thanks for the info. I see the issue, is this a hard-blocker then? Or would another approach like making `TemplateWrapper` non-final be an option?

My use case is that I have a unit test for a service which uses the `Twig\Environment` - and I'd rather not create an actual instance of this class - so I'm trying to verify the behaviour of the interactions really (higher level integration tests then boot the real versions). Not sure how else to handle it...

stof:
@rodnaph as I said, verifying the behavior of the interactions between your code and a mock of `Twig\Environment` will provide absolutely no guarantee that the code is correct. That's because it verifies them against your own idea of the way `Twig\Environment` should be used (as encoded in your mock) rather than the way `Twig\Environment` actually works. This means that such test provides only a false sense of safety (you think your code is tested, but you still don't know whether your code works, and even less whether it will keep working when updating Twig).
The proper way to test code interacting with a third-party library is to actually run the library (which makes it an integration test rather than a pure unit test).

rodnaph:
Ok no problem, sounds like there's no future here then.

Thanks anyway!

PR #3481, opened 2021-02-02
Add some return types

nicolas-grekas:
These return types preemptively tell consumers that Twig 4 / PHP 9 might add real return types here.

For `Profile`, we can add the return types right away because the class is `@final`.

Spotted by DebugClassLoader comparing return types of internal PHP base types / magic methods.


fabpot:
Thank you @nicolas-grekas.

PR #3484, opened 2021-02-06
Allowing compat with Twig 2 by conditionally calling method

weaverryan:
This is an annoying, edge-case issue. The original fix was #3463. The problem, discovered by MakerBundle's tests (I'm not sure why it didn't pop up sooner) is if you're using PHP 7.1... for example with Symfony 4.4 LTS. In that situation, if you run:

```
composer require symfony/twig-pack
```

You will get:

* `twig/twig` v2 (since v3 required PHP 7.2)
* `twig/twig-extra-bundle` 3.2.0 (3.2.1 doesn't allow `twig/twig` v2, but 3.2.0 does).

The result is the container explodes:

```
!!  
!!  [critical] Uncaught Error: Call to undefined method Twig\Environment::registerUndefinedTokenParserCallback()
!!  
!!  
!!  In srcApp_KernelDevDebugContainer.php line 1176:
!!                                                                                 
!!    Attempted to call an undefined method named "registerUndefinedTokenParserCa  
!!    llback" of class "Twig\Environment".                                         
!!    Did you mean to call e.g. "registerUndefinedFilterCallback" or "registerUnd  
!!    efinedFunctionCallback"?   
```

Is this worth fixing? If not, we'll need to work around in the MakerBundle tests.

Thanks!

### DIFF
--- a/extra/twig-extra-bundle/DependencyInjection/Compiler/MissingExtensionSuggestorPass.php
+++ b/extra/twig-extra-bundle/DependencyInjection/Compiler/MissingExtensionSuggestorPass.php
@@ -14,17 +14,23 @@ namespace Twig\Extra\TwigExtraBundle\DependencyInjection\Compiler;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Reference;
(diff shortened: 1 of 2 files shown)


jderusse (approved):

stof (approved):

fabpot:
Thank you @weaverryan.

Issue #3477, opened 2021-01-27
website: filter "format_datetime": arguments


EarthDweller:
Maybe on page "https://twig.symfony.com/doc/3.x/filters/format_datetime.html" add info that for filter `format_datetime` is possible to use argument `timezone`.

 exp.:
```
{{ "2020-02-02"|format_datetime(pattern="YYYY-MM-dd HH:mm (VV)") }}
<br>
{{ "2020-02-02"|format_datetime(pattern="YYYY-MM-dd HH:mm (VV)", timezone='Pacific/Enderbury') }}
```

fabpot:
Can you submit a PR to add that?

EarthDweller:
👨‍🔬…
https://github.com/twigphp/Twig/pull/3486

But I'm not a native English speaker :)

PR #3486, opened 2021-02-08
New info: Documents / filters / format_datetime: added info and examp…

EarthDweller:
…le of using argument `timezone`

issues 3477:
https://github.com/twigphp/Twig/issues/3477

### DIFF
--- a/doc/filters/format_datetime.rst
+++ b/doc/filters/format_datetime.rst
@@ -44,6 +44,30 @@ By default, the filter uses the current locale. You can pass it explicitly:
     {# 7 août 2019 23:39:12 #}
     {{ '2019-08-07 23:39:12'|format_datetime(locale='fr') }}
 
+It is possible to set TimeZone (see `WiKi: List of tz database time zones
+<https://en.wikipedia.org/wiki/List_of_tz_database_time_zones>`_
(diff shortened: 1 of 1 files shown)


cypherbits:
Yes, just discovered this with this PR... I have been using a custom function for that, but now I see this filter exists.

EarthDweller:
> Yes, just discovered this with this PR... I have been using a custom function for that, but now I see this filter exists.

I was too lazy to search and invent something, it turned out that this filter has such an argument :)

doc/filters/format_datetime.rst
```diff
@@ -40,6 +40,30 @@ By default, the filter uses the current locale. You can pass it explicitly:
     {# 7 août 2019 23:39:12 #}
     {{ '2019-08-07 23:39:12'|format_datetime(locale='fr') }}
 
+It is possible to set TimeZone (see `WiKi: List of tz database time zones
+<https://en.wikipedia.org/wiki/List_of_tz_database_time_zones>`_
+). Date and time for timezone set via argument calculated from default timezone, default timezone depends from `system settings
+<https://twig.symfony.com/doc/1.x/filters/date.html#timezone>`_
```
stof, line 50: this should be using a `:ref:` link so that it looks to the same version of the Twig doc

fabpot:
Thank you @EarthDweller.

PR #3488, opened 2021-02-22
[Doc] Update the Twig version where spaceless filter was added

javiereguiluz:


fabpot:
Thank you @javiereguiluz.

PR #3492, opened 2021-03-15
Update divisibleby.rst

HarryThomas57:
Fixed Typo with space between divisible by


fabpot:
The current version is correct. divisibleby is also supported in 1.x, but not in 2.x/3.x.

PR #3498, opened 2021-03-31
Allow installation of league/html-to-markdown 5.x

colinodell:
Release notes: https://github.com/thephpleague/html-to-markdown/releases/tag/5.0.0

There were no changes that would impact how this Twig Markdown Extension calls the html-to-markdown library, meaning that both 4.8+ and 5.x should be fully compatible.


fabpot:
Thank you @colinodell.

PR #3499, opened 2021-04-06
Fix: A template name cannot be an absolute path

### LINKED ISSUE #3209: File check and PHP open_basedir warning
In this line https://github.com/twigphp/Twig/blob/ac0f512553fab5a685aa0ed1e357d20a644b81bf/src/Loader/FilesystemLoader.php#L212 PHP throws a warning about accessing file outside `open_basedir` on Windows if:
1. PHP `open_basedir` is set
2. `$name` is an absolute path

example:
```
ErrorException: is_file(): open_basedir restriction in effect.
File(C:/dev\ps_17\app/Resources/views/C:/dev/ps_17/mails/themes/modern/core/account.html.twig)
is not within the allowed path(s): (C:/dev)
in C:/dev\ps_17\vendor\twig\twig\src\Loader\FilesystemLoader.php:233
```

Issue presented in all major versions (1.x - 3.x).

### LINKED ISSUE #20709: Install fails with open_basedir and NTS PHP on Windows
related with previous issue https://github.com/PrestaShop/PrestaShop/issues/16736

## how to replicate
issue present when installing from CLI using NTS PHP with `open_basedir` set, steps to replicate:

1. download NTS PHP for Windows (ex https://windows.php.net/downloads/releases/php-7.4.9-nts-Win32-vc15-x64.zip )
2. clone the latest PrestaShop develop and run composer install
4. install from CLI using `php -d open_basedir="PrestaShop_directory" admin-dev/index_cli.php`

currently, the install does not succeed and it emits a php warning:
```
Warning: is_file(): open_basedir restriction in effect. File(C:\xxx\ps_php8\ps_17_php8\app/Resources/views/C:/xxx/ps_php8/ps_17_php8/mails/themes/modern/core/account.html.twig) is not within the allowed path(s): (C:\xxx\ps_php8)
```

some template path is built wrongly, ie. absolute file path is prefixed with a directory path

## local workaround

(to make install passing for development purposes)

delete `mails/themes/modern/` folder

mvorisek:
fixes #3209, https://github.com/PrestaShop/PrestaShop/issues/20709

The source of this is in https://github.com/twigphp/Twig/blob/v3.3.0/src/Loader/FilesystemLoader.php#L212 which NEVER accepts an absolute name (path is always prefixed with path). As discussed in https://github.com/twigphp/Twig/issues/3209#issuecomment-814816613.

The original code emits php warning when `open_basedir` is set on Windows on NTS PHP build (because of PHP inconsistency between TS and NTS, see https://bugs.php.net/bug.php?id=78939). This PR fixes this issue and no warning is emitted.

This fix is important as Symfony relies on exception when absolute path is given in https://github.com/symfony/symfony/blob/v3.4.47/src/Symfony/Bundle/TwigBundle/Loader/FilesystemLoader.php#L82.

Please merge and issue a bugfix release for every major Twig version.

### DIFF
--- a/src/Loader/FilesystemLoader.php
+++ b/src/Loader/FilesystemLoader.php
@@ -292,6 +292,10 @@ class FilesystemLoader implements LoaderInterface, ExistsLoaderInterface, Source
             throw new LoaderError('A template name cannot contain NUL bytes.');
         }
 
+        if ($this->isAbsolutePath($name)) {
+            throw new LoaderError(sprintf('A template name cannot be an absolute path (%s).', $name));
+        }
+
         $name = ltrim($name, '/');
         $parts = explode('/', $name);
         $level = 0;
--- a/tests/Loader/FilesystemTest.php
+++ b/tests/Loader/FilesystemTest.php
@@ -32,11 +32,13 @@ class FilesystemTest extends \PHPUnit\Framework\TestCase
     {
         $loader = new FilesystemLoader([__DIR__.'/../Fixtures']);
 
+        $this->expectException(LoaderError::class);
         try {
             $loader->getCacheKey($template);
-            $this->fail();
         } catch (LoaderError $e) {
             $this->assertStringNotContainsString('Unable to find template', $e->getMessage());
+
+            throw $e;
         }
     }
 
@@ -62,6 +64,14 @@ class FilesystemTest extends \PHPUnit\Framework\TestCase
             ['filters\\\\..\\\\..\\\\AutoloaderTest.php'],
             ['filters\\//../\\/\\..\\AutoloaderTest.php'],
             ['/../AutoloaderTest.php'],
+            ['/AutoloaderTest.php'],
+            ['\\AutoloaderTest.php'],
+            ['//AutoloaderTest.php'],
+            ['\\\\AutoloaderTest.php'],
+            ['/./AutoloaderTest.php'],
+            ['\\.\\AutoloaderTest.php'],
+            ['C:/AutoloaderTest.php'],
+            ['C:\\AutoloaderTest.php'],
         ];
     }
 


fabpot (approved):

fabpot:
Thank you @mvorisek.

mvorisek:
@fabpot can patch release be made for all major versions?

mvorisek:
@fabpot can bugfix release be made for all major versions?

mvorisek:
@fabpot can bugfix release be made for all major versions?

fabpot:
I've just released new versions.

brusch:
@fabpot This breaks existing templates with absolute paths (as it worked before), don't know if this is anyhow relevant in terms of SemVer 🤔

mvorisek:
@brusch it was a bug, discussed here: https://github.com/twigphp/Twig/issues/3209#issuecomment-814816613 use `Symfony\Bundle\TwigBundle\Loader\FilesystemLoader` which can handle absolute paths, see: https://github.com/symfony/symfony/blob/v3.4.47/src/Symfony/Bundle/TwigBundle/Loader/FilesystemLoader.php#L82

brusch:
@mvorisek Well, probably depends on the point of view 😉 I'm happy with that, just wanted to double check 😊

the-conqueror-97:
Well i just got "Uncaught Twig\Error\LoaderError: A template name cannot be an absolute path " now after update.
I have to go back to 3.3.0, issue is on the 3.3.1.

Working in a non symfony environment (custom framework)

mvorisek:
@FatihGucuko do you use "absolute filesystem path" or you pass path with just leading slash (but relative to some registered directory)?

the-conqueror-97:
@mvorisek I use Absolute path like this "/media/user/project/templates", auto generated thanks to dirname(__DIR__)

mvorisek:
@FatihGucuko `findTemplate` does not support absolute filesystem paths, partial support was provided by a hidden bug

if you use absolute filesystem paths solely, skip lookup with `findTemplate` and load the template directly like in https://github.com/symfony/symfony/blob/v3.4.47/src/Symfony/Bundle/TwigBundle/Loader/FilesystemLoader.php#L82

the-conqueror-97:
An usefull bug then

mvorisek:
> 
> 
> An usefull bug then

but it breaks other things, see desc

the-conqueror-97:
So do you know how to fix it ?
I have this :  $loader = new FilesystemLoader("/media/user/project/templates");
just like in the Tiwg's site basic section:  https://twig.symfony.com/doc/3.x/api.html

mvorisek:
`FilesystemLoader` is file paths resolving component. With absolute path, you do not need that, simply use the path directly without path resolver.

the-conqueror-97:
Well i guess i found the problem !
We can load this way:  $loader = new FilesystemLoader("/media/user/project/templates"); Even in the way i did before.

BUT now i have to change ALL controller's 
$this->render("/auth/index.html.twig")
To  $this->render("auth/index.hmlt.twig")

AND in Twig file
{% include "/auth/widget.html.twig %}
TO
{% include "auth/widget.html.twig" %}

Which fixes my problem, and i can tell you, i was tired to call templates files with "/path/file.html.twig" prefer "path/file.html.twig"

berridgeab:
> Well i guess i found the problem !
> We can load this way: $loader = new FilesystemLoader("/media/user/project/templates"); Even in the way i did before.
> 
> BUT now i have to change ALL controller's
> $this->render("/auth/index.html.twig")
> To $this->render("auth/index.hmlt.twig")
> 
> AND in Twig file
> {% include "/auth/widget.html.twig %}
> TO
> {% include "auth/widget.html.twig" %}
> 
> Which fixes my problem, and i can tell you, i was tired to call templates files with "/path/file.html.twig" prefer "path/file.html.twig"

I'm in the same boat, its broken ALOT of files, all for the sake of the ltrim which worked fine previously. Its not clearly documented that a string passed to load() cannot contain a leading slash. 

I use the URI route name as a mapping for the TWIG load name i.e. 
/my/url/somewhere
becomes 
/my/url/somewhere.twig.

Now I have to run ltrim on that path before passing to Twig or extend the existing loader to restore the previous behaviour.

Using the \Symfony\Bundle\TwigBundle\Loader\FilesystemLoader is not an option as its not part of the 'twig/twig' package.

I get that its fixing a bug but this kind of change should be saved for a major release i.e. 4.X.X

MvEerd:
fwiw, I also find this being a patch release questionable.

You are now throwing an exception for something which was explicitly accounted for since line 229 actually does an `ltrim` to remove the first slash.
It would have been appreciated to see a deprecation warning before this was merged into a patch release. It now caused things to break on composer update.
Admittedly the first slash should have not been passed to the render function in the first place, but to see the behavior that accounts for this change from 3.3.0 to 3.3.1 was quite unexpected.

stof:
@MvEerd support for passing template names with a leading `/` while treating them like as relative paths for the filesystem loader has **never** been something officially supported. AFAICT, none of the maintainers were even aware that it worked before.

berridgeab:
@stof If that was the case the ltrim removing the forward slash should never of existed in the first place. The fact that it was there suggests someone did know about it and sanitized the string accordingly to prevent issues.

MvEerd:
@stof Like @berridgeab says this was specifically accounted for and has worked like this for 9 years

It was added in the following PR as recommended during the review https://github.com/twigphp/Twig/pull/1026 it may have been a side effect but the leading `/` being stripped is a behavior some code has relied on since then.

Apologies to all participants for the pings, I'll drop it now as I don't intend to question the maintainers intention and knowledge of the codebase.
Just trying to share that it had a real world impact and was unexpected in the hopes this can be considered next time sometime like this gets tagged as 'patch'.
Thanks for all the effort put into maintaining twig

Metrol:
Perhaps the thinking was that nobody was using absolute paths.  As this, and some other issues up here are hopefully showing, it is simply not the case.  It did work, and we are using this feature.

My composer.json now has Twig locked to 3.3.0 as going to 3.3.1 literally breaks multiple sites for me.

fabpot:
This PR has been reverted in all branches as this is a BC break as shown by the many people relying on this behavior before.
New releases have also been made accordingly.
That being said, I would still recommend people upgrade their code.

PR #3518, opened 2021-05-11
Fix return type of Environment::render

Warxcell:


Warxcell:
Sorry, totally mis-looked. It's param, not return.

PR #3524, opened 2021-05-18
fix file permissions

romainmenke:
`drupal_test.sh` was executable and when installing the plugin with `composer` it retains these permissions.

Better to call this script as `bash ./drupal_test.sh` and have `100644` as permissions.

I could only find one reference to this script and it was commented out.
Updated it anyway as it might be revived at some point.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,3 @@
 /tests export-ignore
 /phpunit.xml.dist export-ignore
+./drupal_test.sh export-ignore


stof:
Well, making the file executable was expected. This file is meant to be an executable script.

However, it might make sense to add it in the `.gitattributes` list of files excluded from archives, as there is indeed no need to ship this executable when installing twig as a dependency.

fabpot:
Thank you @romainmenke.

PR #3525, opened 2021-05-18
Replace secure.php.net with www.php.net

MauricioFauth:
It is no longer necessary to use the _secure_ subdomain.

See: https://news-web.php.net/php.internals/104533.

> Instead of having mirrors, we are moving all of PHP.net to HTTPS (and get rid of https://secure.php.net)


stof (approved):

fabpot:
Thank you @mauriciofauth.

PR #3526, opened 2021-05-20
Allow Symfony 6

nicolas-grekas:


composer.json
```diff
@@ -29,7 +29,7 @@
         "symfony/polyfill-ctype": "^1.8"
     },
     "require-dev": {
-        "symfony/phpunit-bridge": "^4.4.9|^5.0.9",
+        "symfony/phpunit-bridge": "^4.4.9|^5.0.9|^6.0",
```
stof: AFAICT, the bridge does not need to allow different versions, as it has a separate handling (Symfony itself always uses the latest version of that bridge when running tests for its older versions)

nicolas-grekas: sure, but it doesn't hurt either :)

nicolas-grekas:
Merged as 1e53d27a

PR #3533, opened 2021-06-01
add support for the class constant on objects

dunglas:
This is especially useful for Symfony Turbo:

Before:

```twig
<div id="books" {{ turbo_stream_listen('App\\Entity\\Book') }}></div>
```

After:

```twig
<div id="books" {{ turbo_stream_listen(constant('class', book) }}></div>
```

Replace https://github.com/php/php-src/pull/6763


dunglas:
The test failure looks unrelated.

fabpot:
Thank you @dunglas.

PR #3535, opened 2021-06-05
Fix implicit float to int casts

derrabus:
PHP 8.1 is picky about implicit float-to-int casts if we lose precision because of that cast. This PR attempts to fix all issues on the 1.x branch.

### DIFF
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -349,7 +349,7 @@ function twig_cycle($values, $position)
 function twig_random(Environment $env, $values = null, $max = null)
 {
     if (null === $values) {
-        return null === $max ? mt_rand() : mt_rand(0, $max);
+        return null === $max ? mt_rand() : mt_rand(0, (int) $max);
     }
 
     if (\is_int($values) || \is_float($values)) {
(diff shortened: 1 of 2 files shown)


fabpot:
Thank you @derrabus.

src/Extension/CoreExtension.php
```diff
@@ -349,7 +349,7 @@ function twig_cycle($values, $position)
 function twig_random(Environment $env, $values = null, $max = null)
 {
     if (null === $values) {
-        return null === $max ? mt_rand() : mt_rand(0, $max);
+        return null === $max ? mt_rand() : mt_rand(0, (int) $max);
```
jvasseur, line 352: The parameter is typed as `int|null` so if it is a float here this would be a user error and maybe we should keep the warning ?

derrabus: But then again, there is a test case that explicitly tests for the behavior of that function when passing a float. That made me think that the silent cast was somewhat expected behavior.

The function is registered for use in Twig templates and I figured it might be better to suppress the warning than to have people do type casts in templates.

But I'm not using this function and I did this change mainly to get the tests green. I might be horribly wrong with my assessment. If we revert here, how shall we proceed with the test case?

jvasseur: Well maybe the fix is just to change the type of the argument to show that we allow floats here, if we have a test with a float value this means such a value is expected.

Issue #3534, opened 2021-06-02
[PHP 8.1] During inheritance of JsonSerializable: Uncaught Declaration of Twig\Markup::jsonSerialize() should be compatible with JsonSerializable::jsonSerialize(): mixed


williamdes:
`twig/twig/src/Markup.php:43`

Error: [on PHP 8.1 CI](https://github.com/phpmyadmin/phpmyadmin/runs/2730389628#step:9:22)
```
PHP Fatal error:  During inheritance of JsonSerializable: Uncaught Declaration of Twig\Markup::jsonSerialize() should be compatible with JsonSerializable::jsonSerialize(): mixed

/home/runner/work/phpmyadmin/phpmyadmin/vendor/twig/twig/src/Markup.php:43
/home/runner/work/phpmyadmin/phpmyadmin/vendor/twig/twig/src/Markup.php:19
/home/runner/work/phpmyadmin/phpmyadmin/tmp/twig/fb/fb1a0d2016c3fc6c9ab5b48d37a6ffd92cd99f658bbb5f157d9ccefa203590e0.php:59
/home/runner/work/phpmyadmin/phpmyadmin/vendor/twig/twig/src/Template.php:394
/home/runner/work/phpmyadmin/phpmyadmin/vendor/twig/twig/src/Template.php:367
/home/runner/work/phpmyadmin/phpmyadmin/vendor/twig/twig/src/Template.php:379
/home/runner/work/phpmyadmin/phpmyadmin/vendor/twig/twig/src/TemplateWrapper.php:40
/home/runner/work/phpmyadmin/phpmyadmin/libraries/classes/Template.php:157
/home/runner/work/phpmyadmin/phpmyadmin/libraries/classes/Header.php:721
/home/runner/work/phpmyadmin/phpmyadmin/libraries/classes/Header.php:194
/home/runner/work/phpmyadmin/phpmyadmin/libraries/classes/Header.php:143
/home/runner/work/phpmyadmin/phpmyadmin/libraries/classes/Response.php:184
/home/runner/work/phpmyadmin/phpmyadmin/libraries/classes/Response.php:215
/home/runner/work/phpmyadmin/phpmyadmin/libraries/classes/Config/PageSettings.php:161
/home/runner/work/phpmyadmin/phpmyadmin/libraries/classes/Config/PageSettings.php:90
/home/runner/work/phpmyadmin/phpmyadmin/test/classes/Config/PageSettingsTest.php:48
[...]

PR #3536, opened 2021-06-05
Add ReturnTypeWillChange to JsonSerializable implementation

derrabus:
On PHP 8.1, implementations of `JsonSerializable` should declare a return type.

In order to avoid a deprecation warning and maintain backwards compatibility, I've added the `ReturnTypeWillChange` attribute to declare that our implementation will add the necessary return type in a future release.

### DIFF
--- a/src/Markup.php
+++ b/src/Markup.php
@@ -40,6 +40,7 @@ class Markup implements \Countable, \JsonSerializable
         return mb_strlen($this->content, $this->charset);
     }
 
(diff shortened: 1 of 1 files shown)


fabpot:
Thank you @derrabus.

MauricioFauth:
Closes #3534

ShadrackKimutai:
I rolled back the server PHP to 8.0.7 and the error disappeared.

PR #3540, opened 2021-06-17
Add needs_source to twig function

shyim:
Hey,

We want to limit the functionality of a twig function in which template it has been used. Currently, it seems not possible to access the current twig file name.

This PR would be made it possible to access the source where the file name is there.

Would you like to have such a thing in Twig? This PR is currently wip as it doesn't have tests or docs. I would provide them if it would be accepted


stof:
Can you describe your use case ?

shyim:
We have found a better not so ugly way. So I will close this then :sweat_smile:

PR #3543, opened 2021-06-21
Fix typo in CacheTokenParser.php

pulzarraider:


fabpot:
Thank you @pulzarraider.

PR #3546, opened 2021-07-05
Add missing throws annotation

VincentLanglet:
Hi

When using a static analysis tool which check for exception, the `@throws` annotation is really important.
It allows the developer to find missing `try/catch` or useless `try/catch`.

In my case I'm using methods like `TokenStream::look` or `TokenStream::next` which can throw an exception but wasn't documented. I saw a lot of `LogicException`, `RuntimException`, `SyntaxError` and `LoaderError` were already in phpdoc of a lot of method, so I assume adding more follow just a consistency pattern.

### DIFF
--- a/src/ExpressionParser.php
+++ b/src/ExpressionParser.php
@@ -206,6 +206,9 @@ class ExpressionParser
         return $token->test(/* Token::OPERATOR_TYPE */ 8) && isset($this->binaryOperators[$token->getValue()]);
     }
 
+    /**
+     * @throws SyntaxError
+     */
     public function parsePrimaryExpression()
     {
         $token = $this->parser->getCurrentToken();
@@ -339,6 +342,9 @@ class ExpressionParser
         return $node;
     }
 
+    /**
+     * @throws SyntaxError
+     */
     public function parseHashExpression()
     {
         $stream = $this->parser->getStream();
@@ -412,6 +418,9 @@ class ExpressionParser
         return $node;
     }
 
+    /**
+     * @throws SyntaxError
+     */
     public function getFunctionNode($name, $line)
     {
         switch ($name) {
@@ -460,6 +469,9 @@ class ExpressionParser
         }
     }
 
+    /**
+     * @throws SyntaxError
+     */
     public function parseSubscriptExpression($node)
     {
         $stream = $this->parser->getStream();
@@ -641,6 +653,9 @@ class ExpressionParser
         return new Node($args);
     }
 
+    /**
+     * @throws SyntaxError
+     */
     public function parseAssignmentExpression()
     {
         $stream = $this->parser->getStream();
@@ -706,6 +721,9 @@ class ExpressionParser
         return new $class($node, $name, $arguments, $this->parser->getCurrentToken()->getLine());
     }
 
+    /**
+     * @throws SyntaxError
+     */
     private function getTest(int $line): array
     {
         $stream = $this->parser->getStream();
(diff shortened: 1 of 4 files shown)


stof:
Those SyntaxError are not meant to be catched when working with the TokenStream.

VincentLanglet:
> Those SyntaxError are not meant to be catched when working with the TokenStream.

Static analysis can also detect useless `@throws` in the phpdoc.
Which means that If I write
```
/** @throws SyntaxError */
public function foo()
{
   ...
   $this->tokenStream->look(42)
   ...
}
```
I end up with a useless `@throws` because the look method is not well documented.

Also, there is already some `@throws SyntaxError` in the code elsewhere. Even if, it's not something which should be often caught, it's seems better to me to document that it can throw an error. I also saw some `LogicException` added to the phpdoc already ; LogicException are typically something which should not be caught, but still it was added to the phpdoc.

In my case, I'm developing a tool for twig, running my tool is using some of these methods.
It's still more user friendly for my tool to catch, at least at the top level, in order to display a friendly error message when the tool is throwing an exception message.

I already got multiple times some of the SyntaxError thrown by twig, so these exceptions are thrown. And it seems better to me to handle them properly.

VincentLanglet:
Friendly ping @nicolas-grekas, can you give your opinion on this ? :)

nicolas-grekas:
> Those SyntaxError are not meant to be catched when working with the TokenStream.

I agree with this statement. `@throw` should be used when an exception is part of the API of an interface. Here, that's not the case: the dev should know about the exception and they should thus not be caught. That reasoning usually applies to any `LogicException`.

VincentLanglet:
> > Those SyntaxError are not meant to be catched when working with the TokenStream.
> 
> I agree with this statement. `@throw` should be used when an exception is part of the API of an interface.

There is already some @throw annotation for SyntaxError:
https://github.com/twigphp/Twig/blob/3.x/src/TokenParser/TokenParserInterface.php#L36

> Here, that's not the case: the dev should know about the exception and they should thus not be caught. That reasoning usually applies to any `LogicException`.

Yes, when you're writing your own template, it make no sens to catch the error, and could be considered as a LogicException. But when you're writing a tool which validate or check the code style of a template provided, it's should be considered a RuntimeException.
BTW, SyntaxError extends Error which extends Exception. So it's not a LogicException.

Twig is already catching some `SyntaxError` in order to have a specific behavior:
- https://github.com/twigphp/Twig/blob/9e6e700f96bafffd0708d700326bcb8520c13e69/src/Parser.php#L89
- https://github.com/twigphp/Twig/blob/13e360972adb846a249301c6059e7958945911bd/src/Util/DeprecationCollector.php#L68

Even in the twig documentation, it's recommended to catch them
https://github.com/twigphp/Twig/blob/4659a99cdf240bce61a8cc152d1fe5cc8df9014a/doc/recipes.rst#validating-the-template-syntax

Since, it's documented to catch them, method throwing a SyntaxError should be documented.

stof:
> Even in the twig documentation, it's recommended to catch them
> https://github.com/twigphp/Twig/blob/4659a99cdf240bce61a8cc152d1fe5cc8df9014a/doc/recipes.rst#validating-the-template-syntax

This only implies we should have `@throws SyntaxError` on some public methods of the Environment, not on all lower level APIs.

VincentLanglet:
> This only implies we should have `@throws SyntaxError` on some public methods of the Environment, not on all lower level APIs.

In my case, I'm implementing the `TokenParserInterface`.
In this one, the `parse` method is documented as possibly throwing `SyntaxError`.

In my implementation of the `parse` method, I'm using the methods of TokenStream, like `next` or `look`.
Would it be ok if I update this PR to only adding the throw tag to the methods of TokenStream then ?

VincentLanglet:
I've only added the SyntaxError I need to catch to have better chance this getting merged.

VincentLanglet:
Is there any chances for those @throws tag to be added ?

PR #3548, opened 2021-07-07
Explicitly specify the encoding for mb_ord in JS escaper

oleg-st:
`mb_ord` should be called with the specified UTF-8 encoding to make it independent of `mb_internal_encoding`.

Related to https://github.com/twigphp/Twig/pull/3397

### DIFF
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -1237,7 +1237,7 @@ function _twig_escape_js_callback($matches)
         return $shortMap[$char];
     }
 
-    $codepoint = mb_ord($char);
+    $codepoint = mb_ord($char, 'UTF-8');
     if (0x10000 > $codepoint) {
         return sprintf('\u%04X', $codepoint);
     }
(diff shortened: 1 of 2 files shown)


tests/Extension/EscaperTest.php
```diff
@@ -182,6 +182,15 @@ public function testJavascriptEscapingConvertsSpecialChars()
         }
     }
 
+    public function testJavascriptEscapingConvertsSpecialCharsWithInternalEncoding()
+    {
+        $twig = new Environment($this->createMock(LoaderInterface::class));
+        mb_internal_encoding('ISO-8859-1');
```
stof: this should be reset to the previous value in a `finally` block, to avoid impacting other tests

stof (changes requested):

tests/Extension/EscaperTest.php
oleg-st: Done

oleg-st:
@stof Fixed

fabpot:
Thank you @oleg-st.

PR #3554, opened 2021-08-10
Add explicit `@return` type next to `#[ReturnTypeWillChange]`

nicolas-grekas:


fabpot:
Thank you @nicolas-grekas.

PR #3557, opened 2021-09-02
Fix strtr call in twig_replace_filter for PHP 8.1

yookoala:
* twig_replace_filter, although documented $str to be string type,
  does not enforce that in typehint. There's chance that null is
  passed into $str. Then it would trigger a "Deprecated: strtr():
  Passing null to parameter 1 ($string) of type string is
  deprecated" warning.

* Cast the $str as string before passing to strtr should fix the
   issue.

### DIFF
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -516,7 +516,7 @@ function twig_replace_filter($str, $from)
         throw new RuntimeError(sprintf('The "replace" filter expects an array or "Traversable" as replace values, got "%s".', \is_object($from) ? \get_class($from) : \gettype($from)));
     }
 
-    return strtr($str, twig_to_array($from));
+    return strtr((string) $str, twig_to_array($from));
 }
 
 /**


stof:
If you pass something that does not respect the contract of the method, the issue is in your code. Silently casting it to string would hide the fact that you misuse the method.

yookoala:
@stof: I totally agree that it was a misuse of the function (although not my own code. I stumbled upon this in another open source project). It's just that the situation is not quite friendly to library users:
1. This was not an issue before PHP 8.1. I don't think people realize this until upgrading to PHP 8.1 or higher.
2. The warning message was for `strtr` and was not very specific as to what exactly gone wrong (which was the use of `twig_replace_filter` or the Twig classes that uses it). It's quite obscure to debug with.

The current situation is right in the middle of nowhere. It does not provide enough information for debug. And it does not work with Twig usages that work before PHP 8.1. I believe this can be improved, It's just a matter of which approach to take.

Twig can be more helpful by:
A. Showing Twig user where exactly the usage of Twig gone wrong; or
B. Silently retain backward compatible with existing code that uses Twig.

The patch I wrote here opted for approach (B), but I'm actually fine with both approaches. For approach (A), Twig can:

1. Change `twig_replace_filter` function signature:
    ```php
    /**
     * Replaces strings within a string.
     *
     * @param string             $str  String to replace in
     * @param array|\Traversable $from Replace values
     *
     * @return string
     */
    function twig_replace_filter(string $str, $from)
    {
       ...
    ```
    or
2. Check and raise some error so user get full backtrace in log:
    ```php
    /**
     * Replaces strings within a string.
     *
     * @param string             $str  String to replace in
     * @param array|\Traversable $from Replace values
     *
     * @return string
     */
    function twig_replace_filter($str, $from)
    {
       if (!is_string($str)) {
           throw new \InvalidArgumentException('$str in twig_replace_filter should be a string.');
       }
       ...
    ```

What do you think?

ruudk:
Also stumbled on this while updating a large application to PHP 8.1.

> If you pass something that does not respect the contract of the method, the issue is in your code. 

I agree, but PHP and Twig never warned about it. As we currently don't have a static analyzer for Twig, it will almost be impossible to check all Twig template files one-by-one and see if null is passed or not.

stof:
adding an exception would *break* the project immediately instead of only triggering a deprecation, which looks worse to me.

> The current situation is right in the middle of nowhere. It does not provide enough information for debug.

The PHP error handler provides a stack trace (that will end up in the compiled twig template rather than the twig file, but that allows identifying the source).
And changing that to an exception in the filter would still require looking at the stack trace to find where this filter was called.

yookoala:
To be very specific, these are the lines that my template had problem with:

https://github.com/GibbonEdu/core/blob/d2eae14a420930e4b25fa0c15dc766eea346e626/resources/templates/components/form.twig.html#L64

https://github.com/GibbonEdu/core/blob/d2eae14a420930e4b25fa0c15dc766eea346e626/resources/templates/components/form.twig.html#L92

There was an issue with some of the `element` passed in the twig template. The method `getClass` of some element class did not return anything at all hence `element.getClass` was evaluated as `null`. Then `null` is passed on to the `replace()` twig filter, and caused the cryptic `strtr()` message all over the log messages:

```
Deprecated: strtr(): Passing null to parameter #1 ($string) of type string is deprecated in /home/runner/work/GibbonEduCore/GibbonEduCore/vendor/twig/twig/src/Extension/CoreExtension.php on line 519
```

Since this is not an Error nor an Exception, there is no trace stack. I have not even a clue which compiled-template (talking about the temporary php file) was responsible. It took me a few days of trail-and-error to figure it all out (being a voluntary work in my off-hour doesn't make it easier). My problem was fixed in GibbonEdu/core#1467.

If the filter wouldn't silently accept the old (working) code, it would have been more productive for the Exception knows where in the twig file it was from. I have the impression that some twig errors does show that (?). If so, how do we make it happen for this error?

yookoala:
The [relevant RFC](https://wiki.php.net/rfc/deprecate_null_to_scalar_internal_arg) specified the plan to make this an error:

> If
> * null is passed to an argument of an internal function,
> * the argument is not nullable,
> * the argument accepts at least one scalar type (one of bool, int, float, or string),
> * coercive typing mode is used, i.e. strict_types=1 is not enabled,
>
> then
> * in PHP < 8.1 the null value will be silently coerced to either false, 0, 0.0 or “”.
> * in PHP >= 8.1 a deprecation error is thrown, but the value is still coerced and the call still takes place.
> * in PHP >= 9.0 a TypeError is thrown, consistent with the behavior of user-defined functions.

So the twig code that currently work will break in PHP 9.0+ if this kind of deprecated message was not dealt with. Right now, it's a pain in the ass to try to figure out what cause the problem. in a twig template.

If the concern was breaking old code, I think we can simply add function that improves the error log. For example:

```php

/**
 * Fomrats a debug_backtrace output.
 *
 * Mimicing the output of debug_print_backtrace() without actually calling it.
 *
 * @param array $stack Supposedly the debug_backtrace() output.
 * @return string The formatted message.
 */
function format_debug_backtrace(array $stack): string {
        return implode("\n", array_map(function ($num, $layer) {
                return "#{$num} {$layer['function']}() called at [{$layer['file']}:{$layer['line']}]";
        }, array_keys($stack), array_values($stack))) ?? '';
}

/**
 * Log deprecated message with stacktrace.
 *
 * With error_log, tell server admin that they have deprecated usage of php
 * with full stacktrace to the file(s) involved.
 *
 * @param string $message The deprecated message.
 * @param int $offset Optional offset to remove unneeded inner layers of the stacktrace.
 * @return void
 */
function twig_log_deprecated_stacktrace(string $message, int $offset=0) {

        // generate stack, then remove the layer created from
        // this function.
        $stack = debug_backtrace();
        if ($offset > sizeof($stack)) return '';
        array_splice($stack, 0, $offset+1);

        // format the message.
        $backtrace_str = format_debug_backtrace($stack);
        $top = array_shift($stack);
        error_log("PHP Deprecated: {$message} in {$top['file']}:{$top['line']}\nStacktrace:\n{$backtrace_str}");
}

...

/**
 * Replaces strings within a string.
 *
 * @param string             $str  String to replace in
 * @param array|\Traversable $from Replace values
 *
 * @return string
 */
function twig_replace_filter($str, $from)
{
   if (!is_string($str)) {
       twig_log_deprecated_stacktrace('$str in twig_replace_filter should be a string. ' . var_export($str, true) . ' given');
   }
   ...

```

fabpot:
Twig is not PHP. I think we need to keep the old behavior and automatically convert `null` to an empty string. See #3617 for my fix.

yookoala:
That was the intended design of this pull request. 😅
Thanks anyway.

PR #3566, opened 2021-09-29
Update markdown-to-html converter missing message

tobias-93:
The `league/html-to-markdown` library does not support Markdown to HTML conversion (anymore?). The `league/commonmark` library is suitable for that.


stof (approved):

fabpot:
Thank you @tobias-93.

PR #3570, opened 2021-10-02
Update .gitattributes to exclude non-dist files

### LINKED ISSUE #3503: Optimize for packagist production
When downloading the Twig package from packagist a lot of files are included which are never used in production or even by the developer itself. Have a look, this is what I get in my `vendor/twig/twig` folder after `composer require "twig/twig:^3.0"`:

```
CHANGELOG
composer.json
doc
drupal_test.sh
.editorconfig
.gitattributes
.github
.gitignore
LICENSE
.php_cs.dist
README.rst
src
```

Files and directories I think can be ignored:

```
.github
doc
drupal_test.sh
.editorconfig
.gitattributes
.gitignore
.php_cs.dist
```

These files together are 640kb, while the whole package is about 1700kb. That's about 48% of the total file size.

Please fix this. Love the package, thank you <3

kaznovac:
resolves #3503 

some extra files can be excluded also:
- changelog
- license
- readme

should this be rebased to 1.x, 2.x?

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,3 @@
+/doc/** export-ignore
 /tests export-ignore
 /phpunit.xml.dist export-ignore


stof:
The license file definitely cannot be excluded.

fabpot:
I don't see the point. These files are do not contribute significantly to the size of the archive and as mentioned by stof, some cannot be removed. Let's close. Thank you for proposing.

kaznovac:
@fabpot

let me cite @stof 
> excluding the doc source code from the archive indeed probably makes sense (that's probably the bulk of these 640KB as others are small files)
https://github.com/twigphp/Twig/issues/3503#issuecomment-815660823

and this pr excludes the doc, along with the git and github speciffic configs

fabpot:
@fabpot Sorry, can you update this PR to "just" exclude the `doc/` directory in addition to what we already have?

kaznovac:
@fabpot np :), excluded "just" `doc/` (in the same manner as `/extra/**`)

stof (approved):

fabpot:
Thank you @kaznovac.

PR #3573, opened 2021-10-12
Start experiment with 'literal-string' type

craigfrancis:
Both [Psalm 4.8](https://github.com/vimeo/psalm/releases/tag/4.8.0) and [PHPStan 0.12.97](https://github.com/phpstan/phpstan/releases/tag/0.12.97) have recently introduced the `literal-string` type.

This allows Twig to note that certain methods expect a `string` to be created by the developer (defined in the source code), so mistakes that lead to a typical Injection Vulnerability can be identified, e.g.

```php
$_GET['name'] = '<script>alert(1)</script>';

$html = $twig->createTemplate('<p>Hi ' . $_GET['name'] . '</p>')->render(); // INSECURE

$html = $twig->createTemplate('<p>Hi {{ name }}</p>')->render(['name' => $_GET['name']]);
```

By setting the `$template` parameter in `createTemplate()` to `literal-string`, for those using use Psalm (level 3 or stricter) or PHPStan (level 5 or stricter), we can check the `$template` HTML does not contain unsafe user data (a mistake often created by junior developers making a "quick edit").

And `literal-string` is quite forgiving, in that it works with variables, and allows string concatenation (assuming both are `literal-string` values), so code like the following works well, e.g.

```php
$template = '<p>Hi ' . ($name ? '<strong>{{ name }}</strong>' : '<span>Guest</span>') . '</p>';

$html = $twig->createTemplate($template)->render(['name' => $name]);
```

That said, I still want to be cautious (I don't want to create issues for the users of Twig). Which is why I'd only like to try this with `->createTemplate()`, so we can see what feedback we get (I'm happy to help with that), and if we get a positive response, we can see what other methods might benefit from this.

### DIFF
--- a/src/Environment.php
+++ b/src/Environment.php
@@ -377,6 +377,9 @@ class Environment
      * @param string $template The template source
      * @param string $name     An optional name of the template to be used in error messages
      *
+     * @phpstan-param literal-string $template
+     * @psalm-param   literal-string $template
+     *
      * @throws LoaderError When the template cannot be found
      * @throws SyntaxError When an error occurred during compilation
      */


fabpot:
I don't get what that does. PHP does not care about these annotations.

craigfrancis:
It's used by static analysis tools (Psalm and PHPStan), so developers using them will be notified when there is a problem in their code (useful for large development teams, where you would have junior developers who often make mistakes).

Also, if this is adopted, there is more of a case to be made for `is_literal()` to be added to PHP itself.

fabpot:
I'm closing as I don't want to manage such specific static analysis features. Let's revisit when it's adopted by PHP itself. Thanks for proposing.

craigfrancis:
You're already providing annotations for Static Analysis? and It's not going to get adopted without interest from the community (fortunately Propel and RedBean have adopted this already).

nicolas-grekas:
I agree with @fabpot's decision. Twig doesn't have any vendor-specific annotation AFAIK, and should remain so.

craigfrancis:
Which Static Analysis tools do you support with these annotations?

If it's just Psalm and PHPStan (who I believe are the most popular), I'm happy to put forward a patch that replaces line 377:

    < @param string $template The template source
    > @param literal-string $template The template source

nicolas-grekas:
> Which Static Analysis tools do you support with these annotations?

My opinion is that neither Twig nor Symfony should be early adopters of newer SA features.

```
< @param string $template The template source
> @param literal-string $template The template source
```

This would work for me, but not today as this is too new. I can't tell when though. Support by phpstorm would help. And waiting a few more months to ensure ppl have upgraded. A year is the bare minimum to me for a reasonable amount of ppl to upgrade their stack. But that's just my opinion :)

craigfrancis:
Thanks, so September 2022 ish?

And I'll see what I can do regarding phpStorm.

craigfrancis:
Maria has added support to PhpStorm last week ([WI-64109](https://youtrack.jetbrains.com/issue/WI-64109/literal-string-support-in-phpdoc)). I assume this will be included in PhpStorm 2022.3, rather than the next patch release, so hopefully December ish?

In the mean time, as this feature has been available in Psalm and PHPStan for over a year, could we consider using `literal-string` for this method?

I'm just concerned by the times I've seen developers use `$twig->createTemplate()` incorrectly, where this provides a fairly simple and effective way to check for these mistakes (I cannot find any other way of solving this issue, and asking developers to read the documentation and not make mistakes doesn't exactly work).

craigfrancis:
PhpStorm 2022.3 now recognises the `literal-string` type.

Do I need to open a new pull request?

PR #3574, opened 2021-10-14
Extension::spaceless: Use better spaceless algorithm.

janbarasek:
In case of template:

```html
{% block content %}
{% apply spaceless %}
<p>
    Open <a href="{{ url }}">Google</a> <strong>for more</strong> info.
</p>
{% endapply %}
{% endblock %}
```

Final old render is:

```html
<p>
    Open <a href="https://google.cz">Google</a><strong>for more</strong> info.
</p>
```

New algorithms will return:

```html
<p> Open <a href="https://google.cz">Google</a> <strong>for more</strong> info. </p>
```

But it breaks space between words.

Snímek obrazovky 2021-10-14 v 15 31 06

I think more better algorithm should be inspired from Latte engine: https://github.com/nette/latte/blob/53e5da9aa2ec20e6e7e4149448e5d0ad2af282ba/src/Latte/Runtime/Filters.php#L309

Thanks.

### DIFF
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -1030,7 +1030,19 @@ function twig_trim_filter($string, $characterMask = null, $side = 'both')
  */
 function twig_spaceless($content)
 {
-    return trim(preg_replace('/>\s+</', '><', $content));
+    $content = (string) preg_replace_callback(
+        '#[ \t\r\n]+|<(/)?(textarea|pre|script)(?=\W)#si',
+        static function ($m) {
+            if (empty($m[2])) {
+                return ' ';
+            }
+
+            return $m[0];
+        },
+        $content
+    );
+
+    return trim($content);
 }
 
 function twig_convert_encoding($string, $to, $from)


stof:
the Latte algorithm is collapsing spaces, which is a weird algorithm for something named `spaceless` as the output contains spaces.

stof:
And this is a BC break. It will break projects that rely on the fact that `spaceless` is actually producing an output without spaces (and so that 2 `inline-block` elements won't be separated by a whitespace in the rendering)

janbarasek:
How can this be fixed to preserve the space between two HTML elements where removing the space would interfere and change the meaning, as in the example above?

stof:
@janbarasek the thing is, your use case is a case where you don't want to remove all spaces. but that does not mean that existing projects using `spaceless` want to apply your use case.

Side note: I don't see any reason to run a filter performing whitespace collapsing server-side with a performance cost (running a regexp on the output is not free, and might get worse if inside a loop, which is why `symfony/form` stopped using spaceless in its rendering years ago to get a huge performance boost for big forms), as browsers are collapsing them anyway during rendering.
The existing `spaceless` filters solves a case where you want to go to 0 spaces, which has a use case (but indeed not when your HTML uses inline elements).

Note that in practice, I found very few use cases for spaceless. In most cases, the need is to remove spaces coming from template indentation, which is better handled using the Twig whitespace control feature (which is free from a performance point of view as that's a compile-time feature)

fabpot:
Closing for the reasons explained by @stof

PR #3578, opened 2021-10-18
Update README.md

noniagriconomie:
Ref https://github.com/twigphp/string-extra/pull/4


fabpot:
I've fixed the URL instead.

PR #3584, opened 2021-11-04
Fix emphasis on compiler to bold styling

jonnyeom:
This is just so the Lexer, Parser, and Compiler has the same styling


fabpot:
Good catch, thanks @jonnyeom.

PR #3585, opened 2021-11-05
[TwigExtraBundle] Add the return type information in getConfigTreeBuilder()

javiereguiluz:
Moved from https://github.com/twigphp/twig-extra-bundle/pull/5

-----

This will fix the following deprecation:

```
  1x: Method "Symfony\Component\Config\Definition\ConfigurationInterface::getConfigTreeBuilder()" might add "TreeBuilder" as a native return type declaration in the future. Do the same in implementation "Twig\Extra\TwigExtraBundle\DependencyInjection\Configuration" now to avoid errors or add an explicit @return annotation to suppress this message.
    1x in AddUserCommandTest::testCreateUserNonInteractive from App\Tests\Command
```

### DIFF
--- a/extra/twig-extra-bundle/DependencyInjection/Configuration.php
+++ b/extra/twig-extra-bundle/DependencyInjection/Configuration.php
@@ -17,6 +17,9 @@ use Twig\Extra\TwigExtraBundle\Extensions;
 
 class Configuration implements ConfigurationInterface
 {
+    /**
+     * @return TreeBuilder
+     */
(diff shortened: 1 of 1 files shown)


fabpot:
I think we need to do the change in 2.x, right?

javiereguiluz:
Yes! I did the change using GitHub's UI ... so rebasing doesn't work as expected ... so I opened instead --> #3586. Let's close this one.

PR #3588, opened 2021-11-06
Use faster hash algorithm (xxh128) on PHP 8.1+

GromNaN:
Twig uses the `hash` function at runtime to convert template paths class names, each time a template is loaded. In a large project with a high granularity of templates (hundreds of `include` per page), this can become the most time consuming function as reported by Blackfire:

To optimise this use-case, [PHP 8.1 supports the xxHash hash algorithms](https://php.watch/versions/8.1/xxHash). `xxh128` is 60x faster that `sha256` according to the [benchmarks](https://php.watch/articles/php-hash-benchmark), and guarantee a very [low risk of collision](http://cyan4973.github.io/xxHash/).

I'm not able to test with PHP 8.1 for now. Before going further I would like to validate performance impact on a large list of file names.

### DIFF
--- a/src/Environment.php
+++ b/src/Environment.php
@@ -298,7 +298,7 @@ class Environment
     {
         $key = $this->getLoader()->getCacheKey($name).$this->optionsHash;
 
-        return $this->templateClassPrefix.hash('sha256', $key).(null === $index ? '' : '___'.$index);
+        return $this->templateClassPrefix.hash(\PHP_VERSION_ID < 80100 ? 'sha256' : 'xxh128', $key).(null === $index ? '' : '___'.$index);
     }
 
     /**
@@ -445,7 +445,7 @@ class Environment
      */
     public function createTemplate($template, string $name = null)
     {
-        $hash = hash('sha256', $template, false);
+        $hash = hash(\PHP_VERSION_ID < 80100 ? 'sha256' : 'xxh128', $template, false);
         if (null !== $name) {
             $name = sprintf('%s (string template %s)', $name, $hash);
         } else {
--- a/src/Cache/FilesystemCache.php
+++ b/src/Cache/FilesystemCache.php
@@ -35,7 +35,7 @@ class FilesystemCache implements CacheInterface
 
     public function generateKey($name, $className)
     {
-        $hash = hash('sha256', $className);
+        $hash = hash(\PHP_VERSION_ID < 80100 ? 'sha256' : 'xxh128', $className);
 
         return $this->directory.$hash[0].$hash[1].'/'.$hash.'.php';
     }
--- a/src/Profiler/NodeVisitor/ProfilerNodeVisitor.php
+++ b/src/Profiler/NodeVisitor/ProfilerNodeVisitor.php
@@ -33,7 +33,7 @@ final class ProfilerNodeVisitor extends AbstractNodeVisitor
     public function __construct(string $extensionName)
     {
         $this->extensionName = $extensionName;
-        $this->varName = sprintf('__internal_%s', hash('sha256', $extensionName));
+        $this->varName = sprintf('__internal_%s', hash(\PHP_VERSION_ID < 80100 ? 'sha256' : 'xxh128', $extensionName));
     }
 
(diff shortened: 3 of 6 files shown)


stof:
This breaks a test assertion

tests/Fixtures/functions/template_from_string_error.test
```diff
@@ -1,5 +1,7 @@
 --TEST--
 "template_from_string" function
+--CONDITION--
+version_compare(phpversion(), '8.1.0-dev', '<')
```
stof: I'd rather have the suffixed file name for old PHP versions than for new ones (the day the min version is bumped to PHP 8.1+, we would then keep only the non-suffixed file)

GromNaN: That's pertinent. I renamed the files.

stof: btw, why not using the same `PHP_VERSION_ID` condition ?

GromNaN: Because I blindly followed other fixtures. Fixed.

src/Compiler.php
```diff
@@ -281,7 +281,7 @@ public function outdent($step = 1)
 
     public function getVarName()
     {
-        return sprintf('__internal_%s', hash('sha256', __METHOD__.$this->varNameSalt++));
+        return sprintf('__internal_%s', hash(\PHP_VERSION_ID < 80100 ? 'sha256' : 'xxh128', __METHOD__.$this->varNameSalt++));
```
GromNaN: It might be possible to remove the `hash` function in all `getVarName` methods and only concat the `varNameSalt`:

```suggestion
        return sprintf('__internal_compiler_%s', $this->varNameSalt++);
```

src/Parser.php
```diff
@@ -67,7 +67,7 @@ public function getEnvironment()
 
     public function getVarName()
     {
-        return sprintf('__internal_%s', hash('sha256', __METHOD__.$this->stream->getSourceContext()->getCode().$this->varNameSalt++));
+        return sprintf('__internal_%s', hash(\PHP_VERSION_ID < 80100 ? 'sha256' : 'xxh128', __METHOD__.$this->stream->getSourceContext()->getCode().$this->varNameSalt++));
```
GromNaN: The `varNameSalt` make it sure we don't generate the same string 2 times. Using a basic counter is enough for that.
```suggestion
        return sprintf('__internal_parser_%s', $this->varNameSalt++);
```

src/Environment.php
```diff
@@ -350,7 +350,7 @@ public function getTemplateClass($name, $index = null)
     {
         $key = $this->getLoader()->getCacheKey($name).$this->optionsHash;
 
-        return $this->templateClassPrefix.hash('sha256', $key).(null === $index ? '' : '___'.$index);
+        return $this->templateClassPrefix.hash(\PHP_VERSION_ID < 80100 ? 'sha256' : 'xxh128', $key).(null === $index ? '' : '___'.$index);
```
GromNaN, line 301: The class name is hashed twice at runtime, for every template load. Once in `FilesystemCache::generateKey` and the second here.

GromNaN (commented):
~Having a different cache in function of the PHP version might be an issue if the cache is pregenerated in a different environment and packaged with the application.~ Cache key change with PHP version.

We can remove calls to the `hash` function is some places to get better performance improvements without cache variation. See https://github.com/twigphp/Twig/pull/3601

src/Compiler.php
GromNaN: Done in an other PR. https://github.com/twigphp/Twig/pull/3601

src/Parser.php
GromNaN: Done in an other PR. https://github.com/twigphp/Twig/pull/3601

fabpot:
@GromNaN Can you rebase this PR?

GromNaN:
@fabpot Rebased on 2.x after #3601.

stof (approved):

fabpot:
Thank you @GromNaN.

Issue #3595, opened 2021-12-09
Serious performance issues when including optional templates


mahagr:
Twig should not throw exceptions when trying to load optional templates such as`{% include ['template1', 'template2', 'template3'] %}` or `{% include 'template' ignore missing %}` .

I was looking into an issue in Grav where forms were very, very slow. Grav uses a lot of nested templates with many overrides, especially with forms. As a result Twig tries to load a lot of missing files, throwing exception every time, thousands of times. During the exception, PHP seems to serialize all the objects in the function parameters by using `__debugInfo()` or if it's not found, falling back to the default serializer. This seems to be repeated for the whole call stack, and in the end the method was called for a single object 10 000 times. Multiply that with all the objects in the Twig context, and I was seeing something like 200 000 `__debugInfo()` calls in total.

This issue exists from Twig 1 to 3. I did a quick and dirty fix in our Environment override: https://github.com/getgrav/grav/blob/develop/system/src/Grav/Common/Twig/TwigEnvironment.php#L43-L46

After applying my quick and dirty fix, I'm seeing page load times decreasing from ~6s to ~0.3s with some complex pages.

Using exceptions to implement a normal workflow in the application sounds like a bug to me...?

stof:
This exception was used because `exists` was not part of the LoaderInterface in 1.x. It has been added through the ExistsLoaderInterface to improve performance in the ChainLoader (and Twig 2.x made it part of LoaderInterface itself).
Using `exists()` in the case of optional includes makes sense to me. This was probably missed because Symfony itself does not contain any such optional include, and so the performance impact was missed.
Can you open a PR against the 2.x branch with your change ?

mahagr:
Thank you for your quick response. I am going to open a PR, but I don't think my approach is the optimal one. There is some local caching going on when loading the template (especially when filesystem checks are turned off) and my quick fix ignores all of that.

I also need to check that we have found most of the cases, as there seem to be still a few exceptions thrown, just not as many as before. Twig 3 simplifies the logic, which may partly fix the issue.

fabpot:
If the logic is different, I would only optimize Twig 3.

mahagr:
The logic is the same, Twig 3 just removes some deprecated checks. 

I took another look into the code and it looks like that my fix is a good one after all, but it only fixes the exceptions with array of includes. Unfortunately `ignore missing` has been implemented by catching exceptions and needs to be fixed locally inside the (each) Node and Template classes.

[...]

PR #3596, opened 2021-12-10
Optimize Environment::resolveTemplate() to be much faster when template overrides do not exist

mahagr:
Fixes partially #3595.

You can test this by adding an object into the context which implements `__debugInfo()` with a counter in it. Make a template with

```twig
{% include ['missing_template1', 'missing_template2', 'missing_template3', 'missing_template4', 'missing_template5'] %}
```

And check how many times an exception was thrown by reading the counter. The issue gets exponentially worse if you have nested templates with overrides in each of them.

### DIFF
--- a/src/Environment.php
+++ b/src/Environment.php
@@ -501,6 +501,7 @@ class Environment
             $names = [$names];
         }
 
+        $count = \count($names);
         foreach ($names as $name) {
             if ($name instanceof Template) {
                 return $name;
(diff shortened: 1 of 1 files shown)


src/Environment.php
```diff
@@ -509,10 +510,15 @@ public function resolveTemplate($names)
                 return $name;
             }
 
+            // Optimization: Avoid throwing an exception when it would be ignored anyway.
+            if (1 !== $count && !$this->getLoader()->exists($name)) {
```
fabpot, line 514: As we are checking the count here we might refactor the code to not catch the exception in case count === 1 (see below).

mahagr: You're right. I was thinking that we still need the catch to display a proper error message, but the code will do it even without try/catch.

GromNaN (approved):

fabpot:
Thank you @mahagr.

PR #3600, opened 2021-12-13
POC: disable output buffering

### LINKED ISSUE #3599: bug: rendering 2 templates concurrently causes twig to return completely wrong results when a non-blocking operation takes place.
ref: symfony/symfony#41520

Twig uses output buffer when rendering the template, this causes issues when rendering multiple templates at once, as one template will capture the output buffer of another.

The fix that needs to happen is to make `doRender` abstract in the `Template` class, and `display` would simply `echo` the result of `doRender`.

meaning that instead of echoing the result in the rendered template, we store it in a variable, and return that instead.

However, one BC-break resulting from this is that twig functions that print/echo content, will no longer work.

---

To reproduce:

```php
<?php

declare(strict_types=1);

use Psl\Async;
use Twig\Environment;
use Twig\Loader\FilesystemLoader;

require __DIR__ . '/../vendor/autoload.php';

final class User {
    public function getUsername(): string {
        return 'azjezz';
    }
}

final class Account {
    public function getUser(): User {
        // simulate a lazy query, the database driver is non-blocking
        Async\sleep(0.02);
        return new User();
    }

    public function getThings(): array {
        // simulate a lazy query, the database driver is non-blocking
        Async\sleep(0.02);
        return [];
    }
}

Async\main(function (): int {

    $loader = new FilesystemLoader(__DIR__ . '/templates');
    $twig = new Environment($loader);

    $account = new Account();

    [$a, $b, $c, $d] = Async\parallel([
        static fn() => $twig->render('user.html.twig', ['account' => $account]),
        static fn() => $twig->render('user.html.twig', ['account' => $account]),
        static fn() => $twig->render('user.html.twig', ['account' => $account]),
        static fn() => $twig->render('user.html.twig', ['account' => $account]),
    ]);

    var_dump($a, $b, $c, $d);

    return 0;
});
```

where `user.html.twig` template contain the following:

```twig
{% extends 'layout.html.twig' %}

{% block cotnent %}
  {% for something in account.things %}

  {% endfor %}
{% endblock %}
```

and `layout.html.twig` contains the following:

```twig
<div>
Hello, {{ account.user.username }}
<hr />
{% block cotnent %}{% endblock %}
</div>
```

The result:

```
> php example/main.php
string(128) "<div>
    Hello, azjezz
    <hr />
        azjezz
    <hr />
        azjezz
    <hr />
        azjezz
    <hr />
        </div>
"
string(24) "<div>
    Hello, </div>
"
string(24) "<div>
    Hello, </div>
"
string(24) "<div>
    Hello, </div>
"
```

What happens in the example above is that when twig starts rendering a template, and reachs a non-blocking I/O call ( in this case, it's the sleep() ), it suspends the current fiber, and `Async\parallel` will switch to another task, `ob_start()` could have been called by 1 task, and another task can finish first and call `ob_get_clean()`, but it will also get all the content that has been outputed.

azjezz:
closes #3599 

---

I don't think i have enough experience with twig to fix the tests, so some help there would be great :)

This PR removes doDisplay, and replaces it with doRender in the generated templates, the template content is now stored in memory instead of output buffer, which fixes the issue mentioned in #3599 ( colliding buffers ).

This PR also is a BC break, any twig function or extension that previously did "echo" will not work as expected.

Using the script in `example/main.php`, you can see the difference between before and after.

before:

```
string(279) "<Layout>
    <Greeting>Hello, azjezz</Greeting>

        <Content>
    azjezz</Greeting>

        <Content>
    azjezz</Greeting>

        <Content>
    azjezz</Greeting>

        <Content>
            <Thing />
            <Thing />
        </Content>
</Layout>
"
string(99) "<Layout>
    <Greeting>Hello,         <Thing />
            <Thing />
        </Content>
</Layout>
"
string(99) "<Layout>
    <Greeting>Hello,         <Thing />
            <Thing />
        </Content>
</Layout>
"
string(99) "<Layout>
    <Greeting>Hello,         <Thing />
            <Thing />
        </Content>
</Layout>
"

```

after:

```
string(144) "<Layout>
    <Greeting>Hello, azjezz</Greeting>

        <Content>
            <Thing />
            <Thing />
        </Content>
</Layout>
"
string(144) "<Layout>
    <Greeting>Hello, azjezz</Greeting>

        <Content>
            <Thing />
            <Thing />
        </Content>
</Layout>
"
string(144) "<Layout>
    <Greeting>Hello, azjezz</Greeting>

        <Content>
            <Thing />
            <Thing />
        </Content>
</Layout>
"
string(144) "<Layout>
    <Greeting>Hello, azjezz</Greeting>

        <Content>
            <Thing />
            <Thing />
        </Content>
</Layout>
"
```

### DIFF
--- a/src/Node/NodeCaptureInterface.php
+++ b/src/Node/NodeCaptureInterface.php
@@ -12,7 +12,7 @@
 namespace Twig\Node;
 
 /**
- * Represents a node that captures any nested displayable nodes.
+ * Represents a node that captures any nested renderable nodes.
  *
  * @author Fabien Potencier <fabien@symfony.com>
  */
--- a/src/Template.php
+++ b/src/Template.php
@@ -136,19 +136,21 @@ abstract class Template
             throw new RuntimeError(sprintf('The template has no parent and no traits defining the "%s" block.', $name), -1, $this->getSourceContext());
         }
     }
-
+    
     /**
-     * Displays a block.
+     * Renders a block.
      *
      * This method is for internal use only and should never be called
      * directly.
      *
-     * @param string $name      The block name to display
+     * @param string $name      The block name to render
      * @param array  $context   The context
      * @param array  $blocks    The current set of blocks
      * @param bool   $useBlocks Whether to use the current set of blocks
+     *
+     * @return string The rendered block
      */
-    public function displayBlock($name, array $context, array $blocks = [], $useBlocks = true, self $templateContext = null)
+    public function renderBlock($name, array $context, array $blocks = [], $useBlocks = true, self $templateContext = null)
     {
         if ($useBlocks && isset($blocks[$name])) {
             $template = $blocks[$name][0];
@@ -168,7 +170,7 @@ abstract class Template
 
         if (null !== $template) {
             try {
-                $template->$block($context, $blocks);
+                return $template->$block($context, $blocks);
             } catch (Error $e) {
                 if (!$e->getSourceContext()) {
                     $e->setSourceContext($template->getSourceContext());
@@ -188,7 +190,7 @@ abstract class Template
                 throw $e;
             }
         } elseif (false !== $parent = $this->getParent($context)) {
-            $parent->displayBlock($name, $context, array_merge($this->blocks, $blocks), false, $templateContext ?? $this);
+            return $parent->renderBlock($name, $context, array_merge($this->blocks, $blocks), false, $templateContext ?? $this);
         } elseif (isset($blocks[$name])) {
             throw new RuntimeError(sprintf('Block "%s" should not call parent() in "%s" as the block does not exist in the parent template "%s".', $name, $blocks[$name][0]->getTemplateName(), $this->getTemplateName()), -1, $blocks[$name][0]->getSourceContext());
         } else {
@@ -221,28 +223,19 @@ abstract class Template
     }
 
     /**
-     * Renders a block.
+     * Displays a block.
      *
      * This method is for internal use only and should never be called
      * directly.
      *
-     * @param string $name      The block name to render
+     * @param string $name      The block name to display
      * @param array  $context   The context
      * @param array  $blocks    The current set of blocks
      * @param bool   $useBlocks Whether to use the current set of blocks
-     *
-     * @return string The rendered block
      */
-    public function renderBlock($name, array $context, array $blocks = [], $useBlocks = true)
+    public function displayBlock($name, array $context, array $blocks = [], $useBlocks = true)
     {
-        if ($this->env->isDebug()) {
-            ob_start();
-        } else {
-            ob_start(function () { return ''; });
-        }
-        $this->displayBlock($name, $context, $blocks, $useBlocks);
-
-        return ob_get_clean();
+        echo $this->renderBlock($name, $context, $blocks, $useBlocks);
     }
 
     /**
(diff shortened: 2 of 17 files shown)


src/Node/NodeCaptureInterface.php
```diff
@@ -12,7 +12,7 @@
 namespace Twig\Node;
 
 /**
- * Represents a node that captures any nested displayable nodes.
+ * Represents a node that captures any nested renderable nodes.
```
stof, line 15: I don't see why this is changed

src/Template.php
```diff
@@ -364,34 +357,22 @@ public function getBlocks()
 
     public function display(array $context, array $blocks = [])
     {
-        $this->displayWithErrorHandling($this->env->mergeGlobals($context), array_merge($this->blocks, $blocks));
+        echo $this->render($context, $blocks);
```
stof, line 360: this will not support streaming the displayed output anymore

src/Template.php
```diff
@@ -364,34 +357,22 @@ public function getBlocks()
...
-            while (ob_get_level() > $level) {
-                ob_end_clean();
-            }
-
-            throw $e;
-        }
-
-        return ob_get_clean();
+        // using func_get_args() allows to not expose the blocks argument
+        // as it should only be used by internal code
```
stof, line 366: why this, which was not there yet ?

src/Template.php
```diff
@@ -413,10 +394,12 @@ protected function displayWithErrorHandling(array $context, array $blocks = [])
     }
 
     /**
-     * Auto-generated method to display the template with the given context.
+     * Auto-generated method to render the template with the given context.
      *
      * @param array $context An array of parameters to pass to the template
      * @param array $blocks  An array of blocks to pass to the template
+     * 
+     * @eturn string, the rendered template.
```
stof, line 402: typo in `@return`

stof (commented):
You still rely on output buffering in `renderParentBlock` (and in a few special nodes).

And your renaming of special nodes in ModuleNode has not been reflected in extensions hooking into them.

azjezz:
> You still rely on output buffering in renderParentBlock (and in a few special nodes).
> And your renaming of special nodes in ModuleNode has not been reflected in extensions hooking into them.

I personally don't have much experience with the code in twig, so i probably need more help here ^^ or if someone else wants to do it, that would great.

since this is a BC break, is there any plans for twig 4? if not, would it make sense to prepare 4.x branch for this patch?

stof:
> since this is a BC break, is there any plans for twig 4? if not, would it make sense to prepare 4.x branch for this patch?

Twig follows the same BC policy than Symfony. This means that a BC break must also produce a continuous migration path (i.e. having a way for other packages to write code using the new approach in Twig 3.x already). As is, this patch does not allow that, and so would be rejected for Twig 4 too.

> I personally don't have much experience with the code in twig, so i probably need more help here ^^ or if someone else wants to do it, that would great.

Well, either don't rename those nodes (one less BC break for which you need to build an upgrade path) or at least search for the old names in the Twig code base itself and change them to the new names (but that won't be enough to have a valid patch due as the second alternative does not account for third-party usages of those extension points)

azjezz:
I'm closing this issue as i personally have no effort to put into working on it.

I think it would a good idea to document that the behavior in a concurrent application is unpredictable, and twig should be avoided in such an environment.

more information: https://twitter.com/azjezz/status/1476295773616848896?s=20

PR #3601, opened 2021-12-13
Hashing is not necessary to generate unique variable names during compilation

GromNaN:
Trying to optimize calls to `hash` function (#3588), I found that some calls are made during compilation that are not necessary at all.

According to the last commits on this functions (57ff88255ef35ad8da325d0cb0bf140e653b132c & 6ab5fe9b8fd52f810fd902652659605b5fcf39a9), the internal variable names must be unique and deterministic. Using a simple sequence is enough.

This avoid CPU cycles during compilation. Which should not have any impact on production; but still interesting for dev&test.

### DIFF
--- a/src/Profiler/NodeVisitor/ProfilerNodeVisitor.php
+++ b/src/Profiler/NodeVisitor/ProfilerNodeVisitor.php
@@ -28,10 +28,12 @@ use Twig\Profiler\Profile;
 final class ProfilerNodeVisitor extends AbstractNodeVisitor
 {
     private $extensionName;
+    private $varName;
 
     public function __construct(string $extensionName)
     {
         $this->extensionName = $extensionName;
+        $this->varName = sprintf('__internal_%s', hash('sha256', $extensionName));
     }
 
     protected function doEnterNode(Node $node, Environment $env)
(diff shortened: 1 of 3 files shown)


src/Profiler/NodeVisitor/ProfilerNodeVisitor.php
```diff
@@ -66,7 +67,7 @@ protected function doLeaveNode(Node $node, Environment $env)
 
     private function getVarName(): string
     {
-        return sprintf('__internal_%s', hash('sha256', $this->extensionName));
+        return sprintf('__internal_profiler_%s', $this->varNameSalt++);
```
GromNaN: Unless this class is used by multiple extensions, I'm quite sure the `varNameSalt` is not necessary.

src/Profiler/NodeVisitor/ProfilerNodeVisitor.php
```diff
@@ -66,7 +67,7 @@ protected function doLeaveNode(Node $node, Environment $env)
 
     private function getVarName(): string
     {
-        return sprintf('__internal_%s', hash('sha256', $this->extensionName));
+        return sprintf('__internal_profiler_%d', $this->varNameSalt++);
```
stof: this now returns different variable names for each call to that method for the same visitor, and won't return different names for different visitors used by different extensions (the Symfony WebProfiler integration relies on that visitor too IIRC).
So this change looks wrong to me.

stof (changes requested):

src/Profiler/NodeVisitor/ProfilerNodeVisitor.php
GromNaN: You are right. I moved the `varName` to a property, computed only 1 time.

stof (approved):

fabpot:
Thank you @GromNaN.

PR #3602, opened 2021-12-14
Store TemplateClass in compiled includes

GromNaN:
Trying to optimize calls to `hash` function (#3588), I found that the template classes could be cached inside the calling templates.

Example with the following template
```twig
{% include '@ResponsiveFront/home/views/metas.html.twig' with { Metadata } only %}
```

Compiled version:
```php
$this->loadTemplate(new TemplateClass("__TwigTemplate_c834197862571e35dc19b42e4a40c8f9826e712d24a46a6b8d08ffee77eab1bc", "@ResponsiveFront/home/views/metas.html.twig"), "@ResponsiveFront/home/views/home.html.twig", 22)->display(twig_to_array(["Metadata" => ($context["Metadata"] ?? null)]));
```

Previous compiled version:
```php
$this->loadTemplate("@ResponsiveFront/home/views/metas.html.twig", "@ResponsiveFront/home/views/home.html.twig", 22)->display(twig_to_array(["Metadata" => ($context["Metadata"] ?? null)]));
```

If the direction is OK, I'd continue with all other types of template calls (extends, embed, use, macro ...)

### DIFF
--- a/src/Template.php
+++ b/src/Template.php
@@ -321,6 +321,10 @@ abstract class Template
                 return $template;
             }
 
+            if ($template instanceof TemplateClass) {
+                return $this->env->loadClass($template->class, $template->name, $index);
+            }
+
             if ($template === $this->getTemplateName()) {
                 $class = static::class;
                 if (false !== $pos = strrpos($class, '___', -1)) {
--- a/src/Node/IncludeNode.php
+++ b/src/Node/IncludeNode.php
@@ -13,7 +13,9 @@
 namespace Twig\Node;
 
 use Twig\Compiler;
+use Twig\Error\LoaderError;
 use Twig\Node\Expression\AbstractExpression;
+use Twig\Node\Expression\ConstantExpression;
 
 /**
  * Represents an include node.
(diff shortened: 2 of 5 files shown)


src/Template.php
```diff
@@ -321,6 +321,10 @@ protected function loadTemplate($template, $templateName = null, $line = null, $
                 return $template;
             }
 
+            if ($template instanceof TemplateClass) {
+                return $this->env->loadClass($template->class, $template->name, $index);
```
GromNaN, line 325: Here is the shortcut.

stof:
I see a big issue here: the output of compiling a template now depends on the template class name of included templates, which makes cache invalidation a lot harder (if you add a new template matching a name in a higher priority location, it will not be consistently used depending on whether the templates referencing that name are already cached with the old class name or no)

GromNaN:
I try to optimize the use-case where the templates are immutable: a typical Symfony application on production. This change should not create any issue in that conditions.

But you are right, that it can create inconsistencies during development or if the loader has variable results. 

An other idea was to add cache to the `getTemplateClass` method. But I'm worried it will have the same drawbacks.

GromNaN:
The performance benefits is not worth the added complexity.

_That was interesting to dive into this code_

Issue #3607, opened 2021-12-22
symfony/translation-contracts:^3 not available for twig/string-extra


michaljusiegaBTN:
Hey!

It would be nice to allowing bump this dependency, because:

```
root@9560bf82c097:/var/www/html# composer prohibits symfony/translation-contracts 3
twig/string-extra  v3.3.4  requires  symfony/translation-contracts (^1.1|^2) 
```

I can prepare PR if you want, let me know ;)

nicolas-grekas:
Please send a PR yes!

PR #3608, opened 2021-12-22
Allow symfony/translation-contracts:^3

michaljusiega:
However, 
I am just wondering why this dependency is required since it's not used anywhere in `twig/string-extra`? Or I am wrong ?

Closing https://github.com/twigphp/Twig/issues/3607

### DIFF
--- a/extra/string-extra/composer.json
+++ b/extra/string-extra/composer.json
@@ -17,7 +17,7 @@
     "require": {
         "php": ">=7.2.5",
         "symfony/string": "^5.0|^6.0",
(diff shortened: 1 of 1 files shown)


stof:
`symfony/translation-contracts` is an optional dependency of `symfony/string` that is needed to use the AsciiSlugger. And twig/string-extra uses the AsciiSlugger.

michaljusiega:
Ok, I got it. Thank for explain @stof

fabpot:
Thank you @michaljusiega.

PR #3610, opened 2021-12-24
Add full stacktrace log to null $str given to twig_replace_filter

yookoala:
Another take on #3557.

* Add a utility class to log message to server log with full stacktrace for debug.
* Use StacktraceLogger to log null or other non-string value given to $str in twig_replace_filter, with full stacktrace for debug.

This is an attempt to address PHP future plan [to make passing null to string parameter as a TypeError](https://wiki.php.net/rfc/deprecate_null_to_scalar_internal_arg). Deprecated messages are generated in 8.1 for the "bad" cases.

Because the nature of twig extensions, the error is far too deep in the library when the deprecated message occur. And since deprecated message does not have stacktrace, the apparent log message is unhelpful for tracking down the culprit(s) in template file.

### DIFF
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -312,6 +312,7 @@ namespace {
     use Twig\Source;
     use Twig\Template;
     use Twig\TemplateWrapper;
+    use Twig\Util\StacktraceLogger;
 
 /**
  * Cycles over a value.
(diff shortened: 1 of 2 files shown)


yookoala:
Potentially, these are the filters that also need similar treatment:
* twig_urlencode_filter
* twig_upper_filter
* twig_lower_filter
* twig_capitalize_string_filter

fabpot:
Closing in favor of #3617.

PR #3611, opened 2021-12-26
Typo fix : "flexiblity" ➤ "flexibility"

Levure:


fabpot:
Good catch, thanks @Levure.

Issue #2817, opened 2019-01-17
html_attr docs are misleading / unclear


timglabisch:
i had a discussion with some of my co workers about escaping strategies and when to use 'html_attr'.

from my point of view, we should add an example when to use `html_attr` instead of `html`.

The docs are saying that:

```
html: escapes a string for the HTML body context.
html_attr: escapes a string for the HTML attribute context.
```

# example 1

following the docs, this wouldn't be best practice, but it seems to be "fine".
https://github.com/twigphp/Twig/issues/2615#issuecomment-370008276

```jinja
{% set untrusted = 'untrusted' %}
<div class="{{ untrusted }}">{{ untrusted }}</a>
```

# example 2

if you stick to the docs this should be best practice?

```jinja
{% set untrusted = 'untrusted' %}
<div class="{{ untrusted | e('html_attr') }}">{{ untrusted }}</a>
```

is there any reason why *example1* is wrong?

stof:
example 1 is quite fine here, because your attribute value is quoted, and the `html` escaping strategy already escapes quotes.

The wrong cases are these one:

- using a dynamic attribute name
- using an unquoted attribute value

That's because in these cases, a lot more thing need to be escaped (spaces and `=` for instance), which is what `html_attr` is doing.

fabpot:
@stof is right. Escaping HTML attributes depends on how you use them (single-quoted, double-quoted, no quotes). I'm closing as there is nothing to do here. I've added a note in #2950 to guide users to some docs about these topics.

PR #3614, opened 2021-12-30
html_attr: do not escape colons

gharlan:
Is it necessary to escape the `:` in `html_attr` context?

For example in this code:
```twig
{% for key, value in attributes %}
    {{ key|escape('html_attr') }}="{{ value }}"
{% endfor %}
```

I think it should be allowed to use attribute keys like `v-on:submit.prevent`.

### DIFF
--- a/src/Runtime/EscaperRuntime.php
+++ b/src/Runtime/EscaperRuntime.php
@@ -259,7 +259,7 @@ final class EscaperRuntime implements RuntimeExtensionInterface
                     throw new RuntimeError('The string to escape is not a valid UTF-8 string.');
                 }
 
-                $string = preg_replace_callback('#[^a-zA-Z0-9,\.\-_]#Su', function ($matches) {
+                $string = preg_replace_callback('#[^a-zA-Z0-9,\.\-_:]#Su', function ($matches) {
                     /**
                      * This function is adapted from code coming from Zend Framework.
                      *
--- a/tests/Runtime/EscaperRuntimeTest.php
+++ b/tests/Runtime/EscaperRuntimeTest.php
@@ -38,6 +38,7 @@ class EscaperRuntimeTest extends TestCase
         '.' => '.',
         '-' => '-',
         '_' => '_',
+        ':' => ':',
         /* Basic alnums excluded */
         'a' => 'a',
         'A' => 'A',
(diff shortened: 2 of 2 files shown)


stof:
The goal is to escape things like `javascript:foo are()` allowing to inject JS execution in href

gharlan:
Good point! 
In https://github.com/twigphp/Twig/issues/2817#issuecomment-455228795 I've learned that it is ok to use `html` escaping in quoted attribute values:

> example 1 is quite fine here, because your attribute value is quoted, and the html escaping strategy already escapes quotes.
>
>The wrong cases are these one:
>
> * using a dynamic attribute name
> * using an unquoted attribute value

So for some attributes (like `href`), `html_attr` escaping should be used for the value, even if it is quoted.

stof:
The href attribute is indeed a bit special due to links allowing to execute JS

gharlan:
Even with `html_attr` the attribute name should never come from user input, because of `onclick` attributes etc.

> The goal is to escape things like `javascript:foo are()` allowing to inject JS execution in href

```twig
{% set attr = "javascript:alert(1)" %}
<a href="{{ attr|e('html_attr') }}">Click</a>
```

will result in:

```html
<a href="javascript&#x3A;alert&#x28;1&#x29;">Click</a>
```
(https://twigfiddle.com/n1rbba)

This is still executable javascript: https://jsfiddle.net/9ekxLy6u/

---

So I'm still not sure whether the `:` should be escaped in `html_attr` context.

Could also be relevant for https://github.com/twigphp/Twig/pull/3760.
Which escaping strategies should the new `html_attributes` function/filter use for keys and values? Should it support attribute names like `v-on:submit.prevent`?

mpdude:
I am not using Vue.js myself, so I cannot tell for sure, but according to https://v2.vuejs.org/v2/guide/syntax.html#v-bind-Shorthand this might not only affect `:`, but also `@` and square brackets `[]`.

The HTML 5 spec does not exclude all those characters from attribute names (https://html.spec.whatwg.org/multipage/syntax.html#attributes-2).

XML processors will treat the colon as the XML namespace separator. 

HTML 5 allows XML only on SVG and MathML elements, and only for pre-defined namespace-prefixes (https://developer.mozilla.org/en-US/docs/Web/API/Attr/localName#:~:text=That%20means%20that%20the%20local,different%20from%20the%20qualified%20name). For other `something:` prefixes, these will simply be passed on as part of the local attribute name.

According to https://engine.sygnal.com/research/html5-attribute-names, all current browser implementations handle at least the colon fine, and the aforementioned Vue.js documentation suggests that this is also the case for `@`.

So, if we'd want to support this, I think it should at least be introduced in a BC way, i. e. with a new escaping strategy name that lets users opt into the new mode with less escaped characters.

PR #3615, opened 2022-01-02
Fix deprecation notices when passing null to the `upper` or `lower` filters with PHP 8.1

DieterHolvoet:


fabpot:
Closing in favor of #3617.

PR #3619, opened 2022-01-03
Allow null for `nl2br`

ruudk:
Same as https://github.com/twigphp/Twig/pull/3617
Fixes `Passing null to parameter #1 ($string) of type string is deprecated` on PHP 8.1

### DIFF
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -213,7 +213,7 @@ final class CoreExtension extends AbstractExtension
             // formatting filters
             new TwigFilter('date', 'twig_date_format_filter', ['needs_environment' => true]),
             new TwigFilter('date_modify', 'twig_date_modify_filter', ['needs_environment' => true]),
-            new TwigFilter('format', 'sprintf'),
+            new TwigFilter('format', 'twig_sprintf'),
             new TwigFilter('replace', 'twig_replace_filter'),
             new TwigFilter('number_format', 'twig_number_format_filter', ['needs_environment' => true]),
             new TwigFilter('abs', 'abs'),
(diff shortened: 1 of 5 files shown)


ruudk:
@fabpot ready for review. The failing test is not related.

tests/Fixtures/filters/nl2br.test
```diff
@@ -3,6 +3,7 @@
 --TEMPLATE--
 {{ "I like Twig.\nYou will like it too.\n\nEverybody like it!"|nl2br }}
 {{ text|nl2br }}
+{{ null|nl2br }}
```
fabpot:
```suggestion
*{{ null|nl2br }}*
```

So that we have `**` in the output (like I did in the referenced PR)

tests/IntegrationTest.php
```diff
@@ -229,7 +229,7 @@ public function nl2br($value, $sep = '<br />')
     {
         // not secure if $value contains html tags (not only entities)
         // don't use
-        return str_replace("\n", "$sep\n", $value);
+        return str_replace("\n", "$sep\n", $value ?? '');
```
fabpot, line 232: Why is it needed?

tests/Fixtures/filters/nl2br.test
fabpot: Can you also add a test with an empty string? (again, for consistency with the referenced PR)

src/Extension/CoreExtension.php
```diff
@@ -231,7 +231,7 @@ public function getFilters()
             new TwigFilter('lower', 'twig_lower_filter', ['needs_environment' => true]),
             new TwigFilter('striptags', 'strip_tags'),
```
fabpot: I suppose we have the same issue with this filter? And maybe all the other ones that call PHP's native function directly with no wrappers?

fabpot (changes requested):

tests/IntegrationTest.php
ruudk: Because of this:
```

  1x: str_replace(): Passing null to parameter #3 ($subject) of type array|string is deprecated
    1x in IntegrationTest::testIntegration from Twig\Tests
```

tests/Fixtures/filters/nl2br.test
ruudk: Done

src/Extension/CoreExtension.php
ruudk: I'll add the following:
* `sprintf`
* `strip_tags`

Should we also do something with `abs`? When null, pass 0?

ruudk:
@fabpot Updated the PR, added missing test for striptags filter, and made sure that striptags and format filters now also accept null inputs.

src/Extension/CoreExtension.php
```diff
@@ -481,6 +481,19 @@ function twig_date_modify_filter(Environment $env, $date, $modifier)
     return $date->modify($modifier);
 }
 
+/**
+ * Return a formatted string.
```
```suggestion
 * Returns a formatted string.
```

src/Extension/CoreExtension.php
```diff
@@ -1080,6 +1105,19 @@ function twig_lower_filter(Environment $env, $string)
     return mb_strtolower($string ?? '', $env->getCharset());
 }
 
+/**
+ * Strip HTML and PHP tags from a string.
```
```suggestion
 * Strips HTML and PHP tags from a string.
```

fabpot (changes requested):

ruudk:
Fixed

fabpot:
Thank you @ruudk.

PR #3559, opened 2021-09-14
Allow CommonMark extensions to easily be added

kbond:
1. I discovered that commonmark 2 works just fine with `markdown-extra` so I added to the `require-dev` of `markdown-extra`'s `composer.json`
2. Added `LeagueMarkdownFactory` to `markdown-extra` as discussed in #3558
3. In `twig-extra-bundle`, if commonmark (and the above factory) is available, wire up the factory with extensions added via the `twig.markdown.league_extension` DI tag

If this PR would be acceptable, I can add some tests.

(closes #3558)

PR #3623, opened 2022-01-04
[extra-bundle] add integration tests

kbond:
Should hopefully prevent a repeat of the #3559 fiasco ;)

I chose to just add integration tests for markdown (and registering custom extensions). The others are pretty straight forward but can add tests for these if desired.


fabpot:
Thank you @kbond.

PR #3626, opened 2022-01-11
The deprecated null value for the method round has been changed to zero.

Parovozzzik:


fabpot (commented):
Maybe we can move 554 as the first line of code instead?
Can you also add some tests (in Fixtures)?

Parovozzzik:
@fabpot, I don't see what you mean about 554...  we can do it like this:

```
function twig_round($value, $precision = 0, $method = 'common')
{
    if (!isset($value)) {
        return 0;
    }
...
}
```

fabpot:
Line is `$value = (float) $value;`. If you move this line to the beginning of the method, then, there is no other change to do as the value will always be a float.

tests/Fixtures/filters/round.test
```diff
@@ -1,6 +1,7 @@
 --TEST--
 "round" filter
 --TEMPLATE--
+{{ null|round }}
```
fabpot: Can you also add two other tests for the `ceil` and `floor`  methods?

stof: `floor` is already tested (that's the last case), but `ceil`  is indeed missing.

Parovozzzik: Done

stof (approved):

fabpot:
Thank you @Parovozzzik.

PR #3628, opened 2022-01-14
Rename variables used in map method

alexander-schranz:
As most people (including me) are lazy people to read text and mostly looks just at the code. Using `key` and `value` as variable make the map faster to understand which argument is which one :)

### DIFF
--- a/doc/filters/map.rst
+++ b/doc/filters/map.rst
@@ -27,7 +27,7 @@ The arrow function also receives the key as a second argument:
         "Alice": "Dupond",
     } %}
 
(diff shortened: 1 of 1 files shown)


GromNaN (approved):
The code has the same semantics.
https://github.com/twigphp/Twig/blob/1231fd4096533a4cbf49430a3eaf8583d94a5a00/src/Extension/CoreExtension.php#L1661-L1663

stof (approved):

fabpot:
Thank you @alexander-schranz.

PR #3629, opened 2022-01-14
Fix map example output

alexander-schranz:
@fabpot sorry little error in my #3628 pull request 🙈


fabpot:
Good catch, thanks @alexander-schranz.

PR #3630, opened 2022-01-17
notice: Return type of Twig\Node\Node::count() should either be compa…

Alexander1000:
Hello! On php 8.1 notice

@fabpot

### DIFF
--- a/src/Node/Node.php
+++ b/src/Node/Node.php
@@ -210,6 +210,7 @@ class Node implements \Twig_NodeInterface
     /**
      * @return int
      */
+    #[\ReturnTypeWillChange]
     public function count()
     {
         return \count($this->nodes);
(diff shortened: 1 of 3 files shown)


stof:
See this note in the Twig 1.x readme: https://github.com/twigphp/Twig/tree/1.x#readme

> **Twig version 1 is not maintained anymore.** Please, upgrade at your earliest convenience.

Such fixes are already applied in Twig 2.x and 3.x

Alexander1000:
Maybe as an exception, another patch release? There are difficulties with the transition of a huge project to a new version, but I'm working on it, I hope I don't have to patch the old version anymore.

fabpot:
Twig 1.x has been maintained for many many years, we need to end it at some point.

stof:
Twig 2.0 was released on January 5, 2017. So you already got 5 years for that transition

PR #3638, opened 2022-01-29
Fix call to deprecated "convertToHtml" method

jderusse:
note: We can not rely on the new interface `ConverterInterface` because this interface already existed in [version 1.0](https://github.com/thephpleague/commonmark/blob/1.0/src/ConverterInterface.php)

### DIFF
--- a/extra/markdown-extra/LeagueMarkdown.php
+++ b/extra/markdown-extra/LeagueMarkdown.php
@@ -16,14 +16,20 @@ use League\CommonMark\CommonMarkConverter;
 class LeagueMarkdown implements MarkdownInterface
 {
     private $converter;
+    private $legacySupport;
 
     public function __construct(CommonMarkConverter $converter = null)
     {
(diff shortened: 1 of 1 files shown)


extra/markdown-extra/LeagueMarkdown.php
```diff
@@ -16,14 +16,20 @@
 class LeagueMarkdown implements MarkdownInterface
 {
     private $converter;
+    private $methodConvert;
```
fabpot, line 19:
```suggestion
    private $convertMethodName;
```
?

fabpot:
```suggestion
    private $legacySupport;
```
?

fabpot:
Thank you @jderusse.

jderusse:
Good catch for the 2.x branch. 
I missed that twig 2.x is compatible with common mark 2.x regardless the require-dev section

PR #3653, opened 2022-02-23
POC - Opt - Add leading slash before function invocation to speed up resolving

SpacePossum:
POC/proposal

Calling functions with their namespace prefix is faster than without. This especially true for compiler optimized function, but all calls have benefit from this.

In some cases people like to either import the function or do not prefix function declared in the global namespace with `\`.
The main reason is readability, however compiled templates are typically not read.

The diff looks way big because of the utests. Wasn't sure about the branch, but maybe first talk concept and than I'm happy to adjust or close if there is no interest.

### DIFF
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -175,65 +175,65 @@ final class CoreExtension extends AbstractExtension
     {
         return [
             // formatting filters
-            new TwigFilter('date', 'twig_date_format_filter', ['needs_environment' => true]),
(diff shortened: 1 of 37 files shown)


nicolas-grekas:
Generated templates are not namespaced, so that this won't improve anything AFAIK.
But maybe I'm wrong? Can you please provide some perf numbers?

SpacePossum:
ah right, that makes sense, thanks :+1:

PR #3654, opened 2022-02-24
do not reuse internally generated variable names during parsing

### LINKED ISSUE #3647: [BC-Break] Twig 2.14.9 broke `filter` tag with `spaceless`
Hi,

we've upgraded our (legacy) code base with the security release Twig 2.14.11 and suddenly had an issue with totally strange rendering on some pages. I've tracked the issue down to the 2.14.9 release.

We've got some old templates using the `filter` tag with the `spaceless` filter:

~~~
{% filter spaceless %}
  something A
{% endfilter %}

{% filter spaceless %}
  something B
{% endfilter %}

{% filter spaceless %}
  something C
{% endfilter %}
~~~

Quote simpliefied but basically multiple tags in a bunch of blocks and template inheritance. The result is that all tags render the content of the **first** tag:

~~~
something A
something A
something A
~~~

I've checked the commits and found nothing obvious which could result in such a strange behaviour. 😕 

Cheers
Matthias

xabbuh:
fixes #3647

### DIFF
--- a/src/Parser.php
+++ b/src/Parser.php
@@ -62,7 +62,7 @@ class Parser
     public function parse(TokenStream $stream, $test = null, $dropNeedle = false)
     {
         $vars = get_object_vars($this);
-        unset($vars['stack'], $vars['env'], $vars['handlers'], $vars['visitors'], $vars['expressionParser'], $vars['reservedMacroNames']);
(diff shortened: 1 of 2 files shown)


fabpot:
Thank you @xabbuh.

PR #3656, opened 2022-02-28
Support lexing numbers with underscores and decimals with no leading zero

fancyweb:
https://github.com/symfony/symfony/pull/44073 for Twig, thanks Nicolas your work :smiley:

### DIFF
--- a/src/Lexer.php
+++ b/src/Lexer.php
@@ -34,6 +34,10 @@ class Lexer
     private $position;
     private $positions;
     private $currentVarBlockLine;
+    /**
+     * @var Token|null
+     */
+    private $lastToken;
 
     public const STATE_DATA = 0;
     public const STATE_BLOCK = 1;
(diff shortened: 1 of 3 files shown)


fabpot:
Looks like it breaks some tests.

fancyweb:
I applied some fixes but it seems that 3.x tests are already broken for another reason.

src/Lexer.php
```diff
@@ -318,12 +322,8 @@ private function lexExpression(): void
             $this->moveCursor($match[0]);
         }
         // numbers
-        elseif (preg_match(self::REGEX_NUMBER, $this->code, $match, 0, $this->cursor)) {
-            $number = (float) $match[0];  // floats
-            if (ctype_digit($match[0]) && $number <= \PHP_INT_MAX) {
-                $number = (int) $match[0]; // integers lower than the maximum
-            }
-            $this->pushToken(/* Token::NUMBER_TYPE */ 6, $number);
+        elseif (preg_match(self::REGEX_NUMBER, $this->code, $match, 0, $this->cursor) && ('.' !== $match[0][0] || !$this->lastToken || !\in_array($this->lastToken->getType(), [/* Token::NAME_TYPE */ 5, /* Token::NUMBER_TYPE */ 6], true))) {
```
fancyweb, line 325: The hard part is to keep the current behavior for array access with number keys:
`foo.0` -> `$foo[0]`
`foo.0.0` -> `$foo[0]`
`foo.0.0.0` -> `$foo[0][0]`

It's edgy because decimals are turned to integers.

fancyweb:
Oh there's still one related failure: `foo.0.bar` :cry:

src/Lexer.php
fabpot: Don't we have the same issue in ExpressionLanguage?

fancyweb: IIRC I checked it and the issue doesn't exist in ExpressionLanguage because `.` can only be used for accessing public properties and for calling methods but both cannot start with a number.

PR #3664, opened 2022-03-13
Rename variable in SecurityPolicyInterface.php

VincentLanglet:
Hi @fabpot

When using psalm, an error is reported when the name of the variable is not the same than the one in the interface.
This also could give issues with named params.

The securityPolicy implementation is using `property` as the name of the variable, which make a lot more sens https://github.com/twigphp/Twig/blob/3.x/src/Sandbox/SecurityPolicy.php#L110.

So I updated the interface. But I wasn't sure on which branch (1.x, 2.x, 3.x) I should have opened the PR.

I also wondering if you want me to rename $obj to $object maybe ?

### DIFF
--- a/src/Extension/SandboxExtension.php
+++ b/src/Extension/SandboxExtension.php
@@ -91,11 +91,11 @@ final class SandboxExtension extends AbstractExtension
         }
     }
 
-    public function checkPropertyAllowed($obj, $method, int $lineno = -1, Source $source = null)
+    public function checkPropertyAllowed($obj, $property, int $lineno = -1, Source $source = null)
     {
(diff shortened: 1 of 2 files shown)


SpacePossum (approved):

fabpot:
Great catch. I think this should be done in 2.x.
Can you also fix `SandboxExtension` which has the same issue?

VincentLanglet:
> Great catch. I think this should be done in 2.x. Can you also fix `SandboxExtension` which has the same issue?

Sure,  it's done ; and I updated the target branch.

fabpot:
Thank you @VincentLanglet.

PR #3670, opened 2022-03-24
Add "some" and "every" filters

Jean-Beru:
This PR adds 2 new tests inspired from Javascript's functions: [some](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) and [every](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every).

Usage :
```twig
Value {{ ([1, 2, 3] has some v => v == 2) ? 'found' : 'not found' }}

Key {{ ([1, 2, 3] has some (v, k) => k == 2) ? 'found' : 'not found' }}

Every values {{ ([1, 2, 3] has every v => 42 > v) ? 'match' : 'do not match' }}

Every keys {{ ([1, 2, 3] has every (v, k) => 42 > k) ? 'match' : 'do not match' }}
```

### DIFF
--- a/src/ExpressionParser.php
+++ b/src/ExpressionParser.php
@@ -80,7 +80,7 @@ class ExpressionParser
             } elseif (isset($op['callable'])) {
                 $expr = $op['callable']($this->parser, $expr);
             } else {
-                $expr1 = $this->parseExpression(self::OPERATOR_LEFT === $op['associativity'] ? $op['precedence'] + 1 : $op['precedence']);
+                $expr1 = $this->parseExpression(self::OPERATOR_LEFT === $op['associativity'] ? $op['precedence'] + 1 : $op['precedence'], true);
                 $class = $op['class'];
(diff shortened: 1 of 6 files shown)


stof:
Something returning a boolean looks like a use case for a Twig test rather than a Twig filter.

Jean-Beru:
You're right. I changed these filter to binary expression (like `starts with`). The `\Twig\ExpressionParser::parseExpression` had to be updated to allow arrow functions in right operator.

fabpot:
Thank you @Jean-Beru.

jdreesen:
This is still missing in the docs.

fabpot:
@jdreesen Would you like to work on some docs? Or maybe @Jean-Beru?

Jean-Beru:
I forgot it, my bad. This is done in https://github.com/twigphp/Twig/pull/3797.

PR #3676, opened 2022-04-07
[Documentation] Add note about simple string matching

Kern046:
During the Symfony Live we mentionned the need to update the documentation around string matching to mention that an approach was possible with the ``ìn`` operator was simpler.

As the documentation part about ``in`` was just below, I wasn't sure that an anchor link was relevant, nor a code example.

Happy to make this first tiny contribution, and available if there are more stuff to do for this PR :).

### DIFF
--- a/doc/templates.rst
+++ b/doc/templates.rst
@@ -664,6 +664,8 @@ string:
         {% if phone matches '/^[\\d\\.]+$/' %}
         {% endif %}
 
+    For simple string comparisons, the containment operator can be used.
+
 Containment Operator
 ~~~~~~~~~~~~~~~~~~~~
 


stof (approved):

doc/templates.rst
```diff
@@ -664,6 +664,8 @@ string:
         {% if phone matches '/^[\\d\\.]+$/' %}
         {% endif %}
 
+    For simple string comparisons, the containment operator can be used.
```
fabpot, line 667: I would also add an example.

GromNaN: There is an example 12 lines down.

Kern046: Yeah I hesitated about that, but the fact that the containment operator section is just below would have seemed kinda duplicated to me. I also thought of an anchor but it didn't seem relevant to me either. But indeed nothing ensures that it won't move someday.

No problem if you'd like me to add one anyway :).

fabpot:
Thank you @Kern046.

PR #3687, opened 2022-04-18
throwing syntax error when the matches regexp is not valid

WebMamba:
During the symfony live, Fabien talk about a PR he did on the Expression Language component (https://github.com/symfony/symfony/pull/45875/files). The goal was to throw a more readable error when using the matchs operator. This think need to be fix in twig to. So as promises, here is my PR ! 
This is my first contrib to symfony so any advice is welcome ! 😁


fabpot:
Thank you @WebMamba.

Issue #3685, opened 2022-04-14
markdown_to_html different/incorrect behavior with leading line-break


Sarke:
I thought I was going mad at one point.  I've simplified the issue to the following code:

```php
$loader = new \Twig\Loader\ArrayLoader([
    'index' => '{{ text|markdown_to_html }}',
]);
$twig = new \Twig\Environment($loader);
$twig->addExtension(new MarkdownExtension());

$twig->addRuntimeLoader(new class implements RuntimeLoaderInterface {
    public function load($class) {
        if (MarkdownRuntime::class === $class) {
            return new MarkdownRuntime(new DefaultMarkdown());
        }
    }
});

$markdown = <<<MD
Paragraph 1

Paragraph 2

1. First

2. Second
MD;

echo $twig->render('index', ['text' => $markdown]);

echo "\n------\n\n";

// only difference is the leading line-break
$markdown = "\n" . $markdown;

echo $twig->render('index', ['text' => $markdown]);
```

### Output

```html
<p>Paragraph 1</p>

<p>Paragraph 2</p>

<ol>
<li><p>First</p></li>
<li><p>Second</p></li>
</ol>

------

<p>Paragraph 1<br />
Paragraph 2<br />
1. First<br />
2. Second</p>
```

---

I originally noticed it when I was getting the second (incorrect) results with this, which seems harmless enough.

```
{% apply markdown_to_html %}

Title
```

Similar effects with other markdown libraries.  Note that passing a leading line-break to the markdown library directly (without Twig) works as expected.

```
michelf/php-markdown               1.9.1
twig/markdown-extra                v3.3.8
twig/twig                          v3.3.9
```

stof:
The filter has a behavior to remove the indentation: https://github.com/twigphp/Twig/blob/b4d6723715da57667cca851051eba3786714290d/extra/markdown-extra/MarkdownRuntime.php#L25-L28

This seems to be what breaks when the markdown start with an empty line, because it will then remove the empty line between `Paragraph 1` and `Paragraph 2`

Sarke:
Well that doesn't seem like a good way of doing it.  Not only for my issue of no indentation, but markdown can legitimately start with an indented code block, followed by non-indented paragraphs, etc.

A workaround for now would be to do `{{ text | trim | markdown_to_html }}`, which solves my problem, but it wouldn't work with the mentioned indented code block.

Sarke:
https://github.com/twigphp/markdown-extra/pull/6

[...]

PR #3688, opened 2022-04-19
Fix for twigphp/Twig/issues/3685 (markdown leading linebreaks and indentation)

Sarke:
Fix for #3685.

Reworked the "remove indentation" to properly account for all lines, work with leading linebreaks, and retain codeblocks at start of markdown.

Reworked the tests, to among other things compare with the markdown generated by the markdown library directly.

I forked `2.x` assuming it can then be merged into `3.x` as well.  Let me know if this is incorrect (I didn't see a `CONTRIBUTING.md`).

### DIFF
--- a/extra/markdown-extra/MarkdownRuntime.php
+++ b/extra/markdown-extra/MarkdownRuntime.php
@@ -22,11 +22,37 @@ class MarkdownRuntime
 
     public function convert(string $body): string
     {
-        // remove indentation
-        if ($white = substr($body, 0, strspn($body, " \t\r\n\0\x0B"))) {
-            $body = preg_replace("{^$white}m", '', $body);
-        }
+        $body = $this->commonWhitespace($body);
+        $body = $this->removeIndentation($body);
 
         return $this->converter->convert($body);
+    }
+
+    protected function commonWhitespace(string $body): string
+    {
+        return str_replace(["\t", "\0", "\x0B"], ['    ', '', ''], $body);
+    }
+
+    protected function removeIndentation(string $body): string
+    {
+        $indent = $this->minIndentations($body);
+        if ($indent > 0) {
+            $body = preg_replace("{^ {{$indent}}}m", '', $body);
+        }
+
+        return $body;
+    }
+
+    protected function minIndentations(string $body): int
+    {
+        $non_empty_lines = preg_split('%(\r|\n)%', $body, -1, PREG_SPLIT_NO_EMPTY);
+
+        $list = [];
+        foreach ($non_empty_lines as $line)
+        {
+            $list[] = strspn($line, " ");
+        }
+
+        return min($list);
     }
 }
--- a/extra/markdown-extra/MarkdownRuntime.php
+++ b/extra/markdown-extra/MarkdownRuntime.php
@@ -28,12 +28,12 @@ class MarkdownRuntime
         return $this->converter->convert($body);
     }
 
-    protected function commonWhitespace(string $body): string
+    private function commonWhitespace(string $body): string
     {
         return str_replace(["\t", "\0", "\x0B"], ['    ', '', ''], $body);
     }
 
-    protected function removeIndentation(string $body): string
+    private function removeIndentation(string $body): string
     {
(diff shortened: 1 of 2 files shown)


extra/markdown-extra/MarkdownRuntime.php
```diff
@@ -22,11 +22,41 @@ public function __construct(MarkdownInterface $converter)
...
 
         return $this->converter->convert($body);
     }
+
+    protected function commonWhitespace(string $body): string
+    {
+        return str_replace(["\t", "\0", "\x0B"], ['    ', '', ''], $body);
+    }
+
+    protected function removeIndentation(string $body): string
```
stof: please make those method protected. There is no reason to make them extension points of the class that would then need to be covered by BC.

stof (changes requested):

stof:
> I forked 2.x assuming it can then be merged into 3.x as well. Let me know if this is incorrect

This is correct. Twig follows the same policy than Symfony: fixes are merged in the lowest maintained branch, and then the maintainers are merging branches up.

extra/markdown-extra/MarkdownRuntime.php
Sarke: I assume you mean private?

stof: yeah indeed

extra/markdown-extra/MarkdownRuntime.php
```diff
@@ -22,11 +22,41 @@ public function __construct(MarkdownInterface $converter)
...
+        if ($indent > 0) {
+            $body = preg_replace("{^ {{$indent}}}m", '', $body);
+        }
+
+        return $body;
+    }
+
+    private function minIndentations(string $body): int
+    {
+        $non_empty_lines = preg_split('%(\r|\n)%', $body, -1, PREG_SPLIT_NO_EMPTY);
```
fabpot, line 48:
```suggestion
        $nonEmptyLines = preg_split('%(\r|\n)%', $body, -1, PREG_SPLIT_NO_EMPTY);
```

extra/markdown-extra/MarkdownRuntime.php
```diff
@@ -22,11 +22,41 @@ public function __construct(MarkdownInterface $converter)
 
     public function convert(string $body): string
     {
-        // remove indentation
-        if ($white = substr($body, 0, strspn($body, " \t\r\n\0\x0B"))) {
-            $body = preg_replace("{^$white}m", '', $body);
-        }
+        $body = $this->commonWhitespace($body);
```
fabpot, line 25: I don't think we should change the content, except for maybe the beginning of the lines. Here, you are removing all occurrences.

extra/markdown-extra/Tests/FunctionalTest.php
```diff
@@ -27,60 +27,79 @@ class FunctionalTest extends TestCase
...
+
+            public function __construct(string $class)
+            {
+                $this->class = $class;
+            }
+
+            public function load($c)
+            {
+                if (MarkdownRuntime::class === $c)
+                {
```
fabpot, line 98: Should be on the previous line

extra/markdown-extra/Tests/FunctionalTest.php
```diff
@@ -27,60 +27,79 @@ class FunctionalTest extends TestCase
...
+                    return new $c(new $this->class());
```
fabpot, line 99: Missing empty line above.

extra/markdown-extra/Tests/FunctionalTest.php
```diff
@@ -27,60 +27,79 @@ class FunctionalTest extends TestCase
...
 =====
 
 Great!
-{% endapply %}
 EOF
             , "<h1>Hello</h1>\n+<p>Great!</p>"],
+
             [<<<EOF
-{% apply markdown_to_html %}
-    Hello
```
fabpot, line 76: Can we keep this example in the tests?

fabpot (changes requested):

PR #3675, opened 2022-04-05
Enable cache invalidation when auto_reload=true

GromNaN:
Related to https://github.com/symfony/symfony-docs/pull/16671

Enabling `auto_reload` is not effective when `opcache.validate_timestamps=0`.

PR #3690, opened 2022-04-20
prevent direct invocation of compiled files under docroot

will0:
When the template compiler stores .php files, many setups allow those .php files to be invoked by external traffic. This may be provably non-problematic here beyond errors in logs, but at least [one managed host](https://wpengine.com/support/wp-engines-security-environment/#Disk_Write_Protection) prevents writes to .php files at runtime because of the large number of vulnerabilities in the category.

### DIFF
--- a/src/Cache/FilesystemCache.php
+++ b/src/Cache/FilesystemCache.php
@@ -33,7 +33,7 @@ class FilesystemCache implements CacheInterface
     {
         $hash = hash(\PHP_VERSION_ID < 80100 ? 'sha256' : 'xxh128', $className);
 
-        return $this->directory.$hash[0].$hash[1].'/'.$hash.'.php';
+        return $this->directory.$hash[0].$hash[1].'/'.$hash.'.compiled';
     }
 
     public function load(string $key): void
--- a/src/Cache/FilesystemCache.php
+++ b/src/Cache/FilesystemCache.php
@@ -22,18 +22,20 @@ class FilesystemCache implements CacheInterface
 
     private $directory;
     private $options;
+    private $extension;
 
-    public function __construct(string $directory, int $options = 0)
+    public function __construct(string $directory, int $options = 0, $extension = 'php')
     {
         $this->directory = rtrim($directory, '\/').'/';
         $this->options = $options;
+        $this->extension = $extension;
     }
 
     public function generateKey(string $name, string $className): string
     {
         $hash = hash(\PHP_VERSION_ID < 80100 ? 'sha256' : 'xxh128', $className);
 
-        return $this->directory.$hash[0].$hash[1].'/'.$hash.'.compiled';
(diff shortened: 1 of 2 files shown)


will0:
I see the confused emoji, and I'll attempt to add some of the explanation covered in a near identical change I proposed in https://github.com/laravel/framework/pull/42066

Not raising a security issue, but a compatibility issue. WP Engine prevents some request contexts from writing `.php` files to disk because file write vulnerabilities are a common class of issue with high severity. WordPress theme and plugin authors commonly use Twig for templating. The combination is that people who deploy those themes and plugins at WP Engine don't have a way to work around the issue of the templating system needing to write `.php` files in the course of serving a request.

In the Laravel thread, a wise suggestion was made to instead make the file suffix configurable and maintain the same default of `.php`.

Does that help?

xabbuh:
This does not look like a good solution for the described problem. If the files are accessible through HTTP, using another extension than `.php` means that the raw template content will be served and thus you will risk exposing sensitive data.

will0:
@xabbuh thats a fair point - I am not deeply versed in the use cases, my assumption was that the template content would typically be the skeleton of page content. The literal thing I'm trying to solve though is that the files are not allowed to be written on the platform in many scenarios, and my understanding this defeats the opcache optimization of using the file cache, and results in repeated recompilation of the source template. It reads like this shouldn't be a direct functional problem, but I can't quite tell.

GromNaN:
The opcache optimization issue should be solved by https://github.com/twigphp/Twig/pull/3675.

Have you tried the proposed change on the problematic environment? Does it solve the issue?

You could use a custom `CacheInterface` instance in the project.

will0:
@GromNaN the change does produce the desired behavior - the compiled templates are written with a different extension, rather than the environment denying write access to the .php file. The opcache doesn't care about the extension for whether it cached the opcache-compiled php content, and so the full performance improvement offered by the cache works.

Your point about a custom CacheInterface is well taken. The reason I'm aiming to contribute upstream is because this issue is encountered by many users.

I took some time to bettter understand how the Twig fits together. I'm going to expand the scope of this PR to make the extension configurable, but remain .php by default. This will require consuming code opt-in to get the behavior change.

will0:
What I'm thinking is to make it easy for [this twig env setup](https://github.com/timber/timber/blob/75bb47223e32b787fd74e92f692f9fc34dec7e74/lib/Loader.php#L172-L174) to construct a FileSystemCache that has the extension configured, without having a customized implementation of the cache - I expect that to be a representative sample of the changes that might be required by consumers to opt in.

fabpot:
Twig cache files should never be stored under the web root directory. If they are, they should be protected from being accessible from the web server. This is a potential security issue.
I don't think having an option to work around this in core is desirable.

will0:
Thanks for a clear call @fabpot, I'll proceed as best I can with that guidance.

PR #3702, opened 2022-05-13
Fix support for named closures

nicolas-grekas:
As spotted by @stof on https://github.com/symfony/symfony/pull/46335#discussion_r872108567

### DIFF
--- a/src/ExtensionSet.php
+++ b/src/ExtensionSet.php
@@ -149,8 +149,6 @@ final class ExtensionSet
             throw new \LogicException(sprintf('Unable to register extension "%s" as it is already registered.', $class));
         }
 
-        // For BC/FC with namespaced aliases
-        $class = (new \ReflectionClass($class))->name;
         $this->extensions[$class] = $extension;
     }
 
--- a/src/Node/Expression/CallExpression.php
+++ b/src/Node/Expression/CallExpression.php
@@ -24,21 +24,21 @@ abstract class CallExpression extends AbstractExpression
     {
         $callable = $this->getAttribute('callable');
 
-        $closingParenthesis = false;
-        $isArray = false;
         if (\is_string($callable) && false === strpos($callable, '::')) {
             $compiler->raw($callable);
         } else {
-            list($r, $callable) = $this->reflectCallable($callable);
-            if ($r instanceof \ReflectionMethod && \is_string($callable[0])) {
-                if ($r->isStatic()) {
+            [$r, $callable] = $this->reflectCallable($callable);
+
+            if (\is_string($callable)) {
+                $compiler->raw($callable);
+            } elseif (\is_array($callable) && \is_string($callable[0])) {
+                if (!$r instanceof \ReflectionMethod || $r->isStatic()) {
                     $compiler->raw(sprintf('%s::%s', $callable[0], $callable[1]));
                 } else {
                     $compiler->raw(sprintf('$this->env->getRuntime(\'%s\')->%s', $callable[0], $callable[1]));
                 }
-            } elseif ($r instanceof \ReflectionMethod && $callable[0] instanceof ExtensionInterface) {
-                // For BC/FC with namespaced aliases
-                $class = (new \ReflectionClass(\get_class($callable[0])))->name;
+            } elseif (\is_array($callable) && $callable[0] instanceof ExtensionInterface) {
+                $class = \get_class($callable[0]);
                 if (!$compiler->getEnvironment()->hasExtension($class)) {
                     // Compile a non-optimized call to trigger a \Twig\Error\RuntimeError, which cannot be a compile-time error
                     $compiler->raw(sprintf('$this->env->getExtension(\'%s\')', $class));
@@ -48,17 +48,11 @@ abstract class CallExpression extends AbstractExpression
 
                 $compiler->raw(sprintf('->%s', $callable[1]));
             } else {
-                $closingParenthesis = true;
-                $isArray = true;
-                $compiler->raw(sprintf('call_user_func_array($this->env->get%s(\'%s\')->getCallable(), ', ucfirst($this->getAttribute('type')), $this->getAttribute('name')));
+                $compiler->raw(sprintf('$this->env->get%s(\'%s\')->getCallable()', ucfirst($this->getAttribute('type')), $this->getAttribute('name')));
             }
         }
 
-        $this->compileArguments($compiler, $isArray);
-
-        if ($closingParenthesis) {
-            $compiler->raw(')');
-        }
+        $this->compileArguments($compiler);
     }
 
     protected function compileArguments(Compiler $compiler, $isArray = false)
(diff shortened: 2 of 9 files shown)


src/ExtensionSet.php
```diff
@@ -149,8 +149,6 @@ public function addExtension(ExtensionInterface $extension)
             throw new \LogicException(sprintf('Unable to register extension "%s" as it is already registered.', $class));
         }
 
-        // For BC/FC with namespaced aliases
-        $class = (new \ReflectionClass($class))->name;
```
nicolas-grekas, line 153: not needed because `$class = \get_class($extension)`, which means aliases are already resolved

src/Node/Expression/CallExpression.php
```diff
@@ -48,17 +45,11 @@ protected function compileCallable(Compiler $compiler)
 
                 $compiler->raw(sprintf('->%s', $callable[1]));
             } else {
-                $closingParenthesis = true;
-                $isArray = true;
-                $compiler->raw(sprintf('call_user_func_array($this->env->get%s(\'%s\')->getCallable(), ', ucfirst($this->getAttribute('type')), $this->getAttribute('name')));
+                $compiler->raw(sprintf('$this->env->get%s(\'%s\')->getCallable()', ucfirst($this->getAttribute('type')), $this->getAttribute('name')));
```
nicolas-grekas, line 51: PHP 7.1 means we can remove using `call_user_func_array()` and use an inline call instead.

src/Node/Expression/CallExpression.php
```diff
@@ -245,10 +236,7 @@ protected function normalizeName($name)
 
     private function getCallableParameters($callable, bool $isVariadic): array
     {
-        list($r) = $this->reflectCallable($callable);
-        if (null === $r) {
```
nicolas-grekas, line 249: a reflection object is now always returned, even for magic methods

stof (changes requested):
This needs to add tests in FunctionTest showing that using Closure::fromCallable on an extension method actually compiles to the optimized `$this->extensions['extension_class_name']->method()` call rather than to `$this->env->getFunction('foo')->getCallable()` which triggers a full initialization of extensions.

nicolas-grekas:
Test added.

stof:
The tests you added don't cover the case of calling extension methods

nicolas-grekas:
Test added :)

stof (approved):

fabpot:
Thank you @nicolas-grekas.

LukeTowers:
@nicolas-grekas this PR has caused an issue with callables that point to Laravel facades in Winter CMS: https://github.com/wintercms/winter/issues/557#issuecomment-1160480419. Do you have any idea what would be causing that and how we can go about fixing it?

nicolas-grekas:
It looks the code should be changed like this:
```diff
--- a/src/Node/Expression/CallExpression.php
+++ b/src/Node/Expression/CallExpression.php
@@ -305,8 +305,8 @@ abstract class CallExpression extends AbstractExpression
             $callable = [$object, $r->name];
             $callableName = (\function_exists('get_debug_type') ? get_debug_type($object) : \get_class($object)).'::'.$r->name;
         } elseif ($class = $r->getClosureScopeClass()) {
-            $callable = [$class, $r->name];
-            $callableName = $class.'::'.$r->name;
+            $callable = [$class->name, $r->name];
+            $callableName = $class->name.'::'.$r->name;
         } else {
```

Would you be available to send this with a test case, please?

LukeTowers:
Off the top of my head, it would probably look something like this:

```php
class CallStaticCallable
{
    public static function __callStatic($name, $arguments)
    {
        if ($name === 'magicMethod') {
            return 'mischief managed';
        }
    }
}

$callable = [CallStaticCallable::class, 'magicMethod'];
```
Is that enough for you to go off of @nicolas-grekas?

nicolas-grekas:
That should definitely help. Would you be up to submit a PR with everything wrapped up?

LukeTowers:
At the moment I don't have the time unfortunately, however perhaps @mjauvin, @damsfx, @webvpf, or @arvislacis would be willing to as the issue is directly impacting them.

bennothommo:
@nicolas-grekas here's a patch that has the full test case:

```diff
diff --git a/tests/Node/Expression/FilterTest.php b/tests/Node/Expression/FilterTest.php
index ef7c82cb..1a86e800 100644
--- a/tests/Node/Expression/FilterTest.php
+++ b/tests/Node/Expression/FilterTest.php
@@ -49,6 +49,7 @@ class FilterTest extends NodeTestCase
                 return [
                     new TwigFilter('foo', \Closure::fromCallable([$this, 'foo'])),
                     new TwigFilter('foobar', \Closure::fromCallable([$this, 'foobar'])),
+                    new TwigFilter('alias_bar', ['Twig_Test_Alias_Class', 'test']),
                 ];
             }
 
@@ -135,6 +136,10 @@ class FilterTest extends NodeTestCase
         $node = $this->createFilter($string, 'foobar');
         $tests[] = [$node, '$this->env->getFilter(\'foobar\')->getCallable()("abc")', $environment];
 
+        // aliased class
+        $node = $this->createFilter($string, 'alias_bar');
+        $tests[] = [$node, '$this->env->getFilter(\'alias_bar\')->getCallable()("abc")', $environment];
+
         return $tests;
     }
 
@@ -190,3 +195,31 @@ function twig_tests_filter_dummy()
 function twig_tests_filter_barbar($context, $string, $arg1 = null, $arg2 = null, array $args = [])
 {
 }
+
+// Sample class with a magic method
+class Twig_Test_Parent_Class
+{
+    public static function getAccessor()
+    {
+        throw new \Exception('Accessor not set');
+    }
+
+    public static function __callStatic($name, $arguments)
+    {
+        $accessor = static::getAccessor();
+
+        if ($name === 'test') {
+            return 'received ' . $arguments[0];
+        }
+    }
+}
+
+class Twig_Test_Child_Class extends Twig_Test_Parent_Class
+{
+    public static function getAccessor()
+    {
+        return 'child';
+    }
+}
+
+class_alias(Twig_Test_Child_Class::class, 'Twig_Test_Alias_Class');
```

It passes in the test case, in the sense that it creates the compiled `getFilter` call just fine - the issue would only manifest if one were to run the actual template.

What seems to happen is that the [`$r->getClosureScopeClass()`](https://github.com/twigphp/Twig/blob/163f07400a27376b3ccaa6d75cf3e163ed6fbf30/src/Node/Expression/CallExpression.php#L301) call here returns the class as `Twig_Test_Parent_Class`. I assume this is because that is the class that will actually fulfill the call. The problem is calling that class directly would fail because of the `getAccessor` method, whereas calling it in the context of the child class (or the alias) would be fine normally. This is similar to how Laravel facades would work.

I hope that makes sense - let me know if you need any further info.

stof:
Could you open an issue with the details instead of commenting on a merged PR ? It would avoid loosing track of the issue.

PR #3704, opened 2022-05-16
Fix optimizing non-public named closures

### LINKED ISSUE #3703: New named callable handling breaks for private methods
As reported in https://github.com/getsentry/sentry-symfony/issues/627 + https://github.com/getsentry/sentry-symfony/pull/628, the Sentry bundle breaks with Twig 3.4.0. This seems related with #3702 and to the fact that our bundle used a closure to invoke a private method: https://github.com/getsentry/sentry-symfony/blob/6177d71f2fd65f4f0f77036a1ca910aa9326767f/src/Twig/SentryExtension.php#L29-L39

Is this intended? Or does it require a fix?

nicolas-grekas:
Fix #3703

### DIFF
--- a/src/Node/Expression/CallExpression.php
+++ b/src/Node/Expression/CallExpression.php
@@ -289,6 +289,7 @@ abstract class CallExpression extends AbstractExpression
             return $this->reflector = [$r, $callable, $r->class.'::'.$r->name];
         }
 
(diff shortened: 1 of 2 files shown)


stof (approved):

fabpot:
Thank you @nicolas-grekas.

PR #3719, opened 2022-06-27
Allow inherited magic method to still run with calling class

### LINKED ISSUE #3716: Aliased/extended classes may not be reflected correctly when creating filter expressions
The class reflection changes that were introduced in https://github.com/twigphp/Twig/commit/bb64a49441a00787ab2ea8f322b450af34f2778f may cause issues with classes that have `__call` or `__callStatic` magic method that belong to a parent class, like Laravel facades.

I've provided a test case as a patch below.

```diff
diff --git a/tests/Node/Expression/FilterTest.php b/tests/Node/Expression/FilterTest.php
index ef7c82cb..1a86e800 100644
--- a/tests/Node/Expression/FilterTest.php
+++ b/tests/Node/Expression/FilterTest.php
@@ -49,6 +49,7 @@ class FilterTest extends NodeTestCase
                 return [
                     new TwigFilter('foo', \Closure::fromCallable([$this, 'foo'])),
                     new TwigFilter('foobar', \Closure::fromCallable([$this, 'foobar'])),
+                    new TwigFilter('alias_bar', ['Twig_Test_Alias_Class', 'test']),
                 ];
             }
 
@@ -135,6 +136,10 @@ class FilterTest extends NodeTestCase
         $node = $this->createFilter($string, 'foobar');
         $tests[] = [$node, '$this->env->getFilter(\'foobar\')->getCallable()("abc")', $environment];
 
+        // aliased class
+        $node = $this->createFilter($string, 'alias_bar');
+        $tests[] = [$node, '$this->env->getFilter(\'alias_bar\')->getCallable()("abc")', $environment];
+
         return $tests;
     }
 
@@ -190,3 +195,31 @@ function twig_tests_filter_dummy()
 function twig_tests_filter_barbar($context, $string, $arg1 = null, $arg2 = null, array $args = [])
 {
 }
+
+// Sample class with a magic method
+class Twig_Test_Parent_Class
+{
+    public static function getAccessor()
+    {
+        throw new \Exception('Accessor not set');
+    }
+
+    public static function __callStatic($name, $arguments)
+    {
+        $accessor = static::getAccessor();
+
+        if ($name === 'test') {
+            return 'received ' . $arguments[0];
+        }
+    }
+}
+
+class Twig_Test_Child_Class extends Twig_Test_Parent_Class
+{
+    public static function getAccessor()
+    {
+        return 'child';
+    }
+}
+
+class_alias(Twig_Test_Child_Class::class, 'Twig_Test_Alias_Class');
```

It passes in the test case, in the sense that it creates the compiled `getFilter` call just fine - the issue would only manifest if one were to run the actual template, as it would throw the `Accessor not set` exception due to the [`$r->getClosureScopeClass()`](https://github.com/twigphp/Twig/blob/163f07400a27376b3ccaa6d75cf3e163ed6fbf30/src/Node/Expression/CallExpression.php#L301) call here returning the class as `Twig_Test_Parent_Class` - ostensibly because this is the class that will actually fulfill the `callable`.

Refs:
- https://github.com/wintercms/winter/issues/557#issuecomment-1160480419
- https://github.com/twigphp/Twig/pull/3702#issuecomment-1160511737

bennothommo:
If a static method cannot be resolved to the calling class, but the calling class has, or inherits, a `__callStatic` handler, this allows the `__callStatic` handler to be used with the calling class, and not the inherited class as would occur with reflection. This allows systems such as Laravel facades to still work.

Fixes https://github.com/twigphp/Twig/issues/3716

### DIFF
--- a/src/Node/Expression/CallExpression.php
+++ b/src/Node/Expression/CallExpression.php
@@ -283,8 +283,15 @@ abstract class CallExpression extends AbstractExpression
             $callable = [substr($callable, 0, $pos), substr($callable, 2 + $pos)];
         }
 
-        if (\is_array($callable) && method_exists($callable[0], $callable[1])) {
-            $r = new \ReflectionMethod($callable[0], $callable[1]);
+        if (\is_array($callable) && \is_callable($callable[0], $callable[1])) {
+            if (!method_exists($callable[0], $callable[1]) && method_exists($callable[0], '__callStatic')) {
+                $r = new \ReflectionMethod($callable[0], '__callStatic');
+                $callable = function (...$args) use ($callable) {
+                    return $callable[0]::__callStatic($callable[1], $args);
+                };
+            } else {
+                $r = new \ReflectionMethod($callable[0], $callable[1]);
+            }
 
             return $this->reflector = [$r, $callable, $r->class.'::'.$r->name];
         }
--- a/tests/Fixtures/functions/inherited_magic_static_call.test
+++ b/tests/Fixtures/functions/inherited_magic_static_call.test
@@ -1,0 +1,10 @@
+--TEST--
+inherited __staticCall calls
+--TEMPLATE--
+{{ 'foo'|inherited_magic_call_string }}
+{{ 'foo'|inherited_magic_call_array }}
+--DATA--
+return []
+--EXPECT--
+inherited_static_magic_child_foo
+inherited_static_magic_child_foo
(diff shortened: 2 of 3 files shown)


bennothommo:
If this fix is acceptable, I'm happy to do a PR to the `2.x` branch as well.

LukeTowers:
@bennothommo are you able to add the test case that fails before this change and passes after too?

bennothommo:
@LukeTowers if you mean my original test patch, that works irrespective of this fix. The issue only manifests when rendering the template. I'm not sure where to put that into a test.

LukeTowers:
@stof @nicolas-grekas any guidance on tests for this PR?

stof:
We do have integration tests that render templates

bennothommo:
Found them - thanks @stof.

nicolas-grekas:
Thanks for the PR. I've ported it to #3722

PR #3720, opened 2022-06-29
Added a loop that uses keys and values

Trkzi-Omar:
If we might need both the keys and the values in a Twig loop. It would be useful.
Solves [this Stackoverflow question](https://stackoverflow.com/questions/58319883/not-able-to-print-the-key-and-values-in-twig-files).


fabpot:
This is not the right place. Anyway, it's already documented here: https://twig.symfony.com/doc/3.x/tags/for.html#iterating-over-keys

PR #3722, opened 2022-07-06
Allow inherited magic method to still run with calling class

nicolas-grekas:
This is #3719 ported to 2.x and improved a bit.

I also reported https://github.com/php/php-src/issues/8932 because the underlying behavior of PHP is broken to me.

If a static method cannot be resolved to the calling class, but the calling class has, or inherits, a `__callStatic` handler, this allows the `__callStatic` handler to be used with the calling class, and not the inherited class as would occur with reflection. This allows systems such as Laravel facades to still work.

Fixes https://github.com/twigphp/Twig/issues/3716


stof (approved):

fabpot:
Thank you @nicolas-grekas.

PR #3724, opened 2022-07-16
Adds dump_json function and documentation

nfreader:
Good morning! 

I built this extension locally last night for a project I'm working on and it occurred to me that someone else might find it useful. 

This PR adds a new debug function, `dump_json` which internally uses PHP's native `json_encode` function and `echo`s the resulting string. Also includes a documentation page.

### DIFF
--- a/doc/functions/dump_json.rst
+++ b/doc/functions/dump_json.rst
@@ -48,7 +48,7 @@ dumped:
 
 .. note::
 
-    Internally, Twig uses the PHP `json_encode` function.
+    Internally, Twig uses the PHP `json_encode` function, which only encodes the public properties of an object.
 
 Arguments
 ---------
--- a/doc/functions/dump_json.rst
+++ b/doc/functions/dump_json.rst
@@ -30,7 +30,7 @@ In an HTML context, wrap the output with a ``<script>`` tag to display it in the
 .. code-block:: html+twig
 
     <script>
-        console.log({{ dump(user) }})
+        console.log({{ dump_json(user) }})
     </script>
 
 You can debug several variables by passing them as additional arguments:
--- a/doc/functions/dump_json.rst
+++ b/doc/functions/dump_json.rst
@@ -1,0 +1,56 @@
+``dump_json``
+========
+
+The ``dump_json`` function dumps information about a template variable. This is
+mostly useful to debug a template that does not behave as expected by
+introspecting its variables:
+
+.. code-block:: twig
+
+    {{ dump_json(user) }}
+
+.. note::
+
+    The ``dump_json`` function is not available by default. You must add the
+    ``\Twig\Extension\DebugExtension`` extension explicitly when creating your Twig
+    environment::
+
+        $twig = new \Twig\Environment($loader, [
+            'debug' => true,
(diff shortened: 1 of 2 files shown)


src/Extension/DebugExtension.php
```diff
@@ -61,4 +63,27 @@ function twig_var_dump(Environment $env, $context, ...$vars)
...
+    if (!$vars) {
+        $vars = [];
+        foreach ($context as $key => $value) {
+            if (!$value instanceof Template && !$value instanceof TemplateWrapper) {
+                $vars[$key] = $value;
+            }
+        }
+        echo json_encode($vars);
+    } else {
+        echo json_encode(...$vars);
```
stof, line 84: why using `echo` with an output buffer instead of just `return` ?

doc/functions/dump_json.rst
```diff
@@ -0,0 +1,56 @@
...
+    Even when enabled, the ``dump_json`` function won't display anything if the
+    ``debug`` option on the environment is not enabled (to avoid leaking debug
+    information on a production server).
+
+In an HTML context, wrap the output with a ``<script>`` tag to display it in the browser console:
+
+.. code-block:: html+twig
+
+    <script>
+        console.log({{ dump(user) }})
```
stof: this uses the wrong function

stof (commented):
I'm not sure this should go in the core. `json_encode` is quite bad at dumping PHP objects, as it only dumps public properties.

src/Extension/DebugExtension.php
nfreader: echo seemed like it made more sense here, but if there's a reason return is preferred I'd love to see it :) Couldn't find much on Google about it.

stof: Your function still returns the data, as you echo it in an output buffer that you then read to return its content. But your implementation is adding useless overhead (and misses the proper error handling to avoid leaking an open output buffer in case an exception happens)

nfreader: Excellent, thank you. That makes a lot of sense :)

GromNaN:
In which situation do you find this function useful, that is not already covered by [`dump`](https://twig.symfony.com/doc/3.x/functions/dump.html) or [`json_encode`](https://twig.symfony.com/doc/3.x/filters/json_encode.html) ?

nfreader:
I like to dump my context during development so I have everything available instead of flipping back and forth between different files. Moving all this into the browser console helps cut down on the clutter.

I can't find a solution to this within the current version of Twig. `var_dump` prints out HTML that `json_encode` can't handle. Is there something I've overlooked?

> I'm not sure this should go in the core. `json_encode` is quite bad at dumping PHP objects, as it only dumps public properties.
I'll update the documentation to reflect this.

Completely understand if this isn't something that will ever get merged though. Just throwing it out there to see if there's interest :)

fabpot:
What about https://twig.symfony.com/doc/3.x/functions/dump.html?
Closing as we won't add this one in core.

GromNaN:
> If you don't pass any value, all variables from the current context are dumped

PR #3726, opened 2022-07-22
Add double quote to render foo

selhabibe:


fabpot:
Closing as it shows that you can also pass a variable.

PR #3729, opened 2022-07-28
Update template to include nested twig file paths so that relative paths can be used

### LINKED ISSUE #3730: Error when including twig files without absolute links
### What steps will reproduce the problem?

Screenshot 2022-07-28 at 15 17 35

With the following file structure. 
If I inside of masterTwig.twig include twig-test/A.twig.
Then, inside of A.twig, I reference B.twig with a relative path `{% include "B.twig" %}`. 

### What's expected?

This should give me the content of B.twig.
It should also looking inside of the folder twigTest, however it looks in the folder twigContainer

### What do you get instead?

The error message that is thrown is:

`Unable to find template "B.twig"(looked into: twigContainer)

### Additional info

| Q                | A
| ---------------- | ---
| Yii version      |  2.0
| Yii Twig version | 2.4
| Twig version     | 
| PHP version      | 8.1
| Operating system | MacOS

developedsoftware:
Closes #3730 
Related to https://github.com/yiisoft/yii2-twig/issues/128

### DIFF
--- a/src/Template.php
+++ b/src/Template.php
@@ -302,7 +302,6 @@ abstract class Template
     protected function loadTemplate($template, $templateName = null, $line = null, $index = null)
     {
         try {
-            
             $this->env->getLoader()->addPath(dirname($this->getSourceContext()->getPath()));
             
             if (\is_array($template)) {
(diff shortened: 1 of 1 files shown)


xabbuh:
This looks like it will introduce a BC break for applications that have templates with the same name in different locations as they will include a different one after this change.

fabpot:
As @xabbuh mentioned, this is not how Twig works. Closing as it would radically change the way we think about template loading.

PR #3733, opened 2022-08-14
Update replace.rst

ankitamin:
foo should be "foo"

### DIFF
--- a/doc/filters/replace.rst
+++ b/doc/filters/replace.rst
@@ -13,7 +13,7 @@ The ``replace`` filter formats a given string by replacing the placeholders
 
     {# using % as a delimiter is purely conventional and optional #}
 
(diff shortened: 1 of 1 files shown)


doc/filters/replace.rst
```diff
@@ -13,7 +13,7 @@ The ``replace`` filter formats a given string by replacing the placeholders
 
     {# using % as a delimiter is purely conventional and optional #}
 
-    {{ "I like this and --that--."|replace({'this': foo, '--that--': "bar"}) }}
+    {{ "I like this and --that--."|replace({'this': "foo", '--that--': "bar"}) }}
```
xabbuh, line 16: `foo` is not quoted intentionally to show how to embed a variable.

fabpot:
See #3734 where I try to clarify the docs.

PR #3736, opened 2022-08-16
fix: add missing initializer for sourceOffset

shyim:
When I create the compiler by own (test case) and call as first `addDebugInfo` it throws a php deprecation

```php
$compiler = new Compiler(new Environment(new ArrayLoader()));

$compiler
            ->addDebugInfo($myNode);
```

PHP Deprecated:  substr_count(): Passing null to parameter #3 ($offset) of type int is deprecated in /Users/shyim/Code/sw6/vendor/twig/twig/src/Compiler.php on line 166

That variable is only set by `compile`, but already used by `addDebugInfo` 🤔

### DIFF
--- a/src/Compiler.php
+++ b/src/Compiler.php
@@ -24,7 +24,7 @@ class Compiler
     private $indentation;
     private $env;
     private $debugInfo = [];
-    private $sourceOffset;
+    private $sourceOffset = 0;
     private $sourceLine;
     private $varNameSalt = 0;
 


fabpot:
Why are you calling `addDebugInfo` outside of compiling nodes?

shyim:
It was inside just to show it easier I removed the code around 😅 

The problem is only the compile of the Compiler class sets that variables

```php
$token = new FeatureCallSilentToken(...);

$compiler = new Compiler(new Environment(new ArrayLoader()));

$token->compile($compiler);
```

but my actual compile in the Node does not use `compile` from the Compiler it just does

```php
    {
        $compiler
            ->addDebugInfo($this)
            ->raw(...)
            ->raw(...);
    }
```

fabpot:
What about https://github.com/twigphp/Twig/pull/3788 as an alternative? That way, you will be able to call `reset` whenever you see fit and never use `compiile()` directly (which is now just a shortcut).

PR #3738, opened 2022-08-24
do not clean up whitespace text nodes inside if tags

xabbuh:


fabpot:
I've updated this PR with another fix. Instead of making a special case in the Parser, I've updated the If node class to support empty nodes instead.

fabpot:
Thank you @xabbuh.

xabbuh:
Thank you for the update Fabien!

PR #3740, opened 2022-08-31
add class_name function to twig

WebMamba:
This PR add a new class_name function, to get the class name of an object

Usage :

`{{class_name(object, true) }}`
 
This is particularly useful when doing content generation with twig.

Use cases : 
In your controller you give an array of object to your template from your controller. In your template your loop over this array, and you wan't to know which class is each object.


stof:
To me, this does not belong to the Twig core. Create a dedicated Twig extension in your project.

And this belongs even less to core when the function does not actually return the class name by default (the namespace is part of the class name in PHP)

fabpot:
I think it's too specialized for core, and should probably be avoided anyway.
Thanks for proposing.

PR #3749, opened 2022-10-02
Twig\Environment::compileSource fail to catch php-native Error's

divinity76:
it will catch Exceptions and Twig\Error\Error's (whatever that is, i haven't investigated), but it won't catch [native PHP Error](https://www.php.net/manual/en/class.error.php)'s and derivatives, for example [ArgumentCountError](https://www.php.net/manual/en/class.argumentcounterror.php)

not sure if the code should be rewritten to 
```
        } catch (Error $e) {
            $e->setSourceContext($source);
            throw $e;
        } catch (\Throwable $e) {
            throw new SyntaxError(sprintf('An exception has been thrown during the compilation of a template ("%s").', $e->getMessage()), -1, $source, $e);
        }
```
or
```
        } catch (\Error $e) {
            $e->setSourceContext($source);
            throw $e;
        } catch (\Exception $e) {
            throw new SyntaxError(sprintf('An exception has been thrown during the compilation of a template ("%s").', $e->getMessage()), -1, $source, $e);
        }
```
but probably 1 of those.. and my best guess is the former.

### DIFF
--- a/src/Environment.php
+++ b/src/Environment.php
@@ -517,7 +517,7 @@ class Environment
         } catch (Error $e) {
             $e->setSourceContext($source);
             throw $e;
-        } catch (\Exception $e) {
+        } catch (\Throwable $e) {
             throw new SyntaxError(sprintf('An exception has been thrown during the compilation of a template ("%s").', $e->getMessage()), -1, $source, $e);
         }
     }


divinity76:
found the issue while investigating a crashing installation of the current stable PhpMyAdmin release, 5.2.0 Released 2022-05-11.
But it happens in PMA's libraries/classes/Twig/RelationExtension.php which is entirely removed in master branch with commit https://github.com/phpmyadmin/phpmyadmin/commit/556804c4e2ba8cd8bab4f75173372342f59f5e62 so.. this issue affect the current PMA release, but will not affect the next PMA release, i think.

nicolas-grekas:
The original error is the following
> "PhpMyAdmin\ConfigStorage\Relation::__construct(): Argument #1 ($dbi) must be of type PhpMyAdmin\DatabaseInterface, null given, called in /srv/http/default/www/db/phpmyadmin/libraries/classes/Twig/RelationExtension.php on line 22"

Why do you think we should wrap those in a SyntaxError? To me it has nothing to do with Twig, and nothing to do with a SyntaxError, so I fail to see why you think Twig should care at all.
What behavior provides this change in the end? Shouldn't the code in phpmyadmin be fixed to provide the proper arguments in the first place?

williamdes:
It was because of phpmyadmin/phpmyadmin#17527 when we where building the twig extension list.

divinity76:
> Why do you think we should wrap those in a SyntaxError? 

for consistency, why should a PDOException or IntlException be converted to a SyntaxError, but not an ArgumentCountError?

>What behavior provides this change in the end?

unclear, but it'll be more consistent

PR #3751, opened 2022-10-04
Update michelf/php-markdown require-dev to allow v2

bobvandevijver:
This new release should work fine: https://github.com/michelf/php-markdown#version-history


fabpot:
Thank you @bobvandevijver.

PR #3760, opened 2022-10-14
Add "html_attributes" twig filter for easiely write attributes as objects

alexander-schranz:
Example usage:

```twig
<!-- button.html.twig -->
{# required #}
{%- set text = text -%}

{# optional #}
{%- set id = id|default(null) -%}
{%- set skin = skin|default('primary') -%}
{%- set type = type|default('button') -%}
{%- set disabled = disabled|default(false) -%}
{%- set href = href|default(null) -%}

{% set attributes = {
    'id': id,
    'class': html_classes(
        'c-button',
        {
            'c-button--primary': skin == 'primary',
            'c-button--secondary': skin == 'secondary',
            'c-button--borderless': skin == 'borderless',
        },
    ),
    'href': href,
    'type': not href ? type : null,
    'disabled': disabled,
} %}

{% set tag = href ? 'a' : 'button' %}

<{{ tag }} {{ attributes|html_attributes }}>
    {{- text -}}
</{{ tag }}>
```

## TODO

 - [x] Escape Value
 - [ ] Add Test Case

### DIFF
--- a/extra/html-extra/HtmlExtension.php
+++ b/extra/html-extra/HtmlExtension.php
@@ -111,7 +111,7 @@ final class HtmlExtension extends AbstractExtension
                 continue;
             }
 
-            $htmlAttributes[] = $key . '="' . twig_escape_filter($environment, $value, 'html_attr') . '"';
+            $htmlAttributes[] = $key . '="' . twig_escape_filter($environment, $value) . '"';
         }
 
         return implode(' ', $htmlAttributes);
--- a/extra/html-extra/HtmlExtension.php
+++ b/extra/html-extra/HtmlExtension.php
@@ -89,7 +89,7 @@ final class HtmlExtension extends AbstractExtension
     }
 
     /**
-     * @param array{string, string|bool|null} $attributes
+     * @param array{string, string|bool|int|float|null} $attributes
      */
     public function htmlAttributes(Environment $environment, array $attributes): string
     {
@@ -106,6 +106,7 @@ final class HtmlExtension extends AbstractExtension
                 continue;
             }
 
+            // null represent no value and should not be outputted for a better DX
             if (null === $value) {
                 continue;
             }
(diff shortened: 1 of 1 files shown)


extra/html-extra/HtmlExtension.php
```diff
@@ -28,6 +28,14 @@ public function getFilters(): array
     {
         return [
             new TwigFilter('data_uri', [$this, 'dataUri']),
+            new TwigFilter(
+                'html_attributes',
```
stof, line 32: this looks more like a function than a filter to me

extra/html-extra/HtmlExtension.php
```diff
@@ -79,6 +87,35 @@ public function dataUri(string $data, string $mime = null, array $parameters = [
...
+
+                continue;
+            }
+
+            // null represent no value and should not be outputted for a better DX
+            if (null === $value) {
+                continue;
+            }
+
+            $htmlAttributes[] = $key . '="' . twig_escape_filter($environment, $value, 'html_attr') . '"';
```
stof: keys must be escaped too

extra/html-extra/HtmlExtension.php
```diff
@@ -79,6 +87,35 @@ public function dataUri(string $data, string $mime = null, array $parameters = [
 
         return $repr;
     }
+
+    /**
+     * @param array{string, string|bool|int|float|null} $attributes
```
stof, line 92: wrong type. `array{string, string|bool|int|float|null}` means an array with 2 elements, with index 0 being a string and index 1 being `string|bool|int|float|null`. Instead, you should use `array<string, string|bool|int|float|null>`

stof (changes requested):

mpdude:
Incidentally I made a very similar suggestion at https://github.com/twigphp/Twig/issues/3907 with a PoC PR at https://github.com/twigphp/Twig/pull/3930.

Would be great if you could 👀 and maybe we can join efforts.

fabpot:
Closing in favor of #3930 (which has the `html_attr_merge` function in addition to this)

alexander-schranz:
Make sense. Thx @mpdude 👍

PR #3761, opened 2022-10-17
Fix php 8.0 compatilibity

topical:
At least with php80 I get something like `Call to undefined function filter_var()` at line 66 of FilesystemCache.php

With the patch applied, it works again.


topical:
Sorry for the noise. It seemed that the software I installed forgot a dependency which lead to this error. 

So I can close this issue.

PR #3763, opened 2022-10-20
Remove internal from ExpressionParser

### LINKED ISSUE #3443: (Partly) remove internal annotation for ExpressionParser
The `Twig\ExpressionParser` class is marked as internal (`@internal` in the DocBlock for the class), which came to my attention because Psalm reported uses of `ExpressionParser::OPERATOR_LEFT` and `ExpressionParser::OPERATOR_RIGHT` in one of my components even though the `ExpressionParser` class is marked as internal, so these usages seemed to be breaking boundaries.

These constants are used in the documentation (see https://twig.symfony.com/doc/3.x/advanced.html#operators , where they are used when showing how to add additional operators to Twig), so it might make sense to at least reduce the usage of `@internal` to some of the public methods of `ExpressionParser`. The way Twig is being extended by custom plugins it seems to me that the `parseExpression` method should also not be internal, as it is often used when creating additional `TokenParser` classes. About the other methods I am not sure, but clarifying some of the Twig boundaries (what to use of `ExpressionParser` and what not) might help both applications and plugins to better know what is safe to use.

VincentLanglet:
In the advanced doc, it's explained how to write TokenParser:
https://github.com/twigphp/Twig/blob/3.x/doc/advanced.rst#defining-a-token-parser

Such an implementation rely on method like
ExpressionParser::parseExpression or Expression::parseMultitargetExpression, 

Currently the ExpressionParser is marked as internal, implying all the method are also internal.
It would be great to promise BC about those methods since they are require/usefull for custom TokenParser.

I dunno if the `@internal` tag is still needed for some methods of the ExpressionParser (like the constructor ?).

This would also
Closes https://github.com/twigphp/Twig/issues/3443

### DIFF
--- a/src/ExpressionParser.php
+++ b/src/ExpressionParser.php
@@ -40,8 +40,6 @@ use Twig\Node\Node;
  * @see https://en.wikipedia.org/wiki/Operator-precedence_parser
  *
  * @author Fabien Potencier <fabien@symfony.com>
- *
- * @internal
  */
 class ExpressionParser
 {


stof (approved):

fabpot (approved):

fabpot:
Thank you @VincentLanglet.

PR #3766, opened 2022-10-24
Suppress tempnam notices

bytestream:
```
tempnam(): file created in the system's temporary directory at /var/www/html/support/vendor/twig/twig/src/Cache/FilesystemCache.php:60)
[stacktrace]
#0 [internal function]: Illuminate\\Foundation\\Bootstrap\\HandleExceptions->handleError()
#1 /var/www/html/support/vendor/twig/twig/src/Cache/FilesystemCache.php(60): tempnam()
#2 /var/www/html/support/vendor/twig/twig/src/Environment.php(349): Twig\\Cache\\FilesystemCache->write()
#3 /var/www/html/support/vendor/twig/twig/src/Environment.php(399): Twig\\Environment->loadTemplate()
```

Notice introduced in PHP 7.1.0-rc.1 by https://bugs.php.net/bug.php?id=69489

### DIFF
--- a/src/Cache/FilesystemCache.php
+++ b/src/Cache/FilesystemCache.php
@@ -63,7 +63,7 @@ class FilesystemCache implements CacheInterface
 
             if (self::FORCE_BYTECODE_INVALIDATION == ($this->options & self::FORCE_BYTECODE_INVALIDATION)) {
                 // Compile cached file into bytecode cache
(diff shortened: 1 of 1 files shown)


fabpot:
These notices are legitimate and should not be hidden.
I don't think this is something we want in core.

PR #3768, opened 2022-11-05
Improve phpdoc of ExtensionInterface/Environment/Expressionparser arrays

VincentLanglet:
Binary and Unary operator have a really specific format
```
array<string, array{precedence: int, class: class-string}>,
array<string, array{precedence: int, class: class-string, associativity: int}>
```
so it's worth to mention that in the phpdoc rather than `array<array>`

### DIFF
--- a/src/Environment.php
+++ b/src/Environment.php
@@ -25,6 +25,8 @@ use Twig\Extension\OptimizerExtension;
 use Twig\Loader\ArrayLoader;
 use Twig\Loader\ChainLoader;
 use Twig\Loader\LoaderInterface;
+use Twig\Node\Expression\Binary\AbstractBinary;
+use Twig\Node\Expression\Unary\AbstractUnary;
 use Twig\Node\ModuleNode;
 use Twig\Node\Node;
 use Twig\NodeVisitor\NodeVisitorInterface;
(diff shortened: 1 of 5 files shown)


src/ExpressionParser.php
```diff
@@ -48,7 +48,9 @@ class ExpressionParser
 
     private $parser;
     private $env;
+    /** @var array<string, array{precedence: int, class: class-string}> */
     private $unaryOperators;
+    /** @var array<string, array{precedence: int, class: class-string, associativity: int}> */
```
stof: `associativity` could even be more precise as `ExpressionParser::OPERATOR_*`

VincentLanglet: I changed to ExpressionParser::OPERATOR_* then

src/Environment.php
```diff
@@ -804,6 +807,8 @@ public function mergeGlobals(array $context): array
 
     /**
      * @internal
+     *
+     * @return array<string, array{precedence: int, class: class-string}>
```
stof: could be `class-string<AbstractUnary>` to be more precise (same for binary operators but using `AbstractBinary`).

VincentLanglet: Done

stof (approved):

fabpot:
Thank you @VincentLanglet.

PR #3769, opened 2022-11-08
Updates CoreExtension::twig_constant to check for definition first to avoid hard crash

AMRoche:
The behaviour of PHP's constant() method has been updated after https://github.com/php/php-src/issues/9905 was accepted and fixed in PHP 8.1 (and previously changed post PHP 8.0) . This PR prevents a fatal error in the case a constant is supplied that doesn't actually exist.


fabpot:
Thank you @AMRoche.

PR #3775, opened 2022-11-22
Update attribute.rst

abdounikarim:
Hello,

I just added examples to help for this function usage.

Cheers 😉

### DIFF
--- a/doc/functions/attribute.rst
+++ b/doc/functions/attribute.rst
@@ -10,6 +10,28 @@ variable:
     {{ attribute(object, method, arguments) }}
     {{ attribute(array, item) }}
 
+If you have a ``Foo`` class with a ``bar`` attribute and a ``getBar()`` method, you can use :
+
+.. code-block:: twig
+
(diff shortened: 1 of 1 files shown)


stof:
I would rather make the note about the resolution algorithm being the same than the `.` operator more visible instead of re-documenting the algorithm partially in a second place (your examples are not covering the whole algorithm)

abdounikarim:
Thanks for your answer @stof. I don't understand your comment, but I let you update the note.  😉

stof:
The full algorithm to resolve attribute is documented at https://twig.symfony.com/doc/3.x/templates.html#variables

Your PR documents a *partial* version of that algorithm in the doc of the `attribute()` function. A better solution is to link to the existing documentation instead, so that we don't duplicate it (and even if we decide to duplicate it, we would have to document the full algorithm as a partial doc would actually be documenting a different algorithm which would be worse)

fabpot:
I'm going to close here as the current doc already explains all of this.
Thank you.

PR #3777, opened 2022-12-12
Fix optimizing closures callbacks

nicolas-grekas:
Follows https://github.com/twigphp/Twig/pull/3722 and https://github.com/php/php-src/issues/8932

Uses a dedicated method added to PHP 8.1.11 (and 8.0.24, but I feel like there is no need to make the check too complex for an outdated version. We just need a marker to make the code simpler when we'll bump to PHP >= 8.1.11.)


stof (approved):

fabpot:
Thank you @nicolas-grekas.

PR #3779, opened 2022-12-15
Cycle function: Add output to example code

kathi-at-datrycs:
At first I couldn't grasp the concept of the cycle function, but after I saw the output in twigfiddle it was clear. I can imagine other readers having the same issue. Therefore I suggest to add the output to the example code.


xabbuh (approved):

stof (approved):

stof:
should be merged in 2.x as it is still maintained, so that both versions of the doc are improved

fabpot:
Thank you @kathi-at-datrycs.

PR #3780, opened 2022-12-15
Add missing argument for the cycle function

stof:


doc/functions/cycle.rst
```diff
@@ -25,4 +25,5 @@ The array can contain any number of values:
 Arguments
 ---------
 
+* ``values``: The list of values to cucle on
```
jdreesen:
```suggestion
* ``values``: The list of values to cycle on
```

fabpot:
Thank you @stof.

PR #3784, opened 2022-12-18
Add `level` argument to `parent()` function

### LINKED ISSUE #3232: [Feature] - {{ parent($level) }} with level
# Feature :
Call `{{ parent($level) }}` with level when having multiple level of extends, closest levels are ignored (like `continue` and `break` in php)
> Ex: `{{ parent(2) }}`

# Example (breadcrumbs) : 
```twig
// base.htm.twig

<ol class="breadcrumb">
{% block breadcrumbs %}
    <li class="breadcrumb-item active"><a href="{{ path("home") }}">Accueil</a></li>
{% endblock %}
</ol>
```
```twig
// missions.htm.twig
{% extends 'base.html.twig' %}

{% block breadcrumbs %}
   {{ parent() }}
    <li class="breadcrumb-item active"><a>Missions</a></li>
{% endblock %}
```
```twig
// mission_details.htm.twig
{% extends 'missions.html.twig' %}

{% block breadcrumbs %}
   {{ parent(2) }}
    <li class="breadcrumb-item active"><a href="{{ path('mission') }}">Missions</a></li>
    <li class="breadcrumb-item active"><a>Mission {{ mission.title }}</a></li>
    <li class="breadcrumb-item active"><a>Details</a></li>
{% endblock %}
```

HypeMC:
Resolves #3232

This PR adds an optional `level` argument to the `parent()` function to allow "skipping" certain parents in case there are multiple levels of inheritance.

### DIFF
--- a/src/Node/Expression/ParentExpression.php
+++ b/src/Node/Expression/ParentExpression.php
@@ -13,6 +13,7 @@
 namespace Twig\Node\Expression;
 
 use Twig\Compiler;
+use Twig\Node\Node;
 
 /**
  * Represents a parent node.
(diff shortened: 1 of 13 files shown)


HypeMC:
The fabbot failure isn't related to any of my changes.

src/Node/Expression/ParentExpression.php
```diff
@@ -21,9 +22,25 @@
  */
 class ParentExpression extends AbstractExpression
 {
-    public function __construct(string $name, int $lineno, string $tag = null)
+    public function __construct(string $name, int $lineno, /* Node */ $level = null, string $tag = null)
```
GromNaN, line 25: Why not adding the new argument last?

src/Node/Expression/ParentExpression.php
```diff
@@ -21,9 +22,25 @@
...
     {
-        parent::__construct([], ['output' => false, 'name' => $name], $lineno, $tag);
+        if (\is_string($level)) {
+            $tag = $level;
+            $level = null;
+
+            @trigger_error(sprintf('Passing $tag as the 3rd argument instead of 4th to the %s constructor is deprecated since Twig 3.5.', __CLASS__), \E_USER_DEPRECATED);
```
stof, line 31: why not putting the level as 4th argument instead, which would avoid that deprecation entirely ?

src/Node/Expression/ParentExpression.php
```diff
@@ -21,9 +22,25 @@
...
+        }
+
+        if (null !== $level && !$level instanceof Node) {
```
stof, line 34: Looking at the way you compile it, you expect an ExpressionNode, not any kind of Node.

stof (commented):
Instead of adding this feature using an integer level, I think it would probably be better to use `{{ block('breadcrumbs', 'base.htm.twig') }}` in your case when trying to render a block from a specific other template, which makes it clear which template is used.

For me, this gets a -1.

fabpot:
Closing in favor of the alternative described by stof.

HypeMC:
Yeah, stof makes a valid point, all my use cases can already be solved with the `block()` function, thanks for the review.

PR #3797, opened 2022-12-28
Add "has some" and "has every" expressions documentation

Jean-Beru:
Add documentation for https://github.com/twigphp/Twig/pull/3670.


fabpot:
Thank you @Jean-Beru.

PR #3798, opened 2022-12-30
Add file and line to error message

ctippler:
When an error occures we currently get the message e.g. 
"An exception has been thrown during the rendering of a template ("Warning: Undefined array key "select"")."

The problem is, that we don't know in which file and where the error occured so it is very hard to find the "cause" of the problem. 
If we add the File and Line it is quite easy to find the problem:

image

### DIFF
--- a/src/Template.php
+++ b/src/Template.php
@@ -405,7 +405,8 @@ abstract class Template
 
             throw $e;
         } catch (\Exception $e) {
-            $e = new RuntimeError(sprintf('An exception has been thrown during the rendering of a template ("%s").', $e->getMessage()), -1, $this->getSourceContext(), $e);
+            $errorText = 'An exception has been thrown during the rendering of a template ("%s"). File: %s Line: %s ';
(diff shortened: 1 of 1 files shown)


fabpot:
Can you give a Twig template that exhibits the issue? This PR does not work as you can see in the tests.

ctippler:
In my case this happens when i call e.g. a twig function and in the twig function i access a undefined array key.

As an example:
The view:
```
{{  "test" | trans }}

{{ app_formatter_test(123)}}
```

The Twig extension looks like this:
```
    /**
     * @return \Twig\TwigFunction[]
     */
    {
        return [
            new TwigFunction('app_formatter_test', [$this, 'test'], ['is_safe' => ['html']]),
        ];
    }

    public function test($number) : string
    {

        $x = [];
        $y = $x['asdf']; //this causes the warning/error
        return $number;
    }
```
image

The twig function is just an example - normaly i have this issue with calls to more complex data models where it is realy hard to find the cause.

Regards,
Christian

fabpot:
The example you gave me emits a warning, which is not cause by Twig. The error message given by PHP gives you the file and line number where it happens, so probably not a good example (at least not one I can use to understand your real-wprld issue).

PR #3800, opened 2023-01-03
Fix error messages in sandboxed mode for has some and has every

stof:
This has been forgotten when changing the PR to implement operators rather than filters.


fabpot:
Good catch, thanks @stof.

PR #3804, opened 2023-01-21
Update docs for filters that use the `calendar` option

phansys:
`IntlExtension` does not accept "Gregorian", just "gregorian" ([ref](https://github.com/twigphp/intl-extra/blob/f5aa1e3fe3286d476ceca5feafd4390564f7cc29/IntlExtension.php#L324)).


stof (approved):

fabpot:
Good catch, thanks @phansys.

PR #3806, opened 2023-01-31
Minor: Fixing language

ThomasLandauer:


doc/filters/striptags.rst
```diff
@@ -1,7 +1,7 @@
 ``striptags``
 =============
 
-The ``striptags`` filter strips SGML/XML tags and replace adjacent whitespace
+The ``striptags`` filter strips SGML/XML tags and replaces adjacent whitespaces
```
fabpot: whitespace is unaccountable, so:
```suggestion
The ``striptags`` filter strips SGML/XML tags and replaces adjacent whitespace
```

ThomasLandauer: Well, but then "adjacent" sounds weird ;-) => So I went for "whitespace characters", as https://www.php.net/manual/en/regexp.reference.escape.php is doing.

fabpot:

PR #3808, opened 2023-02-02
Add the new PHP 8.0 IntlDateFormatter::RELATIVE_* constants for date formatting.

sukei:
Since PHP 8.0, IntlDateFormatter get new constants to format relative dates. This PR add those constants.

### DIFF
--- a/doc/filters/format_datetime.rst
+++ b/doc/filters/format_datetime.rst
@@ -26,6 +26,12 @@ You can tweak the output for the date part and the time part:
 
 Supported values are: ``none``, ``short``, ``medium``, ``long``, and ``full``.
 
+.. versionadded:: 3.6
+
+    ``relative_short``, ``relative_medium``, ``relative_long``, and ``relative_full`` are also supported when running on
+    PHP 8.0 and superior or when using a polyfill that define the ``IntlDateFormatter::RELATIVE_*`` constants and
+    associated behavior.
+
 For greater flexibility, you can even define your own pattern
 (see the `ICU user guide`_ for supported patterns).
 
--- a/extra/intl-extra/IntlExtension.php
+++ b/extra/intl-extra/IntlExtension.php
@@ -26,7 +26,36 @@ use Twig\TwigFunction;
 
 final class IntlExtension extends AbstractExtension
 {
-    private const DATE_FORMATS = [
+    private static function availableDateFormats(): array
+    {
+        static $formats = null;
+
+        if (null !== $formats) {
+            return $formats;
+        }
+
+        $formats = [
+            'none' => \IntlDateFormatter::NONE,
+            'short' => \IntlDateFormatter::SHORT,
+            'medium' => \IntlDateFormatter::MEDIUM,
+            'long' => \IntlDateFormatter::LONG,
+            'full' => \IntlDateFormatter::FULL,
+        ];
+
+        // Assuming that each `RELATIVE_*` constant are defined when one of them is.
+        if (\defined('IntlDateFormatter::RELATIVE_FULL')) {
+            $formats = array_merge($formats, [
+                'relative_short' => \IntlDateFormatter::RELATIVE_SHORT,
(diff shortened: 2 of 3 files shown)


extra/intl-extra/IntlExtension.php
```diff
@@ -32,6 +32,17 @@ final class IntlExtension extends AbstractExtension
         'medium' => \IntlDateFormatter::MEDIUM,
         'long' => \IntlDateFormatter::LONG,
         'full' => \IntlDateFormatter::FULL,
+        'relative_short' => \IntlDateFormatter::RELATIVE_SHORT,
+        'relative_medium' => \IntlDateFormatter::RELATIVE_MEDIUM,
+        'relative_long' => \IntlDateFormatter::RELATIVE_LONG,
+        'relative_full' => \IntlDateFormatter::RELATIVE_FULL,
```
stof: Those must be used conditionally, to avoid breaking things on PHP 7.x

stof (changes requested):

extra/intl-extra/IntlExtension.php
GromNaN: Ideally, it would thrown an exception when this formats are used with PHP 7, telling this formats require PHP8+ and PHPxx is used.

stof: Note that the check should be based on constants being defined, not on the PHP versions, to avoid issues with the `symfony/polyfill-intl-icu` package (which does not define those constants in all its versions, and I don't even know whether the latest version defines them)

GromNaN: They have been added by https://github.com/symfony/polyfill/pull/343, available since [v1.24.0](https://github.com/symfony/polyfill-intl-icu/blob/v1.24.0/IntlDateFormatter.php#L71-L74).

doc/filters/format_datetime.rst
```diff
@@ -24,7 +24,9 @@ You can tweak the output for the date part and the time part:
     {# mercredi 7 août 2019 23:39:12 UTC #}
     {{ '2019-08-07 23:39:12'|format_datetime('full', 'full', locale='fr') }}
 
-Supported values are: ``none``, ``short``, ``medium``, ``long``, and ``full``.
+Supported values are: ``none``, ``short``, ``medium``, ``long``, and ``full``. ``relative_short``, ``relative_medium``,
+``relative_long``, and ``relative_full`` are also supported when running on PHP 8.0 and superior or when using a
+polyfill that define the ``IntlDateFormatter::RELATIVE_*`` constants and associated behavior.
```
fabpot: The support part should be extract in a `.. versionadded::` directive as done elsewhere in the docs.

extra/intl-extra/Tests/Fixtures/format_date.test
```diff
@@ -19,6 +21,8 @@ Aug 7, 2019, 11:39:12 PM
 23:39
 07/08/2019
 mercredi 7 août 2019 à 23:39:12 temps universel coordonné
+07/08/2019
+mercredi 7 août 2019 à 23:39:12 temps universel coordonné
```
fabpot: This looks suspicious to me as I don't see anything "relative" here. Does it mean that it's not implemented for French? We should find an example that returns something different.

fabpot (changes requested):
Tests are still broken with PHP 7.x versions.

extra/intl-extra/Tests/Fixtures/format_date.test
sukei: It's limited to `yesterday`, `today`, and `tomorrow`. Just added a conditional test for it.

doc/filters/format_datetime.rst
sukei: Wich version should I use as a target? It depends on the running PHP version (8.0) or the usage of a polyfill. Others `.. versionadded::` targets a Twig release. Maybe `3.6` with a proper explanation in the block?

fabpot:
Thank you @sukei.

Issue #3801, opened 2023-01-12
Wrong argument for twig_matches


jensschaerer:
After upgrading to TWIG v3.5.0 (coming from v3.4.3) I get a strange error and as my knowledge of the TWIG codebase is limited, I'm not able to get whats the fault.

Error:
`twig_matches(): Argument #2 ($str) must be of type string, null given`

Calling template part:
```
{% if not result is empty %}
    {% include 'my-template.html.twig' with {"result":result} %}
{% endif %}
```

result is an array passed to a sub template. This works in v3.4.3 but not in v3.5.0

jdreesen:
It comes from the usage of the `matches` operator, like `{% if someVar matches '/some-regex/' %}`. 
If `someVar` is empty, Twig throws an error since `v3.5`.

jensschaerer:
the variable "result" shouldn't be empty in any way because of the if statement around the include block

stof:
I don't see any usage of `matches` in your code snippet. So I'm quite sure you are showing the wrong part of your template.

jensschaerer:
If I remove the code snippet shown above the template renders as expected. There is no other code involved

stof:
Then please share the content of the included template. The error is probably somewhere inside it.

jensschaerer:
The included template was made empty to test if the error is inside it, so the error is within the include

stof:
Then please share the stack trace of the exception to see where this `twig_matches` call is.

jensschaerer:
After investigating the error and rollback to TWIG 3.4.3 and now re-upgrade to TWIG 3.5.0 the error seems gone. Strange behavior anyway. I cleared cache, deleted all cache related files, so something went wrong while installing TWIG 3.5.0 over 3.4.3

Sorry for the back and forth and this false report. I think this issue can be closed. If I encounter any more errors after deployment I'll get back

jensschaerer:
@stof I now have same error. Please see stack trace:

```
TypeError
--
[...]

PR #3809, opened 2023-02-07
Restores the leniency of the `matches` twig comparison

la5942:
Restores the leniency of the `matches` twig comparison, allowing null subject to result in a non-match.

Resolves BC break introduced in PR https://github.com/twigphp/Twig/pull/3687 ref in https://github.com/twigphp/Twig/issues/3801#issuecomment-1413345131

As per pattern in https://github.com/twigphp/Twig/pull/3617

### DIFF
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -1021,19 +1021,19 @@ function twig_compare($a, $b)
 
 /**
  * @param string $pattern
- * @param string $subject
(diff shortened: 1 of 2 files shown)


la5942:
This does raise the question of whether tests should be introduced to catch any other `null` problems like this for PHP 8.1 and above

fabpot:
Thank you @la5942.

PR #3813, opened 2023-02-15
Added compatibility with PHPUnit 10

### LINKED ISSUE #3810: IntegrationTestCase: Compatibility with PHPUnit 10
PHPUnit 10 has breaking changes regarding data providers. These must be static now.

However, `\Twig\Test\IntegrationTestCase` uses non-static data providers.

The errors look like this:

```                                                   
* Data Provider method MyTest::getTests() is not static                                                                                  
                                                                                                                                                                                                                   
* Data Provider method MyTest::getTests() expects an argument    
```

Unfortunately, the needed change looks like a breaking change. `IntegrationTestCase::getTests` depends on `$this->getFixturesDir()` which is abstract. Consumers would need to change this as well then.

SiebelsTim:
fixes #3810

### DIFF
--- a/src/Test/ASTNodeTestCase.php
+++ b/src/Test/ASTNodeTestCase.php
@@ -17,51 +17,48 @@ use Twig\Environment;
 use Twig\Loader\ArrayLoader;
 use Twig\Node\Node;
 
-/**
- * @deprecated use \Twig\Test\ASTNodeTestCase instead
- */
-abstract class NodeTestCase extends TestCase
+abstract class ASTNodeTestCase extends TestCase
 {
-    abstract public function getTests();
+    abstract public static function getTests();
 
     /**
      * @dataProvider getTests
      */
     public function testCompile($node, $source, $environment = null, $isPattern = false)
     {
-        $this->assertNodeCompilation($source, $node, $environment, $isPattern);
+        static::assertNodeCompilation($source, $node, $environment, $isPattern);
     }
 
-    public function assertNodeCompilation($source, Node $node, Environment $environment = null, $isPattern = false)
+    public static function assertNodeCompilation($source, Node $node, Environment $environment = null, $isPattern = false)
     {
-        $compiler = $this->getCompiler($environment);
+        $compiler = static::getCompiler($environment);
         $compiler->compile($node);
 
         if ($isPattern) {
-            $this->assertStringMatchesFormat($source, trim($compiler->getSource()));
+            static::assertStringMatchesFormat($source, trim($compiler->getSource()));
         } else {
-            $this->assertEquals($source, trim($compiler->getSource()));
+            static::assertEquals($source, trim($compiler->getSource()));
         }
     }
 
-    protected function getCompiler(Environment $environment = null)
+    protected static function getCompiler(Environment $environment = null)
     {
-        return new Compiler(null === $environment ? $this->getEnvironment() : $environment);
+        return new Compiler(null === $environment ? static::getEnvironment() : $environment);
     }
 
-    protected function getEnvironment()
+    protected static function getEnvironment()
     {
         return new Environment(new ArrayLoader([]));
     }
 
-    protected function getVariableGetter($name, $line = false)
+    protected static function getVariableGetter($name, $line = false)
     {
         $line = $line > 0 ? "// line $line\n" : '';
 
         return sprintf('%s($context["%s"] ?? null)', $line, $name);
     }
 
-    protected function getAttributeGetter()
+    protected static function getAttributeGetter()
     {
         return 'twig_get_attribute($this->env, $this->source, ';
     }
--- a/src/Test/ASTNodeTestCase.php
+++ b/src/Test/ASTNodeTestCase.php
@@ -26,12 +26,12 @@ abstract class ASTNodeTestCase extends TestCase
      */
     public function testCompile($node, $source, $environment = null, $isPattern = false)
     {
-        static::assertNodeCompilation($source, $node, $environment, $isPattern);
+        $this->assertNodeCompilation($source, $node, $environment, $isPattern);
     }
 
-    public static function assertNodeCompilation($source, Node $node, Environment $environment = null, $isPattern = false)
+    public function assertNodeCompilation($source, Node $node, Environment $environment = null, $isPattern = false)
     {
-        $compiler = static::getCompiler($environment);
+        $compiler = $this->getCompiler($environment);
         $compiler->compile($node);
 
         if ($isPattern) {
@@ -41,12 +41,12 @@ abstract class ASTNodeTestCase extends TestCase
         }
     }
 
-    protected static function getCompiler(Environment $environment = null)
+    protected function getCompiler(Environment $environment = null)
     {
-        return new Compiler(null === $environment ? static::getEnvironment() : $environment);
+        return new Compiler(null === $environment ? $this->getEnvironment() : $environment);
     }
 
-    protected static function getEnvironment()
+    protected function getEnvironment()
     {
         return new Environment(new ArrayLoader([]));
     }
--- a/src/Test/BaseTemplateIntegrationTestCase.php
+++ b/src/Test/BaseTemplateIntegrationTestCase.php
@@ -1,0 +1,223 @@
+<?php
+
+namespace Twig\Test;
+
+use PHPUnit\Framework\TestCase;
+use Twig\Environment;
+use Twig\Error\Error;
+use Twig\Extension\ExtensionInterface;
+use Twig\Loader\ArrayLoader;
+use Twig\RuntimeLoader\RuntimeLoaderInterface;
+use Twig\TwigFilter;
+use Twig\TwigFunction;
+use Twig\TwigTest;
+
+abstract class BaseTemplateIntegrationTestCase extends TestCase
+{
+    /**
+     * @return RuntimeLoaderInterface[]
+     */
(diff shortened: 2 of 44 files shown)


SiebelsTim:
This is a breaking change. Are we allowed to do these on 3.x? Do we have something like an UPGRADE.md that I can add this to?

I did not modify Twig's own tests as we can do that when/if we upgrade to PHPUnit 10.  However, this is needed so that Twig's users can upgrade themselves to PHPUnit 10, when they are using `IntegrationTestCase`

src/Test/IntegrationTestCase.php
```diff
@@ -32,7 +32,7 @@ abstract class IntegrationTestCase extends TestCase
     /**
      * @return string
      */
-    abstract protected function getFixturesDir();
+    abstract protected static function getFixturesDir();
```
stof: This change is a BC break. It cannot be done in a minor version as is.

src/Test/IntegrationTestCase.php
```diff
@@ -211,7 +217,7 @@ protected function doIntegrationTest($file, $message, $condition, $templates, $e
                     return;
                 }
 
-                throw new Error(sprintf('%s: %s', \get_class($e), $e->getMessage()), -1, null, $e);
+                throw new Error(sprintf('"%s": "%s".', \get_class($e), $e->getMessage()), -1, null, $e);
```
stof: this change looks wrong to me.

src/Test/IntegrationTestCase.php
SiebelsTim: Correct. I already mentioned it in the issue and in this PR. I was told to open a PR against 3.x. Can you guide me to a better solution?

I don't think we can fix this without a breaking change.

src/Test/IntegrationTestCase.php
SiebelsTim: I agree! I wasn't careful enough to merge fabbot.io's changes. Sorry!

src/Test/IntegrationTestCase.php
SiebelsTim: What do you think of providing a second `IntegrationTestCase` which supports PHPUnit10 and deprecating the old one? If you agree, do you have suggestions for a name?

jdreesen: But do you need to fix it? Non-static data providers are only deprecated but should still work in PHPUnit 10 right?

So changing it could be postponed to the next major release of Twig.

stof: they are deprecated in 9.6 and unsupported in 10

jdreesen: Well, then I must have misunderstood this line: https://github.com/sebastianbergmann/phpunit/blob/main/ChangeLog-10.0.md?plain=1#L99

stof: ah indeed. I confused with the Test suffix for abstract classes.

SiebelsTim: They are still working in PHPUnit 10. It shows a deprecations notice at the end. 
So yes, this would be unsupported in PHPUnit 11.

SiebelsTim:
I've pushed a different concept. I dislike that it is quite complex.

I'm open to suggestions, please tell me what you think.

src/Test/IntegrationTestCasePhpUnit10.php
```diff
@@ -0,0 +1,51 @@
...
+
+namespace Twig\Test;
+
+/**
+ * Integration test helper.
+ *
+ * @author Fabien Potencier <fabien@symfony.com>
+ * @author Karma Dordrak <drak@zikula.org>
+ */
+abstract class IntegrationTestCasePhpUnit10 extends BaseIntegrationTestCase
```
SiebelsTim: I really don't like this class' name.

src/Test/IntegrationTestCase.php
SiebelsTim: fabbot.io still wants this to be changed. How do I deal with this?

stof: you ignore it. this exception message formatting check has some false positives when we have special needs

src/Test/BaseIntegrationTestCase.php
```diff
@@ -0,0 +1,223 @@
...
+use Twig\Environment;
+use Twig\Error\Error;
+use Twig\Extension\ExtensionInterface;
+use Twig\Loader\ArrayLoader;
+use Twig\RuntimeLoader\RuntimeLoaderInterface;
+use Twig\TwigFilter;
+use Twig\TwigFunction;
+use Twig\TwigTest;
+
+abstract class BaseIntegrationTestCase extends TestCase
```
stof: The base class should be tagged as `@internal` to allow us to remove it again in the future when we don't need it anymore.

src/Test/IntegrationTestCasePhpUnit10.php
stof: I don't have a good idea for the name, but this one is indeed not good. This class works on *all* PHPUnit versions, not just on PHPUnit 10 (static data providers have always been supported by PHPUnit since data providers exist). And it is the class that we will want projects to use (the other one should be marked as deprecated at some point)

src/Test/IntegrationTestCase.php
SiebelsTim: I didn't realize that this was an option. All right!

SiebelsTim:
I think I resolved most of the comments. Naming things is hard though.

Furthermore, I noticed that there is also `NodeTestCase`. That is much harder to fix. Every method in that class needs to be static. I only see the option to copy the class and I even might be fine with copying, as the class is much smaller compared to `IntegrationTestCase`.

But even then, upgrading to it is much harder, as we use `$this->createMock()` in the data provider which we want to turn static.

stof:
> But even then, upgrading to it is much harder, as we use `$this->createMock()` in the data provider which we want to turn static.

The replace test case must be refactored in a way where the mock is created in the test, not in the data provider.

src/Test/IntegrationTestCase.php
```diff
@@ -11,69 +11,21 @@
...
-use Twig\TwigFunction;
-use Twig\TwigTest;
-
 /**
  * Integration test helper.
  *
  * @author Fabien Potencier <fabien@symfony.com>
  * @author Karma Dordrak <drak@zikula.org>
+ *
+ * @deprecated use \Twig\Test\TemplateTestCase instead
```
GromNaN: The referenced class needs to be updated (but I like this shorter name).

src/Test/BaseTemplateIntegrationTestCase.php
```diff
@@ -0,0 +1,223 @@
...
+use Twig\Environment;
+use Twig\Error\Error;
+use Twig\Extension\ExtensionInterface;
+use Twig\Loader\ArrayLoader;
+use Twig\RuntimeLoader\RuntimeLoaderInterface;
+use Twig\TwigFilter;
+use Twig\TwigFunction;
+use Twig\TwigTest;
+
+abstract class BaseTemplateIntegrationTestCase extends TestCase
```
GromNaN: We don't want to keep this class when the deprecated class `IntegrationTestCase` will be removed. Neither does the `doIntegrationTest` method.

For cleaner code, I'd suggest converting this class into a trait `Twig\Test\Internal\IntegrationTestTrait` with `@internal` annotation and make `doIntegrationTest` private, so that it cannot be used or extended.

src/Test/ASTNodeTestCase.php
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * @dataProvider getTests
+     */
+    public function testCompile($node, $source, $environment = null, $isPattern = false)
+    {
+        static::assertNodeCompilation($source, $node, $environment, $isPattern);
+    }
+
+    public static function assertNodeCompilation($source, Node $node, Environment $environment = null, $isPattern = false)
```
GromNaN: Could you check why this method needs to be static, same for `getCompiler` and `getEnvironment`?

stof: keeping `getEnvironment` non-static would be more flexible (test cases could then use mocks in it if they need it). so I would keep all these non-static

src/Test/IntegrationTestCase.php
SiebelsTim: I've updated the FQCN.

I decided against the shorter name because I thought this was too ambiguous in this project.

src/Test/ASTNodeTestCase.php
SiebelsTim: When doing so, you cannot use them in data providers anymore. I thought it would be more valueable to allow them to be used in data providers.
We use `getCompiler` in one test internally.

I don't have a strong opinion on this though. 

I've pushed a commit turning this non-static again.

GromNaN: Can you explain de bit more how you use `getCompiler` in the dataprovider ?

SiebelsTim: `\Twig\Tests\Node\DeprecatedTest` uses `Compiler::getVarName()` to build the expected php source code that is compiled from a `Node`.

I suppose we should rather hardcode `__internal_compile_0` into the expected string. What do you think?
I pushed a commit doing that on top of this branch in case you agree.

fabpot:
Closing here as we are now supporting PHPUnit 10 and 11. The approach from @derrabus was a bit different than the approach here.

PR #3815, opened 2023-02-27
Add inflector pluralize/singularize support

seb-jean:
Hi,

I add inflector pluralize/singularize support.

This PR add 2 new filters: `pluralize` and `singularize`.

### DIFF
--- a/doc/filters/pluralize.rst
+++ b/doc/filters/pluralize.rst
@@ -1,0 +1,34 @@
+``pluralize``
+========
+
+The ``pluralize`` filter transforms a given singular string into an array of plural words
+
+Here is an example:
+
+.. code-block:: twig
+
+    {{ 'person'|pluralize|join(', ') }}
+    persons, people
+
+.. note::
+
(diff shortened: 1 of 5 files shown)


stof:
I'm not sure about the usefulness of those filters. The symfony/string inflector is returning several candidates, with only one of them being the correct singular/plural (hopefully, you won't receive only invalid ones, but that might happen in case of weird cases that are not handled). So this is not really suitable for display IMO.

seb-jean:
This is indeed its purpose as the [documentation](https://symfony.com/doc/current/components/string.html#inflector) says:
> The value returned by both methods is always an array because sometimes it's not possible to determine a unique singular/plural form for the given word.


I'm using this filter for properties that are plural (ex: roles), and I need to singularize them dynamically.

seb-jean:
> I'm not sure about the usefulness of those filters. The symfony/string inflector is returning several candidates, with only one of them being the correct singular/plural (hopefully, you won't receive only invalid ones, but that might happen in case of weird cases that are not handled). So this is not really suitable for display IMO.

You're right

PR #3816, opened 2023-03-02
Shorten and improve sentence structure

foxtrotcharlie:
This is more succinct and easier to read.

When I initially read it it took me a bit of time to comprehend due to the sentence structure. I feel this reads better, and says what it means in less words.

### DIFF
--- a/doc/tags/extends.rst
+++ b/doc/tags/extends.rst
@@ -242,8 +242,8 @@ Let's take another example: a block included within an ``if`` statement:
     {% endif %}
 
 Contrary to what you might think, this template does not define a block
-conditionally; it just makes overridable by a child template the output of
-what will be rendered when the condition is ``true``.
+conditionally; when the condition is ``true``, the rendered output of the 
+block may be overridden by a child template.
 
 If you want the output to be displayed conditionally, use the following
 instead:


doc/tags/extends.rst
```diff
@@ -242,8 +242,8 @@ Let's take another example: a block included within an ``if`` statement:
     {% endif %}
 
 Contrary to what you might think, this template does not define a block
-conditionally; it just makes overridable by a child template the output of
-what will be rendered when the condition is ``true``.
+conditionally; when the condition is ``true``, the rendered output of the 
+block may be overridden by a child template.
```
stof, line 246: this is not what the previous sentence meant. The block is always overridden in the child template, not just when the condition is true

foxtrotcharlie: Thanks for the feedback and the correction @stof ! You're right of course: I did some testing and I see that irrespective of the condition, the child template renders that block content. So the original language still seems confusing to me - because it mentions _**when** the condition is "true"_` but doesn't say "not only".

The way you said it is much clearer to me. So how about we update it as follows:

```
Contrary to what you might think, this template does not define a block conditionally. 
The block is always overridden in the child template, not only when the condition is true.
```

Or, a bit longer, but perhaps even clearer:

```
Contrary to what you might think, this template does not define a block conditionally.
Whether the `if` condition evaluates to "true" or "false" makes no difference - the block 
in the child template will always be defined and will be used in place of the parent block.
```

foxtrotcharlie: If you're keen on any of these LMK and I'll update the PR. If you don't think any of this is useful/necessary, feel free to close this.

PR #3822, opened 2023-03-21
Use twig_ord instead of mb_ord (1.x)

cweiske:
.. so that twig still works users without mbstring extension.

Twig provides its own wrapper function for `mb_ord` called `twig_ord`, but does not use it. This breaks installations without mbstring.


stof:
Twig 1.x is EOL 
Twig 2+ relies on symfony/polyfill-mbstring so it is not affected.

fabpot:
Closing as 1.x is not maintained anymore.

PR #3826, opened 2023-03-30
[Extra][Markdown] Allow more variant of MD converter

lyrixx:
I want to inject `GithubFlavoredMarkdownConverter` but ATM, it's not possible

ref: https://github.com/thephpleague/commonmark#-installation--basic-usage


stof:
it would be even better to use the interfaces for that

lyrixx:
But we need to deal with legacy code I guess :/

stof:
Well, then allow both the new interface and the legacy one.

lyrixx:
I miss some time to finish it. Sorry

PR #3829, opened 2023-04-05
Allow customizations of $specialVars

i350:
Set `$specialVars` variable as static, to allow customization or disabling one of special variables for security

### DIFF
--- a/src/Node/Expression/NameExpression.php
+++ b/src/Node/Expression/NameExpression.php
@@ -16,7 +16,7 @@ use Twig\Compiler;
 
 class NameExpression extends AbstractExpression
 {
-    private $specialVars = [
+    static private $specialVars = [
         '_self' => '$this->getTemplateName()',
         '_context' => '$context',
         '_charset' => '$this->env->getCharset()',
@@ -52,7 +52,7 @@ class NameExpression extends AbstractExpression
                 ;
             }
         } elseif ($this->isSpecial()) {
-            $compiler->raw($this->specialVars[$name]);
+            $compiler->raw(static::$specialVars[$name]);
         } elseif ($this->getAttribute('always_defined')) {
             $compiler
                 ->raw('$context[')
@@ -87,7 +87,7 @@ class NameExpression extends AbstractExpression
 
     public function isSpecial()
     {
-        return isset($this->specialVars[$this->getAttribute('name')]);
+        return isset(static::$specialVars[$this->getAttribute('name')]);
     }
 
     public function isSimple()


src/Node/Expression/NameExpression.php
```diff
@@ -52,7 +52,7 @@ public function compile(Compiler $compiler): void
                 ;
             }
         } elseif ($this->isSpecial()) {
-            $compiler->raw($this->specialVars[$name]);
+            $compiler->raw(static::$specialVars[$name]);
```
stof, line 55: accessing a private property with `static` is invalid. It would break if the class is extended.

stof (commented):
For now, I'm -1 on this. Please describe your actual use case for changing the way Twig compiles those.

src/Node/Expression/NameExpression.php
i350: My friend building a SAAS platform, and searching for something similar to [Liquid](https://github.com/Shopify/liquid) but for PHP. I suggested `Twig` to him, but he found that special vars cannot be disabled. These variables expose some data that shouldn't be shown for his clients.

In short, the use case is stated in the first point of liquid's pros:
> - You want to allow your users to edit the appearance of your application but don't want them to run insecure code on your server.
... or see private data like (The full path of view, the context object) 

Using static variables is a way followed by some other packages to allow customization, like [Tenancy For Laravel](https://tenancyforlaravel.com/docs/v3/configuration/) package

> ... whenever the class you're using has a public static property, it's intended to be configured.

stof: Which data is it that should not be shown ?
Blocking access to `_context` won't prevent access to the content of the context (they can try to find the available variables instead of looping over this special variable to discover them all). The Twig context is just the associative array of variables available in the template. If they contain things that users should not see, you should not inject them in the template in the first place as they can access the normal variable anyway.
The `_charset` variable is not really sensitive (in 99% of cases, this will give the string `UTF-8`).
`_self` return the template **name**, not the full path (you could of course register a FilesystemLoader with its root directory being the root of the filesystem, in which case the name would indeed be the full path, but this is then an issue on your side as `source('...')` would then allow reading any file. That's not the recommended setup, especially not if you have to deal with untrusted templates).

> Using static variables is a way followed by some other packages to allow customization

I see 2 issues here:
- static variables are a global state, so this is a global configuration, which would leak to other usages of Twig (you cannot have 2 Twig Environment with different static variables)
- your quote talks about *public* properties. Private static properties cannot be configured.

fabpot:
Closing for the reasons explained by stof.

sstok:
Better to use the Sandbox, as there might be other things you don't want users to do.

PR #3830, opened 2023-04-07
Catch errors thrown during template rendering

richards-square:
Some errors, like not providing a function the proper number of arguments or division by zero, extend from `\Error` rather than `\Exception`. This PR catches these types of errors during template rendering and throws a `RuntimeError` in order to provide better debugging information.

### DIFF
--- a/src/Error/Error.php
+++ b/src/Error/Error.php
@@ -53,7 +53,7 @@ class Error extends \Exception
      * @param int         $lineno  The template line where the error occurred
      * @param Source|null $source  The source context where the error occurred
      */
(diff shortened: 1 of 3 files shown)


stof:
I'm not sure we need separate catch block (instead of catching `Throwable`) just to say `An error` vs `An exception` at the beginning of the message.

bryanashley (approved):

bryanashley:
@stof Any chance this can be merged/re-reviewed? Would be really helpful in the work I'm doing.

fabpot:
Thank you @richards-square.

PR #3832, opened 2023-04-14
return annotation to suppress deprecation warning

PhilETaylor:
(finally PRed in the right repo - sorry about that) 

> !!  2023-04-14T10:59:02+00:00 [info] User Deprecated: Method "Symfony\Component\HttpKernel\Bundle\Bundle::build()" might add "void" as a native return type declaration in the future. Do the same in child class "Twig\Extra\TwigExtraBundle\TwigExtraBundle" now to avoid errors or add an explicit `@return` annotation to suppress this message.


stof (approved):

fabpot:
Thank you @PhilETaylor.

PR #3836, opened 2023-05-02
Add ReturnTypeWillChange to mute deprecation warnings

zethzeth:
Without this change, version 1.44.7 will show this warning:

warning

An alternative fix to this would be something like this:

```
    /**
     * @return int
     */
    #[\ReturnTypeWillChange]
    public function count()
    {
        return \count($this->nodes);
    }

    /**
     * @return \Traversable
     */

    #[\ReturnTypeWillChange]
    public function getIterator()
    {
        return new \ArrayIterator($this->nodes);
    }
```

But I considered this cleaner.

### DIFF
--- a/src/Node/Node.php
+++ b/src/Node/Node.php
@@ -207,7 +207,8 @@ class Node implements \Twig_NodeInterface
         unset($this->nodes[$name]);
     }
 
-    public function count(): int
+    #[\ReturnTypeWillChange]
+    public function count()
     {
         return \count($this->nodes);
     }
--- a/src/Node/Node.php
+++ b/src/Node/Node.php
@@ -276,4 +276,4 @@ class Node implements \Twig_NodeInterface
 class_alias('Twig\Node\Node', 'Twig_Node');
 
 // Ensure that the aliased name is loaded to keep BC for classes implementing the typehint with the old aliased name.
-class_exists('Compiler::class');
+class_exists(Compiler::class);
--- a/src/Node/Node.php
+++ b/src/Node/Node.php
@@ -207,6 +207,9 @@ class Node implements \Twig_NodeInterface
         unset($this->nodes[$name]);
     }
 
+    /**
+     * @return int
+     */
     #[\ReturnTypeWillChange]
     public function count()
     {
--- a/src/Node/Node.php
+++ b/src/Node/Node.php
@@ -78,7 +78,7 @@ class Node implements \Twig_NodeInterface
     }
 
     /**
-     * @deprecated since 1.16.1 (to be removed in 2.0)
+     * @deprecated since Symfony 1.16.1 (to be removed in 2.0)
      */
     public function toXml($asDom = false)
     {
@@ -125,7 +125,7 @@ class Node implements \Twig_NodeInterface
     }
 
     /**
-     * @deprecated since 1.27 (to be removed in 2.0)
+     * @deprecated since Symfony 1.27 (to be removed in 2.0)
      */
     public function getLine()
     {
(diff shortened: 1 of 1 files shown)


src/Node/Node.php
```diff
@@ -207,18 +207,12 @@ public function removeNode($name)
         unset($this->nodes[$name]);
     }
 
-    /**
-     * @return int
-     */
-    public function count()
+    public function count(): int
```
jdreesen: This is a breaking change, thus I think we have to use `#[\ReturnTypeWillChange]` here.

zethzeth: I see. I'm not entirely sure why (and what will break). 
I just corrected it to what you suggested, though. 

Should I do the same for getIterator?

... changing  

```
    public function getIterator(): \Traversable
```

to

```
    #[\ReturnTypeWillChange]
    public function getIterator()
```

?

jdreesen: Yes, it's the same for `getIterator()`.

jdreesen: The problem is, that adding a return type in a parent class forces child classes to also add them. So if anyone extends this class, it'll break until they add the return type there, too.

src/Node/Node.php
```diff
@@ -207,17 +207,13 @@ public function removeNode($name)
         unset($this->nodes[$name]);
     }
 
-    /**
-     * @return int
-     */
+    #[\ReturnTypeWillChange]
```
jdreesen, line 213:
```suggestion
    /**
     * @return int
     */
    #[\ReturnTypeWillChange]
```

src/Node/Node.php
zethzeth: Ahh! I see. 

I just committed the change to `getIterator()` as well. 

Thanks a lot for the explanation. <3

src/Node/Node.php
```diff
@@ -207,17 +207,13 @@ public function removeNode($name)
...
     public function count()
     {
         return \count($this->nodes);
     }
 
-    /**
-     * @return \Traversable
-     */
+    #[\ReturnTypeWillChange]
```
jdreesen, line 222:
```suggestion
    /**
     * @return \Traversable
     */
    #[\ReturnTypeWillChange]
```

src/Node/Node.php
```diff
@@ -277,4 +273,4 @@ public function getFilename()
 class_alias('Twig\Node\Node', 'Twig_Node');
 
 // Ensure that the aliased name is loaded to keep BC for classes implementing the typehint with the old aliased name.
-class_exists('Twig\Compiler');
+class_exists('Compiler::class');
```
jdreesen:
```suggestion
class_exists(Compiler::class);
```

src/Node/Node.php
jdreesen: Here's some more information: https://www.php.net/manual/en/language.oop5.variance.php

> Covariance allows a child's method to return a more specific type than the return type of its parent's method.

But not the other way around! So, child class return types may be *more* specific, but not *less* specific.

zethzeth:
Thanks again for your suggestions/edits/comments, @jdreesen .  
I just committed it all.

stof:
Twig 1.x is not maintained anymore.
And Twig 2.x and 3.x have already been updated to support PHP 8.1+ without deprecation warnings.

ronaldoaires:
Is it possible to add the return type?

```
    public function count(): int
    {
        return \count($this->nodes);
    }
```

xabbuh:
@ronaldoaires `#[\ReturnTypeWillChange]` was added to that method in 2.x and 3.x. Adding a real return type would be a BC break.

PR #3838, opened 2023-05-04
Fix PHP 8.1 deprecation warnings on branch 1.x

TomA-R:
## What
Fix PHP 8.1 deprecation warnings, e.g.:

```
Deprecated: Return type of Twig\Node\Node::count() should either be compatible with Countable::count(): int, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in twig/twig/src/Node/Node.php on line 213
Deprecated: Return type of Twig\Node\Node::getIterator() should either be compatible with IteratorAggregate::getIterator(): Traversable, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in twig/twig/src/Node/Node.php on line 221
```

## Why
To fix deprecation warnings on PHP 8.1

I realise 1.x is quite old but it still has some usage

### DIFF
--- a/src/Markup.php
+++ b/src/Markup.php
@@ -32,10 +32,7 @@ class Markup implements \Countable
         return $this->content;
     }
 
-    /**
-     * @return int
-     */
-    public function count()
+    public function count(): int
     {
         return \function_exists('mb_get_info') ? mb_strlen($this->content, $this->charset) : \strlen($this->content);
     }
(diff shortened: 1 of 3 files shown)


fabpot:
Twig 1 is not maintained anymore. These changes are already part of versions 2 and 3.

stof:
quoted from the [1.x readme](https://github.com/twigphp/Twig/tree/1.x#readme):

> **Twig version 1 is not maintained anymore.**

PR #3840, opened 2023-05-06
Suppress native return type deprecation msg

PhilETaylor:
> Method "Symfony\Component\DependencyInjection\Extension\ExtensionInterface::load()" might add "void" as a native return type declaration in the future. Do the same in implementation "Twig\Extra\TwigExtraBundle\DependencyInjection\TwigExtraExtension" now to avoid errors or add an explicit `@return` annotation to suppress this message.


fabpot:
Thank you @PhilETaylor.

Issue #3842, opened 2023-05-26
ParsedownExtra with MarkdownExtension?


joeworkman:
Is it possible to use [erusev/parsedown-extra](https://github.com/erusev/parsedown-extra/) with the MarkdownExtension?

stof:
The [ErusevMarkdown class](https://github.com/twigphp/Twig/blob/833e51ea358e152d41ad46f30a7a4cbb38c44aed/extra/markdown-extra/ErusevMarkdown.php#L16) implementing the integration of erusev/parsedown with the MarkdownExtension supports injecting the Parsedown instance in its constructor. So it is possible (not sure the twig-extra-bundle allows you to configure the extension this way easily in a Symfony project without a compiler pass modifying services though)

joeworkman:
I guess I would need to submit a PR to make this happen. Let me see what I can do...

fabpot:
#3737 can probably be used as a source of inspiration.

PR #3843, opened 2023-05-26
Parsedown Extra support

joeworkman:
This adds support for [ParsedownExtra](https://github.com/erusev/parsedown-extra) to be used with Twig. 

related to #3842

### DIFF
--- a/extra/markdown-extra/ErusevMarkdownExtra.php
+++ b/extra/markdown-extra/ErusevMarkdownExtra.php
@@ -1,0 +1,29 @@
+<?php
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) Fabien Potencier
+ *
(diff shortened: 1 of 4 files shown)


extra/markdown-extra/ErusevMarkdownExtra.php
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * (c) Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Twig\Extra\Markdown;
+
+class ErusevMarkdownExtra implements MarkdownInterface
```
stof, line 14: Why creating a new ErusevMarkdownExtra ? The existing ErusevMarkdown class supports ParsedownExtra already.

PR #3844, opened 2023-06-05
Fix IntlExtension::formatDateTime use of date formatter prototype

drjayvee:
See https://github.com/twigphp/intl-extra/pull/6 for more details

### DIFF
--- a/extra/intl-extra/Tests/IntlExtensionTest.php
+++ b/extra/intl-extra/Tests/IntlExtensionTest.php
@@ -18,6 +18,18 @@ use Twig\Loader\ArrayLoader;
 
 class IntlExtensionTest extends TestCase
 {
+    public function testFormatterWithoutProto()
+    {
+        $ext = new IntlExtension();
+        $env = new Environment(new ArrayLoader());
+
+        $this->assertSame('12.346', $ext->formatNumber('12.3456'));
+        $this->assertSame(
+            'Feb 20, 2020, 1:37:00 PM',
+            $ext->formatDateTime($env, new \DateTime('2020-02-20T13:37:00+00:00'))
+        );
+    }
+
     public function testFormatterProto()
     {
         $dateFormatterProto = new \IntlDateFormatter('fr', \IntlDateFormatter::FULL, \IntlDateFormatter::FULL, new \DateTimeZone('Europe/Paris'));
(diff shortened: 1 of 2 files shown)


stof:
It would be great to have tests to prevent regressions.

drjayvee:
> It would be great to have tests to prevent regressions.

Voilà!

extra/intl-extra/Tests/IntlExtensionTest.php
```diff
@@ -12,17 +12,25 @@
...
-        $dateFormatterProto = new \IntlDateFormatter('fr', \IntlDateFormatter::FULL, \IntlDateFormatter::FULL);
+        $dateFormatterProto = new \IntlDateFormatter('fr', \IntlDateFormatter::FULL, \IntlDateFormatter::FULL, new \DateTimeZone('Europe/Paris'));
         $numberFormatterProto = new \NumberFormatter('fr', \NumberFormatter::DECIMAL);
         $numberFormatterProto->setTextAttribute(\NumberFormatter::POSITIVE_PREFIX, '++');
         $numberFormatterProto->setAttribute(\NumberFormatter::FRACTION_DIGITS, 1);
         $ext = new IntlExtension($dateFormatterProto, $numberFormatterProto);
+        $env = new Environment(new ArrayLoader());
+        
         $this->assertSame('++12,3', $ext->formatNumber('12.3456'));
+        $this->assertSame(
```
stof, line 43: I suggest splitting the assertion for the date formatter prototype into a separate tests to make the intent clearer. This would also make it easier to have _multiple_ tests for it to cover the various cases:

- passing a locale explicitly (should not use the prototype one) vs using the prototype locale
- passing `false` as timezone vs passing a timezone vs passing `null` as timezone
- in #3845, the different cases regarding the pattern.

drjayvee: Good points. If only those tests had been written from the get-go 😉

fabpot:
Thank you @drjayvee.

fabpot:
Merged

keulinho:
This change seems to break the fallback of using the timezone set in twigs `CoreExtension`.
The issue seems to be in this line:
```php
        if (false === $formatterTimezone) {
            $formatterTimezone = $date->getTimezone();
```

`$formatterTimezone` is basically never false, as if it is not provided directly it is null, so the timezone from the converted date (which is based on the one configured in the CoreExtension) is never used.

Quickfix would be to use:
```php
        if (null === $formatterTimezone) {
            $formatterTimezone = $date->getTimezone();
```

Issue #3568, opened 2021-09-30
ext-intl's use of IntlDateFormatter prototype has several bugs


drjayvee:
`\Twig\Extra\Intl\IntlExtension`'s constructor takes an optional `IntlDateFormatter` prototype. That works great for settings various defaults (`dateType, timeType, pattern`), but it doesn't enable automatic timezone conversion.

# IntlDateFormatter
Consider this plain PHP example:
```php
echo (new IntlDateFormatter(
	'nl_NL',
	IntlDateFormatter::MEDIUM, IntlDateFormatter::MEDIUM,
	new DateTimeZone('Europe/Amsterdam')
))->format(new DateTime()) . "\n";
```
This automatically formats the UTC datetime using Amsterdam time. Neat!

# twig/intl-extra
`new IntlExtension(new IntlDateFormatter(/* ... */))` however doesn't apply the timezone, unfortunately.

## Why not?
[`IntlExtension::formatDateTime`](https://github.com/twigphp/intl-extra/blob/2.x/IntlExtension.php#L280) lets [`\twig_date_converter`](https://github.com/twigphp/Twig/blob/3.x/src/Extension/CoreExtension.php#L458) create a `DateTime` using `CoreExtension::getTimezone()` and passes its timezone to `IntlExtension::createDateFormatter()` - even when no timezone was passed to `formatDateTime`.

Passing in an explicit `$timezone` to `createDateFormatter()` makes the method use that value instead of `$this->dateFormatterPrototype->getTimeZone()`. (which is actually broken - see https://github.com/twigphp/intl-extra/pull/6 ;p).

# Quirky workaround
*Maybe* users are supposed to
```php
$twig->getExtension(CoreExtension::class)->setTimeZone('Europe/Amsterdam');
```
In that case though, date *strings* are not converted at all:
```
{{ '2021-09-30 13:37'|format_datetime() }} {# still 13:37 #}
```
`DateTime` objects *are* converted.

stof:
> In that case though, date _strings_ are not converted at all:

Well, if your string does not contain a timezone offset, which conversion should be applied ? There is no info about the existing timezone.

drjayvee:
Wow that's a fast reply!

>There is no info about the existing timezone.

There is, actually. We set
```php
ini_set('date.timezone', 'UTC');
```
`CoreExtension::getTimezone()` uses `date_default_timezone_get()` to get a default value, which in this case is `UTC`.

drjayvee:
[...]

PR #3845, opened 2023-06-05
Fix IntlExtension::formatDateTime uses prototype pattern if set

drjayvee:
See https://github.com/twigphp/intl-extra/pull/7 for details

### DIFF
--- a/extra/intl-extra/IntlExtension.php
+++ b/extra/intl-extra/IntlExtension.php
@@ -423,7 +423,13 @@ final class IntlExtension extends AbstractExtension
             $timeFormatValue = $timeFormatValue ?: $this->dateFormatterPrototype->getTimeType();
             $timezone = $timezone ?: $this->dateFormatterPrototype->getTimeType();
             $calendar = $calendar ?: $this->dateFormatterPrototype->getCalendar();
-            $pattern = $pattern ?: $this->dateFormatterPrototype->getPattern();
+            if ('' === $pattern) {
+                if (!$dateFormat && !$timeFormat) {    // no explicit values for either format given
+                    $pattern = $this->dateFormatterPrototype->getPattern();
+                } else {
+                    $pattern = null;
+                }
+            }
         }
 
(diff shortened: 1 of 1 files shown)


admdly:
Is there any chance this could get reviewed/merged? @fabpot 

We've been having this issue over at [FOSSBilling/FOSSBilling](https://github.com/FOSSBilling/FOSSBilling/) for a while, and it would be great to get a fix in place.

drjayvee:
Hey Adam,

Could you please write a test for this? My other PR (#3844) was blocked until test coverage was added (and rightly so).

It should be really easy to copy one of the tests and adjust it slightly.

BelleNottelling:
Sadly this patch goes and breaks the usage of the `IntlDateFormatter` as in my testing the options set there are no longer respected. Appling the patch and using `format_datetime` with the date format set to `none` still displays a date. Likewise setting the time format to `none` still shows a time.

So it "fixes" the error by instead breaking it which is a shame as the current problem where `format_date` and `format_datetime` both behave identically by only following the set pattern is quite frustrating.

So with the existing twig releases `format_date` will still display a time if a time format is set. We also can't workaround that by manually specifying a format for a given filter as it's ignored. With this patch that behavior goes away, but seemingly because what's set via the `IntlDateFormatter` is completely ignored.

drjayvee:
@BelleNottelling I'm afraid I don't quite follow. It's been a while since I've looked at this bug/PR. In fact, [my original explanation](https://github.com/twigphp/Twig/issues/3568#issuecomment-931382860) is from September 2021. It's possible the behavior has changed since then.

Can you show some actual code that breaks only with this PR's code? Looking at the diff, it only affects behavior when

1. a `dateFormatterPrototype` is configured, and
2. neither `dateFormat` or `timeFormat` arguments are provided (to either filter)

> Likewise setting the time format to `none` still shows a time.

I don't see how this PR's diff affects formatting in that case.

fabpot:
See #4859 for my take on this. I think I've covered comments as well.

drjayvee:
Bonjour Fabien,

Thanks for looking into this. Your PR looks great, so I'll close this one.

PR #3846, opened 2023-06-06
Add void return type

simonsolutions:
add void-return type to functions in order to prevent depreciation message in Symfony 6.3


stof:
The `@return void` already avoids the deprecation message.

simonsolutions:
I tried it with Symfony 6.3, but it didn't

stof:
My guess is that you are using the stable release of the bundle. those `@return void` are only in the dev version for now.

xabbuh:
I think the reason is that #3840 wasn't part of a release yet.

stof:
3.6.1 is now released.

Issue #929, opened 2012-12-11
Named arguments for macros


hason:
```
--TEST--
macro
--TEMPLATE--
{% from _self import test %}

{% macro test(a, b = 'bar') -%}
{{ a }}{{ b }}
{%- endmacro %}

{{ test('foo') }}
{{ test('bar', 'foo') }}
{{ test(b = 'bar', a = 'foo') }}
--DATA--
return array();
--EXPECT--
foobar
barfoo
foobar
```

Twig_Error_Syntax: Arguments must be separated by a comma. Unexpected token "operator" of value "=" ("punctuation" expected with value ",") in "C:\public_html\Twig\test\Twig\Tests\Fixtures\macros\default_values.test" at line 10

ninsuo:
@chetzof just discovered that named arguments are already implemented and do work very well.
http://twigfiddle.com/w1dwmd
Anyway, if you have lots and lots of arguments, the second solution I given looks cleaner.

zimmermanj42:
@ninsuo Just to clarify, named arguments for calls to macros still are not implemented. The example you link to does not used named function in the calls. This would be a more proper example, and as you can see, it fails even with the latest Twig code at this time.
http://twigfiddle.com/w1dwmd/2

ninsuo:
I do not understand the notation:

```jinja
{{ _self.test(1, b=42) }}
```

Do you want this to be read as "pass `42` to test and `set b = 42` ? 

IMHO, this would introduce some ugly practices in templates.

zimmermanj42:
@ninsuo The modification I made wasn't intended to be a good use of the feature, but just to show that the feature being discussed here is still not implemented (as it throws the exception), as you had stated in an earlier comment that it was.  In hindsight I probably should have taken the time to build a more relevant example.  I came to this thread from a search and didn't want people to possibly be confused and thinking that this should be working for them.

In terms of the feature, in a way I can see it being more of something like "code smell", but I see no difference in been able to define arguments by name in the macro call or using the array of options methodology.  They essentially do the same thing.

However, I see named arguments as a better solution as enforcement of those arguments are defined in the macro definition and handled via the engine vs. needing to write more twig code in the macro to make sure things are set, defaulted, etc.

Despite this, I understand that supporting this is a much deeper issue with the code, so the array of options is an OK workaround until this may be able to be integrated into the software.

alpacagh:
@ninsuo your link shows the "default values", not a "named arguments".
[...]

PR #3849, opened 2023-06-09
Mention where named arguments are supported

mpdude:
This follows up on #929. I tend to forget this detail, so it confuses me over and over again.

### DIFF
--- a/doc/templates.rst
+++ b/doc/templates.rst
@@ -205,6 +205,8 @@ built-in functions.
 Named Arguments
 ---------------
 
+Named arguments are supported in functions, filters and tests.
+
 .. code-block:: twig
 
     {% for i in range(low=1, high=10, step=2) %}


doc/templates.rst
```diff
@@ -204,6 +204,9 @@ built-in functions.
 Named Arguments
 ---------------
 
+Named arguments are supported in functions,
+filters and tests.
```
stof: the line break should be removed. It breaks much too early.

stof:
the 1.x branch is not maintained anymore.

mpdude:
Rebased onto `2.x` and review comments addressed.

I should have known `1.x` is unsupported. I was mislead because in the documentation, it says "... switch to documentation for Twig 1.x, Twig 2.x` in a way that suggests also 1.x receives (documentation) updates.

stof (approved):

fabpot:
Thank you @mpdude.

Issue #3598, opened 2021-12-13
attribute() not looking for camelCase method


linuxprocess:
Hi,
I think it's a sort of bug, but it may be classified as a whish.

I'm using Symfony. I've added the non public attribute $event_login to an entity class. 
I meet a twig error when using {{ attribute(item, field) }} : Neither the property "event_login" nor one of the methods "event_login()", "getevent_login()"/"isevent_login()"/"hasevent_login()" or "__call()" exist and have public access in class "App\Entity\Event".
So I add the getevent_login() getter ...
Then I meet Symfony\Component\PropertyAccess\Exception\NoSuchPropertyException : Can't get a way to read the property "event_login" in class "App\Entity\Event".
So I add the getEventLogin() getter ...

Why cannot they use the same getter ?

After a few exchanges, man states that a method should always been camelCase() (see https://www.php-fig.org/psr/psr-1/ ).
Note that using underscores in attributes is allowed.
That's why PropertyAccess use getEventLogin() ... 

So, why does the twig attribute() function use getevent_login() instead of getEventLogin() ?

Should not it at least also check for getEventLogin() existence ?

Regards

stof:
The error you get is from the PropertyAccess component, not from Twig.

stof:
and Twig supports using either of them for the getter

linuxprocess:
Just to be sure : you're saying the error  **Neither the property event_login" nor one of the methods "event_login()", "getevent_login()"/"isevent_login()"/"hasevent_login()" or "__call()" exist and have public access in class "App\Entity\Event".** is not a twig error ?
It's comming from vendor/twig/twig/src/Extension/CoreExtension.php ...

stof:
ah no indeed. Your post contains 2 error messages, not 1. And I was wrong when thinking that Twig converts snake case names into camel case ones for getters.

linuxprocess:
Should not you reopen ?

Cruiser13:
It looks as if the attribute function is indeed broken here (tested on 3.4.1).
I do have an entity "Inhabitant" which has a "getPerson()" method which does have a "getName()" method.
Sample code:
```twig
                    {% set check = 'person.name' %}
                    {{ attribute(targetEntity, check) }}
                    {{ targetEntity.person.name }}
```
Running second line will give me the following error:

[...]

PR #3851, opened 2023-06-15
Using camelcase method for accessing snakecase property

nsetyo:
When we access a property using snake_case twig will not look for a camelCase getter as in #3598 . This pull request will make twig also looking for the camelCase getter for accessing a property.


GromNaN:
The `twig_get_attribute` function is called for every single property access. We must be very careful to the performance when adding feature to that function.

If we decide that this feature has a place in Twig, the implementation will have to be reworked to avoid a `preg_match` every time a property is read.

Edit: I spoke a little rapidly, this code is only executed once per class.

PR #3853, opened 2023-06-19
Improve error message for `constant(FQCN::class)`

MalteWunsch:
It is not obvious that you cannot use the Twig function `constant` to get the special `::class` constants of classes like so: `{{ constant('Twig\\Extension\\CoreExtension::class') }}`, due to the underlying PHP function `\constant()`.

The previous error message
> Constant "Twig\Extension\CoreExtension::class" is undefined.

was generally helpful, but in this special case it was a bit misleading.

An alternative might be to determine and return the FCQN (e.g. via `substr($constant, 0, -7)`), but I'm not sure if this would be too much of a guess.

### DIFF
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -1456,6 +1456,10 @@ final class CoreExtension extends AbstractExtension
         }
 
         if (!\defined($constant)) {
+            if ('::class' === strtolower(substr($constant, -7))) {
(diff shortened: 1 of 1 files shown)


MalteWunsch:
[These fabbot issues](https://fabbot.io/report/twigphp/Twig/3853/582f30d71b8e66f8a97ce7e4c02e6e89f0ff1fb8) seem to be unrelated.

fabpot:
Thank you @MalteWunsch.

PR #3855, opened 2023-07-01
Update signature to acknowledge a TemplateWrapper

richardhj:
`$name` legitimately can be of type `Twig\TemplateWrapper`. This is already considered in line 1331. This PR updates the method signature in the PHPDoc.

This PR is necessary because other developers may rely on the signature, see contao/contao#6169.


richardhj:
The fabbot.io complaints look unrelated.

fabpot:
Thank you @richardhj.

PR #3856, opened 2023-07-05
Fix callable phpdoc for twig elements

VincentLanglet:
Hi @fabpot 
```
[MyRuntime::class, 'myMethod]
```
is not a valid callable in PHP 8 https://3v4l.org/tWqPs

But this syntax is supported because twig has a special behavior
```
$compiler->raw(sprintf('$this->env->getRuntime(\'%s\')->%s', $callable[0], $callable[1]));
```

The PHPDoc should be updated in order to reflect the real behavior and solve static analysis errors.
A phpstan discussion can be found here: https://github.com/phpstan/phpstan/discussions/9567

### DIFF
--- a/src/TwigFilter.php
+++ b/src/TwigFilter.php
@@ -29,7 +29,7 @@ final class TwigFilter
     private $arguments = [];
 
     /**
-     * @param callable|null $callable A callable implementing the filter. If null, you need to overwrite the "node_class" option to customize compilation.
+     * @param callable|array<class-string, string>|null $callable A callable implementing the filter. If null, you need to overwrite the "node_class" option to customize compilation.
      */
     public function __construct(string $name, $callable = null, array $options = [])
     {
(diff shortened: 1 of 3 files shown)


fabpot:
Maybe that's something we want to deprecate?

xabbuh:
@fabpot Isn't it a requirement for lazy loading?

stof:
Why would we want to deprecate it ? this is what supports the runtime feature, enabling lazy-loading.

stof (approved):

jordisala1991:
Are you sure is a class-string and not a service id?

fabpot:
Thank you @VincentLanglet.

VincentLanglet:
> Are you sure is a class-string and not a service id?

Yes @jordisala1991, the signature is
´´´
public function getRuntime(string $class)
´´´

VincentLanglet:
@fabpot Thanks for the merge.

Currently, all phpstan-bleeding-edge users will have static analysis build failing since this PR is currently not release.
I was wondering what is the twig planning for release, and if it was possible to make a patch release of twig 3.x if none was planned soon. Thanks :)

fabpot:
Sorry for the delay @VincentLanglet; this one fell through the cracks. I've just made a new release.

PR #3858, opened 2023-07-18
Update Template.php

Jugolo:
When use {%include somting%} it will be conveted to  $this->loadTemplate(($context["somting"] ?? null), "<>", 1)

When context["somting"] is not exists it will be null. and a errror happens at $tempelate cant be null.  This will fix it. I dont know if it a propper way

### DIFF
--- a/src/Template.php
+++ b/src/Template.php
@@ -301,6 +301,9 @@ abstract class Template
      */
     protected function loadTemplate($template, $templateName = null, $line = null, $index = null)
     {
+        if($tempelate == null){
+            throw new RuntimeError("Unknown tempelate", $line, $this->getSourceContext());
+        }
         try {
             if (\is_array($template)) {
                 return $this->env->resolveTemplate($template);


src/Template.php
```diff
@@ -301,6 +301,9 @@ public function getBlockNames(array $context, array $blocks = [])
      */
     protected function loadTemplate($template, $templateName = null, $line = null, $index = null)
     {
+        if($tempelate == null){
```
GromNaN, line 304: The variable name is incorrect.

GromNaN (changes requested):
Your PR will require an integration test to demonstrate the bug fixed and ensure that there will be no regression.

fabpot:
Having a `null` template is not supported. I don't think we need to "improve" the error message as it will not be supported in 4.x when we will have proper type hinting on `Template::loadTemplate`.

See https://github.com/twigphp/Twig/pull/3943/files#diff-0a40cf21ba3009b815b6646d4e2d2ad13e93c9603b6ea66919b1994cc828116bR300

Issue #3828, opened 2023-04-05
Add sequence and mapping tests


pdureau:
Jinja has:

- the iterable test: https://jinja.palletsprojects.com/en/3.1.x/templates/#jinja-tests.iterable
- the sequence test: https://jinja.palletsprojects.com/en/3.1.x/templates/#jinja-tests.sequence
- the mapping test: https://jinja.palletsprojects.com/en/3.1.x/templates/#jinja-tests.mapping

Twig has only the [iterable test]( https://github.com/twigphp/Twig/pull/700), which returns true for any PHP array or any object implementing the Traversable interface :  https://twig.symfony.com/doc/3.x/tests/iterable.html

Sometimes, it is necessary to differentiate between sequences and mappings. 

For example, Drupal templates are using "render arrays"  (renderable objects injected into the template as a mapping variable) and testing "is iterable" is not enough. 

Is is conceivable to add those 2 tests?

pudovmaxim:
What do you mean by "sequences" and "mappings"?

It seems like in Drupal "render array" is simple array with some conventions:

> **Render arrays**
The core structure of the Render API is the render array, which is a hierarchical **associative array** containing data to be rendered and properties describing how the data should be rendered.
https://api.drupal.org/api/drupal/core!lib!Drupal!Core!Render!theme.api.php/group/theme_render/8.2.x#arrays

PHP have classes Sequence and Map in SPL. But they are rarely used. And it can be tested with additional extensions implement method of "instanceof" like here: https://gist.github.com/xthiago/ad5bce2f7d39bdd54b6f

pdureau:
Hi @pudovmaxim , 

> What do you mean by "sequences" and "mappings"?

Those terms come from Jinja terminology and are used many  times in Twig documentation without being properly defined. Examples:

- "The slice filter extracts a slice of a sequence, a mapping, or a string". https://twig.symfony.com/doc/3.x/filters/slice.html
- "The first filter returns the first "element" of a sequence, a mapping, or a string". https://twig.symfony.com/doc/3.x/filters/first.html

A `sequence` is called a `list` in Pyhton or an `array` in JSON. PHP  8.1 introduced a similar concept in [array_is_list](https://www.php.net/manual/en/function.array-is-list.php):

> Determines if the given array is a list. An array is considered a list if its keys consist of consecutive numbers from 0 to count($array)-1. 

A `mapping` is called a `dict` in Python or an `object` in JSON. In PHP, it would be any array which is not a list.

[...]

PR #3859, opened 2023-07-20
Issue #3828: Add sequence and mapping tests

pdureau:
Add sequence and mapping tests from Jinja.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -3,9 +3,9 @@
  * Deprecate the second argument of `Twig\Node\Expression\CallExpression::compileArguments()`
  * Deprecate `Twig\ExpressionParser\parseHashExpression()` in favor of
    `Twig\ExpressionParser::parseMappingExpression()`
-
  * Deprecate `Twig\ExpressionParser\parseArrayExpression()`` in favor of
    `Twig\ExpressionParser::parseSequenceExpression()`
+ * Add `sequence` and `mapping` tests.
 
 # 3.10.3 (2024-05-16)
 
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,8 @@
         "symfony/polyfill-php80": "^1.22",
         "symfony/deprecation-contracts": "^2.5|^3",
         "symfony/polyfill-mbstring": "^1.3",
-        "symfony/polyfill-ctype": "^1.8"
+        "symfony/polyfill-ctype": "^1.8",
+        "symfony/polyfill-php81": "^1.29"
     },
     "require-dev": {
         "symfony/phpunit-bridge": "^5.4.9|^6.4|^7.0",
--- a/doc/tests/mapping.rst
+++ b/doc/tests/mapping.rst
@@ -1,0 +1,14 @@
+``mapping``
+===========
+
+``mapping`` checks if a variable is a mapping:
+
+.. code-block:: twig
+
+    {% set users = {alice: "Alice Dupond", bob: "Bob Smith"} %}
+    {# evaluates to true if the users variable is a mapping #}
+    {% if users is mapping %}
+        {% for key, user in users %}
+            {{ key }}: {{ user }};
+        {% endfor %}
+    {% endif %}
--- a/doc/tests/sequence.rst
+++ b/doc/tests/sequence.rst
@@ -1,0 +1,14 @@
+``sequence``
+============
+
+``sequence`` checks if a variable is a sequence:
+
+.. code-block:: twig
+
+    {% set users = ["Alice", "Bob"] %}
+    {# evaluates to true if the users variable is a sequence #}
+    {% if users is sequence %}
+        {% for user in users %}
+            Hello {{ user }}!
+        {% endfor %}
+    {% endif %}
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -260,6 +260,8 @@ final class CoreExtension extends AbstractExtension
             new TwigTest('constant', null, ['node_class' => ConstantTest::class]),
             new TwigTest('empty', [self::class, 'testEmpty']),
             new TwigTest('iterable', 'is_iterable'),
+            new TwigTest('sequence', [self::class, 'testSequence']),
+            new TwigTest('mapping', [self::class, 'testMapping']),
         ];
     }
 
@@ -1283,6 +1285,56 @@ final class CoreExtension extends AbstractExtension
         }
 
         return '' === $value || false === $value || null === $value || [] === $value;
+    }
+
+    /**
+     * Checks if a variable is a sequence.
+     *
+     *    {# evaluates to true if the foo variable is a sequence #}
+     *    {% if foo is sequence %}
+     *        {# ... #}
+     *    {% endif %}
+     *
+     * @param mixed $value
+     *
+     * @internal
+     */
+    public static function testSequence($value): bool
+    {
+        if ($value instanceof \ArrayObject) {
+            $value = $value->getArrayCopy();
+        }
+
+        if ($value instanceof \Traversable) {
+            $value = iterator_to_array($value);
+        }
+
+        return \is_array($value) && array_is_list($value);
+    }
+
+    /**
+     * Checks if a variable is a mapping.
+     *
+     *    {# evaluates to true if the foo variable is a mapping #}
+     *    {% if foo is mapping %}
+     *        {# ... #}
+     *    {% endif %}
+     *
+     * @param mixed $value
+     *
+     * @internal
+     */
+    public static function testMapping($value): bool
+    {
+        if ($value instanceof \ArrayObject) {
+            $value = $value->getArrayCopy();
+        }
+
+        if ($value instanceof \Traversable) {
+            $value = iterator_to_array($value);
+        }
+
+        return (\is_array($value) && !array_is_list($value)) || \is_object($value);
     }
 
     /**
--- a/doc/tests/iterable.rst
+++ b/doc/tests/iterable.rst
@@ -5,7 +5,7 @@
 
 .. code-block:: twig
 
-    {# evaluates to true if the foo variable is iterable #}
+    {# evaluates to true if the users variable is iterable #}
     {% if users is iterable %}
         {% for user in users %}
             Hello {{ user }}!
--- a/tests/Fixtures/tests/mapping.test
+++ b/tests/Fixtures/tests/mapping.test
@@ -1,0 +1,38 @@
+--TEST--
+"mapping" test
+--TEMPLATE--
+{{ empty is mapping ? 'ok' : 'ko' }}
+{{ sequence is mapping ? 'ok' : 'ko' }}
+{{ empty_array_obj is mapping ? 'ok' : 'ko' }}
+{{ sequence_array_obj is mapping ? 'ok' : 'ko' }}
+{{ mapping_array_obj is mapping ? 'ok' : 'ko' }}
+{{ obj is mapping ? 'ok' : 'ko' }}
+{{ mapping is mapping ? 'ok' : 'ko' }}
+{{ string is mapping ? 'ok' : 'ko' }}
+--DATA--
+return [
+    'empty' => [],
+    'sequence' => [
+        'foo',
+        'bar',
+        'baz'
+    ],
+    'empty_array_obj' => new \ArrayObject(),
+    'sequence_array_obj' => new \ArrayObject(['foo', 'bar']),
+    'mapping_array_obj' => new \ArrayObject(['foo' => 'bar']),
+    'obj' => new \stdClass(),
+    'mapping' => [
+        'foo' => 'bar',
+        'bar' => 'foo'
+    ],
+    'string' => 'test',
+]
+--EXPECT--
+ko
+ko
+ko
+ko
+ok
+ok
+ok
+ko
--- a/tests/Fixtures/tests/sequence.test
+++ b/tests/Fixtures/tests/sequence.test
@@ -1,0 +1,38 @@
+--TEST--
+"sequence" test
+--TEMPLATE--
+{{ empty is sequence ? 'ok' : 'ko' }}
+{{ sequence is sequence ? 'ok' : 'ko' }}
+{{ empty_array_obj is sequence ? 'ok' : 'ko' }}
+{{ sequence_array_obj is sequence ? 'ok' : 'ko' }}
+{{ mapping_array_obj is sequence ? 'ok' : 'ko' }}
+{{ obj is sequence ? 'ok' : 'ko' }}
+{{ mapping is sequence ? 'ok' : 'ko' }}
+{{ string is sequence ? 'ok' : 'ko' }}
+--DATA--
+return [
+    'empty' => [],
+    'sequence' => [
+        'foo',
+        'bar',
+        'baz'
+    ],
+    'empty_array_obj' => new \ArrayObject(),
+    'sequence_array_obj' => new \ArrayObject(['foo', 'bar']),
+    'mapping_array_obj' => new \ArrayObject(['foo' => 'bar']),
+    'obj' => new \stdClass(),
+    'mapping' => [
+        'foo' => 'bar',
+        'bar' => 'foo'
+    ],
+    'string' => 'test',
+]
+--EXPECT--
+ok
+ok
+ok
+ok
+ko
+ko
+ko
+ko


src/Extension/CoreExtension.php
```diff
@@ -1300,6 +1302,65 @@ function twig_test_iterable($value)
...
+ *        {# ... #}
+ *    {% endif %}
+ *
+ * @param mixed $value A variable
+ *
+ * @return bool true if the value is a mapping
+ */
+function twig_test_mapping($value)
+{
+    return \is_array($value) && (empty($value) || !\array_is_list($value));
```
fabpot: We tried never to use `empty`.

src/Extension/CoreExtension.php
```diff
@@ -1300,6 +1302,65 @@ function twig_test_iterable($value)
...
+}
+
+/**
+ * Checks whether a given array is a list
```
```suggestion
 * Checks whether a given array is a list.
```

fabpot (changes requested):

src/Extension/CoreExtension.php
pdureau: Thanks, I did replace empty($value) by (0 === \count($value)) in commit 508bc72

src/Extension/CoreExtension.php
pdureau: Thanks, I added the dot in commit https://github.com/twigphp/Twig/commit/508bc72c99bea1bfb69284610aaa154a3abeaf9a

pdureau:
Not 100% sure but I believe the fails reported by https://fabbot.io/report/twigphp/Twig/3859/508bc72c99bea1bfb69284610aaa154a3abeaf9a doesn't come from the code I have changed.

src/Extension/CoreExtension.php
```diff
@@ -1300,6 +1302,65 @@ function twig_test_iterable($value)
...
+ *        {# ... #}
+ *    {% endif %}
+ *
+ * @param mixed $value A variable
+ *
+ * @return bool true if the value is a sequence
+ */
+function twig_test_sequence($value)
+{
+    return $value instanceof \Traversable || (\is_array($value) && \array_is_list($value));
```
GromNaN: Not all `Traversable` are sequences, they can return any key. Example https://3v4l.org/LJ0R2

The implementation of `array_is_list` could accept any `iterable`, a new name will be required as this is not identical to the PHP 8.1 function.

src/Extension/CoreExtension.php
```diff
@@ -1300,6 +1302,65 @@ function twig_test_iterable($value)
...
+ *        {# ... #}
+ *    {% endif %}
+ *
+ * @param mixed $value A variable
+ *
+ * @return bool true if the value is a mapping
+ */
+function twig_test_mapping($value)
+{
+    return \is_array($value) && ((0 === \count($value)) || !\array_is_list($value));
```
GromNaN: Any `stdClass` is a mapping. That's the type returned by `json_decode` by default for objects. It's also the only way to have an empty mapping.

This test would be more useful if it was different than a `<var> is not sequence`.

Given this [definition of this test in Jinja](https://jinja.palletsprojects.com/en/2.11.x/templates/#mapping), the implementation should check that none of the keys are numeric.
> mapping(value)
> Return true if the object is a mapping (dict etc.).

src/Extension/CoreExtension.php
```diff
@@ -1300,6 +1302,65 @@ function twig_test_iterable($value)
...
+ *
+ * Polyfill for PHP 7.4 to 8.0.
+ * From https://www.php.net/manual/en/function.array-is-list.php#127044
+ *
+ * @param mixed $array The array being evaluated.
+ *
+ * @return bool true if array is a list, false otherwise
+ */
+if (!function_exists('array_is_list')) {
+    function array_is_list(array $array)
```
GromNaN: `symfony/polyfill-php81` should be used instead of creating the polyfill here.
But that will not be necessary if you change the implementation to support any `iterable` (previous comment).

smustgrave:
+1 for this feature.  Working on a Drupal site and this causes issues for rendering lists.

src/Extension/CoreExtension.php
stof: The issue for an iterator is that this requires consuming the iterator, which might not be rewindable to consume it a second time (see generators)

src/Extension/CoreExtension.php
stof: numeric keys that are not sequentials would still be a mapping IMO.

src/Extension/CoreExtension.php
stof: this should indeed depend on `symfony/polyfill-php81` instead of providing its own untested polyfill.

pdureau:
I have rebased my work with current 3.x branch, so my MR has changed.

composer.json
```diff
@@ -28,7 +28,8 @@
         "symfony/polyfill-php80": "^1.22",
         "symfony/deprecation-contracts": "^2.5|^3",
         "symfony/polyfill-mbstring": "^1.3",
-        "symfony/polyfill-ctype": "^1.8"
+        "symfony/polyfill-ctype": "^1.8",
+        "symfony/polyfill-php81": "1.x-dev"
```
FlorentTorregrosa: Should it be `^1.29`?

pdureau:
Hello all, my PR is ready for review.

xabbuh:
fabbot suggests some changes that should be address: https://fabbot.io/report/twigphp/Twig/3859/5629adca007fc77574def9d5a309f713a95610ce

pdureau:
Like last time, the 2 feedbacks from fabbot are not related to the code I have added or changed. Could a human do the review?

src/Extension/CoreExtension.php
```diff
@@ -1285,6 +1287,73 @@ public static function testEmpty($value): bool
...
+     * @param mixed $value A variable
+     *
+     * @return bool true if the value is a sequence
+     */
+    public static function testSequence($value)
+    {
+        if ($value instanceof \ArrayObject) {
+            $value = $value->getArrayCopy();
+        }
+        // array_is_list() is provided by symfony/polyfill-php81.
```
xabbuh: missing blank line before the comment

src/Extension/CoreExtension.php
```diff
@@ -1285,6 +1287,73 @@ public static function testEmpty($value): bool
...
+        return \is_array($value) && \array_is_list($value);
```
xabbuh:
```suggestion
        return \is_array($value) && array_is_list($value);
```

src/Extension/CoreExtension.php
```diff
@@ -1285,6 +1287,73 @@ public static function testEmpty($value): bool
...
+     * @param mixed $value A variable
+     *
+     * @return bool true if the value is a mapping
+     */
+    public static function testMapping($value)
+    {
+        if ($value instanceof \ArrayObject) {
+            $value = $value->getArrayCopy();
+        }
+        // array_is_list() is provided by symfony/polyfill-php81.
```
xabbuh: missing blank line before comment

src/Extension/CoreExtension.php
```diff
@@ -1285,6 +1287,73 @@ public static function testEmpty($value): bool
...
+        return (\is_array($value) && !\array_is_list($value)) || \is_object($value);
```
xabbuh:
```suggestion
        return (\is_array($value) && !array_is_list($value)) || \is_object($value);
```

xabbuh:
@fabpot Can you approve the workflow so that all jobs on GHA are triggered as well?

pdureau:
thanks you @xabbuh  I have added a new commit with:
- Missing blank lines before comments
- array_is_list() is not always part of the global namespace

fabpot:
> @fabpot Can you approve the workflow so that all jobs on GHA are triggered as well?

Done

src/Extension/CoreExtension.php
```diff
@@ -1285,6 +1287,75 @@ public static function testEmpty($value): bool
...
+     * @return bool true if the value is traversable
+     *
+     * @deprecated since Twig 3.8, to be removed in 4.0 (use the native "is_iterable" function instead)
+     *
+     * @internal
+     */
+    public static function testIterable($value)
+    {
+        return is_iterable($value);
+    }
```
fabpot: This method should be removed, right?

pdureau: I would keep `iterable` test in Twig, because:

- `mapping` & `sequence` tests are not replacing it. Jinja has all three: https://jinja.palletsprojects.com/en/3.0.x/templates/#builtin-tests
- it also covers strings AFAIK, so `mapping` + `sequence` != `iterable`
- removing it will introduce a compatibility break

stof: the comment is not about removing the Twig test. It is about removing the internal method used to implement it, as `is_iterable` can be used directly as the implementation of the Twig test (and it is already used directly btw).

src/Extension/CoreExtension.php
```diff
@@ -1285,6 +1287,75 @@ public static function testEmpty($value): bool
...
+     *
+     * @return bool true if the value is a mapping
+     */
+    public static function testMapping($value)
+    {
+        if ($value instanceof \ArrayObject) {
+            $value = $value->getArrayCopy();
+        }
+
+        // array_is_list() is provided by symfony/polyfill-php81.
```
stof: I suggest removing this comment entirely (for PHP 8.1+, the function is provided by PHP itself, so the comment is not even correct)

src/Extension/CoreExtension.php
```diff
@@ -1285,6 +1287,75 @@ public static function testEmpty($value): bool
...
+     *    {# evaluates to true if the foo variable is a sequence #}
+     *    {% if foo is sequence %}
+     *        {# ... #}
+     *    {% endif %}
+     *
+     * @param mixed $value A variable
+     *
+     * @return bool true if the value is a sequence
+     */
+    public static function testSequence($value)
```
stof: those new methods should be marked as `@internal` so that they are not part of the public API of Twig but only an internal implementation detail of the Twig tests.

src/Extension/CoreExtension.php
```diff
@@ -1285,6 +1287,75 @@ public static function testEmpty($value): bool
...
+     * Traversables must be converted before being tested.
+     *
+     *    {# evaluates to true if the foo variable is a sequence #}
+     *    {% if foo is sequence %}
+     *        {# ... #}
+     *    {% endif %}
+     *
+     * @param mixed $value A variable
+     *
+     * @return bool true if the value is a sequence
```
stof: this should be a native return type instead.

src/Extension/CoreExtension.php
```diff
@@ -1285,6 +1287,75 @@ public static function testEmpty($value): bool
...
+     */
+    public static function testIterable($value)
+    {
+        return is_iterable($value);
+    }
+
+    /**
+     * Checks if a variable is a sequence.
+     *
+     * Traversables must be converted before being tested.
```
stof: this behavior is not consistent with the behavior of other Twig APIs, and I don't think we expose the `toArray` method to Twig code, so it is probably impossible to satisfy this requirement easily.

stof (changes requested):

src/Extension/CoreExtension.php
pdureau: ok, got it. It is already marked "@deprecated since Twig 3.8, to be removed in 4.0 (use the native "is_iterable" function instead)"

src/Extension/CoreExtension.php
pdureau: Done in https://github.com/twigphp/Twig/pull/3859/commits/3d23fe1c1aaca508021e4ca2f372e8bdccd648aa

src/Extension/CoreExtension.php
pdureau: Done in https://github.com/twigphp/Twig/pull/3859/commits/3d23fe1c1aaca508021e4ca2f372e8bdccd648aa

src/Extension/CoreExtension.php
stof: well, introducing a new method that is already deprecated does not make sense. A deprecation is meant to avoid breaking existing code that uses that method, delaying the removal until the next major version where BC breaks are allowed.
For a new method, there is no existing code that could break, as the method does not exist at all in previous versions

pdureau: OK, understood. It may be the leftover of a rebase. Removed in https://github.com/twigphp/Twig/pull/3859/commits/db3753d8464804c96b067dff8e7619152b21ad33

src/Extension/CoreExtension.php
pdureau: I am afraid I don't understand this feedback

src/Extension/CoreExtension.php
pdureau: @dunglas can we talk about this part? I may have misunderstood an advice shared in Rennes late March

src/Extension/CoreExtension.php
fabpot: Remove the phpdoc line, and add the type hint directly to the method signature (`public static function testSequence($value): bool
`)

src/Extension/CoreExtension.php
fabpot: Have a look at the `CoreExtension::toArray()` code, it is used everywhere in the file to easily convert a sequence to an array when possible so that this operation is transparent for users.

src/Extension/CoreExtension.php
```diff
@@ -1285,6 +1287,54 @@ public static function testEmpty($value): bool
...
+     *        {# ... #}
+     *    {% endif %}
+     *
+     * @param mixed $value A variable
+     *
+     * @return bool true if the value is a sequence
+     *
+     * @internal
+     */
+    public static function testSequence($value)
```
```suggestion
    public static function testSequence($value): bool
```

src/Extension/CoreExtension.php
```diff
@@ -1285,6 +1287,54 @@ public static function testEmpty($value): bool
...
+     *        {# ... #}
+     *    {% endif %}
+     *
+     * @param mixed $value A variable
+     *
+     * @return bool true if the value is a mapping
+     *
+     * @internal
+     */
+    public static function testMapping($value)
```
```suggestion
    public static function testMapping($value): bool
```

fabpot (changes requested):

src/Extension/CoreExtension.php
pdureau: Thanks a lot. Done in https://github.com/twigphp/Twig/pull/3859/commits/798c8ffa10543aebe5e534fc043b5ba7c0580198

src/Extension/CoreExtension.php
pdureau: Thansk a lot. Done in https://github.com/twigphp/Twig/pull/3859/commits/798c8ffa10543aebe5e534fc043b5ba7c0580198

src/Extension/CoreExtension.php
pdureau: Thanks a lot. Done in https://github.com/twigphp/Twig/pull/3859/commits/798c8ffa10543aebe5e534fc043b5ba7c0580198

src/Extension/CoreExtension.php
pdureau: Thanks. I didn't know [iterator_to_array](https://www.php.net/manual/fr/function.iterator-to-array.php)

pdureau: Done in https://github.com/twigphp/Twig/pull/3859/commits/95f6983b671cfbf2d8b0ed096de57cbdbc0e172f  

Two tests use [ArrayIterator](https://www.php.net/manual/fr/class.arrayiterator.php) which implements [SeekableIterator](https://www.php.net/manual/fr/class.seekableiterator.php) which extends [Traversable](https://www.php.net/manual/fr/class.traversable.php) (through [Iterator](https://www.php.net/manual/fr/class.iterator.php)) and they are still passing OK

src/Extension/CoreExtension.php
```diff
@@ -1285,6 +1287,56 @@ public static function testEmpty($value): bool
...
 
+    /**
+     * Checks if a variable is a sequence.
+     *
+     *    {# evaluates to true if the foo variable is a sequence #}
+     *    {% if foo is sequence %}
+     *        {# ... #}
+     *    {% endif %}
+     *
+     * @param mixed $value A variable
```
```suggestion
     * @param mixed $value
```

src/Extension/CoreExtension.php
```diff
@@ -1285,6 +1287,56 @@ public static function testEmpty($value): bool
...
+
+    /**
+     * Checks if a variable is a mapping.
+     *
+     *    {# evaluates to true if the foo variable is a mapping #}
+     *    {% if foo is mapping %}
+     *        {# ... #}
+     *    {% endif %}
+     *
+     * @param mixed $value A variable
```
```suggestion
     * @param mixed $value
```

pdureau: Done in https://github.com/twigphp/Twig/pull/3859/commits/ee5dcd6444218f6640825c73266f124c9013f3fe

src/Extension/CoreExtension.php
pdureau: Done in https://github.com/twigphp/Twig/pull/3859/commits/ee5dcd6444218f6640825c73266f124c9013f3fe

src/Extension/CoreExtension.php
```diff
@@ -260,6 +260,8 @@ public function getTests(): array
             new TwigTest('constant', null, ['node_class' => ConstantTest::class]),
             new TwigTest('empty', [self::class, 'testEmpty']),
             new TwigTest('iterable', 'is_iterable'),
+            new TwigTest('sequence', [self::class, 'testSequence']),
+            new TwigTest('mapping', [self::class, 'testMapping']),
```
fabpot, line 264: I've just checked Twig docs and code, and we are using "sequence" and "hash".
So, we have 2 options here:
 * Rename "mapping" to "hash" to be consistent with Twig
 * Keep "mapping" here for consistency with Jinja and rename "hash" to "mapping" in current docs and code (that would not break anything).

I'd go with consistency with Jinja here.

What do others think? @stof @GromNaN @xabbuh

dunglas: +1 for consistency with Jinja.

xabbuh: I'd also go for consistency with Jinja 👍

pdureau: "Mapping" is already used in some pages of the documentation. Examples:

https://twig.symfony.com/doc/3.x/filters/map.html
"The map filter applies an arrow function to the elements of a sequence or a mapping."

https://twig.symfony.com/doc/3.x/filters/first.html
"The first filter returns the first "element" of a sequence, a mapping, or a string"

fabpot:
@pdureau We're almost there! Thank you for your patience and your hard work here. The last step before being able to merge (now that the names won't change anymore) is to add a note in CHANGELOG and documentation under doc/.

pdureau:
@fabpot thanks for your message. I added  a commit with  a note in CHANGELOG and documentation under doc/. Very simple and straightforward, maybe too simple.

https://github.com/twigphp/Twig/pull/3859/commits/5c27e1a95d232014ceab651532df768ca38d77de

CHANGELOG
```diff
@@ -1,6 +1,6 @@
 # 3.10.4 (2024-XX-XX)
 
- * n/a
+ * Add sequence & mapping tests.
```
```suggestion
 * Add `sequence` and `mapping` tests
```

doc/tests/mapping.rst
```diff
@@ -0,0 +1,13 @@
+``mapping``
+============
```
fabpot:
```suggestion
===========
```

doc/tests/mapping.rst
```diff
@@ -0,0 +1,13 @@
...
+``mapping`` checks if a variable is a mapping:
+
+.. code-block:: twig
+
+    {# evaluates to true if the users variable is a mapping #}
+    {% if users is mapping %}
+        {% for key, user in users %}
+            {{ key }}: {{ user }};
+        {% endfor %}
+    {% endif %}
```
fabpot, line 14: Maybe we can a note about how we define a mapping?

doc/tests/sequence.rst
```diff
@@ -0,0 +1,13 @@
...
+``sequence`` checks if a variable is a sequence:
+
+.. code-block:: twig
+
+    {# evaluates to true if the users variable is a sequence #}
+    {% if users is sequence %}
+        {% for user in users %}
+            Hello {{ user }}!
+        {% endfor %}
+    {% endif %}
```
fabpot, line 14: Maybe we can a note about how we define a sequence?

fabpot (changes requested):

doc/tests/mapping.rst
pdureau: sure

pdureau: Done in https://github.com/twigphp/Twig/pull/3859/commits/9ecaf164fcf9de0a2f782153dbfa235d88a4efb6

doc/tests/sequence.rst
pdureau: Done in https://github.com/twigphp/Twig/pull/3859/commits/9ecaf164fcf9de0a2f782153dbfa235d88a4efb6

doc/tests/mapping.rst
pdureau: Done in https://github.com/twigphp/Twig/pull/3859/commits/9ecaf164fcf9de0a2f782153dbfa235d88a4efb6

CHANGELOG
pdureau: Done in https://github.com/twigphp/Twig/pull/3859/commits/9ecaf164fcf9de0a2f782153dbfa235d88a4efb6

fabpot:
Thank you @pdureau.

PR #3861, opened 2023-07-21
Introduce ImplicitKeyExpression in ArrayExpression

nicolas-grekas:
A better fix than #3860


nicolas-grekas:
This changes the AST.

PR #3862, opened 2023-07-27
Fix callable phpdoc for twig elements

VincentLanglet:
Hi @fabpot 
I made a stupid mistake in https://github.com/twigphp/Twig/pull/3856

The correct phpdoc is `array{class-string, string}` and not `array<class-string, string>`.

Code is used this way
```
$compiler->raw(sprintf('$this->env->getRuntime(\'%s\')->%s', $callable[0], $callable[1]));
```
which means the key 0 is a class string, and the key 1 is a string.

`array{class-string, string}` is same as `array{0: class-string, 1: string}`
`array<class-string, string>` is saying all the keys are class-string, and values are string.

### DIFF
--- a/src/TwigFilter.php
+++ b/src/TwigFilter.php
@@ -29,7 +29,7 @@ final class TwigFilter
     private $arguments = [];
 
     /**
-     * @param callable|array<class-string, string>|null $callable A callable implementing the filter. If null, you need to overwrite the "node_class" option to customize compilation.
+     * @param callable|array{class-string, string}|null $callable A callable implementing the filter. If null, you need to overwrite the "node_class" option to customize compilation.
      */
     public function __construct(string $name, $callable = null, array $options = [])
     {
@@ -57,7 +57,7 @@ final class TwigFilter
     /**
      * Returns the callable to execute for this filter.
      *
-     * @return callable|array<class-string, string>|null
+     * @return callable|array{class-string, string}|null
      */
     public function getCallable()
     {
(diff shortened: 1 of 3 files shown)


VincentLanglet:
cc @stof which reviewed the last PR

src/TwigFilter.php
```diff
@@ -29,7 +29,7 @@ final class TwigFilter
     private $arguments = [];
 
     /**
-     * @param callable|array<class-string, string>|null $callable A callable implementing the filter. If null, you need to overwrite the "node_class" option to customize compilation.
+     * @param callable|array{class-string, string}|null $callable A callable implementing the filter. If null, you need to overwrite the "node_class" option to customize compilation.
```
xabbuh, line 32: Is `array{class-string, string}` as a shortcut for `array{0: class-string, 1: string}` understood by all important static analyzers (Psalm, PhpStan, PhpStorm)?

jdreesen: AFAIK yes

xabbuh (approved):

src/TwigFilter.php
VincentLanglet: AFAIK too, yes.

But if you prefer (to be explicit ?) I can use `array{0: class-string, 1: string}`

fabpot:
Thank you @VincentLanglet.

abozhinov:
I do it because PHPStan check. It failed if I don't fixed it.

VincentLanglet:
Hi @fabpot

Could a patch release be considered to solve StaticAnalysis build failure on lot of project using twig ? :)

Thanks

fabpot:
Release done

PR #3872, opened 2023-08-22
Add `enum_cases` function

### LINKED ISSUE #3681: Support reading enums from PHP 8.1
Hi,

Since PHP 8.1 we can use Enums. Which in my opinion are a great asset to PHP.
But unfortunate there is not an elegant way of retrieving values inside Twig templates.

I currently use this in my Symfony project;

```twig
{{ constant('App\\...\\Enum::Key').value }}
```

Which will return the value of the key in de the Enum.

Maybe it's good to add some specific functions for this in the Twig core?

Regards, Bert

HeahDude:
May fix #3681.

Allows:
```twig
{% for enum in enum_cases('App\\MyBackedEnum') %}
    {{ enum.value }}
{% endfor %}
```
________
Use case:
We need to display supported locales for a language switcher in the menu.
The block is included in a header template which is included in the base. It is cumbersome to pass `Locale::cases()` in the context from all controllers rendering a template that inherits from the base, or to cascade the variable from template to template.

Simplified implementation:
```php
enum Locale: string
{
    case EN = 'en';
    case FR = 'fr';

    public function flag(): string
    {
        return match ($this) {
            self::EN => '🇬🇧',
            self::FR => '🇫🇷',
        };
    }
}
```
```twig
{% for locale in enum_cases('App\\Locale') %}
    <option value="{{ locale.value }}"{% if locale.value == app.locale %} selected{% endif %}>
        {{ locale.flag }}
    </option>
{% endfor %}
```

### DIFF
--- a/src/Extension/EnumExtension.php
+++ b/src/Extension/EnumExtension.php
@@ -1,0 +1,40 @@
+<?php
+
+namespace Twig\Extension;
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Twig\TwigFunction;
+
+class EnumExtension extends AbstractExtension
+{
+    public function getFunctions(): array
+    {
+        return [
+            new TwigFunction('enum_cases', __CLASS__.'::enumCases'),
+        ];
+    }
+
+    /**
+     * @template T of \BackedEnum
+     *
+     * @param class-string<T> $backedEnum
+     *
+     * @return T[]
+     */
+    public static function enumCases(string $backedEnum): array
+    {
+        if (!is_a($backedEnum, \BackedEnum::class, true)) {
+            throw new \InvalidArgumentException(sprintf('The enum must be a "\BackedEnum", "%s" given.', $backedEnum));
+        }
+
+        return $backedEnum::cases();
+    }
+}
--- a/tests/Extension/EnumExtensionTest.php
+++ b/tests/Extension/EnumExtensionTest.php
@@ -1,0 +1,50 @@
+<?php
+
+namespace Twig\Tests\Extension;
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use PHPUnit\Framework\TestCase;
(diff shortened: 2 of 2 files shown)


HeahDude:
I've updated the description to show a basic use case.

src/Extension/EnumExtension.php
```diff
@@ -0,0 +1,40 @@
...
+     * @template T of \BackedEnum
+     *
+     * @param class-string<T> $backedEnum
+     *
+     * @return T[]
+     */
+    public static function enumCases(string $backedEnum): array
+    {
+        if (!is_a($backedEnum, \BackedEnum::class, true)) {
+            throw new \InvalidArgumentException(sprintf('The enum must be a "\BackedEnum", "%s" given.', $backedEnum));
```
GromNaN, line 35: The restriction on `BackedEnum` is not necessary. As we can see in you `Locale::flag` example, the enum case can be used as argument to call a function, or directly to call a method on it.

tests/Extension/EnumExtensionTest.php
```diff
@@ -0,0 +1,50 @@
...
+    {
+        $this->expectException(\InvalidArgumentException::class);
+        $this->expectExceptionMessage('The enum must be a "\BackedEnum", "Twig\Tests\Extension\MyUnitEnum" given.');
+
+        EnumExtension::enumCases(MyUnitEnum::class);
+    }
+}
+
+if (80100 <= \PHP_VERSION_ID) {
+    enum MyBackedEnum: string
```
GromNaN, line 39: Even if you put an `if` around it, `enum` is a syntax error for PHP < 8.1. The `@requires PHP 8.1` doesn't help either since the file can't be compiled on PHP 8.0 and bellow. That should be OK if you move this enum declarations in other files.

fabpot:
Any news @HeahDude?

stof:
for reference, I implemented a similar function at Incenteev, with a specificity: I'm using a custom node type to perform a special compilation when the argument is a string literal:

- I validate that the argument is a valid enum class at compile time (which conveniently helps when you forget to escape backslashes and end up not having them in the actual string)
- I compile the code directly to a called to `MyEnum::cases()` without the wrapper function, avoiding to repeat the enum validation at runtime.

Do you want me to submit my implementation as a PR ?

stof:
And I'm wondering whether it should be in a dedicated extension or in CoreExtension.

fabpot:
@stof It would be great if you could submit a PR. Having it in CoreExtension sounds good to me.

PR #3873, opened 2023-08-30
Fix premature loop exit in Security Policy lookup of allowed methods/properties

YSaxon:
The current security policy logic exits too soon when checking permissions for allowed classes and their methods/properties, causing false negatives in situations involving classes related by inheritance.

Consider the following configuration:

```
'methods' => [
  'App\BasicCollection' => ['sortAlphabetically'],
  'App\AdvancedCollection'=> ['sortByTimestamp'],
],
```
where `AdvancedCollection` is a subclass of `BasicCollection`, and `mylist` is an instance of `AdvancedCollection`

If you try to call `{{ mylist.sortByTimestamp() }}`, the current code will first match `mylist` against `App\BasicCollection`. Since `sortByTimestamp` is not an allowed method for `App\BasicCollection`, the code will exit the loop and incorrectly deny access. It will never get to checking `App\AdvancedCollection`.

Note that reordering classes in the config can't solve this issue. If you flipped the order, then it would fail for `{{ mylist.sortAlphabetically() }}` instead.

This pull request fixes the issue by only exiting the loop early when both the class and method/property match.

### DIFF
--- a/src/Sandbox/SecurityPolicy.php
+++ b/src/Sandbox/SecurityPolicy.php
@@ -94,9 +94,8 @@ final class SecurityPolicy implements SecurityPolicyInterface
         $allowed = false;
         $method = strtr($method, 'ABCDEFGHIJKLMNOPQRSTUVWXYZ', 'abcdefghijklmnopqrstuvwxyz');
         foreach ($this->allowedMethods as $class => $methods) {
-            if ($obj instanceof $class) {
-                $allowed = \in_array($method, $methods);
-
+            if ($obj instanceof $class && \in_array($method, $methods)) {
+                $allowed = true;
                 break;
             }
         }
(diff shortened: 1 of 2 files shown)


YSaxon:
I just modified the pull request to point to the 2.x branch instead of the 3.x branch.

YSaxon:
@fabpot if you can take a look at this one as well please

fabpot:
@YSaxon Can you add a test to ensure that we won't introduce a regression later on.

src/Sandbox/SecurityPolicy.php
```diff
@@ -101,7 +101,7 @@ public function checkMethodAllowed($obj, $method)
         }
 
         if (!$allowed) {
-            $class = \get_class($obj);
+            $class = $obj::class;
```
fabpot: Can you revert the changes unrelated to this PR? Sometimes fabbot is just wrong ;)
Basically, reverting this commit.

fabpot (changes requested):

src/Sandbox/SecurityPolicy.php
YSaxon: That's why I split the commit in two, to make sure it would be easy to revert if necessary.

fabpot:
Thank you @YSaxon.

PR #3878, opened 2023-09-11
Fix CI

fabpot:


tests/drupal_test.sh
```diff
@@ -18,7 +18,7 @@ wget https://get.symfony.com/cli/installer -O - | bash
 export PATH="$HOME/.symfony5/bin:$PATH"
 symfony server:start -d --no-tls
 
-curl -OLsS https://get.blackfire.io/blackfire-player.phar
+curl -OLsS https://get.blackfire.io/blackfire-player-v1.31.0.phar
```
stof: instead of `-O`, you should use `-o blackfire-player.phar` so that the file created on the file system does not have the version number in it (which will avoid having to change all usages each time the version is updated)

PR #3883, opened 2023-10-08
Deprecate internal extension functions in favor of methods on the extension classes

fabpot:
~~PoC on a simple extension for now.~~

### DIFF
--- a/doc/deprecated.rst
+++ b/doc/deprecated.rst
@@ -10,3 +10,10 @@ Functions
 
  * The `twig_test_iterable` function is deprecated; use the native
    `is_iterable` instead.
+
+Extensions
+----------
+
+* All functions defined in Twig extensions are marked as internal as of Twig
+  3.9.0, and will be removed in Twig 4.0. They have been replaced by internal
+  methods on their respective extension classes.
--- a/extra/html-extra/HtmlExtension.php
+++ b/extra/html-extra/HtmlExtension.php
@@ -9,8 +9,10 @@
  * file that was distributed with this source code.
  */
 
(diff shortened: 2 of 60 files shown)


doc/deprecated.rst
```diff
@@ -4,3 +4,10 @@ Deprecated Features
 This document lists deprecated features in Twig 3.x. Deprecated features are
 kept for backward compatibility and removed in the next major release (a
 feature that was deprecated in Twig 3.x is removed in Twig 4.0).
+
+Extensions
+----------
+
+* All functions defined in Twig extensions are marked as internal as of Twig
+  3.8.0, and will be removed in Twig 4.0. They have been replaced by methods on
+  their respective extension classes.
```
GromNaN: Make it clear that they can't be used directly.

```suggestion
* All functions defined in Twig extensions are marked as internal as of Twig
  3.8.0, and will be removed in Twig 4.0. They have been replaced by internal
  methods on their respective extension classes.
```

extra/html-extra/HtmlExtension.php
```diff
@@ -34,7 +36,7 @@ public function getFilters(): array
     public function getFunctions(): array
     {
         return [
-            new TwigFunction('html_classes', 'twig_html_classes'),
+            new TwigFunction('html_classes', [static::class, 'htmlClasses']),
```
GromNaN: There is no need for the late static binding, the class is final.
```suggestion
            new TwigFunction('html_classes', [self::class, 'htmlClasses']),
```

fabpot:
Not WIP anymore.
Ready for review.

PR #3881, opened 2023-10-07
Drop support for PHP < 8.2 (aka version 4.0)

fabpot:

PR #3884, opened 2023-10-08
Use PHP 8.0 functions with polyfill

GromNaN:
Backport some changes from #3881 into 3.x using `symfony/polyfill-php80`.

- `str_starts_with`
- `str_ends_with`
- `str_contains`
- `get_debug_type`

Benefits:
- Less conflicts when both 3.x and 4.x branches will be maintained.
- Performance gain for PHP 8.0+ ([83,7% of installs](https://packagist.org/packages/twig/twig/php-stats#3))

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,6 +25,7 @@
     ],
     "require": {
         "php": ">=7.2.5",
+        "symfony/polyfill-php80": "^1.22",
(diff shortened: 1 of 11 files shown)


fabpot:
Thank you @GromNaN.

PR #3886, opened 2023-10-08
Remove `TemplateWrapper::render()` 2nd parameter not used

GromNaN:
The 2nd virtual parameter of `TemplateWrapper::render()` was added by #2805.
`Template::render()` don't define and don't use this 2nd parameter.

Running the test suite, this method never get more than 1 parameter (I assumed it could be used by the compiled Template via the backtrace, but it isn't).


fabpot:
Thank you @GromNaN.

PR #3888, opened 2023-10-08
Remove unused variables and unreachable code

GromNaN:
I did some archaeology to find the origin of each useless line of code. Details in comments.

### DIFF
--- a/src/Environment.php
+++ b/src/Environment.php
@@ -345,7 +345,6 @@ class Environment
                 $this->cache->load($key);
             }
 
-            $source = null;
             if (!class_exists($cls, false)) {
                 $source = $this->getLoader()->getSourceContext($name);
                 $content = $this->compileSource($source);
--- a/src/ExpressionParser.php
+++ b/src/ExpressionParser.php
@@ -506,10 +506,6 @@ class ExpressionParser
             }
 
             if ($node instanceof NameExpression && null !== $this->parser->getImportedSymbol('template', $node->getAttribute('name'))) {
-                if (!$arg instanceof ConstantExpression) {
-                    throw new SyntaxError(sprintf('Dynamic macro names are not supported (called on "%s").', $node->getAttribute('name')), $token->getLine(), $stream->getSourceContext());
-                }
-
                 $name = $arg->getAttribute('value');
 
                 $node = new MethodCallExpression($node, 'macro_'.$name, $arguments, $lineno);
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -364,7 +364,6 @@ function twig_random(Environment $env, $values = null, $max = null)
             }
         } else {
             $min = $values;
-            $max = $max;
         }
 
         return mt_rand((int) $min, (int) $max);
@@ -669,7 +668,7 @@ function twig_slice(Environment $env, $item, $start, $length = null, $preserveKe
         return \array_slice($item, $start, $length, $preserveKeys);
     }
 
-    return (string) mb_substr((string) $item, $start, $length, $env->getCharset());
+    return mb_substr((string) $item, $start, $length, $env->getCharset());
 }
 
 /**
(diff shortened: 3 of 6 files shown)


src/Environment.php
```diff
@@ -345,7 +345,6 @@ public function loadTemplate(string $cls, string $name, int $index = null): Temp
                 $this->cache->load($key);
             }
 
-            $source = null;
```
GromNaN, line 348: Initialized in the `if` and never used outside.

src/ExpressionParser.php
```diff
@@ -506,10 +506,6 @@ public function parseSubscriptExpression($node)
             }
 
             if ($node instanceof NameExpression && null !== $this->parser->getImportedSymbol('template', $node->getAttribute('name'))) {
-                if (!$arg instanceof ConstantExpression) {
```
GromNaN, line 509: `$arg` is always a `ConstantExpression`, initialized line 496.

This condition should have been removed when the code was moved in https://github.com/twigphp/Twig/commit/9f67d588c8a39a07ce0d78db56d9ef6a7d4b07f2.

src/Extension/CoreExtension.php
```diff
@@ -364,7 +364,6 @@ function twig_random(Environment $env, $values = null, $max = null)
             }
         } else {
             $min = $values;
-            $max = $max;
```
GromNaN, line 367: > The left-hand and right-hand sides of the assignment are equal.

Added by [#2698](https://github.com/twigphp/Twig/pull/2698/files#diff-2fa56c635b23777069e2e6336d62e65fa50dee4e6dc76ca60ab9383c39b50ef9R356), perhaps to be explicit about the value of `$max`.

src/Extension/CoreExtension.php
```diff
@@ -669,7 +668,7 @@ function twig_slice(Environment $env, $item, $start, $length = null, $preserveKe
         return \array_slice($item, $start, $length, $preserveKeys);
     }
 
-    return (string) mb_substr((string) $item, $start, $length, $env->getCharset());
```
GromNaN, line 672: [`mb_substr`](https://www.php.net/manual/en/function.mb-substr.php#refsect1-function.mb-substr-returnvalues) always return a string.

src/Node/Node.php
```diff
@@ -27,7 +27,6 @@ class Node implements \Countable, \IteratorAggregate
     protected $lineno;
     protected $tag;
 
-    private $name;
```
GromNaN, line 30: Left-over from #2967

src/NodeVisitor/EscaperNodeVisitor.php
```diff
@@ -183,12 +183,11 @@ private function isSafeFor(string $type, Node $expression, Environment $env): bo
         return \in_array($type, $safe) || \in_array('all', $safe);
     }
 
-    private function needEscaping(Environment $env)
```
GromNaN, line 186: Parameter not used since [`1f00c33`](https://github.com/twigphp/Twig/commit/1f00c33acde77c1c3eb5c2c13c292f5e5afcffda#diff-f94a61cb3a67823e5fe2dd545a52b47962d3486d8eb9919106c69dd7910d12b5L138)

src/NodeVisitor/OptimizerNodeVisitor.php
```diff
@@ -125,7 +125,7 @@ private function optimizeRawFilter(Node $node, Environment $env): Node
     /**
      * Optimizes "for" tag by removing the "loop" variable creation whenever possible.
      */
-    private function enterOptimizeFor(Node $node, Environment $env): void
```
GromNaN, line 128: The `$env` parameter have never been used since this method was created https://github.com/twigphp/Twig/commit/bd5629b3f24d47463691afa22a6da5cad0b1a108

fabpot:
Thank you @GromNaN.

PR #3892, opened 2023-10-14
range example leads to Array to string conversion exception

devojifr:
In templates documentation, a range example is displayed and leads to an Array to string conversion exception without a join filter.


stof (approved):

fabpot:
Thank you @devojifr.

PR #3893, opened 2023-10-17
Add SourcePolicyInterface to selectively enable the Sandbox based on a template's Source

YSaxon:
This is needed to patch some downstream vulnerabilities which I won't describe here.
I wrote @fabpot an email about this with more details a few weeks ago.

Generally the Sandbox can be enabled for a given template in either of two ways:
* Globally
* Including the template from another template in which we use the sandbox tag or parameter

This pull request adds a third way
* Using a SourcePolicy to selectively sandbox templates based on their Source object


src/Extension/SandboxExtension.php
```diff
@@ -50,16 +53,21 @@ public function disableSandbox()
...
-        return $this->sandboxedGlobally || $this->sandboxed;
+        return $this->sandboxedGlobally || $this->sandboxed || $this->isSourceSandboxed($source);
     }
 
     public function isSandboxedGlobally()
     {
         return $this->sandboxedGlobally;
     }
 
+    public function isSourceSandboxed($source)
```
stof: This should have parameter and return type (for new methods, there is no backward compatibility issue preventing us to add it immediately rather than in the next major version)

stof: is there any reason to make this method public ?

src/Sandbox/SourcePolicyInterface.php
```diff
@@ -0,0 +1,24 @@
...
+use Twig\Source;
+
+/**
+ * Interface for a class that can optionally enable the sandbox mode based on a template's Twig\Source.
+ *
+ * @author Yaakov Saxon
+ */
+interface SourcePolicyInterface
+{
+    public function enableSandbox(Source $source = null): bool;
```
stof: why making the argument optional ?

YSaxon: Good point. My intention was to mark it as nullable, since the upstream method Source comes from is marked as = null. I didn't realize PHP has a distinct syntax for marking as required but nullable. I'll fix that.

src/Extension/SandboxExtension.php
YSaxon: No specific reason, just that all the rest of the functions in the class are public..

YSaxon: Done

stof: That's because other methods are part of the public API.

src/Extension/SandboxExtension.php
```diff
@@ -50,16 +53,21 @@ public function disableSandbox()
         $this->sandboxed = false;
     }
 
-    public function isSandboxed()
+    public function isSandboxed($source = null)
```
stof: this is missing the parameter type

src/Extension/SandboxExtension.php
```diff
@@ -50,16 +53,21 @@ public function disableSandbox()
...
-        return $this->sandboxedGlobally || $this->sandboxed;
+        return $this->sandboxedGlobally || $this->sandboxed || $this->isSourceSandboxed($source);
     }
 
     public function isSandboxedGlobally()
     {
         return $this->sandboxedGlobally;
     }
 
+    public function isSourceSandboxed(?Source $source) : bool
```
stof: This should be a private method unless there is a good reason to expose that publicly (once something is public, we cannot remove it anymore)

src/Sandbox/SourcePolicyInterface.php
```diff
@@ -0,0 +1,24 @@
...
+use Twig\Source;
+
+/**
+ * Interface for a class that can optionally enable the sandbox mode based on a template's Twig\Source.
+ *
+ * @author Yaakov Saxon
+ */
+interface SourcePolicyInterface
+{
+    public function enableSandbox(?Source $source): bool;
```
stof: should we even support null in this interface ? I don't know how an implementation would take a decision in this case. We should rather handle that case in the SandboxExtension, calling the source policy only when we have a source (which should always be the case in calls done in compiled templates rather than custom callers anyway)

YSaxon: I wasn't entirely sure under what circumstances Source could actually be null, but my thought was to allow the implementer to decide how to handle the null cases - essentially a question of whether to whitelist acceptable Sources (and block nulls) or blacklist unacceptable ones (and accept nulls)

YSaxon: But in any case I just added commit da79bdc1f5aeb8135701271049aab8b9b1f02309 as an implementation of what you said

YSaxon: (Post rebase, that's now commit cbbf38b7db8b768eec0aca1928aef2f2df80de64)

YSaxon:
I did a rebase onto the squash commit from my other pull request to solve a merge conflict.

YSaxon:
@stof

YSaxon:
@stof, I believe this is ready for review, having addressed the points you raised

YSaxon:
@fabpot

YSaxon:
@fabpot just a friendly reminder that I'm still waiting for a merge here so I can fix the downstream vulns. I believe the only "X" is a mistaken fabbot.io request to change things in existing code.

fabpot:
Thank you @YSaxon.

YSaxon:
Thanks @fabpot

Could we get a minor version release for this on 2.x?

fabpot:
@YSaxon Done now

PR #3894, opened 2023-10-20
fix `NumberFormatter::TYPE_CURRENCY` being deprecated in PHP 8.3

mbolli:
It was never used anyway (by Twig and by PHP), so removing it should not have any side-effects.


fabpot:
Thank you @mbolli.

OrestisZag:
@fabpot Can you please tag this? PHP 8.3 will be released in 2 days. Thanks a lot! 😄

PR #3898, opened 2023-10-25
[Twig Extra Bundle] : Add return type for functions

ghost:
Hi there,

In order to make some deprecation disappear, I'd like to submit this small PR.

Changed the return type of three function from phpdoc to native return types.

_Deprecations :_ 

`Method "Symfony\Component\DependencyInjection\Extension\ExtensionInterface::load()" might add "void" as a native return type declaration in the future. Do the same in implementation "Twig\Extra\TwigExtraBundle\DependencyInjection\TwigExtraExtension" now to avoid errors or add an explicit @return annotation to suppress this message.`

`Method "Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface::process()" might add "void" as a native return type declaration in the future. Do the same in implementation "Twig\Extra\TwigExtraBundle\DependencyInjection\Compiler\MissingExtensionSuggestorPass" now to avoid errors or add an explicit @return annotation to suppress this message.`

`User Deprecated: Method "Symfony\Component\HttpKernel\Bundle\Bundle::build()" might add "void" as a native return type declaration in the future. Do the same in child class "Twig\Extra\TwigExtraBundle\TwigExtraBundle" now to avoid errors or add an explicit @return annotation to suppress this message.`

Cheers,

Laurent

### DIFF
--- a/extra/twig-extra-bundle/DependencyInjection/Compiler/MissingExtensionSuggestorPass.php
+++ b/extra/twig-extra-bundle/DependencyInjection/Compiler/MissingExtensionSuggestorPass.php
@@ -18,8 +18,7 @@ use Twig\Environment;
 
 class MissingExtensionSuggestorPass implements CompilerPassInterface
 {
-    /** @return void */
-    public function process(ContainerBuilder $container)
+    public function process(ContainerBuilder $container): void
     {
         if ($container->getParameter('kernel.debug')) {
             $twigDefinition = $container->getDefinition('twig');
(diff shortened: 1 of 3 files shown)


stof:
there is already an explicit `@return void` so the deprecation should already not be triggered.

Are you sure your version of Twig is uptodate ?

lolo-orsu:
Thanks for your quick reply @stof !

Yes I'm up to date, got the 3.7.1 installed... I've cleared the cache, done everything I could, but the deprecation still there

stof:
then something is definitely wrong in your setup. I don't get these deprecations locally.

Could it be that you have `opcache.save_comments` configured to `0`, which would make opcache remove all doc comments before caching opcodes ? If yes, that could explain why the tooling does not recognize that there is a `@return void` in phpdoc in the doc comment of the method (because the code being analyzed does not have them anymore).

PR #3901, opened 2023-10-27
Security policy wildcard support for methods/properties

YSaxon:
Allows for flexible wildcard support in allowedMethods and allowedProperties in SecurityPolicy.
 - Class can be specified as wildcard, `* => ['foo',...]` in order to allow those methods/properties for all classes.
 - Method/property can be specified as wildcard eg. `\Foo\Bar\Baz => '*'` in order to allow **all** methods/properties for that class.
 - Method/property can also be specified with a trailing wildcard to allow all methods/properties with a certain prefix, eg. `\Foo\Bar\Baz => ['get*', ...]` in order to allow all Baz methods/properties that start with `get`.
 
 
Here are some real examples of security policy items that previously had to be hardcoded that can now be expressed with wildcards:
* Twig's hardcoded [security exceptions for methods in Template and Markup classes](https://github.com/twigphp/Twig/blob/3ae6fb8723776f4fc9d8ab490fcabcd3173da8e3/src/Sandbox/SecurityPolicy.php#L90) can be expressed as 
```
'Twig\Template' => '*',
'Twig\Markup' => '*'
```

* Drupal's [$allowed_classes](https://github.com/drupal/drupal/blob/6efb70cd568335e9b292cde6530cc21c2931c448/core/lib/Drupal/Core/Template/TwigSandboxPolicy.php#L49C7-L49C7) can now be expressed as 
```
Drupal\Core\Template\Attribute => '*'
```

* Drupal's [$allowed_methods](https://github.com/drupal/drupal/blob/6efb70cd568335e9b292cde6530cc21c2931c448/core/lib/Drupal/Core/Template/TwigSandboxPolicy.php#L57) and [$allowed_prefixes](https://github.com/drupal/drupal/blob/6efb70cd568335e9b292cde6530cc21c2931c448/core/lib/Drupal/Core/Template/TwigSandboxPolicy.php#L69) can be expressed together as 
```
'*' => ['get*', 'has*', 'is*', '__toString', 'toString', 'id', 'label', 'bundle']
```

### DIFF
--- a/src/Sandbox/MemberMatcher.php
+++ b/src/Sandbox/MemberMatcher.php
@@ -16,7 +16,7 @@ namespace Twig\Sandbox;
  * - Class can be specified as wildcard `* => [...]` in order to allow those methods/properties for all classes.
  * - Method/property can be specified as wildcard eg. `\DateTime => '*'` in order to allow all methods/properties for that class.
  * - Method/property can also be specified with a trailing wildcard to allow all methods/properties with a certain prefix, eg. `\DateTime => ['get*', ...]` in order to allow all methods/properties that start with `get`.
- * 
+ *
  * @author Yaakov Saxon <ysaxon@gmail.com>
  */
 final class MemberMatcher
@@ -28,16 +28,16 @@ final class MemberMatcher
     {
         $normalizedMembers = [];
         foreach ($allowedMembers as $class => $members) {
-            if (!is_array($members)) {
+            if (!\is_array($members)) {
                 $normalizedMembers[$class][] = strtolower($members);
-            }
-            else foreach ($members as $index => $member) {
-                $normalizedMembers[$class][$index] = strtolower($member);
+            } else {
+                foreach ($members as $index => $member) {
+                    $normalizedMembers[$class][$index] = strtolower($member);
+                }
             }
         }
         $this->allowedMembers = $normalizedMembers;
     }
-
 
     public function isAllowed($obj, string $member): bool
     {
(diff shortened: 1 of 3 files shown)


src/Sandbox/MemberMatcher.php
```diff
@@ -0,0 +1,79 @@
...
+namespace Twig\Sandbox;
+
+/**
+ * Allows for flexible wildcard support in allowedMethods and allowedProperties in SecurityPolicy.
+ * - Class can be specified as wildcard `* => [...]` in order to allow those methods/properties for all classes.
+ * - Method/property can be specified as wildcard eg. `\DateTime => '*'` in order to allow all methods/properties for that class.
+ * - Method/property can also be specified with a trailing wildcard to allow all methods/properties with a certain prefix, eg. `\DateTime => ['get*', ...]` in order to allow all methods/properties that start with `get`.
+ *
+ * @author Yaakov Saxon <ysaxon@gmail.com>
+ */
```
fabpot, line 21: Let's mark this class as`@internal`.

src/Sandbox/MemberMatcher.php
```diff
@@ -0,0 +1,79 @@
...
+final class MemberMatcher
+{
+    private $allowedMembers;
```
fabpot, line 24:
```suggestion
    private array $allowedMembers;
```

src/Sandbox/MemberMatcher.php
```diff
@@ -0,0 +1,79 @@
...
+    private $cache = [];
```
fabpot, line 25:
```suggestion
    private array $cache = [];
```

src/Sandbox/MemberMatcher.php
```diff
@@ -0,0 +1,79 @@
...
+                    $normalizedMembers[$class][$index] = strtolower($member);
+                }
+            }
+        }
+        $this->allowedMembers = $normalizedMembers;
+    }
+
+    public function isAllowed($obj, string $member): bool
+    {
+        $cacheKey = get_class($obj) . "::" . $member;
```
fabpot, line 44:
```suggestion
        $cacheKey = get_class($obj).'::'.$member;
```

src/Sandbox/MemberMatcher.php
```diff
@@ -0,0 +1,79 @@
...
+            return true;
+        }
+
+        $member = strtolower($member); // normalize member name
+
+        foreach ($this->allowedMembers as $class => $members) {
+            if ('*' === $class || $obj instanceof $class) {
+                foreach ($members as $allowedMember) {
+                    if ('*' === $allowedMember) {
+                        $this->cache[$cacheKey] = true;
```
fabpot, line 57:
```suggestion
                        return $this->cache[$cacheKey];
```

Same below

src/Sandbox/MemberMatcher.php
```diff
@@ -0,0 +1,79 @@
...
+
+                        return true;
+                    }
+                    if ($allowedMember === $member) {
+                        $this->cache[$cacheKey] = true;
+
+                        return true;
+                    }
+                    // if allowedMember ends with a *, check if the member starts with the allowedMember
```
fabpot, line 66: Why do we only allow `*`to be at the end?

fabpot (changes requested):
That will be for 3.x as 2.x EOL is reached.
_I know I've merged your other PR in 2.x, but as it was security related, I accepted it_

fabpot:
I forgot to mention that overall I'm not sold yet this is something I want in core. I very much prefer white-listing **explicitly** what is allowed. With `*`, future methods/properties/classes will be automatically "accepted", which can be problematic.

fabpot:
Or maybe that's fine but we need to make it very clear in the docs that this should be used with caution (can you some docs for this new feature?).

PR #3902, opened 2023-10-30
Optional defaults for SecurityPolicy

YSaxon:
This is an implementation of Defaults for the Twig Security Policy.

The idea is that there should be a set of known safe values that a user can whitelist. The goals for the pull request were:
* It should be easy to use the defaults
* It should be easy to NOT use the defaults
* It should be easy to include all the values in the defaults plus some of your own. I've seen implementations of defaults where you need to take it or leave it; if you like the defaults but want to add some extra values in your project, you have no choice but to reject the defaults and just copy them into your own policy.

I've implemented this via a special value or "token", meant to be referred to as SecurityPolicyDefaults::INCLUDE_DEFAULTS, which, if included as a value in any of the whitelist config arguments, will be replaced and merged with the actual hardcoded default values.

So for instance if you want to trust all of the default filters, plus your own custom filters `format_number` and `format_currency` and 'format_currency', you can initialize your SecurityPolicy with `$allowedFilters = ['format_number', 'format_currency', SecurityPolicyDefaults::INCLUDE_DEFAULTS]`, and the actual allowedFilters will end up as `['format_number', 'format_currency','abs','batch','capitalize', 'date', 'date_modify', ...]`.

Note that I haven't done extensive research for the present list of defaults. They are intended as a proof-of-concept sample for the purpose of the pull request, and should definitely be reviewed thoroughly before actually accepting and merging.

### DIFF
--- a/src/Sandbox/SecurityPolicy.php
+++ b/src/Sandbox/SecurityPolicy.php
@@ -29,25 +29,26 @@ final class SecurityPolicy implements SecurityPolicyInterface
 
     public function __construct(array $allowedTags = [], array $allowedFilters = [], array $allowedMethods = [], array $allowedProperties = [], array $allowedFunctions = [])
     {
-        $this->allowedTags = $allowedTags;
-        $this->allowedFilters = $allowedFilters;
+        $this->setAllowedTags($allowedTags);
+        $this->setAllowedFilters($allowedFilters);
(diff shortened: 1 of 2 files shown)


fabpot:
I'm closing here as each use case of using the sandbox is specific.

PR #3903, opened 2023-11-03
Fix timezone fallback to CoreExtension in IntlExtension

keulinho:
This is probably a regression from #3844

Refer to my comment on the original MR: https://github.com/twigphp/Twig/pull/3844#issuecomment-1792464040

### DIFF
--- a/extra/intl-extra/IntlExtension.php
+++ b/extra/intl-extra/IntlExtension.php
@@ -371,7 +371,7 @@ final class IntlExtension extends AbstractExtension
         $date = twig_date_converter($env, $date, $timezone);
 
         $formatterTimezone = $timezone;
-        if (false === $formatterTimezone) {
+        if (null === $formatterTimezone) {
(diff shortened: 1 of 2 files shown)


fabpot:
Thank you @keulinho.

petski:
This PR breaks the behavior of `format_datetime` with `timezone` set to `false`. Calling `format_datetime` with `timezone` to `false` [is documented as](https://twig.symfony.com/doc/3.x/filters/format_datetime.html):

> If the date is already a DateTime object, and if you want to keep its current timezone, pass false as the timezone value:
> {{ datetime|format_datetime(locale='en', timezone=false) }}

The exception thrown if the above feature is used:

```
Twig\Extra\Intl\IntlExtension::createDateFormatter(): Argument #5 ($timezone) must be of type ?DateTimeZone, false given, called in /var/www/vendor/twig/intl-extra/IntlExtension.php on line 379
```

xabbuh:
@petski see #3965

PR #3913, opened 2023-11-22
Add strict_properties, array_methods options

danog:
* `strict_properties`: if true, treats property accesses in templates only as property accesses, without ever trying to invoke methods with the same name if the property does not exist (defaults to false).
* `array_methods`: if true, treats method calls on callable array elements as if they were object method calls.

Both rules came in extremely handy @ nicelocal, as the first one increases strictness, also avoiding issues in our automated migration from smarty to twig, while the second one allows the usage of some specific abstractions.

### DIFF
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -1469,7 +1469,7 @@ function twig_array_batch($items, $size, $fill = null, $preserveKeys = true)
 function twig_get_attribute(Environment $env, Source $source, $object, $item, array $arguments = [], $type = /* Template::ANY_CALL */ 'any', $isDefinedTest = false, $ignoreStrictCheck = false, $sandboxed = false, int $lineno = -1)
 {
     // array
-    if (/* Template::METHOD_CALL */ 'method' !== $type) {
+    if (/* Template::METHOD_CALL */ 'method' !== $type || $env->isArrayMethods()) {
         $arrayItem = \is_bool($item) || \is_float($item) ? (int) $item : $item;
 
         if (((\is_array($object) || $object instanceof \ArrayObject) && (isset($object[$arrayItem]) || \array_key_exists($arrayItem, (array) $object)))
@@ -1479,7 +1479,13 @@ function twig_get_attribute(Environment $env, Source $source, $object, $item, ar
                 return true;
             }
 
-            return $object[$arrayItem];
+            if ($type === 'method') {
+                if (is_callable($object[$arrayItem])) {
+                    return $object[$arrayItem](...$arguments);
+                }
+            } else {
+                return $object[$arrayItem];
+            }
         }
 
         if (/* Template::ARRAY_CALL */ 'array' === $type || !\is_object($object)) {
@@ -1553,6 +1559,18 @@ function twig_get_attribute(Environment $env, Source $source, $object, $item, ar
             }
 
             return $object->$item;
+        }
+
+        if ($env->isStrictProperties()) {
+            if ($isDefinedTest) {
+                return false;
+            }
+
+            if ($ignoreStrictCheck || !$env->isStrictVariables()) {
+                return;
+            }
+
+            throw new RuntimeError(sprintf('The property "%1$s" does not exist in class "%2$s".', $item, get_class($object)), $lineno, $source);
         }
     }
 
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -1481,6 +1481,31 @@ function twig_get_attribute(Environment $env, Source $source, $object, $item, ar
 
             if ($type === 'method') {
                 if (is_callable($object[$arrayItem])) {
+                    if ($sandboxed) {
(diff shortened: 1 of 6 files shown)


src/Extension/CoreExtension.php
```diff
@@ -1479,7 +1479,13 @@ function twig_get_attribute(Environment $env, Source $source, $object, $item, ar
                 return true;
             }
 
-            return $object[$arrayItem];
+            if ($type === 'method') {
+                if (is_callable($object[$arrayItem])) {
+                    return $object[$arrayItem](...$arguments);
```
stof, line 1509: This is not properly integrated with the sandbox system.

src/Extension/CoreExtension.php
```diff
@@ -1554,6 +1560,18 @@ function twig_get_attribute(Environment $env, Source $source, $object, $item, ar
 
             return $object->$item;
         }
+
+        if ($env->isStrictProperties()) {
+            if ($isDefinedTest) {
+                return false;
+            }
+
+            if ($ignoreStrictCheck || !$env->isStrictVariables()) {
```
stof, line 1594: Why a check on strict variables here ?

stof (commented):
I'm not a fan of this `strict_properties` option.
The only way for a package to use Twig while being compatible with any project would then be to never rely on the implicit getter behavior, which would be a major breakage for the ecosystem (for instance, many things in Symfony are implemented as getters, not as public properties, as this allows computing things on-demand).
I don't think it is a good idea to force all shared packages to write non-idiomatic Twig code.

src/Extension/CoreExtension.php
danog: This avoids falling through to the method call logic below, without throwing an exception (for consistency with the chosen `strict_variables` setting).

stof: The `strict_variables` setting has a well-defined semantic: deciding whether undefined values are returned as `null` or throwing an exception.

danog:
Hi @stof,

> The only way for a package to use Twig while being compatible with any project would then be to never rely on the implicit getter behavior, which would be a major breakage for the ecosystem (for instance, many things in Symfony are implemented as getters, not as public properties, as this allows computing things on-demand).
> I don't think it is a good idea to force all shared packages to write non-idiomatic Twig code.

@ work we have a very large amount of twig templates, but we don't actually use any libraries from the symfony ecosystem (apart from twig), thus the fallback-to-methodcall behavior for properties was causing us issues, without any valid reasons to keep it enabled.

`strict_properties` is set to false by default, thus this does not affect the default behavior for anyone.

stof:
@danog the issue is that a shared package cannot assume that the current config is the default config. It must work in all modes. Otherwise, we create a split ecosystem.

src/Extension/CoreExtension.php
danog: This logic aims precisely at maintaining this behavior, regardless of the value of strict_properties.

src/Extension/CoreExtension.php
danog: Done!

danog:
@stof I now understand your point, however twig is not used exclusively with Symfony libraries: it is a standalone templating engine, that can be used even without symfony libraries; I see no real reason to offer an option to disable a behavior that, while idiomatic for twig templates *in the context of Symfony*, is not mandatory in contexts outside of Symfony (and in our specific case the default behavior is actually harmful, since it causes breakage and potentially unwanted behavior for thousands of templates).  

A (subjective) argument can even be made that the default behavior brings actually more harm than good, as it introduces unexpected side effects for what should be just a simple property fetch, and this opinion is shared by my peers and CTO @ work.   

Regardless, this PR does not aim to remove this behavior altogether, but rather provide an option to increase strictness, which frankly can only do good :)

I've added some additional docs to clarify that `strict_properties` should not be used in conjunction with Symfony libraries.

stof:
@danog being a standalone library makes its ecosystem event *larger* than the Symfony ecosystem, which only means that splitting the ecosystem between packages compatible with `strict_properties` or not compatible with it is even worse. It is not limited to the Symfony ecosystem.

Introducing this option in Twig means introducing the splitting of the ecosystem, which has a huge cost for the Twig community.

danog:
@stof The same argument of an ecosystem split can be made for `strict_variables`, `autoescape` with `html`, `js` values, yet these options exist and can be enabled or disabled at will; especially the `html`/`js` `autoescape` strategies, which were added at a later time in Twig 1.8, not directly at release, just like the proposed `strict_properties`.

danog:
Could I also get a second opinion from @fabpot as well? I strongly feel that the `strict_properties` option is an overall useful addition to Twig :)

stof:
@danog For `strict_variables`, the best practice is anyway to deal properly with the structure of your variables (making you compatible with `strict_variables` being turned on). Writing code compatible with `strict_variables` won't be non-idiomatic Twig.

fabpot:
Closing for reasons explained by @stof

PR #3916, opened 2023-11-26
Create attributes `AsTwigFilter`, `AsTwigFunction` and `AsTwigTest` to ease extension development

GromNaN:
One drawback to writing extensions at present is that the declaration of functions/filters/tests is not directly adjacent to the methods. It's worse for runtime extensions because they need to be in 2 different classes. See [`SerializerExtension`](https://github.com/symfony/symfony/blob/7.0/src/Symfony/Bridge/Twig/Extension/SerializerExtension.php) and [`SerializerRuntime`](https://github.com/symfony/symfony/blob/7.0/src/Symfony/Bridge/Twig/Extension/SerializerRuntime.php) as an example.

By using attributes for filters, functions and tests definition, we can make writing extensions more expressive, and use reflection to detect particular options (`needs_environment`, `needs_context`, `is_variadic`).

Example if we implemented the `formatDate` filter: https://github.com/twigphp/Twig/blob/aeeec9a5e907a79e50a6bb78979154599401726e/extra/intl-extra/IntlExtension.php#L392-L395

By using the `AsTwigFilter` attribute, it is not necessary to create the `getFilters()` method. The `needs_environment` option is detected from method signature. The name is still required as the method naming convention (camelCase) doesn't match with Twig naming convention (snake_case).

```php
use Twig\Extension\Attribute\AsTwigFilter;

class IntlExtension
{
    #[AsTwigFilter(name: 'format_date')]
    public function formatDate(Environment $env, $date, ?string $dateFormat = 'medium', string $pattern = '', $timezone = null, string $calendar = 'gregorian', string $locale = null): string
    {
        return $this->formatDateTime($env, $date, $dateFormat, 'none', $pattern, $timezone, $calendar, $locale);
    }
}
```

This approach does not totally replace the current definition of extensions, which is still necessary for advanced needs. It does, however, make for more pleasant reading and writing. 

This makes writing lazy-loaded runtime extension the easiest way to create Twig extension in Symfony: https://github.com/symfony/symfony/pull/52748

Related to https://github.com/symfony/symfony/issues/50016

Is there any need to cache the parsing of method attributes? They are only read at compile time, but that can have a performance impact during development or when using dynamic templates.

### DIFF
--- a/doc/advanced.rst
+++ b/doc/advanced.rst
@@ -814,6 +814,107 @@ The ``getTests()`` method lets you add new test functions::
         // ...
     }
 
+Using PHP Attributes to define Extensions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 3.21
+
+    The attribute classes were added in Twig 3.21.
+
+You can add the ``#[AsTwigFilter]``, ``#[AsTwigFunction]``, and ``#[AsTwigTest]``
+attributes to public methods of any class to define filters, functions, and tests.
+
+Create a class using these attributes::
+
+    use Twig\Attribute\AsTwigFilter;
+    use Twig\Attribute\AsTwigFunction;
+    use Twig\Attribute\AsTwigTest;
+
+    class ProjectExtension
+    {
+        #[AsTwigFilter('rot13')]
+        public static function rot13(string $string): string
+        {
+            // ...
+        }
+
+        #[AsTwigFunction('lipsum')]
+        public static function lipsum(int $count): string
+        {
+            // ...
+        }
+
+        #[AsTwigTest('even')]
+        public static function isEven(int $number): bool
+        {
+            // ...
+        }
+    }
+
+Then register the ``Twig\Extension\AttributeExtension`` with the class name::
+
+    $twig = new \Twig\Environment($loader);
+    $twig->addExtension(new \Twig\Extension\AttributeExtension(ProjectExtension::class));
+
+If all the methods are static, you are done. The ``ProjectExtension`` class will
+never be instantiated and the class attributes will be scanned only when a template
+is compiled.
+
+Otherwise, if some methods are not static, you need to register the class as
+a runtime extension using one of the runtime loaders::
+
+    use Twig\Attribute\AsTwigFunction;
+
+    class ProjectExtension
+    {
+        // Inject hypothetical dependencies
+        public function __construct(private LipsumProvider $lipsumProvider) {}
+
+        #[AsTwigFunction('lipsum')]
+        public function lipsum(int $count): string
+        {
+            return $this->lipsumProvider->lipsum($count);
+        }
+    }
+
+    $twig = new \Twig\Environment($loader);
+    $twig->addExtension(new \Twig\Extension\AttributeExtension(ProjectExtension::class);
+    $twig->addRuntimeLoader(new \Twig\RuntimeLoader\FactoryLoader([
+        ProjectExtension::class => function () use ($lipsumProvider) {
+            return new ProjectExtension($lipsumProvider);
+        },
+    ]));
+
+If you want to access the current environment instance in your filter or function,
+add the ``Twig\Environment`` type to the first argument of the method::
+
+    class ProjectExtension
+    {
+        #[AsTwigFunction('lipsum')]
+        public function lipsum(\Twig\Environment $env, int $count): string
+        {
+            // ...
+        }
+    }
+
+``#[AsTwigFilter]`` and ``#[AsTwigFunction]`` support variadic arguments
+automatically when applied to variadic methods::
+
+    class ProjectExtension
+    {
+        #[AsTwigFilter('thumbnail')]
+        public function thumbnail(string $file, mixed ...$options): string
+        {
+            // ...
+        }
+    }
+
+The attributes support other options used to configure the Twig Callables:
+
+ * ``AsTwigFilter``: ``needsCharset``, ``needsEnvironment``, ``needsContext``, ``isSafe``, ``isSafeCallback``, ``preEscape``, ``preservesSafety``, ``deprecationInfo``
+ * ``AsTwigFunction``: ``needsCharset``, ``needsEnvironment``, ``needsContext``, ``isSafe``, ``isSafeCallback``, ``deprecationInfo``
+ * ``AsTwigTest``: ``needsCharset``, ``needsEnvironment``, ``needsContext``, ``deprecationInfo``
+
 Definition vs Runtime
 ~~~~~~~~~~~~~~~~~~~~~
 
--- a/src/Attribute/AsTwigFilter.php
+++ b/src/Attribute/AsTwigFilter.php
@@ -1,0 +1,56 @@
+<?php
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Twig\Attribute;
+
+use Twig\DeprecatedCallableInfo;
+use Twig\TwigFilter;
+
+/**
+ * Registers a method as template filter.
+ *
+ * If the first argument of the method has Twig\Environment type-hint, the filter will receive the current environment.
+ * Additional arguments of the method come from the filter call.
+ *
+ *     #[AsTwigFilter(name: 'foo')]
+ *     function fooFilter(Environment $env, $string, $arg1 = null, ...) { ... }
+ *
+ *    {{ 'string'|foo(arg1) }}
+ *
+ * @see TwigFilter
+ */
+#[\Attribute(\Attribute::TARGET_METHOD | \Attribute::IS_REPEATABLE)]
+final class AsTwigFilter
+{
+    /**
+     * @param non-empty-string $name The name of the filter in Twig.
+     * @param bool|null $needsCharset Whether the filter needs the charset passed as the first argument.
+     * @param bool|null $needsEnvironment Whether the filter needs the environment passed as the first argument, or after the charset.
+     * @param bool|null $needsContext Whether the filter needs the context array passed as the first argument, or after the charset and the environment.
+     * @param string[]|null $isSafe List of formats in which you want the raw output to be printed unescaped.
+     * @param string|array|null $isSafeCallback Function called at compilation time to determine if the filter is safe.
+     * @param string|null $preEscape Some filters may need to work on input that is already escaped or safe
+     * @param string[]|null $preservesSafety Preserves the safety of the value that the filter is applied to.
+     * @param DeprecatedCallableInfo|null $deprecationInfo Information about the deprecation
+     */
+    public function __construct(
+        public string $name,
+        public ?bool $needsCharset = null,
+        public ?bool $needsEnvironment = null,
+        public ?bool $needsContext = null,
+        public ?array $isSafe = null,
+        public string|array|null $isSafeCallback = null,
+        public ?string $preEscape = null,
+        public ?array $preservesSafety = null,
+        public ?DeprecatedCallableInfo $deprecationInfo = null,
+    ) {
+    }
+}
--- a/src/Attribute/AsTwigTest.php
+++ b/src/Attribute/AsTwigTest.php
@@ -1,0 +1,48 @@
+<?php
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Twig\Attribute;
+
+use Twig\DeprecatedCallableInfo;
+use Twig\TwigTest;
+
+/**
+ * Registers a method as template test.
+ *
+ * The first argument is the value to test and the other arguments are the
+ * arguments passed to the test in the template.
+ *
+ *     #[AsTwigTest(name: 'foo')]
+ *     public function fooTest($value, $arg1 = null) { ... }
+ *
+ *     {% if value is foo(arg1) %}
+ *
+ * @see TwigTest
+ */
+#[\Attribute(\Attribute::TARGET_METHOD | \Attribute::IS_REPEATABLE)]
+final class AsTwigTest
+{
+    /**
+     * @param non-empty-string $name The name of the test in Twig.
+     * @param bool|null $needsCharset Whether the test needs the charset passed as the first argument.
+     * @param bool|null $needsEnvironment Whether the test needs the environment passed as the first argument, or after the charset.
+     * @param bool|null $needsContext Whether the test needs the context array passed as the first argument, or after the charset and the environment.
+     * @param DeprecatedCallableInfo|null $deprecationInfo Information about the deprecation
+     */
+    public function __construct(
+        public string $name,
+        public ?bool $needsCharset = null,
+        public ?bool $needsEnvironment = null,
+        public ?bool $needsContext = null,
+        public ?DeprecatedCallableInfo $deprecationInfo = null,
+    ) {
+    }
+}
--- a/src/Extension/AttributeExtension.php
+++ b/src/Extension/AttributeExtension.php
@@ -1,0 +1,174 @@
+<?php
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Twig\Extension;
+
+use Twig\Attribute\AsTwigFilter;
+use Twig\Attribute\AsTwigFunction;
+use Twig\Attribute\AsTwigTest;
+use Twig\Environment;
+use Twig\TwigFilter;
+use Twig\TwigFunction;
+use Twig\TwigTest;
+
+/**
+ * Define Twig filters, functions, and tests with PHP attributes.
+ *
+ * @author Jérôme Tamarelle <jerome@tamarelle.net>
+ */
+final class AttributeExtension extends AbstractExtension
+{
+    private array $filters;
+    private array $functions;
+    private array $tests;
+
+    /**
+     * Use a runtime class using PHP attributes to define filters, functions, and tests.
+     *
+     * @param class-string $class
+     */
+    public function __construct(private string $class)
+    {
+    }
+
+    /**
+     * @return class-string
+     */
+    public function getClass(): string
+    {
+        return $this->class;
+    }
+
+    public function getFilters(): array
+    {
+        if (!isset($this->filters)) {
+            $this->initFromAttributes();
+        }
+
+        return $this->filters;
+    }
+
+    public function getFunctions(): array
+    {
+        if (!isset($this->functions)) {
+            $this->initFromAttributes();
+        }
+
+        return $this->functions;
+    }
+
+    public function getTests(): array
+    {
+        if (!isset($this->tests)) {
+            $this->initFromAttributes();
+        }
+
+        return $this->tests;
+    }
+
+    public function getLastModified(): int
+    {
+        return max(
+            filemtime(__FILE__),
+            is_file($filename = (new \ReflectionClass($this->getClass()))->getFileName()) ? filemtime($filename) : 0,
+        );
+    }
+
+    private function initFromAttributes(): void
+    {
+        $filters = $functions = $tests = [];
+        $reflectionClass = new \ReflectionClass($this->getClass());
+        foreach ($reflectionClass->getMethods() as $method) {
+            foreach ($method->getAttributes(AsTwigFilter::class) as $reflectionAttribute) {
+                /** @var AsTwigFilter $attribute */
+                $attribute = $reflectionAttribute->newInstance();
+
+                $callable = new TwigFilter($attribute->name, [$reflectionClass->name, $method->getName()], [
+                    'needs_context' => $attribute->needsContext ?? false,
+                    'needs_environment' => $attribute->needsEnvironment ?? $this->needsEnvironment($method),
+                    'needs_charset' => $attribute->needsCharset ?? false,
+                    'is_variadic' => $method->isVariadic(),
+                    'is_safe' => $attribute->isSafe,
+                    'is_safe_callback' => $attribute->isSafeCallback,
+                    'pre_escape' => $attribute->preEscape,
+                    'preserves_safety' => $attribute->preservesSafety,
+                    'deprecation_info' => $attribute->deprecationInfo,
+                ]);
+
+                if ($callable->getMinimalNumberOfRequiredArguments() > $method->getNumberOfParameters()) {
+                    throw new \LogicException(sprintf('"%s::%s()" needs at least %d arguments to be used AsTwigFilter, but only %d defined.', $reflectionClass->getName(), $method->getName(), $callable->getMinimalNumberOfRequiredArguments(), $method->getNumberOfParameters()));
+                }
+
+                $filters[$attribute->name] = $callable;
+            }
+
+            foreach ($method->getAttributes(AsTwigFunction::class) as $reflectionAttribute) {
+                /** @var AsTwigFunction $attribute */
+                $attribute = $reflectionAttribute->newInstance();
+
+                $callable = new TwigFunction($attribute->name, [$reflectionClass->name, $method->getName()], [
+                    'needs_context' => $attribute->needsContext ?? false,
+                    'needs_environment' => $attribute->needsEnvironment ?? $this->needsEnvironment($method),
+                    'needs_charset' => $attribute->needsCharset ?? false,
+                    'is_variadic' => $method->isVariadic(),
+                    'is_safe' => $attribute->isSafe,
+                    'is_safe_callback' => $attribute->isSafeCallback,
+                    'deprecation_info' => $attribute->deprecationInfo,
+                ]);
+
+                if ($callable->getMinimalNumberOfRequiredArguments() > $method->getNumberOfParameters()) {
+                    throw new \LogicException(sprintf('"%s::%s()" needs at least %d arguments to be used AsTwigFunction, but only %d defined.', $reflectionClass->getName(), $method->getName(), $callable->getMinimalNumberOfRequiredArguments(), $method->getNumberOfParameters()));
+                }
+
+                $functions[$attribute->name] = $callable;
+            }
+
+            foreach ($method->getAttributes(AsTwigTest::class) as $reflectionAttribute) {
+
+                /** @var AsTwigTest $attribute */
+                $attribute = $reflectionAttribute->newInstance();
+
+                $callable = new TwigTest($attribute->name, [$reflectionClass->name, $method->getName()], [
+                    'needs_context' => $attribute->needsContext ?? false,
+                    'needs_environment' => $attribute->needsEnvironment ?? $this->needsEnvironment($method),
+                    'needs_charset' => $attribute->needsCharset ?? false,
+                    'is_variadic' => $method->isVariadic(),
+                    'deprecation_info' => $attribute->deprecationInfo,
+                ]);
+
+                if ($callable->getMinimalNumberOfRequiredArguments() > $method->getNumberOfParameters()) {
+                    throw new \LogicException(sprintf('"%s::%s()" needs at least %d arguments to be used AsTwigTest, but only %d defined.', $reflectionClass->getName(), $method->getName(), $callable->getMinimalNumberOfRequiredArguments(), $method->getNumberOfParameters()));
+                }
+
+                $tests[$attribute->name] = $callable;
+            }
+        }
+
+        // Assign all at the end to avoid inconsistent state in case of exception
+        $this->filters = array_values($filters);
+        $this->functions = array_values($functions);
+        $this->tests = array_values($tests);
+    }
+
+    /**
+     * Detect if the first argument of the method is the environment.
+     */
+    private function needsEnvironment(\ReflectionFunctionAbstract $function): bool
+    {
+        if (!$parameters = $function->getParameters()) {
+            return false;
+        }
+
+        return $parameters[0]->getType() instanceof \ReflectionNamedType
+            && Environment::class === $parameters[0]->getType()->getName()
+            && !$parameters[0]->isVariadic();
+    }
+}
--- a/src/ExtensionSet.php
+++ b/src/ExtensionSet.php
@@ -18,6 +18,7 @@ use Twig\ExpressionParser\InfixAssociativity;
 use Twig\ExpressionParser\InfixExpressionParserInterface;
 use Twig\ExpressionParser\PrecedenceChange;
 use Twig\ExpressionParser\Prefix\UnaryOperatorExpressionParser;
+use Twig\Extension\AttributeExtension;
 use Twig\Extension\ExtensionInterface;
 use Twig\Extension\GlobalsInterface;
 use Twig\Extension\LastModifiedExtensionInterface;
@@ -142,7 +143,11 @@ final class ExtensionSet
 
     public function addExtension(ExtensionInterface $extension): void
     {
-        $class = $extension::class;
+        if ($extension instanceof AttributeExtension) {
+            $class = $extension->getClass();
+        } else {
+            $class = $extension::class;
+        }
 
         if ($this->initialized) {
             throw new \LogicException(\sprintf('Unable to register extension "%s" as extensions have already been initialized.', $class));
--- a/tests/Extension/Fixtures/ExtensionWithAttributes.php
+++ b/tests/Extension/Fixtures/ExtensionWithAttributes.php
@@ -1,0 +1,112 @@
+<?php
+
+namespace Twig\Tests\Extension\Fixtures;
+
+use Twig\Attribute\AsTwigFilter;
+use Twig\Attribute\AsTwigFunction;
+use Twig\Attribute\AsTwigTest;
+use Twig\DeprecatedCallableInfo;
+use Twig\Environment;
+
+class ExtensionWithAttributes
+{
+    #[AsTwigFilter(name: 'foo', isSafe: ['html'])]
+    public function fooFilter(string|int $string)
+    {
+    }
+
+    #[AsTwigFilter('with_context_filter', needsContext: true)]
+    public function withContextFilter(array $context, string $string)
+    {
+    }
+
+    #[AsTwigFilter('with_env_filter')]
+    public function withEnvFilter(Environment $env, string $string)
+    {
+    }
+
+    #[AsTwigFilter('with_env_and_context_filter', needsContext: true)]
+    public function withEnvAndContextFilter(Environment $env, array $context, array $data)
+    {
+    }
+
+    #[AsTwigFilter('variadic_filter')]
+    public function variadicFilter(string ...$strings)
+    {
+    }
+
+    #[AsTwigFilter('deprecated_filter', deprecationInfo: new DeprecatedCallableInfo('foo/bar', '1.2'))]
+    public function deprecatedFilter(string $string)
+    {
+    }
+
+    #[AsTwigFilter('pattern_*_filter')]
+    public function patternFilter(string $string)
+    {
+    }
+
+    #[AsTwigFunction(name: 'foo', isSafe: ['html'])]
+    public function fooFunction(string|int $string)
+    {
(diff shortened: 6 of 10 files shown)


src/Extension/Extension.php
```diff
@@ -0,0 +1,100 @@
...
+use Twig\TwigFilter;
+use Twig\TwigFunction;
+use Twig\TwigTest;
+
+/**
+ * Abstract class for extension using the new PHP 8 attributes to define filters, functions, and tests.
+ *
+ * @author Jérôme Tamarelle <jerome@tamarelle.net>
+ */
+abstract class Extension extends AbstractExtension
```
GromNaN: Is it better to modify the existing `AbstractExtension`, create a new abstract class or a new trait? The trait would have to be used with the `AbstractExtension` or `ExtensionInterface`.

src/Extension/Extension.php
```diff
@@ -0,0 +1,100 @@
...
+                    $param = $method->getParameters()[$firstParam] ?? null;
+                    $options['needs_context'] = $param && 'context' === $param->getName() && 'array' === $param->getType()->getName();
+                }
+                $firstParam += $options['needs_context'] ? 1 : 0;
+                if (!\array_key_exists('is_variadic', $options)) {
+                    $param = $method->getParameters()[$firstParam] ?? null;
+                    $options['is_variadic'] = $param && $param->isVariadic();
+                }
+
+                yield new TwigFilter($attribute->name ?? $method->getName(), [$this, $method->getName()], $options);
```
GromNaN: Should the default name be converted from camelCase to snake_case, to match the Twig convention? This makes it more difficult to find the declaration of a twig function/filter/test in PHP source code. But it does ensure that developers follow the naming convention by default.

src/Extension/Attribute/AsTwigFilter.php
```diff
@@ -0,0 +1,29 @@
...
+    public function __construct(
+        /**
+         * The name of the filter in Twig (defaults to the method name).
+         *
+         * @var non-empty-string|null $name
+         */
+        public ?string $name = null,
+
+        /**
+         * @var array{needs_environment?:bool, needs_context?:bool, is_variadic?:bool, is_safe?:array|null, is_safe_callback?:callable|null, pre_escape?:string|null, preserves_safety?:array|null, node_class?:class-string, deprecated?:bool|string, alternative?:string}
```
GromNaN: Since we can detect automatically if the function is variadic, needs the environment of the context, this option could be undocumented.
I'm not sure for the `node_class` either.

GromNaN: We could flatten the options in the class properties, which would undoubtedly be easier to read and benefit from native typing.

src/Extension/Attribute/AsTwigTest.php
```diff
@@ -0,0 +1,34 @@
...
+
+        /**
+         * @var array{is_variadic?:bool, node_class?:class-string, deprecated?:bool|string, alternative?:string, one_mandatory_argument?:bool}
+         */
+        public array $options = [],
+
+        /**
+         * @var array<int, mixed>
+         */
+        public array $arguments = [],
```
GromNaN: I'm not sure what `$arguments` is for.

src/Extension/Attribute/AsTwigFilter.php
```diff
@@ -0,0 +1,29 @@
...
+         * The name of the filter in Twig (defaults to the method name).
+         *
+         * @var non-empty-string|null $name
+         */
+        public ?string $name = null,
+
+        /**
+         * @var array{is_safe?:array|null, is_safe_callback?:callable|null, pre_escape?:string|null, preserves_safety?:array|null, deprecated?:bool|string, alternative?:string}
+         */
+        public array $options = [],
```
derrabus: All those options should be dedicated properties on the attribute.

src/Extension/Extension.php
derrabus: From a DX perspective: If I used the new attributes, why would I still want to extend `AbstractExtension` directly or though this class? If we introduce those attributes, I believe we should make them work on standalone classes.

GromNaN: Right. I can create a wrapper class that implements the ExtensionInterface and scan one or many object methods, passed to the constructor, to find filters/functions/tests. That would be much better. 

We need a marker to hint Symfony DI that a class contains Twig extension attributes, to avoid scanning attributes of all methods of all classes in the project.

GromNaN:
I'll rework the implementation after reading discussions on symfony/symfony#50016

src/Extension/Extension.php
GromNaN: The marker will be `RuntimeExtensionInterface`. This interface will soon find its purpose in Twig.

src/Extension/Extension.php
stof: Given that Twig function names must be globally unique while method names are unique only to the class, I'm not even sure we want this automatic name. I'd rather make the name mandatory in the attribute.

Thus, this name is a public API of the Twig extension, so it means that any guessing logic we introduce is then set in stone as any change to it as a BC break.

src/Extension/AttributeExtension.php
```diff
@@ -0,0 +1,179 @@
...
+
+                    $name = $attribute->name ?? $method->getName();
+                    if (isset($filters[$name])) {
+                        throw new \LogicException(sprintf('Multiple definitions of the "%s" filter.', $name));
+                    }
+
+                    $parameters = $method->getParameters();
+                    $needsEnvironment = isset($parameters[0]) && Environment::class === $parameters[0]->getType()?->getName();
+                    $firstParam = $needsEnvironment ? 1 : 0;
+                    $needsContext = isset($parameters[$firstParam]) && 'context' === $parameters[$firstParam]->getName() && 'array' === $parameters[$firstParam]->getType()?->getName();
```
stof: this special behavior of parameter named `$context` needs to be documented on the attributes.

src/Extension/Extension.php
derrabus: If we introduce all new attributes in this PR, can't the marker be, you know, an attribute? 🥺

src/Extension/Extension.php
GromNaN: I updated to make the function name required. A default value can always be set later.

src/Extension/AttributeExtension.php
GromNaN: I added the doc and example in the attribute class header.

src/Extension/WithLastModified.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace Twig\Extension;
+
+interface WithLastModified
```
GromNaN: This is also a new feature required to detect when the class that defines attributes have been updated.

src/Extension/Extension.php
GromNaN: I created the `AsTwigExtension` attribute as marker.

src/Extension/Attribute/AsTwigExtension.php
```diff
@@ -0,0 +1,20 @@
...
+/*
+ * This file is part of Twig.
+ *
+ * (c) Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Twig\Extension\Attribute;
```
GromNaN: Should it move to `Twig\Attribute` namespace? I see other libraries like Monolog and Symfony HttpKernel have the `Attribute` namespace at the root of their namespace.

src/Extension/AttributeExtension.php
GromNaN: The [Symfony `serialize` filter](https://github.com/symfony/symfony/blob/2d88c405ba0de17aebcf5df3de79b12f3fdb297a/src/Symfony/Bridge/Twig/Extension/SerializerRuntime.php#L29) accepts an `array $context` as 3rd argument. This will not conflict but this means developers could have an unexpected behavior if they don't know about this special argument handling.

src/Extension/WithLastModified.php
GromNaN: Alternatively, we can make passing a class name to the ExtensionSet, that will be scanned by it. This has slightly more impact on existing classes.

src/Extension/AttributeExtension.php
```diff
@@ -0,0 +1,179 @@
...
+            }
+
+            foreach ($reflectionClass->getMethods() as $method) {
+                // Filters
+                foreach ($method->getAttributes(AsTwigFilter::class) as $attribute) {
+                    $attribute = $attribute->newInstance();
+
+                    $name = $attribute->name;
+                    if (isset($filters[$name])) {
+                        throw new \LogicException(sprintf('Multiple definitions of the "%s" filter.', $name));
```
smnandre: Could we have a unique centralized system for that ? Because not all functions/filters will come from here i guess ?

src/Extension/AttributeExtension.php
```diff
@@ -0,0 +1,179 @@
...
+                    if (isset($tests[$name])) {
+                        throw new \LogicException(sprintf('Multiple definitions of the "%s" test.', $name));
+                    }
+
+                    $parameters = $method->getParameters();
+                    $isVariadic = isset($parameters[$firstParam]) && end($parameters)->isVariadic();
+
+                    $tests[$name] = new TwigTest($name, [$objectOrClass, $method->getName()], [
+                        'is_variadic' => $isVariadic,
+                        'deprecated' => $attribute->deprecated,
```
smnandre: \+ options `node_class` and `one_mandatory_argument` ?

 (if TwigTest keeps it [current signature](https://github.com/twigphp/Twig/blob/aeeec9a5e907a79e50a6bb78979154599401726e/src/TwigTest.php#L38))

src/Extension/Attribute/AsTwigExtension.php
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ * Identifies a class that uses PHP attributes to define filters, functions, or tests.
+ */
+#[\Attribute(\Attribute::TARGET_CLASS)]
+class AsTwigExtension
```
smnandre: What is the purpose of this attribute ? I don't see (yet) how it is / will be used

src/Extension/AttributeExtension.php
GromNaN: - `node_class` option is exclusive with the callback. It cannot used for an attribute of a method. It we extends the attributes to the class, that would be the class itself.
- I must check how arguments works with tests.

src/Extension/Attribute/AsTwigExtension.php
GromNaN: It's similar to `Twig\Extension\RuntimeExtensionInterface`, it's a marker for classes that provide extension features, to be automatically tagged by Symfony dependency injection.

src/Extension/AttributeExtension.php
smnandre: I did not realize they were 100% exclusive, thanks for the explanation !

tests/Extension/Fixtures/ExtensionWithAttributes.php
```diff
@@ -0,0 +1,98 @@
...
+    #[AsTwigTest(name: 'foo')]
+    public function fooTest(string $string)
+    {
+    }
+
+    #[AsTwigTest('variadic_test')]
+    public function variadicTest(string ...$strings)
+    {
+    }
+
```
smnandre, line 92: I don't see how this class would change anything (as it's [internal to ExtensionSet](https://github.com/twigphp/Twig/blob/aeeec9a5e907a79e50a6bb78979154599401726e/src/ExtensionSet.php#L169))... but maybe just to be sure could we add a "pattern" Filter/Function/Test  ?

`pattern_*_filter`

src/Extension/Attribute/AsTwigExtension.php
GromNaN: Moved.

src/Extension/AttributeExtension.php
GromNaN: This is a feature of Twig. You can replace an existing filter/function/test by declaring the same name in an extension.
I'm thinking about removing this check.

tests/Extension/Fixtures/ExtensionWithAttributes.php
GromNaN: Test added.

src/Extension/WithLastModified.php
GromNaN: @stof I can't replace the modification date by a hash, the `Environment::isFresh` relies on a date comparison between the ExtensionSet and the cache file mtime.

https://github.com/twigphp/Twig/blob/aeeec9a5e907a79e50a6bb78979154599401726e/src/Environment.php#L416-L419

doc/advanced.rst
```diff
@@ -756,6 +756,156 @@ The ``getTests()`` method lets you add new test functions::
...
+Then register the class using ``Twig\Extension\AttributeExtension``::
+
+    $twig = new \Twig\Environment($loader);
+    $twig->addExtension(new \Twig\Extension\AttributeExtension([
+        Project_Twig_Extension::class,
+    ]));
+
+.. note::
+
+    The ``\Twig\Extension\AttributeExtension`` can be added only once to an environment.
```
GromNaN: There is an exception if we try to register the extension a second time.

GromNaN (commented):
Ready for review @fabpot.

smnandre (approved):

src/Extension/Attribute/AsTwigExtension.php
stof: Given that extensions must implement the ExtensionInterface anyway, there is no need for a marker attribute. The Symfony dependency injection component is also able to perform auto-configuration based on implemented interfaces (and TwigBundle already does it for the Twig ExtensionInterface

GromNaN: It's more a question of DX. It's more meaningful to use an attribute than an interface as a marker as the developer will use attributes in the class. Implementing an interface without method is a deviation from the interface system.

src/Extension/WithLastModified.php
nicolas-grekas: `LastModifiedAwareInterface`? What's the existing convention in Twig?

GromNaN: Renamed to `ModificationAwareInterface`. The convention if to suffix with `Interface`.

GromNaN:
In all honesty, I'm not totally convinced by this implementation. The fact that extensions have to be registered in 2 places (as a runtime extension and for definition in AttributeExtension) makes them difficult to use with Twig standalone, this is totally hidden for Symfony users with the TwigBundle.

I would like to add a new `Environment::add(object $extension)` method that would accept any class with attributes. Invocable classes could be used to define filters and functions.

nicolas-grekas:
What is a bit strange is having one single instance of AttributeExtension be responsible for registering many runtimes.
Maybe we should instead have one AttributeExtension per runtime?
Then it might be easier to use this standalone : `addExtension(new AttributeExtension(new AttributeBasedRuntime)))` (and such runtimes could have a static factory to make this even easier to create).
Note that this suggestion might be wrong as I didn't think of how runtimes should be made lazy-instantiated.

kbond:
For reference, this is my attempt/experiment to solve the problem of apps having to create a bunch of twig extensions for things: https://github.com/zenstruck/twig-service-bundle

src/Environment.php
```diff
@@ -616,14 +616,17 @@ public function getRuntime(string $class)
         throw new RuntimeError(sprintf('Unable to load the "%s" runtime.', $class));
     }
 
-    public function addExtension(ExtensionInterface $extension)
+    /**
+     * @param ExtensionInterface|object|class-string $extension
+     */
+    public function addExtension(object|string $extension)
```
GromNaN: This is a breaking change.

GromNaN (commented):
> What is a bit strange is having one single instance of AttributeExtension be responsible for registering many runtimes. Maybe we should instead have one AttributeExtension per runtime? Then it might be easier to use this standalone : `addExtension(new AttributeExtension(new AttributeBasedRuntime)))` (and such runtimes could have a static factory to make this even easier to create). Note that this suggestion might be wrong as I didn't think of how runtimes should be made lazy-instantiated.

In the `ExtensionSet`, extensions are identified by their class name. The same extension class cannot be registered multiple times with distinct configuration.

I updated this PR so that extension can be registered directly using any class name or object that have the `#[AsTwigExtension]` attribute (the attribute becomes required).

```diff
    $twig = new \Twig\Environment($loader);
-    $twig->addExtension(new \Twig\Extension\AttributeExtension([
-        ProjectExtension::class,
-    ]));
+    $twig->addExtension(ProjectExtension::class);
    $twig->addRuntimeLoader(new \Twig\RuntimeLoader\FactoryLoader([
        ProjectExtension::class => function () use ($lipsumProvider) {
            return new ProjectExtension($lipsumProvider);
        },
    ]));
```

To use standalone (non-lazy-loaded):

```php
$twig->addExtension(new ProjectExtension($lipsumProvider));
```

src/Extension/AttributeExtension.php
```diff
@@ -0,0 +1,162 @@
...
+                        'alternative' => $attribute->alternative,
+                    ]);
+                }
+            }
+        }
+
+        // Assign all at the end to avoid inconsistent state in case of exception
+        $this->filters = $filters;
```
stof: you should do `array_values` on those arrays instead of returning names as keys

src/Extension/Attribute/AsTwigExtension.php
GromNaN: I've made this attribute mandatory to prevent any object from being passed by mistake. In practice, it will be mandatory anyway for Symfony autoconfiguration.

src/Attribute/AsTwigTest.php
```diff
@@ -0,0 +1,50 @@
...
+namespace Twig\Attribute;
+
+use Twig\TwigTest;
+
+/**
+ * Registers a method as template test.
+ *
+ * If the first argument of the method has Twig\Environment type-hint, the test will receive the current environment.
+ * If the next argument of the method is named $context and has array type-hint, the test will receive the current context.
+ * The last argument of the method is the value to be tested, if any.
```
stof: this is wrong. It is not the last argument. It is the following argument. A test can have as many arguments as you want after the value being tested, which become arguments in the Twig signature.

src/Extension/WithLastModified.php
GromNaN: With the last change, I removed this interface. `ExtensionSet` is now aware of the extension classes that use attributes, and check their modification date as well.

src/Attribute/AsTwigFilter.php
```diff
@@ -0,0 +1,78 @@
...
+         */
+        public ?string $preEscape = null,
+
+        /**
+         * Preserves the safety of the value that the filter is applied to.
+         */
+        public ?array $preservesSafety = null,
+
+        /**
+         * Set to true if the filter is deprecated.
```
stof: what is the string type about then ?

src/Attribute/AsTwigFilter.php
```diff
@@ -0,0 +1,78 @@
...
+        /**
+         * Some filters may need to work on input that is already escaped or safe, for
+         * example when adding (safe) HTML tags to originally unsafe output. In such a
+         * case, set preEscape to an escape format to escape the input data before it
+         * is run through the filter.
+         */
+        public ?string $preEscape = null,
+
+        /**
+         * Preserves the safety of the value that the filter is applied to.
```
stof: this misses the doc for the type of array values

src/Extension/AttributeExtension.php
```diff
@@ -0,0 +1,162 @@
...
+            } catch (\ReflectionException $e) {
+                throw new \LogicException(sprintf('"%s" class requires a list of objects or class name, "%s" given.', __CLASS__, get_debug_type($objectOrClass)), 0, $e);
+            }
+
+            $attributes = $reflectionClass->getAttributes(AsTwigExtension::class);
+            if (!$attributes) {
+                throw new \LogicException(sprintf('Extension class "%s" must have the attribute "%s" in order to use attributes', is_string($objectOrClass) ? $objectOrClass : get_debug_type($objectOrClass), AsTwigExtension::class));
+            }
+
+            foreach ($reflectionClass->getMethods() as $method) {
```
stof: thinking about that, I'm wondering whether those `asTwigFilter`, `AsTwigFunction` and `AsTwigTest` attributes should be read by that extension at all.

An alternative solution would be to have an extension in which you inject this list of test, filter and function metadata, and letting TwigBundle perform all this attribute reading at build-time instead. 
This would even removed the need for the `AsTwigExtension` attribute entirely (which would be quite confusing as the class having this attribute is _not_ a twig extension but a twig runtime) because the autoconfiguration system is able to apply autoconfiguration based on method-level attributes already.

src/ExtensionSet.php
```diff
@@ -424,9 +433,16 @@ private function initExtensions(): void
         $this->unaryOperators = [];
         $this->binaryOperators = [];
 
+        $classes = [];
         foreach ($this->extensions as $extension) {
-            $this->initExtension($extension);
+            if ($extension instanceof ExtensionInterface) {
+                $this->initExtension($extension);
+            } else {
+                $classes[] = $extension;
```
stof: making `$this->extensions` contain a mix of extensions and of runtimes (but not all of them) looks weird to me

stof: if this is a way to solve the case of `getLastModified`, I'd rather add a way to make ExtensionSet aware of a **separate** list of runtimes needing to be checked (which would allow solving the case where we don't invalidate the cache today if you change the argument names in a runtime while it actually impacts the template compilation if you use named parameters in Twig, by making TwigBundle inject the name of _all_ runtimes in the list).

doc/advanced.rst
```diff
@@ -813,6 +813,145 @@ The ``getTests()`` method lets you add new test functions::
         // ...
     }
 
+Using PHP Attributes to define extensions
```
fabpot:
```suggestion
Using PHP Attributes to define Extensions
```

doc/advanced.rst
```diff
@@ -813,6 +813,145 @@ The ``getTests()`` method lets you add new test functions::
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 3.9
```
fabpot:
```suggestion
.. versionadded:: 3.19
```

doc/advanced.rst
```diff
@@ -813,6 +813,145 @@ The ``getTests()`` method lets you add new test functions::
...
+
+    The attribute classes were added in Twig 3.9.
```
fabpot:
```suggestion
    The attribute classes were added in Twig 3.19.
```

doc/advanced.rst
```diff
@@ -813,6 +813,145 @@ The ``getTests()`` method lets you add new test functions::
...
+
+From PHP 8.0, you can use the attributes ``#[AsTwigFilter]``, ``#[AsTwigFunction]``,
```
fabpot: As we now require PHP 8, the version mention can be removed:

```suggestion
You can use the attributes ``#[AsTwigFilter]``, ``#[AsTwigFunction]``,
```

doc/advanced.rst
```diff
@@ -813,6 +813,145 @@ The ``getTests()`` method lets you add new test functions::
...
+            // ...
+        }
+    }
+
+Then register the extension class::
+
+    $twig = new \Twig\Environment($loader);
+    $twig->addExtension(ProjectExtension::class);
+
+If all the methods are static, you are done. The ``Project_Twig_Extension`` class will
```
fabpot:
```suggestion
If all the methods are static, you are done. The ``ProjectExtension`` class will
```

src/ExtensionSet.php
```diff
@@ -424,6 +434,7 @@ private function initExtensions(): void
         foreach ($this->extensions as $extension) {
             $this->initExtension($extension);
         }
+
```
fabpot: Can be reverted

tests/Extension/Fixtures/FilterWithoutValue.php
```diff
@@ -0,0 +1,13 @@
...
+
+use Twig\Attribute\AsTwigFilter;
+
+class FilterWithoutValue
+{
+    #[AsTwigFilter('my_filter')]
+    public function myFilter()
+    {
+    }
+}
```
fabpot: missing new line

tests/Extension/Fixtures/TestWithoutValue.php
```diff
@@ -0,0 +1,13 @@
...
+
+use Twig\Attribute\AsTwigTest;
+
+class TestWithoutValue
+{
+    #[AsTwigTest('my_test')]
+    public function myTest()
+    {
+    }
+}
```
fabpot: missing new line

fabpot (changes requested):
Not an extensive review, just things I've spotted while reading the code quickly.

src/Extension/AttributeExtension.php
GromNaN: I think that the logic for reading attributes should be in Twig,and the extension could receive the configuration already extracted. And Symfony TwigBundle will do the job of serializing and caching the configuration.

The `serializing` may be problematic in the future if closures are used in attributes for `safeCallback` ([RFC for PHP 8.5](https://wiki.php.net/rfc/closures_in_const_expr))

src/Extension/AttributeExtension.php
GromNaN: Removed this fragile detection.

src/Extension/Attribute/AsTwigExtension.php
GromNaN: Symfony autoconfiguration can work on method attributes, so I removed the `AsTwigExtension` attribute.

GromNaN:
> What is a bit strange is having one single instance of AttributeExtension be responsible for registering many runtimes. Maybe we should instead have one AttributeExtension per runtime? Then it might be easier to use this standalone : `addExtension(new AttributeExtension(new AttributeBasedRuntime)))` (and such runtimes could have a static factory to make this even easier to create). Note that this suggestion might be wrong as I didn't think of how runtimes should be made lazy-instantiated.

I refactored once again to get 1 instance of `AttributeExtension` for each class. This allows to list the encapsulated class in the `ExtensionSet::getSignature` to invalidate the cache when an extension class is removed.

Also, I removed the ability to pass an instance of an object to `AttributeExtension`, this would require proxying calls to `ExtensionSet::getExtension('MyClass')->functionCall()` to the wrapped object. This can be implemented in a follow-up PR.

The doc is up-to-date on how to use it standalone. The Symfony implementation need to be updated.

src/Extension/AttributeExtension.php
GromNaN: > thinking about that, I'm wondering whether those `asTwigFilter`, `AsTwigFunction` and `AsTwigTest` attributes should be read by that extension at all.

Here is an other approach where attribute class can create the twig callable themself.
https://github.com/twigphp/Twig/blob/36a0732ee66bf405c2164282626fbdc8c23ae81a/src/Attribute/AsTwigFilter.php

nicolas-grekas:
git rebase FTW ;)

src/ExtensionSet.php
```diff
@@ -142,7 +143,11 @@ public function getLastModified(): int
 
     public function addExtension(ExtensionInterface $extension): void
     {
-        $class = $extension::class;
+        if ($extension instanceof AttributeExtension) {
+            $class = $extension->getClass();
+        } else {
+            $class = $extension::class;
+        }
```
GromNaN, line 150: This is for cache invalidation when an extension is added/removed.

Instead of relying on this class check, I can introduce an interface: `DelegatingExtensionInterface` with the `getClass` method.

GromNaN (commented):
The state of my reflection is that we would want a different implementation in Symfony TwigBundle that reads the attributes during container compilation and inject a pre-compiled instance of `AttributeExtension`.

This can be done in 2 ways:
- by serializing the `AttributeExtension` (I'm not sure for the performances),
- or by exposing an API to convert attributes to TwigCallable. Maybe creating a public method on the attribute classes `AsTwigFilter::getTwigCallable(\ReflectionFunctionAbstract $reflection): TwigFilter`

doc/advanced.rst
```diff
@@ -814,6 +814,107 @@ The ``getTests()`` method lets you add new test functions::
...
+        public static function isEven(int $number): bool
+        {
+            // ...
+        }
+    }
+
+Then register the ``Twig\Extension\AttributeExtension`` with the class name::
+
+    $twig = new \Twig\Environment($loader);
+    $twig->addExtension(new \Twig\Extension\AttributeExtension(ProjectExtension::class);
```
fabpot:
```suggestion
    $twig->addExtension(new \Twig\Extension\AttributeExtension(ProjectExtension::class));
```

src/Attribute/AsTwigFilter.php
```diff
@@ -0,0 +1,57 @@
...
+     * @param bool|null $needsEnvironment Whether the filter needs the environment passed as the first argument, or after the charset.
+     * @param bool|null $needsContext Whether the filter needs the context array passed as the first argument, or after the charset and the environment.
+     * @param string[]|null $isSafe List of formats in which you want the raw output to be printed unescaped.
+     * @param string|array|null $isSafeCallback Function called at compilation time to determine if the filter is safe.
+     * @param string|null $preEscape Some filters may need to work on input that is already escaped or safe
+     * @param string[]|null $preservesSafety Preserves the safety of the value that the filter is applied to.
+     * @param DeprecatedCallableInfo|null $deprecationInfo Information about the deprecation
+     */
+    public function __construct(
+        public string $name,
```
fabpot, line 45: Maybe we can make it optional and use the method name if no name is provided?

GromNaN: There is a conflict in name convention as method names are [camelCase](https://symfony.com/doc/current/contributing/code/standards.html#naming-conventions) and Twig function names are [snake_case](https://twig.symfony.com/doc/3.x/coding_standards.html).

This was discussed with @stof in https://github.com/twigphp/Twig/pull/3916#discussion_r1405929739.

But I agree that this is duplicate information.

fabpot: Those are just conventions ;) And for single word names, that works well.

GromNaN: You're right, I made the name optional again.

doc/advanced.rst
```diff
@@ -814,6 +814,119 @@ The ``getTests()`` method lets you add new test functions::
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 3.21
+
+    The attribute classes were added in Twig 3.21.
+
+You can use the attributes ``#[AsTwigFilter]``, ``#[AsTwigFunction]``, and
+``#[AsTwigTest]`` on any method of any class to define filters, functions, and tests.
+
+Create a class using this attributes::
```
kbond:
```suggestion
Create a class using these attributes::
```

doc/advanced.rst
```diff
@@ -814,6 +814,119 @@ The ``getTests()`` method lets you add new test functions::
...
 
+Using PHP Attributes to define Extensions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. versionadded:: 3.21
+
+    The attribute classes were added in Twig 3.21.
+
+You can use the attributes ``#[AsTwigFilter]``, ``#[AsTwigFunction]``, and
+``#[AsTwigTest]`` on any method of any class to define filters, functions, and tests.
```
kbond:
```suggestion
You can add the ``#[AsTwigFilter]``, ``#[AsTwigFunction]``, and
``#[AsTwigTest]`` attributes to public methods of any class to define filters, functions, and tests.
```

doc/advanced.rst
```diff
@@ -814,6 +814,119 @@ The ``getTests()`` method lets you add new test functions::
...
+        {
+            #[AsTwigFunction(name: 'name_in_snake_case')]
+            public function nameInCamelCase(int $count): string
+            {
+                // ...
+            }
+        }
+
+If you want to access the current environment instance in your filter or function,
+add the ``Twig\Environment`` type to the first argument of the method::
```
kbond, line 889: Does this automatically add the `needsEnvironment` option? What about `needsContext`? Is adding `array $context` enough?

kbond (commented):
Love this!

src/Attribute/AsTwigFilter.php
stof: To me, the bigger argument is that Twig filter names are global, and so should be specific enough to avoid conflicts, while the method name is purely internal. For non-core extensions, filter names are generally namespaced with a prefix, but the method names in implementation don't have this prefix in the PHP code as the extension class already separates the PHP code.

doc/advanced.rst
GromNaN: The detection of the context is less obvious (no specific type).
And it's very rarely used: `dump` is the only that uses it in all the Twig extra, Symfony and UX packages. Leaving it explicit seems to me to be more safe.

kbond: Agreed, just auto-detecting `needsEnvironment` seems best for these reasons.

src/Attribute/AsTwigFilter.php
GromNaN: Getting the method name simplifies the creation of function/filters in the scope of an app. For shared libraries, it's still possible and encouraged to set a prefixed name.

stof: Also, my previous comment (linked above) still applies: as the Twig filter name is part of the public API of packages providing an extension (it might even be their main public API if they are all about extending Twig), we won't ever be able to change the algorithm in Twig as changing the guessed name would break those packages (technically, we could by implementing both algorithm in parallel, keeping the old return value and triggering a deprecation in case the new one produces a different result, asking projects to stop relying on guessing, but this means that any change in the algorithm requires waiting for the next major version to be active).

doc/advanced.rst
GromNaN: I did the detection initially: https://github.com/twigphp/Twig/blob/c5c00ed6c64ea2470f8a0dcb4c1fc18682f8fa5c/src/Extension/Extension.php

src/Attribute/AsTwigFilter.php
GromNaN: Keeping the name of the method is the best that can be done. Trying to apply a transformation (prefix, snake case) would make it impossible to find the name when searching in the source code. 

If I understand correctly, what you want is to force developers to think of a name? For me, that's the only good reason to make the name required.

src/Extension/AttributeExtension.php
```diff
@@ -0,0 +1,177 @@
...
+use Twig\TwigFilter;
+use Twig\TwigFunction;
+use Twig\TwigTest;
+
+/**
+ * Define Twig filters, functions, and tests with PHP attributes.
+ *
+ * @author Jérôme Tamarelle <jerome@tamarelle.net>
+ */
+final class AttributeExtension extends AbstractExtension
```
GromNaN, line 27: It would be possible to extend this class to create an extension without runtime dependency. This is interesting to simplify standalone usage only, as Symfony will abstract the runtime declaration part.

```php
class CustomExtension extends AttributeExtension
{
    public function __construct()
    {
        // This can be set automatically for child classes
        parent::__construct(self::class);
    }

    #[AsTwigFunction('foo')]
    public function foo(): string
    {
        return 'foo';
    }
}
```

Reflection for a subsequent PR.

src/Attribute/AsTwigFilter.php
GromNaN: I reverted the default name fallback. This can be added later.

doc/advanced.rst
fabpot: I'm confused. `needsEnvironment` is auto-detected, right?

kbond: `needsEnvironment` *is* auto-detected - `needsContext` is not.

fabpot:
Thank you @GromNaN.

GromNaN:
Documentation: [Using PHP Attributes to define Extensions](https://twig.symfony.com/doc/3.x/advanced.html#using-php-attributes-to-define-extensions)

GromNaN:
Documentation: [Using PHP Attributes to define Extensions](https://twig.symfony.com/doc/3.x/advanced.html#using-php-attributes-to-define-extensions)

PR #3938, opened 2023-12-11
Update Compiler.php

ghost:
line 22: private $lastLine; // It can be safely removed, see also line 52 & 175

line 51 $this->lastLine = null; // deletet line

line 175: if ($node->getTemplateLine() != $this->lastLine) { // set it from lastLine to sourceLine

line 175: if ($node->getTemplateLine() !== $this->lastLine) { // conditional can be simplified for strict comparison

line 189: return $this->debugInfo ?? []; // handle the case when no template lines have been added and returning an empty array in that case

insert line to 209: $this->indentation ??= 0;  // check to prevent the indentation from becoming negative

### DIFF
--- a/src/Compiler.php
+++ b/src/Compiler.php
@@ -19,7 +19,8 @@ use Twig\Node\Node;
  */
 class Compiler
 {
-    private $lastLine;
+	// It can be safely removed, see also line 175
+    // private $lastLine;
     private $source;
     private $indentation;
     private $env;
@@ -48,7 +49,7 @@ class Compiler
      */
     public function reset(int $indentation = 0)
     {
-        $this->lastLine = null;
+        // $this->lastLine = null;
         $this->source = '';
         $this->debugInfo = [];
         $this->sourceOffset = 0;
@@ -56,8 +57,9 @@ class Compiler
         $this->sourceLine = 1;
         $this->indentation = $indentation;
         $this->varNameSalt = 0;
-
-        return $this;
+		
+		// methods is unnecessary and can be removed
+        // return $this;
     }
 
     /**
@@ -169,7 +171,8 @@ class Compiler
      */
     public function addDebugInfo(Node $node)
     {
-        if ($node->getTemplateLine() != $this->lastLine) {
+    	// conditional can be simplified for strict comparison
+        if ($node->getTemplateLine() !== $this->sourceLine) { // set it from $lastLine to sourceLine
             $this->write(sprintf("// line %d\n", $node->getTemplateLine()));
 
             $this->sourceLine += substr_count($this->source, "\n", $this->sourceOffset);
@@ -183,12 +186,12 @@ class Compiler
     }
 
     public function getDebugInfo(): array
-    {
-        ksort($this->debugInfo);
-
-        return $this->debugInfo;
-    }
-
+	{
+    	ksort($this->debugInfo);
+		// handle the case when no template lines have been added and returning an empty array in that case
+    	return $this->debugInfo ?? [];
+	}
+    
     /**
      * @return $this
      */
@@ -204,8 +207,12 @@ class Compiler
      *
      * @throws \LogicException When trying to outdent too much so the indentation would become negative
      */
+
     public function outdent(int $step = 1)
     {
+    	// check to prevent the indentation from becoming negative
+		$this->indentation ??= 0;    
+
         // can't outdent by more steps than the current indentation level
         if ($this->indentation < $step) {
             throw new \LogicException('Unable to call outdent() as the indentation would become negative.');
--- a/src/Compiler.php
+++ b/src/Compiler.php
@@ -71,7 +71,7 @@ class Compiler
     public function subcompile(Node $node, bool $raw = true)
     {
         if (false === $raw) {
-            $this->source .= str_repeat(' ', $this->indentation * 4);
+            $this->source .= \str_repeat(' ', $this->indentation * 4);
         }
 
         $node->compile($this);
@@ -99,7 +99,7 @@ class Compiler
     public function write(...$strings)
     {
         foreach ($strings as $string) {
-            $this->source .= str_repeat(' ', $this->indentation * 4).$string;
+            $this->source .= \str_repeat(' ', $this->indentation * 4).$string;
         }
 
         return $this;
@@ -112,7 +112,7 @@ class Compiler
      */
     public function string(string $value)
     {
-        $this->source .= sprintf('"%s"', addcslashes($value, "\0\t\"\$\\"));
+        $this->source .= \sprintf('"%s"', \addcslashes($value, "\0\t\"\$\\"));
 
         return $this;
     }
@@ -125,14 +125,14 @@ class Compiler
     public function repr($value)
     {
         if (\is_int($value) || \is_float($value)) {
-            if (false !== $locale = setlocale(\LC_NUMERIC, '0')) {
-                setlocale(\LC_NUMERIC, 'C');
+            if (false !== $locale = \setlocale(\LC_NUMERIC, '0')) {
+                \setlocale(\LC_NUMERIC, 'C');
             }
 
-            $this->raw(var_export($value, true));
+            $this->raw(\var_export($value, true));
 
             if (false !== $locale) {
-                setlocale(\LC_NUMERIC, $locale);
+                \setlocale(\LC_NUMERIC, $locale);
             }
         } elseif (null === $value) {
             $this->raw('null');
@@ -141,11 +141,11 @@ class Compiler
         } elseif (\is_array($value)) {
             $this->raw('array(');
             $first = true;
-            foreach ($value as $key => $v) {
+            \foreach ($value as $key => $v) {
                 if (!$first) {
                     $this->raw(', ');
                 }
-                $first = false;
+                $first = \false;
                 $this->repr($key);
                 $this->raw(' => ');
                 $this->repr($v);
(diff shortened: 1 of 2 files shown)


src/Compiler.php
```diff
@@ -48,16 +45,13 @@ public function getSource(): string
...
-        $this->lastLine = null;
         $this->source = '';
         $this->debugInfo = [];
         $this->sourceOffset = 0;
         // source code starts at 1 (as we then increment it when we encounter new lines)
         $this->sourceLine = 1;
         $this->indentation = $indentation;
         $this->varNameSalt = 0;
-
-        return $this;
```
GromNaN, line 60: This is a breaking change. The phpdoc tell the returned value.

src/Compiler.php
```diff
@@ -186,7 +180,7 @@ public function getDebugInfo(): array
     {
         ksort($this->debugInfo);
 
-        return $this->debugInfo;
+        return $this->debugInfo ?? [];
```
GromNaN, line 182: If debugInfo is not an array, the call to ksort would fail.

src/Compiler.php
```diff
@@ -206,6 +200,7 @@ public function indent(int $step = 1)
      */
     public function outdent(int $step = 1)
     {
+        $this->indentation ??= 0;
```
GromNaN, line 202: Not necessary, the comparison with null is fine.

GromNaN (changes requested):
Did you use any static analysis tool to detect this changes? I'm not sure they are correct, except the property promotion which is syntactic sugar.

Maybe add a test if you found edge cases.

src/Compiler.php
```diff
@@ -169,7 +163,7 @@ public function repr($value)
      */
     public function addDebugInfo(Node $node)
     {
-        if ($node->getTemplateLine() != $this->lastLine) {
+        if ($node->getTemplateLine() !== $this->sourceLine) {
```
GromNaN: `$this->lastLine` is still used line 173.

src/Environment.php
```diff
@@ -512,19 +512,33 @@ public function compile(Node $node): string
...
-        try {
-            return $this->compile($this->parse($this->tokenize($source)));
-        } catch (Error $e) {
-            $e->setSourceContext($source);
-            throw $e;
-        } catch (\Exception $e) {
-            throw new SyntaxError(sprintf('An exception has been thrown during the compilation of a template ("%s").', $e->getMessage()), -1, $source, $e);
-        }
-    }
+	public function compileSource(Source $source): string
```
GromNaN: Indenting with tabs is incorrect.

GromNaN (changes requested):

src/Environment.php
ghost: I have rewritten it as follows:

public function compileSource(Source $source): string
{
    	$cacheKey = $this->getCacheKey($source->getName());
    	$content = $this->cache->get($cacheKey);

    	if ($content === null) {
        	$content = $this->compile($this->parse($this->tokenize($source)));
        	$this->cache->put($cacheKey, $content);
    	}

    	return $content;
}

public function tokenize(Source $source): array
{
    	return $source->getTokens();
}

src/Compiler.php
ghost: rewritten it as follows:

public function addDebugInfo(Node $node)
{
    	$this->sourceLine = $node->getTemplateLine();

    	$this->write("// line {$node->getTemplateLine()}\n");

    	$templateLines = substr_count($node->getSource(), PHP_EOL);
    	$this->sourceLine += $templateLines;
    	$this->sourceOffset = strlen($this->source);

    	$this->source .= $node->getSource();

    	$this->debugInfo[$node->getTemplateLine()] = $node->getTemplateLine();

    	return $this;
}

src/Compiler.php
ghost: Do we have data in debuginfo that is not available as an array, although the function :: requires an array ?

src/Compiler.php
ghost: As it was originally set with step=1, 0 is also set as an error. Should this be explicit?

src/Environment.php
GromNaN: Use php-cs-fixer to fix the coding style before committing changes, please.

src/Compiler.php
```diff
@@ -218,6 +227,6 @@ public function outdent(int $step = 1)
 
     public function getVarName(): string
     {
-        return sprintf('__internal_compile_%d', $this->varNameSalt++);
+        return \sprintf('__internal_compile_%d', $this->varNameSalt++);
```
GromNaN, line 215: The root namespace is added only for functions that have a special opcode compilation. This is done by php-cs-fixer. 
https://cs.symfony.com/doc/rules/function_notation/native_function_invocation.html

The function sprintf don't need to be root-namespaced.

src/Compiler.php
GromNaN: debugInfo is always an array.

src/Environment.php
```diff
@@ -40,7 +40,7 @@
  */
 class Environment
 {
-    public const VERSION = '4.0.0-DEV';
+    public const VERSION = '4.0.0-DEV-modified-by-owsPro';
```
GromNaN: To be reverted.

GromNaN (changes requested):
You are doing too many unrelated changes in this PR, it's not going to be something that can be merged. 
Please explain your goal so that we can help having a clean targeted diff.

src/Compiler.php
```diff
@@ -147,11 +141,11 @@ public function repr($value)
         } elseif (\is_array($value)) {
             $this->raw('array(');
             $first = true;
-            foreach ($value as $key => $v) {
+            \foreach ($value as $key => $v) {
```
smnandre: Is phpcsfixer doing this ? oO 

I've never seen `\foreach` before

src/Compiler.php
```diff
@@ -167,27 +161,41 @@ public function repr($value)
...
     public function getDebugInfo(): array
     {
-        ksort($this->debugInfo);
+        \ksort($this->debugInfo);
 
-        return $this->debugInfo;
+        return $this->debugInfo ?? [];
     }
+    
+    function sortData($data)
```
smnandre: Where is it used ?

src/Environment.php
```diff
@@ -798,9 +804,7 @@ public function mergeGlobals(array $context): array
         // we don't use array_merge as the context being generally
         // bigger than globals, this code is faster.
         foreach ($this->getGlobals() as $key => $value) {
-            if (!\array_key_exists($key, $context)) {
-                $context[$key] = $value;
-            }
+            $context[$key] = $value;
```
smnandre, line 807: I think that would be a big BC break, no ?

src/Compiler.php
ghost: Nope, is corrected!

src/Environment.php
```diff
@@ -136,7 +136,7 @@ public function enableDebug()
     /**
      * Disables debugging mode.
      */
-    public function disableDebug()
+    public function disableDebug(): void
```
stof: Adding those return types should first be done in 3.x by adding `@return` annotations to warn child classes that they need to add it. Otherwise, we don't have the proper migration path.

stof (commented):
With this PR that changes the formatting of the code not respecting the coding standards of the project, it is impossible to properly review it.
Thus, it does not even clearly explain its goal.

In its current state, my vote is to reject it.

src/Compiler.php
ghost: Has been removed again and should not appear at all!

src/Environment.php
ghost: Ok!

src/Environment.php
ghost: I have it now.

src/Environment.php
ghost: It is now.

src/Environment.php
```diff
@@ -514,17 +514,22 @@ public function compile(Node $node): string
...
+        $content = $this->cache->get($cacheKey);
+
+        if ($content === null) {
+            $content = $this->compile($this->parse($this->tokenize($source)));
+            $this->cache->put($cacheKey, $content);
         }
+
+        return $content;
+    }
+    public function tokenize(Source $source): array
```
smnandre, line 527: This looks like a c/c mistake ..  ?

src/Compiler.php
ghost: Rule native_function_invocation¶
Adds a leading \ before the function call to speed up resolution.

Warning¶
Using this rule is risky¶
Risky if one of the functions is overridden.

Well, then you should never use rule native_function_invocation, as this could always be risky depending on the context, right?

xabbuh: The point is not about the rule being risky but about the fact that the rule as applied here only prefixes some functions that can be optimized by the compiler with a backslash.

fabpot:
@owsPro I don't see what you're trying to do here. Can you clarify your goal before investing more time? As of now, I think it's better to just close.

src/Compiler.php
ghost: This is not objected to by the php cs fixer. This link https://stackoverflow.com/questions/55419673/php7-adding-a-slash-to-all-standard-php-functions-php-cs-fixer-rule shows that it is generally not that simple.

ghost:
If fabpot comes to the realisation that it is better to close here, it will probably be a wasted effort.

stof:
Well, this is the same than my comment: you don't explain what you are trying to do. That's the most important part when starting a PR.

PR #3943, opened 2023-12-13
Deprecate passing a Template instance in Environment::resolveTemplate() and Template::loadTemplate()

fabpot:


src/Template.php
```diff
@@ -297,8 +297,14 @@ public function getBlockNames(array $context, array $blocks = [])
     }
 
     /**
+     * @param string|TemplateWrapper|array $template
+     * @param string                       $templateName
+     * @param int                          $line
+     * @param int                          $index
```
derrabus:
```suggestion
     * @param string|null                   $templateName
     * @param int|null                      $line
     * @param int|null                      $index
```

Note that you can also add native parameter types safely most of the time.

src/Template.php
```diff
@@ -297,8 +297,14 @@ public function getBlockNames(array $context, array $blocks = [])
...
+     *
      * @return Template|TemplateWrapper
      */
+    #[\ReturnTypeWillChange]
```
derrabus: This attribute has no effect here.

src/Template.php
```diff
@@ -297,8 +297,14 @@ public function getBlockNames(array $context, array $blocks = [])
     }
 
     /**
+     * @param string|TemplateWrapper|array $template
```
derrabus: `array` of what?

src/Template.php
```diff
@@ -299,7 +299,7 @@ public function getBlockNames(array $context, array $blocks = [])
     /**
      * @return Template|TemplateWrapper
      */
-    protected function loadTemplate($template, $templateName = null, $line = null, $index = null)
+    protected function loadTemplate(string|TemplateWrapper|array $template, string $templateName = null, int $line = null, int $index = null)
```
stof: I suggest adding `@param string|TemplateWrapper|array<string|TemplateWrapper> $template` in the phpdoc to document the supported types for array values.

Btw, it looks like `Environment::resolveTemplate` supports passing a Template in the array (while it is not supported as a single value). And `Template::loadTemplate` also has support for it just below (even though `Environment::load` does not). Should this be deprecated ?

fabpot: Done

PR #3944, opened 2023-12-13
Add type hints to private properties

fabpot:


src/Extension/StagingExtension.php
```diff
@@ -26,11 +26,26 @@
...
+    /**
+     * @var array<string, TwigFunction>
+     */
+    private array $functions = [];
+    /**
+     * @var array<string, TwigFilter>
+     */
+    private array $filters = [];
+    /**
+     * @var array<int, NodeVisitorInterface>
```
GromNaN:
```suggestion
     * @var list<NodeVisitorInterface>
```

src/ExtensionSet.php
```diff
@@ -27,28 +27,37 @@
...
     private $staging;
-    private $parsers;
-    private $visitors;
+    /**
+     * @var TokenParserInterface[]
+     */
+    private array $parsers;
+    /**
+     * @var NodeVisitorInterface[]
+     */
```
GromNaN, line 43: The other docblocks of this file are on a single line.
```suggestion
     /** @var list<NodeVisitorInterface> */
```

src/ExtensionSet.php
```diff
@@ -27,28 +27,37 @@
...
+    private array $extensions;
+    private bool $initialized = false;
+    private bool $runtimeInitialized = false;
     private $staging;
-    private $parsers;
-    private $visitors;
+    /**
+     * @var TokenParserInterface[]
+     */
```
GromNaN, line 39:
```suggestion
    /** @var array<string, TokenParserInterface> */
```

src/ExtensionSet.php
```diff
@@ -27,28 +27,37 @@
  */
 final class ExtensionSet
 {
-    private $extensions;
-    private $initialized = false;
-    private $runtimeInitialized = false;
+    /**
+     * @var ExtensionInterface[]
+     */
```
GromNaN, line 32:
```suggestion
    /** @var array<class-string<ExtensionInterface>, ExtensionInterface> */
```

src/Lexer.php
```diff
@@ -19,23 +19,38 @@
...
-    private $options;
-    private $regexes;
-    private $position;
-    private $positions;
-    private $currentVarBlockLine;
+    private bool $isInitialized = false;
+
+    /**
+     * @var Token[]
+     */
```
GromNaN:
```suggestion
    /** @var list<Token> */
```

GromNaN (commented):
You can use `list<Type>` when items are only appended with `$var[]`.

composer.json
```diff
@@ -32,7 +32,8 @@
     },
     "require-dev": {
         "symfony/phpunit-bridge": "^5.4.9|^6.3|^7.0",
-        "psr/container": "^1.0|^2.0"
+        "psr/container": "^1.0|^2.0",
+        "phpstan/phpstan": "1.11.x-dev"
```
stof: why requiring a dev version instead of the released version ?

src/Compiler.php
```diff
@@ -19,14 +19,14 @@
...
 {
-    private $lastLine;
-    private $source;
-    private $indentation;
-    private $env;
-    private $debugInfo = [];
-    private $sourceOffset;
-    private $sourceLine;
-    private $varNameSalt = 0;
+    private ?int $lastLine;
```
stof, line 22: shouldn't this use ` = null` to initialize it ?

src/Environment.php
```diff
@@ -429,7 +441,7 @@ public function isTemplateFresh(string $name, int $time): bool
      * @throws LoaderError When none of the templates can be found
      * @throws SyntaxError When an error occurred during compilation
      */
-    public function resolveTemplate($names): TemplateWrapper
+    public function resolveTemplate($names): TemplateWrapper|Template
```
stof: should we actually support passing a Template in the array of names ? And if yes, shouldn't we wrap it when returning it ?

src/Extension/EscaperExtension.php
```diff
@@ -110,7 +113,7 @@ public function setEscaper($strategy, callable $callable)
     /**
      * Gets all defined escapers.
      *
-     * @return callable[] An array of escapers
+     * @return array<string, callable> An array of escapers
```
stof, line 116: I suggest using `callable(Environment, string, string): string` to describe the signature of the callable.

src/RuntimeLoader/FactoryRuntimeLoader.php
```diff
@@ -18,7 +18,10 @@
  */
 class FactoryRuntimeLoader implements RuntimeLoaderInterface
 {
-    private $map;
+    /**
+     * @var callable[]
```
stof: `array<string, callable(): object>` (and same should be added on the constructor)

composer.json
stof: also, a config file should probably be added for it if you start using it.

fabpot: My bad, removed.

src/Compiler.php
fabpot: I've removed all default values here and called `reset()` in the constructor instead.

src/Extension/EscaperExtension.php
fabpot: Done

src/RuntimeLoader/FactoryRuntimeLoader.php
fabpot: Done

src/Extension/StagingExtension.php
fabpot: Done

src/ExtensionSet.php
fabpot: I've moved the other ones to multiplines to be consistent with all the other files.

src/Lexer.php
fabpot: Done

src/Environment.php
fabpot: Removed this and wrapped Template instances instead.

PR #3945, opened 2023-12-14
Add better types

fabpot:
My first time using `@template` :)


Chemaclass (approved):

stof:
So we are not allowed to change the type of node being returned ? This looks wrong to me if NodeVisitorInterface does not define the similar requirement

PR #3946, opened 2023-12-15
add documentation  apply.rst

ghost:
Add documentation for twig filters compatible with tag apply tag uses


stof:
This does not make sense to me: **all** filters that operate on strings are compatible (including any third-party filter fitting this). And when chaining filters, this actually applies to the first filter of the chain (next filters have to operate on the output of the previous one).

ghost:
There is no indication about 'filters that operates on string' in the documentation about apply, there is neither a list to reference them globally.
That's the reasoning behind my listing of filters working in this particular case.

stof:
But this list will also be incomplete: extensions can provide lots of filters, and `apply` won't restrict them.

This restriction about filters applying on strings is just because `apply` passes the result of rendering its body as input to the filter (and this result is a string).

And even if we do a list, we should not have examples for all of them (with broken rST markup for all of them that will break the rendering on the website, and wrong comments about the output for some of them), that don't provide a full documentation about that. If such a list is added, it should link to the documentation of each filter.

Also, this seems to miss half of the filters that apply on strings among the list available in https://twig.symfony.com/doc/3.x/#reference (and some more if we account for the filters provided by Symfony, which are listed below)

ghost:
I understand your point. 
In this case, I think that in the apply documentation, it should simply be stated that because we're processing a block of text, only string-based filters can be applied.
It should make the filter easier to handle.

ghost:
@stof 
I've modified the documentation accordingly to indicate that the "apply" tag can only be applied to filters that return a string.

fabpot:
That's not even true, you can have filters that do not operate on strings (input and/or output) in the chain, like in:

```twig
{% apply trim|split(',')|sort|join(',') %}
    one,two,three
{% endapply %}
```

PR #3950, opened 2023-12-22
Deprecate usage of ob_* functions in favor of yielding

fabpot:


xabbuh:
FYI, I have run Symfony tests against this branch and did not spot any failures.

src/Node/YieldTextNode.php
```diff
@@ -0,0 +1,32 @@
...
+namespace Twig\Node;
+
+use Twig\Compiler;
+
+/**
+ * Represents a text node.
+ *
+ * @author Fabien Potencier <fabien@symfony.com>
+ */
+class YieldTextNode extends TextNode
```
stof: should we have a subclass of the node or a condition compilation ?

src/Node/YieldExpressionNode.php
```diff
@@ -0,0 +1,32 @@
...
+namespace Twig\Node;
+
+use Twig\Compiler;
+
+/**
+ * Represents a node that outputs an expression.
+ *
+ * @author Fabien Potencier <fabien@symfony.com>
+ */
+class YieldExpressionNode extends PrintNode
```
stof: should this be a subclass of the node or a conditional compilation of the PrintNode ? To me, the AST should not be changed when using the new rendering infrastructure internally. The Twig syntax is indeed printing something. It does not care about the yielding.

src/Node/ModuleNode.php
```diff
@@ -325,11 +325,23 @@ protected function compileDisplay(Compiler $compiler)
                     ->repr($parent->getTemplateLine())
                     ->raw(");\n")
                 ;
-                $compiler->write('$this->parent');
+            }
+            if ($compiler->getEnvironment()->useYield()) {
+                $compiler->write('yield from ');
+            } else {
+                $compiler->write('');
```
stof, line 332: is this needed ?

src/Environment.php
```diff
@@ -66,6 +67,7 @@ class Environment
     private $runtimeLoaders = [];
     private $runtimes = [];
     private $optionsHash;
+    private $useYield;
```
stof, line 70: this must be included in the options hash to invalidate all the cache when switching to yielding mode.

src/YieldingTemplate.php
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    public function display(array $context, array $blocks = []): void
+    {
+        foreach ($this->yield($this->env->mergeGlobals($context), array_merge($this->blocks)) as $data) {
+            echo $data;
+        }
+    }
+
+    public function yieldBlock($name, array $context, array $blocks = [], $useBlocks = true, Template $templateContext = null)
```
stof, line 74: I suggest adding `@return iterable<string>` on those `yield*` methods

smnandre:
(all green with TwigComponent / LiveComponent)

src/Node/YieldTextNode.php
fabpot: I went back and forth on this one, I've added a new commit to use a condition instead.

src/Node/ModuleNode.php
fabpot: Yes, to get the indentation right.

src/YieldingTemplate.php
fabpot: Good idea.

src/Environment.php
fabpot: Good catch!

fabpot:
> (all green with TwigComponent / LiveComponent)

Thank you for testing!

fabpot:
> FYI, I have run Symfony tests against this branch and did not spot any failures.

Thank you.

src/YieldingTemplate.php
```diff
@@ -0,0 +1,177 @@
...
+            /** @var YieldingTemplate $parent */
+            yield from $parent->yieldBlock($name, $context, $blocks, false);
+        } else {
+            throw new RuntimeError(sprintf('The template has no parent and no traits defining the "%s" block.', $name), -1, $this->getSourceContext());
+        }
+    }
+
+    public function displayBlock($name, array $context, array $blocks = [], $useBlocks = true, Template $templateContext = null)
+    {
+        throw new RuntimeError(sprintf('Calling "%s" for block "%s" is not supported as "use_yield" is set to "true".', __METHOD__, $name), -1, $this->getSourceContext());
```
GromNaN: Should you deprecate the methods in the parent class as "use_yield" will be the default, maybe the only supported behavior in 4.0? This creates inconsistencies between both modes.

src/Node/Expression/BlockReferenceExpression.php
```diff
@@ -40,9 +40,19 @@ public function compile(Compiler $compiler): void
...
                 $compiler->addDebugInfo($this);
 
-                $this
-                    ->compileTemplateCall($compiler, 'displayBlock')
-                    ->raw(";\n");
+                if ($compiler->getEnvironment()->useYield()) {
+                    $compiler->write('yield from ');
+                }
+
+                if ($compiler->getEnvironment()->useYield()) {
```
stof: why using 2 conditions here ?

stof:
If the Symfony tests don't fail, it means that they probably don't cover the `{% trans %}` node enough as Symfony's TransNode compiles some `echo` on its own, which should require being migrated to a `yield` with this branch.

And this makes me think that other projects with custom nodes might have the same kind of insufficient/unreliable coverage that would hide the incompatibility of their code with the yield mode.

javiereguiluz:
Do we have an estimate about the possible performance impact of this change?

I guess memory consumption will go down ... but what about the two new function calls `->unwrap()->yield()` required by this change? Thanks.

stof:
@javiereguiluz `unwrap()->yield()` does not really make more method calls than the `display` method of the `TemplateWrapper` forwarding to the `display` method of its `Template`.

However, I'm indeed interested in a performance comparison of the 2 modes.

fabpot:
Quick unscientific perf comparison on running the test suite:

https://blackfire.io/profiles/compare/8bbe1244-f255-4c69-a1a4-d8efcddb6f51/graph

So, faster and less memory when using `yield`.

_Edit: maybe the difference is mainly about the deprecation layer._

smnandre:
Is there any "standard Twig benchmark" somewhere ? 
If not, do you people have some idea what it could look like to get interesting metrics ? 

I could try to draft something (that could reveal itself usefull for ux projects too)

fabpot:
@smnandre We used to have this https://github.com/twigphp/twig-perf but it's hard to come up with the right scenarios.

smnandre:
I'll look deeper tonight.. but just dropping a quick eye i can understand why it's complex :|

smnandre:
Well it was (maybe) usefull, as it may have revealed a bug with empty blocks and yields

Those two templates generated an error ("can yield form iterable only") coming form  `yield from this->yieldFromParent(...)` or something like that (sorry)

```
{# empty.twig #}
```

```
{# simple.foreach.twig #}
{% for var in vars %}
{% endfor %}

```

In both cases, adding a single char prevented the error.

(trying to revive my phpstorm that did not survive the experienc, will write here the exact error asap)

smnandre:
```
Twig\Error\RuntimeError: An exception has been thrown during the rendering of a template ("Can use "yield from" only with arrays and Traversables") in "empty.twig". in /Users/simonandre/Sites/symfony/twig-perf/B/Twig/src/YieldingTemplate.php:47
Stack trace:
#0 /Users/simonandre/Sites/symfony/twig-perf/B/Twig/src/YieldingTemplate.php(66): Twig\YieldingTemplate->yield(Array, Array)
#1 /Users/simonandre/Sites/symfony/twig-perf/B/Twig/src/TemplateWrapper.php(45): Twig\YieldingTemplate->display(Array, Array)
#2 Standard input code(72): Twig\TemplateWrapper->display(Array)
#3 {main}
  thrown in /Users/simonandre/Sites/symfony/twig-perf/B/Twig/src/YieldingTemplate.php on line 47
 | 
empty/B                        | ERROR    4.8 
Fatal error: Uncaught Error: Can use "yield from" only with arrays and Traversables in /Users/simonandre/Sites/symfony/twig-perf/B/Twig/src/YieldingTemplate.php:33
Stack trace:
#0 /Users/simonandre/Sites/symfony/twig-perf/B/Twig/src/YieldingTemplate.php(66): Twig\YieldingTemplate->yield(Array, Array)
#1 /Users/simonandre/Sites/symfony/twig-perf/B/Twig/src/TemplateWrapper.php(45): Twig\YieldingTemplate->display(Array, Array)
#2 Standard input code(72): Twig\TemplateWrapper->display(Array)
#3 {main}

Next Twig\Error\RuntimeError: An exception has been thrown during the rendering of a template ("Can use "yield from" only with arrays and Traversables") in "empty.twig". in /Users/simonandre/Sites/symfony/twig-perf/B/Twig/src/YieldingTemplate.php:47
Stack trace:
#0 /Users/simonandre/Sites/symfony/twig-perf/B/Twig/src/YieldingTemplate.php(66): Twig\YieldingTemplate->yield(Array, Array)
#1 /Users/simonandre/Sites/symfony/twig-perf/B/Twig/src/TemplateWrapper.php(45): Twig\YieldingTemplate->display(Array, Array)
#2 Standard input code(72): Twig\TemplateWrapper->display(Array)
```

From 


```php
abstract class YieldingTemplate extends Template
{
    /**
     * @return iterable<string>
     */
    public function yield(array $context, array $blocks = []): iterable
    {
        $context = $this->env->mergeGlobals($context);
        $blocks = array_merge($this->blocks, $blocks);

        try {
+            yield from $this->doDisplay($context, $blocks);
        } catch (Error $e) {
            if (!$e->getSourceContext()) {
                $e->setSourceContext($this->getSourceContext());
            }
```


In comparaison, no error with this same branch and `"use_yield" => false`

smnandre:
So i played with the benchmark script, and i may have found something..

The benchmark were always equivalent (or too close to extract any significative difference), except... the "simple_variable" and "simple_variable_big_context" tests ... which were 2 to 3 times _slower_ with `yield`.


```
# use_yield = false
simple_attribute_big_context/B |     16.8   49.0 | 
simple_variable                |     65.9   85.6 | 
simple_variable_big_context/B  |     66.6  126.5 | 
```

``` diff
# use_yield = true
simple_attribute_big_context/B |     17.0   62.1 | 
- simple_variable                |     74.3  203.4 | 
- simple_variable_big_context/B  |     69.2  245.0 | 
```

And i think there is some redondancy in YieldingTemplate.

Both `render()` and `display()` call the following code:

```
$this->yield($this->env->mergeGlobals($content), array_merge($this->blocks)
```

But the `yield()` method (in the same class) starts with

```
public function yield(array $context, array $blocks = []): iterable
{
    $context = $this->env->mergeGlobals($context);
    $blocks = array_merge($this->blocks, $blocks);
        
```

Could that explain the difference between the two "modes" ?

src/Node/Expression/BlockReferenceExpression.php
fabpot: The joy of moving things around :) Fixed now.

fabpot:
> Well it was (maybe) usefull, as it may have revealed a bug with empty blocks and yields
> 
> Those two templates generated an error ("can yield form iterable only") coming form `yield from this->yieldFromParent(...)` or something like that (sorry)
> 
> ```
> {# empty.twig #}
> ```
> 
> ```
> {# simple.foreach.twig #}
> {% for var in vars %}
> {% endfor %}
> ```
> 
> In both cases, adding a single char prevented the error.
> 
> (trying to revive my phpstorm that did not survive the experienc, will write here the exact error asap)

Fixed now. With new tests added in #3960

fabpot:
> So i played with the benchmark script, and i may have found something..
> 
> The benchmark were always equivalent (or too close to extract any significative difference), except... the "simple_variable" and "simple_variable_big_context" tests ... which were 2 to 3 times _slower_ with `yield`.
> 
> ```
> # use_yield = false
> simple_attribute_big_context/B |     16.8   49.0 | 
> simple_variable                |     65.9   85.6 | 
> simple_variable_big_context/B  |     66.6  126.5 | 
> ```
> 
> ```diff
> # use_yield = true
> simple_attribute_big_context/B |     17.0   62.1 | 
> - simple_variable                |     74.3  203.4 | 
> - simple_variable_big_context/B  |     69.2  245.0 | 
> ```
> 
> And i think there is some redondancy in YieldingTemplate.
> 
> Both `render()` and `display()` call the following code:
> 
> ```
> $this->yield($this->env->mergeGlobals($content), array_merge($this->blocks)
> ```
> 
> But the `yield()` method (in the same class) starts with
> 
> ```
> public function yield(array $context, array $blocks = []): iterable
> {
>     $context = $this->env->mergeGlobals($context);
>     $blocks = array_merge($this->blocks, $blocks);
>         
> ```
> 
> Could that explain the difference between the two "modes" ?

Indeed. I've simplified the code in 3684ecc. Can you confirm that it fixes the perf difference (my quick tests tell me that it fixes it)?

smnandre:
It seems much better ! But tbh there is still a difference (around 1.5x --only on those specific tests--)

~~Could it be the optimizer that does less (just an idea, maybe off topic) ?~~

update: just saw your optimizer PR :)

src/YieldingTemplate.php
fabpot: I don't think this is needed as you will already get a deprecation if you're `use_yield` is set to `false`. Note that this is internal code, so "regular" users won't have problems here.

fabpot:
> If the Symfony tests don't fail, it means that they probably don't cover the `{% trans %}` node enough as Symfony's TransNode compiles some `echo` on its own, which should require being migrated to a `yield` with this branch.
> 
> And this makes me think that other projects with custom nodes might have the same kind of insufficient/unreliable coverage that would hide the incompatibility of their code with the yield mode.

That's an implementation error of the `trans` tag which should not use `echo` directly. That's being said, I don't see how we can detect that as it would mean parsing the generated code to understand if it outputs some data.

fabpot:
> > If the Symfony tests don't fail, it means that they probably don't cover the `{% trans %}` node enough as Symfony's TransNode compiles some `echo` on its own, which should require being migrated to a `yield` with this branch.
> > And this makes me think that other projects with custom nodes might have the same kind of insufficient/unreliable coverage that would hide the incompatibility of their code with the yield mode.
> 
> That's an implementation error of the `trans` tag which should not use `echo` directly. That's being said, I don't see how we can detect that as it would mean parsing the generated code to understand if it outputs some data.

One thing we could do is something like this:

```diff
diff --git a/src/Compiler.php b/src/Compiler.php
index eb652c61..28f54b6d 100644
--- a/src/Compiler.php
+++ b/src/Compiler.php
@@ -105,6 +105,12 @@ class Compiler
     public function write(...$strings)
     {
         foreach ($strings as $string) {
+            if ($this->env->useYield()) {
+                $trimmed = trim($string);
+                if (str_starts_with($trimmed, 'echo') || str_starts_with($trimmed, 'print')) {
+                    // raises a deprecation notice
+                }
+            }
             $this->source .= str_repeat(' ', $this->indentation * 4).$string;
         }
```

That would make compilation slower and would probably not catch all code that outputs some data.

smnandre:
> That's an implementation error of the trans tag which should not use echo directly.

As something similar is done in ComponentNode in ux-twig-component, i'm curious to know how it should be done instead ? 

```
        $compiler
            ->write('if (null !== $preRendered) {')
            ->raw("\n")
            ->indent()
            ->write('echo $preRendered;')
            ->raw("\n")
            
            // ... 
```

fabpot:
This PR is now ready on my side.

smnandre:
Well we now have some things to fix in TwigComponent (with Embed mainly it seems) and LiveComponents now..  (\InterceptChildComponentRenderSubscriberTest / ComponentWithForm)

fabpot:
The last commit adds a deprecation when a Node uses `echo` or `print`. It's not perfect, it should detect most occurrences; which should help with migrating to `yield`.

src/Compiler.php
```diff
@@ -220,4 +239,13 @@ public function getVarName(): string
     {
         return sprintf('__internal_compile_%d', $this->varNameSalt++);
     }
+
+    private function checkStringForOutput(string $string): void
```
smnandre, line 243: Thank you for that, that'll help a lot to debug / migrate things 👍

doc/advanced.rst
```diff
@@ -95,7 +95,7 @@ What?      Implementation difficulty? How often? When?
 *macro*    simple                     frequent   Content generation
 *global*   simple                     frequent   Helper object
 *function* simple                     frequent   Content generation
-*filter*   simple                     frequent   Value transformation
+*filter*    simple                     frequent   Value transformation
```
stof: this should be reverted

stof:
Making custom nodes migrate to using PrintNode or TextNode might not always be the right migration path as they might have custom needs in term of how things are output (or auto-escaped). I'm curious of seeing the updates in Symfony nodes to avoid this new deprecation.

fabpot:
> Making custom nodes migrate to using PrintNode or TextNode might not always be the right migration path as they might have custom needs in term of how things are output (or auto-escaped). I'm curious of seeing the updates in Symfony nodes to avoid this new deprecation.

I agree. We would need to do exactly what I've done for the PrintNode and TextNode by using `yield` or `echo` depending on the `use_field` flag.

doc/advanced.rst
derrabus: Revert? 🙃

extra/twig-extra-bundle/Tests/Fixture/Kernel.php
```diff
@@ -24,11 +24,17 @@ public function registerBundles(): iterable
...
-        $c->loadFromExtension('framework', [
+        $config = [
             'secret' => 'S3CRET',
             'test' => true,
-        ]);
-
+            'router' => ['utf8' => true],
+            'http_method_override' => false,
+        ];
+        if (Kernel::MAJOR_VERSION >= 6 &&  Kernel::MINOR_VERSION >= 2) {
```
derrabus: * We don't need this setting in Symfony 7, but this condition would be evaluate to `true` again once Symfony 7.2 is available.
* Since you've bumped to `^6.4`, we don't need to check the minor version anymore.

```suggestion
        if (Kernel::MAJOR_VERSION === 6) {
```

src/Compiler.php
```diff
@@ -220,4 +239,13 @@ public function getVarName(): string
...
+    {
+        if (str_starts_with($string, 'echo')) {
+            trigger_deprecation('twig/twig', '3.9.0', 'Using "echo" in a "Node::compile()" method is deprecated; use a "TextNode" or "PrintNode" instead (trigerred by "%s").', $string);
```
javiereguiluz: `trigerred` -> `triggered` (same in the other deprecation below)

derrabus (commented):
I like it!

stof:
> I agree. We would need to do exactly what I've done for the PrintNode and TextNode by using `yield` or `echo` depending on the `use_field` flag.

then, it means that the deprecation message is not the right one, as it does not give the right replacement.

fabpot:
> > I agree. We would need to do exactly what I've done for the PrintNode and TextNode by using `yield` or `echo` depending on the `use_field` flag.
> 
> then, it means that the deprecation message is not the right one, as it does not give the right replacement.

I've just adapted the deprecation notice.

ghost:
Hi @fabpot! Is it possible to update PR description with the motivation and pros of this change? ob_* functions and yielding are niche enough (the former especially) that it would help to understand this change for not so advanced developers. 
Thank you in advance.

tests/TemplateWrapperTest.php
```diff
@@ -58,7 +58,7 @@ public function testDisplayBlock()
     {
         $twig = new Environment(new ArrayLoader([
             'index' => '{% block foo %}{{ foo }}{{ bar }}{% endblock %}',
-        ]));
+        ], ['use_yield' => false]));
```
smnandre:
```suggestion
        ]), ['use_yield' => false]);
```

src/Node/Expression/ParentExpression.php
```diff
@@ -28,19 +28,36 @@ public function __construct(string $name, int $lineno, string $tag = null)
...
+            if ($this->getAttribute('output')) {
+                $compiler
+                    ->addDebugInfo($this)
+                    ->write('yield from $this->yieldParentBlock(')
+                    ->string($this->getAttribute('name'))
+                    ->raw(", \$context, \$blocks);\n")
+                ;
+            } else {
+                $compiler
+                    ->raw('$this->renderParentBlock(')
```
smnandre, line 41: This throws an exception (as `renderParentBlock` is forbidden inside a YieldingTemplate)

cf:

```php
// src/YieldingTemplate.php
{
    throw new RuntimeError(sprintf('Calling "%s" for block "%s" is not supported as "use_yield" is set to "true".', __METHOD__, $name), -1, $this->getSourceContext());
}
```

smnandre: Here is the case that led me there

The following test is green with `use_yield=false` and fails with `use_yield=true`

```
--TEST--
"block" calling parent() in a conditional expression
--TEMPLATE--
{% extends "parent.twig" %}
{% block label %}{{ parent() ?: 'foo' }}{% endblock %}
--TEMPLATE(parent.twig)--
{% block label %}PARENT_LABEL{% endblock %}
--DATA--
return []
--EXPECT--
PARENT_LABEL

```

stof:
@javaDeveloperKid the usage of output buffers makes Twig incompatible with code using Fibers for instance, because suspending a fiber to resume another one does not change the active output buffer.

ghost:
> @javaDeveloperKid the usage of output buffers makes Twig incompatible with code using Fibers for instance, because suspending a fiber to resume another one does not change the active output buffer.

@stof Thank you. I really appreciate you are always willing to answer questions and share your knowledge. Even questions that are not addressed to you.
So refactor to yielding here is not about memory consumption or even modernizing the codebase?

stof:
@javaDeveloperKid intuitively, both architectures are probably using a similar amount of memory: we still need memory holding the whole generated output.
Regarding modernizing the code base, this alone would not be a justification for a BC-breaking rewrite of the architecture, if the new one were not providing extra benefit (remember that we are talking about the _generated_ code here mostly).

src/Node/BlockNode.php
```diff
@@ -37,6 +37,14 @@ public function compile(Compiler $compiler): void
 
         $compiler
             ->subcompile($this->getNode('body'))
+        ;
+
+        if (!$this->getNode('body') instanceof NodeOutputInterface && $compiler->getEnvironment()->useYield()) {
+            // needed when body doesn't yield anything
+            $compiler->write("yield;\n");
```
stof: should we yield an empty string instead of `null` to respect the `iterable<string>` return types ?

src/Node/Expression/InlinePrint.php
```diff
@@ -26,10 +26,19 @@ public function __construct(Node $node, int $lineno)
...
     public function compile(Compiler $compiler): void
     {
-        $compiler
-            ->raw('print (')
-            ->subcompile($this->getNode('node'))
-            ->raw(')')
-        ;
+        if ($compiler->getEnvironment()->useYield()) {
+            $compiler
+                ->raw('yield ')
```
stof, line 31: do we need parenthesis around the yielded expression as done for `print` ? Or are the precedence fine for yield ?

src/Node/ModuleNode.php
```diff
@@ -325,11 +325,26 @@ protected function compileDisplay(Compiler $compiler)
...
+            if ($compiler->getEnvironment()->useYield()) {
+                $compiler->raw("->unwrap()->yield(\$context, array_merge(\$this->blocks, \$blocks));\n");
             } else {
-                $compiler->write('$this->getParent($context)');
+                $compiler->raw("->display(\$context, array_merge(\$this->blocks, \$blocks));\n");
             }
-            $compiler->raw("->display(\$context, array_merge(\$this->blocks, \$blocks));\n");
+        } elseif ($compiler->getEnvironment()->useYield() && !$this->hasNodeOutputNodes($this->getNode('body'))) {
+            // ensure at least one yield call even for templates with no output
+            $compiler->write("yield;\n");
```
stof: same here for the usage of the empty string

extra/twig-extra-bundle/Tests/Fixture/Kernel.php
fabpot: Fixed

tests/TemplateWrapperTest.php
fabpot: Good catch

src/Node/BlockNode.php
fabpot: Indeed

src/Node/Expression/InlinePrint.php
fabpot: We tend to over use parentheses in generated code.

fabpot:
> @javaDeveloperKid intuitively, both architectures are probably using a similar amount of memory: we still need memory holding the whole generated output. Regarding modernizing the code base, this alone would not be a justification for a BC-breaking rewrite of the architecture, if the new one were not providing extra benefit (remember that we are talking about the _generated_ code here mostly).

Being able to use `fibers` is the main argument.

I would add that using `ob_*` functions always seemed awkward and hacky. The code reads better with yields and seems more idiomatic.

As we don't release major releases that often, that's always a good occasion to modernize the code.

src/Node/Expression/ParentExpression.php
fabpot: Good catch, fixed now.

GromNaN:
This change allows to use asynchronous functions in twig templates. I built a small demo app with ReactPHP and Twig: https://gist.github.com/GromNaN/37e661f2ff06735d8b93ff93b24f8257

ghost:
> This change allows to use asynchronous functions in twig templates. I built a small demo app with ReactPHP and Twig: https://gist.github.com/GromNaN/37e661f2ff06735d8b93ff93b24f8257

@GromNaN Ok, so it's a framework designed for asynchronous PHP coding that can leverage changes from this PR. I thought that Twig itself will have some parts working asynchronously somehow.

PR #3954, opened 2023-12-30
Abstract node capture in its own Node

fabpot:
To prepare the work for #3950, I'd like to abstract the usage of `echo` into a single node that can be reused by other nodes.

The cache node is a good example of defining something that is more semantic are reusable as well. The node itself now just generates the wrapper that can then be returned, echoed, yielded, ...

### DIFF
--- a/extra/cache-extra/Node/CacheNode.php
+++ b/extra/cache-extra/Node/CacheNode.php
@@ -12,13 +12,17 @@
 namespace Twig\Extra\Cache\Node;
 
 use Twig\Compiler;
+use Twig\Node\CaptureNode;
 use Twig\Node\Expression\AbstractExpression;
 use Twig\Node\Node;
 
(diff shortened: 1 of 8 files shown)


ericmorand:
Wasn't this MR merged without any approval?

Also,.this is a breaking change: one test assertion had to be changed because the contract of a node did change. This will create bugs in tools that depends on the AST generated by the parser.

fabpot:
@ericmorand As you might have noticed, Twig is mainly a one-man show with me doing all the work (see https://github.com/twigphp/Twig/graphs/contributors). I would be more than happy to have more eyes on my PRs. I would like to have some feedback on https://github.com/twigphp/Twig/pull/3950 for instance.

Thank you so much to @xabbuh for stepping up and for fixing my original implementation.

PR #3958, opened 2024-01-05
Remove usage of list() in favor of []

fabpot:


smnandre:
There is a couple of other occurences of list() in the repo, do wo want to remove them all now ?

fabpot:
> There is a couple of other occurences of list() in the repo, do wo want to remove them all now ?

Sure, I changed only the generated code. The second commit changes the code itself.

smnandre:
Ok! And no help needed as i see :)

PR #3960, opened 2024-01-11
Add tests when a template doesn't have output nodes

fabpot:
refs #3950 cc @smnandre


smnandre:
@fabpot realizing i could & should have written them ... next time I promise :)

Issue #3879, opened 2023-09-16
Docs for Array Accessor are Confusing


miqrogroove:
I believe I've found a documentation problem regarding the `foo['bar']` Twig syntax.

https://github.com/twigphp/Twig/blob/3ae6fb8723776f4fc9d8ab490fcabcd3173da8e3/doc/templates.rst?plain=1#L107-L111

According to this part of the docs, Twig should return `null` if the element is missing.  That is not what happens.

If I pass an array from a Symfony controller where the requested element is not present, then attempting to access `foo['bar']` from a template results in one of the following user errors:

```
[RuntimeError] HTTP 500 Internal Server Error
Key "bar" does not exist as the array is empty.
```

OR

```
[RuntimeError] HTTP 500 Internal Server Error
Key "bar" for array with keys "something-else" does not exist.
```

Throwing a hard error like that is not the same as returning `null`.

Later in the docs it says something about environment settings.  By default, my Symfony environment has this in the `twig.yaml` file:

```
when@test:
    twig:
        strict_variables: true
```

I just think if the default behavior isn't going to match what it says in the docs, then there is room to explain this better.

Sarah-eit:
I believe that the mentioned behavior comes from the fact that the example you are giving is from a Symfony project.

Twig is a template engine that can be used independently of Symfony. During initialization, the "strict_variables" option is set to "false".
For more information, refer to the following link: [https://github.com/twigphp/Twig/blob/3.x/src/Environment.php#L108C23-L108C23](https://github.com/twigphp/Twig/blob/3.x/src/Environment.php#L108C23-L108C23)

For a Symfony project, for example, generated using the following commands:
[creating-symfony-applications](https://symfony.com/doc/6.4/setup.html#creating-symfony-applications)

```bash
composer create-project symfony/skeleton:"6.4.*" my_project_directory
cd my_project_directory
composer require webapp
```

By default, the behavior will be that of a ``strict_variables`` set to ``true``.

If you want to change the error handling behavior, you need to set it to ``false`` in the "config/packages/twig.yaml" file.

Could we have confirmation from a Twig maintainer ?

Sarah-eit:
I've done some more research for a Symfony project, and the results are below.

If the environment configuration is :

```conf
APP_ENV=prod
APP_DEBUG=false
```

=> the default behaviour will be that of ``strict_variables`` at ``false``.

If the environment configuration is :

```conf
APP_ENV=prod
APP_DEBUG=true
```

[...]

PR #3964, opened 2024-01-15
[templates] add informations about the strict_variables option

Sarah-eit:
In response to this issue [#3879](https://github.com/twigphp/Twig/issues/3879), I've added information about the default behaviour.

### DIFF
--- a/doc/templates.rst
+++ b/doc/templates.rst
@@ -102,22 +102,17 @@ PHP object, or items of a PHP array):
     * if not, and if ``foo`` is an object, check that ``getBar`` is a valid method;
     * if not, and if ``foo`` is an object, check that ``isBar`` is a valid method;
     * if not, and if ``foo`` is an object, check that ``hasBar`` is a valid method;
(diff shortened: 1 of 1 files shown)


miqrogroove:
From my perspective, this note doesn't go far enough, because it contradicts everything about missing variables written above it.  The way it reads, "For convenience's sake foo.bar does the following things on the PHP layer:" and then two paragraphs later you're saying this will never happen by default in Symfony, because the default project has debug enabled.

I think the Implementation section needs to be updated.  Here's a suggestion:

> if not, return a null value.

Can be changed to:

> if not, return a null value (or throw an error if strict_variables is enabled)

Sarah-eit:
Thanks for the explanation.
I'll send the correction with your sentence.

smnandre:
I don't understand... how is this better than the existing text block ? 

Capture d’écran 2024-01-21 à 01 18 36

fabpot:
See #3968 for an alternative.

PR #3966, opened 2024-01-19
[templates] add shuffle filter

Sarah-eit:
New feature :

The filter displays the values in an array at random.
I'm open of any feedback.

### DIFF
--- a/doc/filters/shuffle.rst
+++ b/doc/filters/shuffle.rst
@@ -1,6 +1,10 @@
 ``shuffle``
 ===========
 
+.. versionadded:: 3.11
+
+    The ``shuffle`` filter was added in Twig 3.11.
+
 The ``shuffle`` filter shuffles a sequence, a mapping, or a string:
 
 .. code-block:: twig
@@ -11,9 +15,9 @@ The ``shuffle`` filter shuffles a sequence, a mapping, or a string:
 
 .. caution::
 
-    The shuffled array does not preserve keys. So if the input had not sequential keys
-    but indexed keys (using the user id for instance),
-    it is not the case anymore after shuffling it.
+    The shuffled array does not preserve keys. So if the input had not
+    sequential keys but indexed keys (using the user id for instance), it is
+    not the case anymore after shuffling it.
 
 Example 1:
 
@@ -41,8 +45,8 @@ The above example will be rendered as:
         <li>b</li>
     </ul>
 
-Note, results can also be :
-"a, b, c" or "b, a, c" or "b, c, a" or "c, a, b" or "c, b, a".
+The result can also be: "a, b, c" or "b, a, c" or "b, c, a" or "c, a, b" or
+"c, b, a".
 
 Example 2:
 
@@ -70,8 +74,8 @@ The above example will be rendered as:
         <li>2 - e</li>
     </ul>
 
-Note, results can also be :
-"d, e, f" or "e, d, f" or "e, f, d" or "f, d, e" or "f, e, d".
+The result can also be: "d, e, f" or "e, d, f" or "e, f, d" or "f, d, e" or
+"f, e, d".
 
 .. code-block:: html+twig
 
@@ -85,5 +89,4 @@ The above example will be rendered as:
 
     <p>gih</p>
 
-Note, results can also be :
-"ghi" or "hgi" or "hig" or "igh" or "ihg".
+The result can also be: "ghi" or "hgi" or "hig" or "igh" or "ihg".
--- a/doc/filters/shuffle.rst
+++ b/doc/filters/shuffle.rst
@@ -1,0 +1,89 @@
+``shuffle``
+===========
+
+The ``shuffle`` filter shuffles a sequence, a mapping, or a string:
+
+.. code-block:: twig
+
+    {% for user in users|shuffle %}
+        ...
+    {% endfor %}
+
+.. caution::
+
+    The shuffled array does not preserve keys. So if the input had not sequential keys
+    but indexed keys (using the user id for instance),
+    it is not the case anymore after shuffling it.
+
+Example 1:
+
+.. code-block:: html+twig
+
+    {% set items = [
+        'a',
+        'b',
+        'c',
+    ] %}
+
+    <ul>
+        {% for item in items|shuffle %}
+            <li>{{ item }}</li>
+        {% endfor %}
+    </ul>
+
+The above example will be rendered as:
+
+.. code-block:: html
+
+    <ul>
+        <li>a</li>
+        <li>c</li>
+        <li>b</li>
+    </ul>
+
+Note, results can also be :
+"a, b, c" or "b, a, c" or "b, c, a" or "c, a, b" or "c, b, a".
+
+Example 2:
+
+.. code-block:: html+twig
+
+    {% set items = {
+        'a': 'd',
+        'b': 'e',
+        'c': 'f',
+    } %}
+
+    <ul>
+        {% for index, item in items|shuffle %}
+            <li>{{ index }} - {{ item }}</li>
+        {% endfor %}
+    </ul>
+
+The above example will be rendered as:
+
+.. code-block:: html
+
+    <ul>
+        <li>0 - d</li>
+        <li>1 - f</li>
+        <li>2 - e</li>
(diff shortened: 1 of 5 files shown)


src/Extension/CoreExtension.php
```diff
@@ -948,6 +949,28 @@ public static function sortFilter(Environment $env, $array, $arrow = null)
         return $array;
     }
 
+    /**
+     * Shuffle an array.
+     * The function does not preverse keys.
+     *
+     * @param array|\Traversable $array
+     *
+     * @return array
```
stof: this should use a native return type instead of phpdoc as there is no BC impact about it.

src/Extension/CoreExtension.php
```diff
@@ -948,6 +949,28 @@ public static function sortFilter(Environment $env, $array, $arrow = null)
...
+     *
+     * @internal
+     */
+    public function shuffle($array)
+    {
+        if (!is_iterable($array)) {
```
stof: I would suggest using a native parameter type `iterable` here instead of a custom validation.

doc/filters/shuffle.rst
```diff
@@ -0,0 +1,76 @@
...
+
+    {% for user in users|shuffle %}
+        ...
+    {% endfor %}
+
+.. caution::
+
+    Internally, Twig uses the PHP `shuffle`_ function.
+    This function assigns new keys to the elements in array. It will remove
+    any existing keys that may have been assigned, rather than just reordering the keys.
```
xabbuh: A Twig filter that modifies a PHP data structure passed to the template does not look like a good idea to me.

stof: note that it does not modify the data passed to the template though. The shuffled array will be a copy of the original one.

src/Extension/CoreExtension.php
Sarah-eit: I'll send the correction with your information.
Thank you.

src/Extension/CoreExtension.php
Sarah-eit: I'll send the correction with your information.
Thank you.

doc/filters/shuffle.rst
xabbuh: If the implementation takes care of it, we should IMO remove this paragraph as it's then irrelevant for users of this filter.

stof: Well, what is relevant for the users is that the shuffled array does not preserve keys. So if your input had not sequential keys (where you often don't care about them when iterating) but indexed keys (using the user id for instance), it is not the case anymore after shuffling it.

Sarah-eit: I've added your explanation.
Thank you.

GromNaN (approved):

src/Extension/CoreExtension.php
```diff
@@ -948,6 +949,20 @@ public static function sortFilter(Environment $env, $array, $arrow = null)
         return $array;
     }
 
+    /**
+     * Shuffle an array.
+     * The function does not preserve keys.
+     *
+     * @internal
+     */
+    public static function arrayShuffle(iterable $array): array
```
smnandre: I think there is no function currently that enforce argument types like this... 

Is this something we want to have in the future ? Right now ? 

I feel that could lead to some unexpected behaviour, as this one would raises a PHP `TypeError` if given a string for example, whereas similar functions/filters would raise a `Twig\Error\RuntimeError`.

WDYT ?

fabpot: Most Twig filters try hard to work on everything you pass them. So, here, we might support strings for instance.

Sarah-eit: I understand both your comments.
I will make sure that this type of error does not occur.
I will modify the filter and add support for the "string" type.

tests/Fixtures/filters/shuffle.test
```diff
@@ -0,0 +1,16 @@
...
+{{ {'a': 'd', 'b': 'e', 'c': 'f'}|shuffle|join()|length }}
+{{ traversable|shuffle|join|length }}
+--DATA--
+return ['traversable' => new \ArrayObject([0 => 3, 1 => 2, 2 => 1])]
+--EXPECT--
+3
+2
+6
+3
+3
```
fabpot: These tests don't really test anything, right?

src/Extension/CoreExtension.php
```diff
@@ -222,6 +222,7 @@ public function getFilters(): array
 
             // string/array filters
             new TwigFilter('reverse', [self::class, 'reverseFilter'], ['needs_environment' => true]),
+            new TwigFilter('shuffle', [self::class, 'shuffleFilter'], ['needs_environment' => true]),
```
```suggestion
            new TwigFilter('shuffle', [self::class, 'shuffleFilter'], ['needs_charset' => true]),
```

src/Extension/CoreExtension.php
```diff
@@ -920,6 +921,38 @@ public static function reverseFilter(Environment $env, $item, $preserveKeys = fa
         return $string;
     }
 
+    /**
+     * Shuffle an array, a \Traversable instance, or a string.
+     * The function does not preserve keys.
+     *
+     * @internal
+     */
+    public static function shuffleFilter(Environment $env, mixed $item): mixed
```
```suggestion
    public static function shuffleFilter(string $charset $item): mixed
```

For `$item`, we can be more precise than `mixed` o the phpdocs.

src/Extension/CoreExtension.php
```diff
@@ -920,6 +921,38 @@ public static function reverseFilter(Environment $env, $item, $preserveKeys = fa
...
+    {
+        if (\is_string($item)) {
+            $charset = $env->getCharset();
```
fabpot: Can be removed

src/Extension/CoreExtension.php
```diff
@@ -920,6 +921,38 @@ public static function reverseFilter(Environment $env, $item, $preserveKeys = fa
         return $string;
     }
 
+    /**
+     * Shuffle an array, a \Traversable instance, or a string.
```
```suggestion
     * Shuffles an array, a \Traversable instance, or a string.
```

fabpot (changes requested):

tests/Fixtures/filters/shuffle.test
Sarah-eit: For the test, given that the result could vary, I couldn't find any other solution to check my results.
I used the "batch_with_zero_elements.test" file as a model in order to remain consistent.

src/Extension/CoreExtension.php
```diff
@@ -895,6 +896,36 @@ public static function reverse(string $charset, $item, $preserveKeys = false)
         return $string;
     }
 
+    /**
+     * Shuffles an array, a \Traversable instance, or a string.
+     * The function does not preserve keys.
+     *
+     * @internal
+     */
+    public static function shuffle(string $charset, array|\Traversable|string|null $item): mixed
```
fabpot: As Twig 3.x supports older versions of PHP, the `array|\Traversable|string|null` type hint should be a phpdoc instead.

src/Extension/CoreExtension.php
```diff
@@ -895,6 +896,36 @@ public static function reverse(string $charset, $item, $preserveKeys = false)
...
+                $item = self::convertEncoding($item, 'UTF-8', $charset);
+            }
+
+            $item = preg_split('/(?<!^)(?!$)/u', $item, -1);
+            shuffle($item);
+            $item = implode('', $item);
+
+            if ('UTF-8' !== $charset) {
+                $item = self::convertEncoding($item, $charset, 'UTF-8');
+            }
```
fabpot, line 924: Let's return here.

fabpot (changes requested):

tests/Fixtures/filters/shuffle.test
smnandre: Maybe you could check if the "items" from the value are the same after the filter

* `"abcd"|suffle` should still contain 4 letters: `a b c` and `d `
* `["ab", "cd"]|suffle` should still contain two elements: `ab`, `cd`
* ...

fabpot: Fixed

fabpot:
Thank you @Sarah-eit.

PR #3968, opened 2024-01-22
Clarify the behavior when a property/method does not exist

fabpot:
Alternative to #3964


miqrogroove:
Seems a bit wordy.  Another alternative is to continue the "if not, and if" refrain.

"If not, and if strict_variables is disabled, returns null..."

fabpot:
> Seems a bit wordy. Another alternative is to continue the "if not, and if" refrain.
> 
> "If not, and if strict_variables is disabled, returns null..."

Indeed, that's a great suggestion, updated accordingly.

PR #3969, opened 2024-01-23
[html-extra] filter classes

kbond:
A small formatting change to avoid double spaces:

```twig
class="{{ html_classes('a', false ? 'b', true ? 'c') }}"

{# before #}
class="a  c"

{# after #}
class="a c"
```

### DIFF
--- a/extra/html-extra/HtmlExtension.php
+++ b/extra/html-extra/HtmlExtension.php
@@ -108,6 +108,6 @@ final class HtmlExtension extends AbstractExtension
             }
         }
 
-        return implode(' ', array_unique($classes));
+        return implode(' ', array_unique(array_filter($classes, static function($v) { return '' !== $v; })));
     }
 }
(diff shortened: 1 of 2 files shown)


extra/html-extra/HtmlExtension.php
```diff
@@ -108,6 +108,6 @@ public static function htmlClasses(...$args): string
             }
         }
 
-        return implode(' ', array_unique($classes));
+        return implode(' ', array_unique(array_filter($classes)));
```
stof: be careful. `0` is empty in PHP and will be removed here.

stof (commented):
This PR clearly needs a description explaining what it expects doing.

extra/html-extra/HtmlExtension.php
```diff
@@ -75,7 +75,7 @@ public function dataUri(string $data, string $mime = null, array $parameters = [
             $repr .= ';'.$key.'='.rawurlencode($value);
         }
 
-        if (0 === strpos($mime, 'text/')) {
+        if (str_starts_with($mime, 'text/')) {
```
fabpot: Should be reverted

fabpot (approved):

fabpot:
Thank you @kbond.

PR #3970, opened 2024-01-23
[html-extra] rename `html_classes()` to `html_attr()`

kbond:
This function can be used for any html attribute (ie `data-action`, `data-controller`).

### DIFF
--- a/extra/html-extra/HtmlExtension.php
+++ b/extra/html-extra/HtmlExtension.php
@@ -36,7 +36,11 @@ final class HtmlExtension extends AbstractExtension
     public function getFunctions(): array
     {
         return [
-            new TwigFunction('html_classes', [self::class, 'htmlClasses']),
+            new TwigFunction('html_attr', [self::class, 'htmlAttr']),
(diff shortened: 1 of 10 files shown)


kbond:
If acceptable, I'll update the docs.

stof:
it can be used for attributes containing a space-separated list. That's not for any attribute out there. So `html_attr` is too generic for what it does (thus, using the same name than the `html_attr` escaping strategy can cause confusion)

kbond:
That's fair, maybe `html_join` or `html_attr_join`? But we'd have to configure the separator which I think would be hard (maybe not worth it).

fabpot:
If we were to rename to function, `html_attr_join` sounds good to me.

fabpot:
@kbond If you're ok with the new name, can you update this PR?

fabpot (changes requested):
Can you add a note in the main CHANGELOG file and in doc/deprecated.rst?

kbond:
I'm going to close this for now. I'm now feeling it's not worth the change.

PR #3980, opened 2024-01-25
Added details to parameters in function doc script_names

FlorianChenot:


stof:
As it looks like you add the same kind of info for each filter related to symfony/intl, it would be easier to keep them all in a single PR in my opinion.

FlorianChenot:
I will make a PR with all of them combined then

PR #3982, opened 2024-01-25
Added php function in map filters doc

FlorianChenot:


doc/filters/map.rst
```diff
@@ -32,3 +32,9 @@ Arguments
 ---------
 
 * ``arrow``: The arrow function
+
+.. note::
+
+    Internally, Twig uses the PHP `array_map`_ function.
```
GromNaN, line 38: Why is this information relevant to add to the documentation? We don't explain the implementation of all the features.

FlorianChenot: Observing the PHP functions associated with reduce and merge filters proved to be very useful for my personal learning experience. 
I think it would be advantageous to have a reference for comparison with another documentation that includes numerous practical examples. 
This would prevent redundancy in this particular documentation, especially considering the versatility of the map filter.

smnandre: Just realizing... it does not use `array_map` 

```php
/**
* @internal
*/
public static function arrayMap(Environment $env, $array, $arrow)
{
  self::checkArrowInSandbox($env, $arrow, 'map', 'filter');

  $r = [];
  foreach ($array as $k => $v) {
      $r[$k] = $arrow($v, $k);
  }

  return $r;
}
```

See [/src/Extension/CoreExtension.php#L1773 ](https://github.com/twigphp/Twig/blob/675f62d60c546a6a2590a837c8d6580093ff0a5a/src/Extension/CoreExtension.php#L1773)

FlorianChenot: It would seem yeah, I should have red the wrong part of the code

Thanks for the comment @smnandre

PR #3985, opened 2024-01-25
Added informations in format_currency and format_number docs

FlorianChenot:


doc/filters/format_currency.rst
```diff
@@ -20,25 +20,85 @@ You can pass attributes to tweak the output:
...
-* ``format_width``;
-* ``padding_position``;
-* ``secondary_grouping_size``;
-* ``significant_digits_used``;
-* ``min_significant_digits_used``;
-* ``max_significant_digits_used``;
-* ``lenient_parse``.
+* ``grouping_used``: Specifies whether to use grouping separator for thousands.
+
+.. code-block:: twig
```
stof: I'm quite sure the way this markup is written means that we don't have 1 list of options anymore but one list between each code blocks. The code blocks should be inside each list item.

doc/filters/format_currency.rst
```diff
@@ -20,25 +20,85 @@ You can pass attributes to tweak the output:
...
+
+    {# €1,234,567.89 #}
```
stof: As the formatting is locale-dependent, this is misleading IMO as you don't specify for which locale this is the result.

doc/filters/format_currency.rst
```diff
@@ -74,4 +134,16 @@ Arguments
 
 * ``currency``: The currency
 * ``attrs``: A map of attributes
-* ``locale``: The locale
+* ``locale``: The locale (only 'en' supported at the moment)
```
stof: this is not true. If you install `ext-intl`, all locales are supported. The restriction of only `en` does not come from the Twig filter. It comes from `symfony/polyfill-intl-icu` which polyfills the `NumberFormatter` but does not have the formatting rules for hundreds of locales.

doc/filters/format_currency.rst
```diff
@@ -74,4 +134,16 @@ Arguments
...
+
+.. note::
+
+    You can find the comprehensive list of available currencies for this filter at the following link:
+
+    https://github.com/symfony/polyfill-intl-icu/blob/master/Resources/currencies.php
+
+.. note::
+
+    Internally, Twig uses the PHP `NumberFormatter::formatCurrency or numfmt_format_currency`_ function.
```
stof:
```suggestion
    Internally, Twig uses the PHP `NumberFormatter::formatCurrency`_ function.
```

doc/filters/format_currency.rst
```diff
@@ -74,4 +134,16 @@ Arguments
...
+
+.. _`numfmt_format_currency`: https://www.php.net/manual/en/numberformatter.formatcurrency.php
```
stof: this is a broken link as it is not the one used above

doc/filters/format_currency.rst
```diff
@@ -74,4 +134,16 @@ Arguments
...
 * ``currency``: The currency
 * ``attrs``: A map of attributes
-* ``locale``: The locale
+* ``locale``: The locale (only 'en' supported at the moment)
+
+.. note::
+
+    You can find the comprehensive list of available currencies for this filter at the following link:
+
+    https://github.com/symfony/polyfill-intl-icu/blob/master/Resources/currencies.php
```
stof: I would not use this file as link. The NumberFormatter API accepts currencies as the 3 letter code of ISO 4217, which is a standard. This is what we should document.

stof (changes requested):

FlorianChenot:
I added the format_number filter in this PR because it is really close to format_currency

doc/filters/format_currency.rst
```diff
@@ -1,7 +1,7 @@
 ``format_currency``
 ===================
 
-The ``format_currency`` filter formats a number as a currency:
+The ``format_currency`` filter formats a number (float) as a currency:
```
smnandre: No, passing an integer is totally valid.

* https://www.php.net/manual/fr/numberformatter.format.php
* https://github.com/twigphp/intl-extra/blob/8a953b1d8d8be4bef57508c4b825d508fcca6efe/IntlExtension.php#L335

doc/filters/format_currency.rst
```diff
@@ -20,25 +20,75 @@ You can pass attributes to tweak the output:
...
+
+        {# €1,23,45,678.00 #}
+        {{ 12345678 | format_currency('EUR', {grouping_size:3, secondary_grouping_size:2}, 'en') }}
+
+* ``rounding_mode``:
+* ``rounding_increment``: Control rounding behavior.
+    .. code-block:: twig
+
+        {# €123.50 #}
+        {{ 123.456 | format_currency('EUR', {rounding_mode:'ceiling', rounding_increment:0.05}, 'en') }}
```
smnandre, line 77: Maybe list the different modes available ?

doc/filters/format_currency.rst
```diff
@@ -20,25 +20,75 @@ You can pass attributes to tweak the output:
...
+
+* ``format_width``:
+* ``padding_position``: Set width and padding for the formatted number.
+    .. code-block:: twig
+
+        {# €123.00 #}
+        {{ 123 | format_currency('EUR', {format_width:10, padding_position:'before_suffix'}, 'en') }}
```
smnandre, line 89: Maybe list the different `padding_position` available ?

doc/filters/format_currency.rst
```diff
@@ -20,25 +20,75 @@ You can pass attributes to tweak the output:
...
+
+        {# €123.4568 #}
+        {{ 123.456789 | format_currency('EUR', {significant_digits_used:true, min_significant_digits_used:4, max_significant_digits_used:7}, 'en') }}
+
+
+* ``lenient_parse``: If true, allows lenient parsing of the input.
+    .. code-block:: twig
+
+        {# €123.00 #}
+        {{ 123 | format_currency('EUR', {lenient_parse:true}, 'en') }}
```
smnandre, line 104: Could be more descriptive of the lenient parsing effect

doc/filters/format_currency.rst
FlorianChenot: You're right, I changed it

doc/filters/format_currency.rst
FlorianChenot: I added it

doc/filters/format_currency.rst
FlorianChenot: I added it too

doc/filters/format_currency.rst
FlorianChenot: I couldn't find a useful example, do you have any ideas ?

smnandre: I'm not sure here, to be honest. As i see it, the "lenient_parse" is a way to allow data that is not perfectly in the expected syntax... 


```php
<?php

$formatter = NumberFormatter::create('fr-FR', NumberFormatter::CURRENCY);

foreach ([false, true] as $lenientParse) {

    echo 'lenientParse: ' . ($lenientParse ? 'true' : 'false') . PHP_EOL;
    $formatter->setAttribute(NumberFormatter::LENIENT_PARSE, $lenientParse);

    echo $formatter->format(-9) . PHP_EOL;
    echo $formatter->format('.9') . PHP_EOL;
    echo $formatter->format(12.34000) . PHP_EOL;

    echo $formatter->format($formatter->parse('56,7800000')) . PHP_EOL;
    echo $formatter->format($formatter->parse('56,78')) . PHP_EOL;
    echo $formatter->format($formatter->parse('56,78€')) . PHP_EOL;
    echo $formatter->format($formatter->parse('56,78 €')) . PHP_EOL;
    echo $formatter->format($formatter->parse('56,78 €')) . PHP_EOL;
}
```

Would render 

```diff
lenientParse: false
 -9,00 €
0,90 €
12,34 €
-0,00 €
-0,00 €
-0,00 €
-0,00 €
56,78 €

lenientParse: true
 -9,00 €
0,90 €
12,34 €
+56,78 €
+56,78 €
+56,78 €
+56,78 €
56,78 €
```


As you see, it's when the `parse` method is called it makes a difference.

But it this specific case (currency), as we do not call "`parse`", and the `format` method only accept `int|float`, i don't see a concrete exemple where it would have an impact.

fabpot:
Thank you @FlorianChenot.

Issue #3951, opened 2023-12-22
What would you like to see as major changes that would lead to a BC break for the next major release of Twig?


fabpot:
We've started working on the next major Twig release. As announced last year, this new version will probably land under the Symfony mono-repository.

For the first time in Twig history, I'd like to fix the major issues that would be BC breaks and that we've never done.

This issue acts as a community wish list for things you want to see in the next major version that could lead to BC breaks (hint: like operators precedence).

Feel free to comment and vote.

nsetyo:
I really like the [askama](https://github.com/djc/askama) approach to make template type safe, and a little out of topic, it would be great if twig had an up-to-date formatter (maybe with a prettier plugin maintained by core team)

GromNaN:
I like that you started removing usage of `echo` and `ob_*` in [#3950](https://github.com/twigphp/Twig/pull/3950). That's a breaking change if some extension use `echo` instead of returning a string.

This will allow concurrent Twig renderings using fibers.

aszenz:
I would like to see a focus on component based templating over overriding/inheritance. 

For that macros need an improvement specifically the need to import them separately in every block.

Would be great to support ux html component syntax natively for creating components.

A type checker plugin (pslam/phpstan) would be nice to have as well.

TheCelavi:
Support for step debugging via source maps would be nice.

JoshuaBehrens:
Plugin / Theme Structure like Shopware built around Twig. We had to built something as well and it feels bad to just copy if this could have been solved directly. And twig is already strong in multiple inheritance.

JoshuaBehrens:
Static type support for generated php to increase performance and predictable execution. I already tried to approach it but optimisation levels were commented as too breaking. So we could break it here. https://github.com/twigphp/Twig/pull/3928

JoshuaBehrens:
Still be able to use Twig without pulling a lot of more dependencies.

E.g when you merge the twig symfony bridge into the twig package and in a small project just want to get some templating, you don't get the whole dependency clutter from the framework bundle. I don't use laravel/blade because you always get so much laravel stuff with it. I just want something small that can do inheritance and autoescape. Otherwise I would just use php right away.

I see benefits in merging some code from expression language as twig has to parse expressions as well and has "just" some nodes on top.

75th:
[...]

Issue #2897, opened 2019-03-13
Feature Request: Allow nested comments. Suggestion: `{##`


ThomasLandauer:
It would be nice to have a quick way to comment out some lines of code that already contain comments:
```twig
Line1 {# comment #}
Line2 {# comment #}
Line3 {# comment #}
Line4 {# comment #}
Line5 {# comment #}
```
One way to achieve this would be one-line comments (like PHP's `//`) - this has been suggested before, see https://github.com/twigphp/Twig/issues/1331

Another (more consistent with the existing Twig syntax) way would be to introduce "second-level" comments. What about `{##`? This could ignore `#}` and only get closed by `##}`.

There are IDE keyboard shortcuts available for this feature in PHP and YAML (by prepending `//` or `#` to each line). Just Twig is missing.

ericmorand:
It think this is simply a bug of the lexer. `{# {# #} #}` should be lexed to a comment token containing `{# #}` - and thus be ignored since the current lexer simply ignore comment tokens.

The same issue applies to the `verbatim `tag:

https://twigfiddle.com/a3nn16

SpacePossum:
it would be nice if these comments do not "eat up" the line break,
so 

```
test  {## foo ##}
bar
```

renders
```
test
bar
```

and not
```
test bar
```
https://twigfiddle.com/srb8mr

stof:
@ericmorand the whole point of comments and verbatim is that they are **not** lexing what is inside them. Changing that would be a big BC break.

And btw, the same is true for comments in all other languages I know about. PHP, JS, CSS, HTML all use the first closing marker for the comment for instance instance, and the list can go on with more languages.

ericmorand:
@stof , I agree totally. This is the way to go. As you say, this is how it works with other languages and this is how this should work with Twig too.

About verbatim, I think this is debatable. There is a real need to be able to display raw verbatim syntaxt inside a verbatim block. What do you think.

7ochem:
If I comment out Twig lines in PhpStorm with <kbd>Ctrl</kbd>+<kbd>/</kbd>, then it takes existing comments into account:

```
    <div>{{ some_var }}</div>
    <div>{# comment #}{{ another_var }}</div>
```
<kbd>Ctrl</kbd>+<kbd>/</kbd> :arrow_down: 
```
{#    <div>{{ some_var }}</div>#}
{#    <div>#}{# comment #}{#{{ another_var }}</div>#}
```
[...]

PR #3986, opened 2024-01-26
[RFC] Multiline {## Comment Block ##} 

smnandre:
I started to work on this feature, following @ThomasLandauer  comment in #3951 (and previously discussed in #2897)

So this PR is a starting point to discuss implementation, any particularities, specific behaviours, etc.

The idea is that anything between  `{##` and `##}`  would be parsed as a comment (and thus ignored)

```twig
Hello {## foo bar 
Foo
Bar {# Bar #}
Foo ##} Hello
```

would render

```html
Hello  Hello
```


tests/LexerTest.php
```diff
@@ -131,6 +131,33 @@ public function testLongComments()
...
+        $template = '{## {# '.str_repeat('*', 100000).' #} ##}';
+
+        $lexer = new Lexer(new Environment($this->createMock(LoaderInterface::class)));
+
+        $res = $lexer->tokenize(new Source($template, 'index'));
+        $this->assertTrue($res->isEOF());
+
+        // add a dummy assertion here to satisfy PHPUnit, the only thing we want to test is that the code above
+        // can be executed without throwing any exceptions
+        $this->addToAssertionCount(1);
```
stof: no need for a dummy assertion here, as there is an actual assertion just above.

tests/LexerTest.php
```diff
@@ -178,6 +205,49 @@ public function testBigNumbers()
...
+            ."baz\n"
+            ."}}\n##}\n";
+
+        $lexer = new Lexer(new Environment($this->createMock(LoaderInterface::class)));
+        $stream = $lexer->tokenize(new Source($template, 'index'));
+
+        $this->assertTrue($stream->isEOF());
+    }
+
+    public function testMultiLineCommentNestedBlock()
```
stof: this test does not seem to have any nested block.

tests/LexerTest.php
```diff
@@ -178,6 +205,49 @@ public function testBigNumbers()
...
+
+        $lexer = new Lexer(new Environment($this->createMock(LoaderInterface::class)));
+        $stream = $lexer->tokenize(new Source($template, 'index'));
+
+        $stream->test(Token::TEXT_TYPE, 'baz');
+        $stream->next();
+        $this->assertTrue($stream->isEOF());
+    }
+
+    public function testMultiLineCommentBlockWithInlineBLocks()
```
stof, line 231: this seems to be about inline comments

stof (changes requested):

smnandre:
Thank you for this first feedback @stof

smnandre:
Is this something really needed / asked for ? 

Just to be sure it is worth the time needed to refine this PR, add documentation, think how to handle BC, ....

fabpot:
As far as I know, this syntax is not supported by any other language similar to Twig (Jinja and the like), so I'm not sure if the use cases are worth the added complexity in core.

smnandre:
Well let's close then :)

c33s:
i would have been happy to see this feature finally. it helps to easily comment out blocks of text which have comments inside and preserving them after you uncomment the block again.

PR #3988, opened 2024-02-01
add details to html_classes function

FlorianChenot:


doc/functions/html_classes.rst
```diff
@@ -12,6 +12,14 @@ names together:
         'pending': object.pending,
     }) }}">How are you doing?</p>
 
+For this example, knowing that object.errored and object.pending are true, this will render:
+
+.. code-block:: html+twig
+
+    <p class="a-class another-class errored pending">
+        How are you doing?
```
drjayvee, line 20: Super pedantic review from a complete stranger here... 

The output isn't quite correct, since there will be no white space between around "How are you doing?"

I doubt anyone would get confused, but maybe change the doc to
> [...] this will render (white space added for clarity):

fabpot:
Closing as there is no more activity.

PR #3989, opened 2024-02-05
add details to parameters array in data_uri filter

FlorianChenot:


doc/filters/data_uri.rst
```diff
@@ -16,7 +16,7 @@ The ``data_uri`` filter generates a URL using the data scheme as defined in
     {# also works with plain text #}
     {{ '<b>foobar</b>'|data_uri(mime="text/html") }}
 
-    {# add some extra parameters #}
+    {# we can add any extra parameters we could put in an url #}
```
smnandre: I feel the "we" is more used for internal / algorithm explanations, and "you" to illustrate user need / usages... wdyt ?  

(not as binary in fact, but i feel this is the general idea)

FlorianChenot: You're right, I changed it

fabpot:
I don't think this is worth it.

PR #3991, opened 2024-02-05
Fix some errors reported by phpstan

fabpot:


stof:
@fabpot should phpstan be added in the CI setup ?

PR #3992, opened 2024-02-06
Implement YieldingTemplate::display*Block()

nicolas-grekas:
This should make the migration a bit easier (Symfony uses displayBlock() directly in the Form component.)

### DIFF
--- a/src/YieldingTemplate.php
+++ b/src/YieldingTemplate.php
@@ -167,12 +167,16 @@ abstract class YieldingTemplate extends Template
 
     public function displayBlock($name, array $context, array $blocks = [], $useBlocks = true, Template $templateContext = null)
     {
-        throw new RuntimeError(sprintf('Calling "%s" for block "%s" is not supported as "use_yield" is set to "true".', __METHOD__, $name), -1, $this->getSourceContext());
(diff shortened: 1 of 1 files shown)


GromNaN:
How that would make the migration easier?

The [call to `displayBlock` in Symfony Form](https://github.com/symfony/symfony/blob/dfe4aa54c5f6e2166dce3ca80c280dddf42a640a/src/Symfony/Bridge/Twig/Form/TwigRendererEngine.php#L39-L50) is surrounded by `ob_start`/`ob_get_clean`. We can fix the calling code in Symfony 5.4+ to check if Twig is using the yielding mode (`method_exists($template, 'yieldBlock')`) and call it instead.

nicolas-grekas:
This improves the migration by not breaking ppl that will update twig but not symfony. This will happen during the transition period.

stof (approved):

GromNaN:
If there is a conflict rule in twig for `symfony/form < 5.4.36` and the fix is in the new version bugfix, that would prevent issues.

If this PR is merged, it deserves a deprecation warning as this methods are removed from 4.0.

nicolas-grekas:
twig shouldn't know about symfony/form - that's be entering the package coupling arena
the deprecation is possible but not worth it IMHO

nicolas-grekas:
Closing in favor of #3999

PR #3993, opened 2024-02-07
fix TwigExtraBundle tests with Symfony 6.0/6.1

xabbuh:


extra/twig-extra-bundle/Tests/Fixture/Kernel.php
```diff
@@ -30,7 +31,7 @@ protected function configureContainer(ContainerBuilder $c, LoaderInterface $load
             'router' => ['utf8' => true],
             'http_method_override' => false,
         ];
-        if (6 === Kernel::MAJOR_VERSION) {
+        if (trait_exists(NotificationAssertionsTrait::class)) {
```
fabpot, line 39: Can we add a comment about what we're trying to achieve here (so that we can remove the check when we change the minimum Symfony supported version)?

xabbuh: done, now we need to find a way to install the String component and Translation contracts in compatible versions

fabpot:
Thank you @xabbuh.

PR #3999, opened 2024-02-12
Add `#[YieldReady]` to allow extensions to tell when they're ready for yielding

nicolas-grekas:
The current opt-in based on "use_yield" doesn't work, because there is no decider to enable the flag:
the flag can be turned on only when all extensions are compatible, and this is fragile, because extensions currently don't have a way to signal that they are ready for yielding.

This PR takes the following approach instead:
- Between each "yield", the output buffer is checked and emptied.
- Extensions need to label their node implementations with `#[YieldReady]`. If they don't, a deprecation is raised.

On Twig 4, we could then just remove this output buffer check and possibly throw when a node is not "YieldReady".

(Classes that extend `AbstractExpression` are implicitly considered "YieldReady".)

Best reviewed [ignoring whitespaces](https://github.com/twigphp/Twig/pull/3999/files?w=1).


src/Node/BlockNode.php
```diff
@@ -37,14 +38,7 @@ public function compile(Compiler $compiler): void
...
             ->subcompile($this->getNode('body'))
-        ;
-
-        if (!$this->getNode('body') instanceof NodeOutputInterface && $compiler->getEnvironment()->useYield()) {
-            // needed when body doesn't yield anything
-            $compiler->write("yield '';\n");
-        }
-
-        $compiler
+            ->write("return; yield '';\n") // needed when body doesn't yield anything
```
nicolas-grekas, line 42: Not all nodes that do generate some content have the `NodeOutputInterface`. This does the trick of forcing an empty generator without any overhead.

src/Template.php
```diff
@@ -371,36 +313,46 @@ public function getBlocks()
...
+        $context = $this->env->mergeGlobals($context);
+        $blocks = array_merge($this->blocks, $blocks);
+        $level = ob_get_level();
+        ob_start();
+
         try {
-            $this->doDisplay($context, $blocks);
+            foreach ($this->doDisplay($context, $blocks) as $data) {
+                if (ob_get_length()) {
+                    $data = ob_get_clean().$data;
```
nicolas-grekas, line 353: This is the meat of the PR

src/Node/YieldReady.php
```diff
@@ -0,0 +1,20 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Twig\Node;
+
+/**
+ * Marks nodes that are ready for using "yield" instead of "echo" or "print()" for rendering.
+ */
+#[\Attribute(\Attribute::TARGET_CLASS)]
+class YieldReady
```
stof: should this class really be in the `Node` namespace ? This namespace only contains nodes until now AFAICT.

src/Compiler.php
```diff
@@ -239,13 +220,4 @@ public function getVarName(): string
     {
         return sprintf('__internal_compile_%d', $this->varNameSalt++);
     }
-
-    private function checkStringForOutput(string $string): void
```
stof, line 243: shouldn't we keep this deprecation ? It helps finding places where you are not yield ready. I fear that packages might mark them nodes as being yield ready without noticing they are not.

stof (commented):
What is the performance impact of the runtime check combined with yield ?

src/Node/BlockNode.php
stof: nodes generating output without implementing NodeOutputInterface should also be deprecated though, as some other parts of Twig rely on this interface. Not sure how to handle that.

src/Template.php
stof: Please add tests covering this behavior though. I don't think there is any right now.

src/NodeVisitor/YieldNotReadyNodeVisitor.php
```diff
@@ -0,0 +1,75 @@
...
+use Twig\Node\TextNode;
+use Twig\Node\YieldReady;
+
+/**
+ * Tries to optimize the AST.
+ *
+ * This visitor is always the last registered one.
+ *
+ * You can configure which optimizations you want to activate via the
+ * optimizer mode.
```
jdreesen: Copy/paste?

stof:
Note that an effect of this change is that support for Fiber-based runtimes will become possible only as of Twig 4, even if all extensions are yield ready, as there is no way anymore to opt-in for the yield-only mode that unlocks that usage.

src/NodeVisitor/YieldNotReadyNodeVisitor.php
nicolas-grekas: oops :)

src/Node/BlockNode.php
nicolas-grekas: I don't see a way to force node implementations to have this interface if they want to output something. The only side-effect of implementing the interface is forbidding output-enabled nodes from being used outside of blocks, so it's not a critical feature either.

src/Node/YieldReady.php
nicolas-grekas: It fits there to me. Of course let me know if you have a better idea.

src/Compiler.php
nicolas-grekas: on it

nicolas-grekas:
> What is the performance impact of the runtime check combined with yield ?

I did a micro-benchmark comparing the time it takes to call ob_get_length() vs the time it takes to call an empty `function c() {}`.
I complete 10M iterations in 180ms for the ob function vs 210ms for the empty one.
This should be fine.

> Note that an effect of this change is that support for Fiber-based runtimes will become possible only as of Twig 4, even if all extensions are yield ready, as there is no way anymore to opt-in for the yield-only mode that unlocks that usage.

I improved this by adding the "use_yield" flag back. When it's turned on, the node visitor will throw if any nodes are not `#[YieldReady]`, and the Template will skip using any ob functions so that things are at full speed.

src/Template.php
nicolas-grekas: Test added

nicolas-grekas:
PR ready (failures are false positives)

src/Environment.php
```diff
@@ -133,6 +130,11 @@ public function __construct(LoaderInterface $loader, $options = [])
 
         $this->addExtension(new CoreExtension());
         $this->addExtension(new EscaperExtension($options['autoescape']));
+        if (\PHP_VERSION_ID >= 80000) {
+            $this->addExtension(new YieldNotReadyExtension($this->useYield));
+        } elseif ($this->useYield) {
+            throw new \LogicException('The "use_yield" option of Twig can only be enabled when running on PHP 8.0 or later.');
```
stof: I suggest updating the phpdoc to describe this requirement when documenting the option.

src/Node/CaptureNode.php
```diff
@@ -27,62 +28,25 @@ public function __construct(Node $body, int $lineno, ?string $tag = null)
...
         if ($this->getAttribute('with_blocks')) {
             $compiler->raw("(function () use (&\$context, \$macros, \$blocks) {\n");
         } else {
             $compiler->raw("(function () use (&\$context, \$macros) {\n");
         }
-        $compiler->indent();
-        if ($compiler->getEnvironment()->isDebug()) {
-            $compiler->write("ob_start();\n");
-        } else {
-            $compiler->write("ob_start(function () { return ''; });\n");
```
stof, line 64: When not in yield-only mode, don't we need to handle the output buffer in this capture node to ensure we are properly capturing everything including legacy output nodes ?

src/Node/YieldReady.php
stof: I suggest putting it in the `Twig\Attribute` namespace instead, which will also be used by https://github.com/twigphp/Twig/pull/3916

stof: this class should be final as the code reading the attribute does not currently support child classes (and there is no reasons to extend it anyway as it has no state)

src/NodeVisitor/YieldNotReadyNodeVisitor.php
```diff
@@ -0,0 +1,60 @@
...
+use Twig\Node\Node;
+use Twig\Node\YieldReady;
+
+/**
+ * @internal to be removed in Twig 4
+ */
+final class YieldNotReadyNodeVisitor implements NodeVisitorInterface
+{
+    private $useYield;
+    private $yieldReadyNodes;
```
stof:
```suggestion
    private $yieldReadyNodes = [];
```

src/NodeVisitor/YieldNotReadyNodeVisitor.php
```diff
@@ -0,0 +1,60 @@
...
+
+    public function enterNode(Node $node, Environment $env): Node
+    {
+        $class = \get_class($node);
+
+        if ($node instanceof AbstractExpression || isset($this->yieldReadyNodes[$class])) {
+            return $node;
+        }
+
+        if (!$this->yieldReadyNodes[$class] = (new \ReflectionClass($class))->getAttributes(YieldReady::class)) {
```
stof: should we really store the array of ReflectionAttribute instances there ? I suggest storing only a boolean in the array.

src/Environment.php
nicolas-grekas: I removed the PHP 8 requirement: it exists just because attributes can be read only on PHP 8, but yield works will on PHP 7, so it's fine to skip the check. We can expect ppl that write new code for PHP 7 to also run it on PHP 8 so their code will be checked anyway.

src/Node/CaptureNode.php
nicolas-grekas: indeed, updated

nicolas-grekas:
All comments addressed.

src/Environment.php
```diff
@@ -854,7 +854,6 @@ private function updateOptionsHash(): void
             self::VERSION,
             (int) $this->debug,
             (int) $this->strictVariables,
-            $this->useYield ? '1' : '0',
```
stof: This should be kept as the compiled template is still different in yield mode so we need to invalidate the cache based on this option.

nicolas-grekas: Because of the CaptureNode indeed. Updated.

stof: technically, also because of the node visitor switching between a deprecation warning and an error (as you would not get the error after enabling the option if the cache was not invalidated)

stof (approved):

src/Node/Node.php
```diff
@@ -82,7 +84,7 @@ public function __toString()
     public function compile(Compiler $compiler)
     {
         foreach ($this->nodes as $node) {
-            $node->compile($compiler);
+            $compiler->subcompile($node);
```
fabpot, line 87: Should be reverted

src/Template.php
```diff
@@ -210,14 +166,12 @@ public function displayBlock($name, array $context, array $blocks = [], $useBloc
      */
     public function renderParentBlock($name, array $context, array $blocks = [])
     {
-        if ($this->env->isDebug()) {
-            ob_start();
-        } else {
-            ob_start(function () { return ''; });
```
fabpot, line 216: You've removed this "feature". I'm fine as we won't be able to replicate it with yield, but that should be noted in the CHANGELOG.

fabpot (changes requested):

src/Template.php
nicolas-grekas: I've read https://github.com/twigphp/Twig/issues/1962 and https://github.com/twigphp/Twig/issues/3058 and I think this is fine now: both issues are handled by using yield instead of output buffering. I'd keep things as is.

src/Node/Node.php
nicolas-grekas: We'd better not: this line allows the compiler to accurately tell which node is leaking via "echo". If I remove it, the error/deprecation message will report about the wrong node (the wrapping one).
Can't we keep this line? Why if not? Calling subcompile instead of compile looks OK to me also, semantically speaking.

src/Template.php
nicolas-grekas: I updated the changelog to tell about what this PR does, but about this specific behavior, I can't figure out what to write. It's too complex to describe to really be a "feature".

fabpot:
Thank you @nicolas-grekas.

src/Node/Node.php
stof: note that `$compiler->subcompile($node);` is already calling `$node->compile($compiler);`

src/Template.php
stof: Maybe CoreExtension should keep that logic in its `captureOutput`. The issue of not leaking output buffers in case of fatal errors still exists when this BC layer is used.

fabpot:
Performance regression fixed in #4216

Issue #3714, opened 2022-06-12
Documented precedence for binary operators does not match actual behavior


nox-4O4:
The [documentation](https://github.com/twigphp/Twig/blob/b4d6723715da57667cca851051eba3786714290d/doc/templates.rst#expressions) states the following:

> The operator precedence is as follows, with the lowest-precedence operators listed first: ?: (ternary operator), b-and, b-xor, b-or, or, and, ==, !=, <=>, <, >, >=, <=, in, matches, starts with, ends with, .., +, -, ~, *, /, //, %, is (tests), **, ??, | (filters), [], and .

When compiling following template:
```twig
{{ 6 b-and 2 or 6 b-and 16 }}
```

This PHP code [gets generated](https://twigfiddle.com/8dupdt):

```php
echo twig_escape_filter($this->env, ((6 & 2) || (6 & 16)), "html", null, true);
```

If `b-and` had indeed had lower precedence than `or`, the result would have been `((6 & (2 || 6)) & 16)`.

The documentation should be adjusted to reflect current behavior.

lacatoire:
@nox-4O4 I proposed a PR to answer it, put it in the right order and added some additional information

nox-4O4:
Thanks for clarifying the current behavior in the documentation! I added some review comments to your PR.

PR #4004, opened 2024-02-29
[Doc] Correct Order & Explanation on precedence

lacatoire:
Following this [issue](https://github.com/twigphp/Twig/issues/3714), i worked on it trying to make it simple and easy to understand

Indeed order of the precedence was not good for ``b-and``, ``b-xor`` and  ``b-and``. 
Code [here ](https://github.com/twigphp/Twig/blob/3.x/src/Extension/CoreExtension.php#L280) to understand. I also added additional explanations.

### DIFF
--- a/doc/templates.rst
+++ b/doc/templates.rst
@@ -514,13 +514,40 @@ Twig allows expressions everywhere.
 
 .. note::
 
+    Twig uses operators to perform various operations within templates. Understanding
+    the precedence of these operators is crucial for writing correct and efficient Twig templates.
+
     The operator precedence is as follows, with the lowest-precedence operators
-    listed first: ``?:`` (ternary operator), ``b-and``, ``b-xor``, ``b-or``,
-    ``or``, ``and``, ``==``, ``!=``, ``<=>``, ``<``, ``>``, ``>=``, ``<=``,
-    ``in``, ``matches``, ``starts with``, ``ends with``, ``has every``, ``has
-    some``, ``..``, ``+``, ``-``,
-    ``~``, ``*``, ``/``, ``//``, ``%``, ``is`` (tests), ``**``, ``??``, ``|``
-    (filters), ``[]``, and ``.``:
+    listed first:
+
+    =============================  =================================== ===================================================
+    Operator                       Score of precedence                 Description
+    =============================  =================================== ===================================================
+    ``?:``                                                             Perfoms a ternary, conditional statement.
+    ``or``                         10                                  Performs a logical OR operation between two
+                                                                       boolean expressions.
+    ``and``                        15                                  Performs a logical AND operation between two
+                                                                       boolean expressions.
+    ``b-or``                       16                                  Performs a bitwise OR operation on integers.
+    ``b-xor``                      17                                  Performs a bitwise XOR operation on integers.
+    ``b-and``                      18                                  Performs a bitwise AND operation on integers
+    ``==``, ``!=``, ``<=>``,       20                                  Comparison Operators: Compare values and check
+    ``<``, ``>``, ``>=``,                                              for containment, pattern matching, etc.
+    ``<=``, ``not in``, ``in``,
+    ``matches``, ``starts with``,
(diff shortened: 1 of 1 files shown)


doc/templates.rst
```diff
@@ -514,13 +514,41 @@ Twig allows expressions everywhere.
 
 .. note::
 
+    Twig uses operators to perform various operations within templates. Understanding
+    the precedence of these operators is crucial for writing correct and efficient Twig templates.
+
     The operator precedence is as follows, with the lowest-precedence operators
-    listed first: ``?:`` (ternary operator), ``b-and``, ``b-xor``, ``b-or``,
```
smnandre, line 518: What about ``?:`` .. ?  :)

doc/templates.rst
```diff
@@ -514,13 +514,41 @@ Twig allows expressions everywhere.
...
+    ``ends with``, ``has some``,
+    ``has every``
+    ``..``                       25                                  Creates a range of values, commonly used in loops.
+    ``+``, ``-``                 30                                  Performs operations on numbers.
+    ``~``                        40                                  Concatenates strings together.
+    ``*``, ``/``, ``//``, ``%``  60                                  Handles arithmetic operations on numbers.
+    ``is``, ``is not`` (tests)   100
+    ``**``                       200                                 Raises a number to the power of another.
+    ``??``                       300                                 Handles cases where a variable might be null.
+    ``|``,``[]``,``.``           300                                 Filters comes last
```
nox-4O4: Highest precedence means they come first, not last.
(You can see this behavior using following snippet: `{{"a" ~ "b" | upper}}`).

So, a better description might be `Filters are evaluated first.`

doc/templates.rst
nox-4O4: The ternary operator has lowest precedence, even lower than `or`. (See for example the output of `{{ 3 ?: 4 or 5 }}` which is `3`, as `or` is part of the ternary expression (and not the other way around).)

The operator is currently missing and should be added to the table.

doc/templates.rst
```diff
@@ -514,13 +514,41 @@ Twig allows expressions everywhere.
...
-    listed first: ``?:`` (ternary operator), ``b-and``, ``b-xor``, ``b-or``,
-    ``or``, ``and``, ``==``, ``!=``, ``<=>``, ``<``, ``>``, ``>=``, ``<=``,
-    ``in``, ``matches``, ``starts with``, ``ends with``, ``has every``, ``has
-    some``, ``..``, ``+``, ``-``,
-    ``~``, ``*``, ``/``, ``//``, ``%``, ``is`` (tests), ``**``, ``??``, ``|``
-    (filters), ``[]``, and ``.``:
+    listed first:
+
+    ===========================  =================================== ===================================================
+    Operator                     Score of precedence                 Description
```
nox-4O4: There might be something off with the markup, the table does not render when using the "rich diff" option in GitHub. Though, perhaps this is only a limitation of GitHub?

doc/templates.rst
```diff
@@ -514,13 +514,41 @@ Twig allows expressions everywhere.
...
+    ``*``, ``/``, ``//``, ``%``  60                                  Handles arithmetic operations on numbers.
+    ``is``, ``is not`` (tests)   100
+    ``**``                       200                                 Raises a number to the power of another.
+    ``??``                       300                                 Handles cases where a variable might be null.
+    ``|``,``[]``,``.``           300                                 Filters comes last
+    ===========================  =================================== ===================================================
+
+    .. code-block:: twig
+
+        This mean that {{ 6 b-and 2 or 6 b-and 16 }} render (6 & 2) || (6 & 16)
```
nox-4O4: This part should not be marked as twig code block.
Rather, inline code is more appropriate: 

```
This means that ``{{ 6 b-and 2 or 6 b-and 16 }}`` results in ``(6 & 2) || (6 & 16)``.
```

fabpot:
Thank you @lacpandore.

Issue #4003, opened 2024-02-29
Add static code analysis and/or linting to improve code quality


drjayvee:
I'd love to use static code analysis and/or linting for our logic-heavy templates to prevent silly mistakes from breaking production. (Asking for a friend 😉)

Compiling templates catches syntax errors, at least. But there are other types of bugs which aren't detected this way.

For example, suppose that a macro uses an undeclared variable (i.e., not a macro argument or declared using `{% set %}`). This is almost certainly a mistake. `strict_variables` will throw an Exception, of course, but only if the code is executed. Without `strict_variables`, Twig will use `null`, but that can easily lead to an Exception when passing it as a non-nullable argument.

I searched for separate linters, but only found tools that either focus on HTML ([curlylint](https://www.curlylint.org/) and [djLint](https://www.djlint.com/docs/linter/)) or are deprecated ([phpstan-twig-rules](https://packagist.org/packages/symplify/phpstan-twig-rules) and [twig-lint](https://github.com/asm89/twig-lint)).

I see two ways of increasing code quality:
1. Separate linter CLI, either included in the main `twig/twig` package or separately
2. An extension which adds a token parser (or node visitor?) and throws exceptions at compile-time

stof:
Static analysis is definitely much more involved than a linter. And this is a whole project on its own. The current Twig team (which is a way of saying "Fabien Potencier" actually, with a bit a help from Nicolas and me for triaging issues) does not have the resources of creating and maintaining a static analysis tool. So it would have to be created by someone else (ready to invest quite some time in such a project).

drjayvee:
Bonjour stof,

Thanks for getting back to me so quickly. I understand this is out of scope for the core maintainer team.

I started writing an extension with a NodeVisitor which already showed me a few issues in our code. I am lacking some knowledge on the internals and could use some help though.

I'll probably open a repo to develop this extension tomorrow. I'll post a link here, of course. Could you help raise some awareness so this project can get some contributions? I'll reach out in our company's network to raise awareness.

drjayvee:
Hey stof and @fabpot,

I've created an extension as discussed above: [TwigQI](https://github.com/alisqi/TwigQI/) 🎉

It already has two working inspections, one of which will forever prevent the very mistake that sparked my interest in static code analysis for Twig templates: use of undeclared/undefined variables inside `macro`s.

One prerequisite for most other envisioned inspections is for the parser to add comment nodes to the AST. I've implemented (and motivated) this in #4009.

[...]

PR #4009, opened 2024-03-16
Add CommentNode to the AST

drjayvee:
These nodes have no effect on compiled templates.

Adding these to the TokenStream and AST enables extensions to add logic in node visitors, such as parsing PHPDoc-style `@var string name` comments.

This in turn is invaluable for static code analysis of Twig templates as briefly discussed in [#4003](https://github.com/twigphp/Twig/issues/4003) and in more detail in [TwigStan](https://github.com/alisqi/twig-stan/)'s README.

### DIFF
--- a/src/Lexer.php
+++ b/src/Lexer.php
@@ -412,7 +412,7 @@ class Lexer
             throw new SyntaxError('Unclosed comment.', $this->lineno, $this->source);
         }
 
-        $text = substr($this->code, $this->cursor, $match[0][1] - $this->cursor) . $match[0][0];
+        $text = substr($this->code, $this->cursor, $match[0][1] - $this->cursor).$match[0][0];
         $this->pushToken(/* Token::COMMENT_TYPE */ 14, trim(substr($text, 0, strrpos($text, '#}'))));
         $this->moveCursor($text);
     }
--- a/src/Lexer.php
+++ b/src/Lexer.php
@@ -412,7 +412,9 @@ class Lexer
             throw new SyntaxError('Unclosed comment.', $this->lineno, $this->source);
         }
 
-        $this->moveCursor(substr($this->code, $this->cursor, $match[0][1] - $this->cursor).$match[0][0]);
+        $text = substr($this->code, $this->cursor, $match[0][1] - $this->cursor) . $match[0][0];
+        $this->pushToken(/* Token::COMMENT_TYPE */ 14, trim(substr($text, 0, strrpos($text, '#}'))));
+        $this->moveCursor($text);
     }
 
     private function lexString(): void
--- a/src/Node/CommentNode.php
+++ b/src/Node/CommentNode.php
@@ -1,0 +1,35 @@
+<?php
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) Fabien Potencier
+ * (c) Armin Ronacher
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Twig\Node;
+
+use Twig\Attribute\YieldReady;
+use Twig\Compiler;
+
+/**
+ * Represents a comment node.
+ *
+ * @author Jeroen Versteeg <jeroen@alisqi.com>
+ */
+#[YieldReady]
+class CommentNode extends Node
+{
+    public function __construct(string $data, int $lineno)
+    {
+        parent::__construct([], ['text' => $data], $lineno);
+    }
+
+    public function compile(Compiler $compiler): void
+    {
+        // skip comments in compilation
+    }
+}
--- a/tests/LexerTest.php
+++ b/tests/LexerTest.php
@@ -353,6 +353,26 @@ bar
         $lexer->tokenize(new Source($template, 'index'));
     }
 
+    public function testCommentValues()
+    {
+        $template = '{# comment #}some text{#another one#}';
+        $lexer = new Lexer(new Environment($this->createMock(LoaderInterface::class)));
+        $stream = $lexer->tokenize(new Source($template, 'index'));
+
+        self::assertEquals(
+            'comment', // assert that whitespace is stripped
+            $stream->expect(Token::COMMENT_TYPE)->getValue() // implicit assertion is that expect() doesn't throw
+        );
+        self::assertEquals(
+            'some text',
+            $stream->expect(Token::TEXT_TYPE)->getValue()
+        );
+        self::assertEquals(
+            'another one', // assert that comment is parsed
+            $stream->expect(Token::COMMENT_TYPE)->getValue()
+        );
+    }
+
     public function testOverridingSyntax()
     {
         $template = '[# comment #]{# variable #}/# if true #/true/# endif #/';
@@ -362,6 +382,7 @@ bar
             'tag_variable' => ['{#', '#}'],
         ]);
         $stream = $lexer->tokenize(new Source($template, 'index'));
+        $stream->expect(Token::COMMENT_TYPE);
         $stream->expect(Token::VAR_START_TYPE);
         $stream->expect(Token::NAME_TYPE, 'variable');
         $stream->expect(Token::VAR_END_TYPE);
--- a/src/Parser.php
+++ b/src/Parser.php
@@ -16,6 +16,7 @@ use Twig\Error\SyntaxError;
 use Twig\Node\BlockNode;
 use Twig\Node\BlockReferenceNode;
 use Twig\Node\BodyNode;
+use Twig\Node\CommentNode;
 use Twig\Node\Expression\AbstractExpression;
 use Twig\Node\MacroNode;
 use Twig\Node\ModuleNode;
@@ -123,6 +124,11 @@ class Parser
                 case /* Token::TEXT_TYPE */ 0:
                     $token = $this->stream->next();
                     $rv[] = new TextNode($token->getValue(), $token->getLine());
+                    break;
+
+                case /* Token::COMMENT_TYPE */ 14:
+                    $token = $this->stream->next();
+                    $rv[] = new CommentNode($token->getValue(), $token->getLine());
                     break;
 
                 case /* Token::VAR_START_TYPE */ 2:
--- a/src/Token.php
+++ b/src/Token.php
@@ -36,6 +36,7 @@ final class Token
     public const INTERPOLATION_END_TYPE = 11;
     public const ARROW_TYPE = 12;
     public const SPREAD_TYPE = 13;
+    public const COMMENT_TYPE = 14;
 
     public function __construct(int $type, $value, int $lineno)
     {
@@ -137,6 +138,9 @@ final class Token
             case self::SPREAD_TYPE:
                 $name = 'SPREAD_TYPE';
                 break;
+            case self::COMMENT_TYPE:
+                $name = 'COMMENT_TYPE';
+                break;
             default:
                 throw new \LogicException(sprintf('Token of type "%s" does not exist.', $type));
         }
@@ -177,6 +181,8 @@ final class Token
                 return 'arrow function';
             case self::SPREAD_TYPE:
                 return 'spread operator';
+            case self::COMMENT_TYPE:
+                return 'comment';
             default:
                 throw new \LogicException(sprintf('Token of type "%s" does not exist.', $type));
         }
--- a/tests/Node/CommentTest.php
+++ b/tests/Node/CommentTest.php
@@ -1,0 +1,32 @@
+<?php
+
+namespace Twig\Tests\Node;
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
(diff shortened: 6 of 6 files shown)


drjayvee:
@fabpot and @stof, do you think this can be added in 3.x? Or do you consider it backward breaking, since new nodes are added to the lexer's token stream and the parser's AST?

Please have a look at [TwigStan](https://github.com/alisqi/twig-stan/)'s `README` for further motivation of the utility of this change.

I really think that TwigStan could be huge boon for Twig users.

drjayvee:
One of the code style issues is:
```
+        $nested = $nested || Node::class !== $node::class;
```
However, the suggested change isn't supported in PHP < 8.0.

fabpot:
@drjay

> One of the code style issues is:
> 
> ```
> -        $nested = $nested || Node::class !== \get_class($node);
> +        $nested = $nested || Node::class !== $node::class;
> ```
> 
> However, the suggested change isn't supported in PHP < 8.0.

You can ignore fabbot when it suggests stupid changes.

fabpot:
I'm wondering if it would be better to define a tag instead of using comments.

Instead of:

```
{# @var user \User #}
{# @var showBadge bool #}
```

We could have:

```
{% var user \User %}
{% var showBadge bool %}
```

which also allows Twig to check that the syntax is ok and would make it available to more than just static analysis tools (IDEs 
for instance).

drjayvee:
Bonjour Fabian,

> I'm wondering if it would be better to define a tag instead of using comments.

A new tag has the big advantage of not needing any change to Twig itself. An extension can define, parse and otherwise process it in any way it sees fit.

However, `@var` comments are already supported by PhpStorm + [Symfony Support plugin](https://plugins.jetbrains.com/plugin/7219-symfony-support):
image

> would make it available to more than just static analysis tools (IDEs for instance)

Agreed, but at least the most popular IDE (correct me if I'm wrong) already supports comments.

I don't know about other project, but our app has 750+ of these comments, and we're adding more every week to declare and document templates' context variables, as well as macro arguments. It works really well!

It also aligns with phpDoc, which also uses comments. (That's for historical reasons, of course. But still, I'm not aware of plans to adopt Attributes as a newer alternative.)

So what I'm saying is that the proposed change align with what (some in) the community have already adopted.

> which also allows Twig to check that the syntax is ok

I'm planning for TwigStan to check the syntax, most likely by using phpDocumentor under the hood. Twig could implement this functionality by itself, of course. All I'm saying is that I don't think this feature is exclusively possible with a new tag.

fabpot:
What you're describing is the community having worked around the system. Here, we're talking about making this feature official, so I think using a proper tag is the way to go.

drjayvee:
Fair enough. I'm actually thrilled to see you take this quite seriously! Twig certainly deserves better support for types. ❤️

However, I'm skeptical a new tag will be adopted in practice anytime soon. Specifically, the Symfony Support plugin is poorly maintained ([500 open issues](https://github.com/Haehnchen/idea-php-symfony2-plugin/issues)), so I doubt this particular project will adapt. (I've reached out to its author, @Haehnchen, a few weeks ago, but haven't received an answer.)

Do you perhaps have contacts at JetBrains (someone like @brendt, perhaps), who could tell us whether PhpStorm would consider adoption of this?

Speaking just for our team, we'd have to duplicate `{# @var #}` and `{% var %}` to have proper support in the IDE and static code analysis for the foreseeable future. Not a fun prospect.

So why not support both? The `var` _tag_ will be the official way to do it, and Twig (itself) will add (deprecated) support for _comments_ as a (temporary) alternative. (Similar to how Routes can be [configured](https://symfony.com/doc/current/routing.html#creating-routes-in-yaml-xml-or-php-files) in various ways.)

What do you think?

drjayvee:
Aside from the type declarations, I can imagine the community adding support for more meta-level features, for example to suppress code inspections (similar to `@phpstan-ignore-next-line`).

To enable the community to add these sorts of features, either Twig needs to add a more generic tag (e.g., `{% annotate %}` or `{% meta %}`), or, you know, use comments. 😉

Again, Twig could do both CommentNodes for flexibility and extensibility, and 1st class support for type declarations.

brendt:
Hey I passed this to the PhpStorm team, let us know if/when this is getting added and we're happy to look into it 👍

Update: I've made an issue on our side about this: https://youtrack.jetbrains.com/issue/WI-76637/Twig-var-tag

Haehnchen:
> Symfony Support plugin is poorly maintained ([500 open issues](https://github.com/Haehnchen/idea-php-symfony2-plugin/issues)), so I doubt this particular project will adapt

open issues does not reflect project maintenance, number is still less the Symfony itself :)

But just for letting you known as soon as any implemention is on the way, it will be implemented together with a migration inspection.

----------

Just some hints

arrays should considered also: `{% var user \User[] %}`

whereas `var` doc comment is widely adopted, there are also ways to define entrypoints to collect incoming variables:

```
{# @controller FooBundle:Bar:index #}
{# @controller Class::method #}
```

overall i am sceptical about a tag unless it provides more support out of the box. e.g. it would be nice having `lint:twig` or something else to check for class existence, profiler support / triggering exceptions in dev mode, ...

stof:
Adding CommentNode in the AST looks weird to me, because it means it would not be an AST anymore but an intermediate between an AST and a CST (a full CST needs to retain all formatting information)

drjayvee:
> But just for letting you known as soon as any implemention is on the way, it will be implemented together with a migration inspection.

Thanks for chiming in. This is great to hear.

> arrays should considered also: `{% var user \User[] %}`

Agreed. I'm hoping for community contributions to TwigStan to extend the inspections.

> there are also ways to define entrypoints to collect incoming variables

Can you point me in the direction of some documentation? I don't quite understand what this means / does, but I am intrigued.

This does support my argument that adding support for comments allows for more generic extensibility than the `var` tag.

Again, maybe the best is to do both, with a `@var` comment as a (temporary) alternative to the `var` tag.

> it would be nice having lint:twig or something else to check for class existence, profiler support / triggering exceptions in dev mode

Linting, in combination with exceptions (actually, errors/warnings/notices) in dev mode is exactly the point of TwigStan. Type check (invalid class, misspelled primitive, etc) is already a planned inspection.

<hr>

> Adding CommentNode in the AST looks weird to me, because it means it would not be an AST anymore but an intermediate between an AST and a CST (a full CST needs to retain all formatting information)

Ah, I wasn't aware of that distinction at all. I do appreciate the desire to keep the compiler parts clean and by the books.

One alternative could be transform comments to node _attributes_. But the question immediately becomes: _which_ nodes should receive these attributes? Plus this would have to happen in the Lexer, which isn't the appropriate class to look around the token stream (which it's building itself). Therefore, this seems a no-go.

How about making this behavior configurable? It's clear that only strict dev environments (and linting) would want this. I'm not sure whether this makes matters worse, though. Nor does it really solve stof's principled objection.

Is there another way we can enable static code analysis other than comments?

Or do we maybe just add comments to the AST anyway, since so much good can come out of it to balance the bad? 😜

JoshuaBehrens:
Can you mention mention me, when this is merged? This way I can use them in this PR https://github.com/twigphp/Twig/pull/3928

drjayvee:
@stof and @fabpot, have you reconsidered your stance on this PR?

To summarize my position: I'm in favor of a first-class `var` tag. Once the Symfony Support plugin (and PhpStorm) support it, developers can replace their `@var` comments. In the meantime, however, support for comments would be a great workaround.

Aside from variable types, I do see other potential uses for comments, so I think making these accessible in TwigExtensions would be really useful. Then again, extensions can already easily add custom tags, so comments aren't needed.

I really want to move ahead on TwigStan because it will significantly increase code quality and prevent bugs. Support for variable typing is essential to most inspections, however.

fabpot:
I haven't changed my mind here. Having a tag is the way to go for having something officially supported by Twig.

drjayvee:
Très bien allons-y alors!

I'd be more than happy to work on this, but I don't think I can do it all on my own.

We should probably start by writing a spec? The syntax isn't a big deal, but should Twig do anything with this tag, or leave it up to extensions? I guess at the very least, Twig should check whether the type exists.

Fabian, do you have any ideas for a spec?

Should I start with a basic implementation to get the conversation going? Do we create a new issue for that and close this one?

fabpot:
I've never thought about such a spec, but others have. So, let's bootstrap a spec doc and let's try to involve people interested in this topic. Let's do it in another PR maybe. When bootstrapped, I can then publish something on the Symfony blog to ask for help. Thank you.

ruudk:
I was curious if there is a specific reason for not supporting comments in the Lexer? It seems a bit odd that these are simply skipped. While I agree that these comments should not become Nodes in the AST, I believe it would be beneficial to have them available as attributes. This would allow for reading all Twig tokens.

This approach is similar to how PHP-Parser handles it, where comments are stored in an attribute. This method enables anyone to read the AST along with all comments and even supports writing the AST back to the original file.

drjayvee:
Bonjour @fabpot et @stof,

I've created a first draft of the spec for the `var` tag in #4165. Please review and share your feedback.

drjayvee:
@stof, how do you feel about @ruudk's alternative of adding comments as attributes to the next node? If that's a 👎 as well, I'll close this PR.

src/Lexer.php
```diff
@@ -412,7 +412,9 @@ private function lexComment(): void
             throw new SyntaxError('Unclosed comment.', $this->lineno, $this->source);
         }
 
-        $this->moveCursor(substr($this->code, $this->cursor, $match[0][1] - $this->cursor).$match[0][0]);
+        $text = substr($this->code, $this->cursor, $match[0][1] - $this->cursor).$match[0][0];
+        $this->pushToken(/* Token::COMMENT_TYPE */ 14, trim(substr($text, 0, strrpos($text, '#}'))));
+        $this->moveCursor($text);
```
fabpot, line 417:
```suggestion
        $text = substr($this->code, $this->cursor, $match[0][1] - $this->cursor);
        $this->pushToken(Token::COMMENT_TYPE, trim($text));
        $this->moveCursor($text.$match[0][0]);
```

src/Node/CommentNode.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) Fabien Potencier
+ * (c) Armin Ronacher
```
fabpot, line 7: Can be removed

src/Node/CommentNode.php
```diff
@@ -0,0 +1,35 @@
...
+class CommentNode extends Node
+{
+    public function __construct(string $data, int $lineno)
+    {
+        parent::__construct([], ['text' => $data], $lineno);
+    }
+
+    public function compile(Compiler $compiler): void
+    {
+        // skip comments in compilation
```
fabpot, line 33:
```suggestion
        // ignore comments in generated code
```

tests/LexerTest.php
```diff
@@ -353,6 +353,26 @@ public function testUnterminatedBlock()
         $lexer->tokenize(new Source($template, 'index'));
     }
 
+    public function testCommentValues()
+    {
+        $template = '{# comment #}some text{#another one#}';
```
fabpot, line 358:
```suggestion
        $template = "{# comment #}some text{#another one#}{# \n \nwhitespace\n #}";
```

tests/LexerTest.php
```diff
@@ -353,6 +353,26 @@ public function testUnterminatedBlock()
...
+        $lexer = new Lexer(new Environment($this->createMock(LoaderInterface::class)));
```
fabpot, line 359:
```suggestion
        $lexer = new Lexer(new Environment(new ArrayLoader()));
```

As this is how it's done in 4.x.

tests/LexerTest.php
```diff
@@ -353,6 +353,26 @@ public function testUnterminatedBlock()
...
+            $stream->expect(Token::COMMENT_TYPE)->getValue() // implicit assertion is that expect() doesn't throw
+        );
+        self::assertEquals(
+            'some text',
+            $stream->expect(Token::TEXT_TYPE)->getValue()
+        );
+        self::assertEquals(
+            'another one', // assert that comment is parsed
+            $stream->expect(Token::COMMENT_TYPE)->getValue()
+        );
```
fabpot, line 373:
```suggestion
        );
        self::assertEquals(
            'whitespace',
            $stream->expect(Token::COMMENT_TYPE)->getValue()
        );
```

fabpot (changes requested):
Even if supporting comment nodes in core is simple as demonstrated by this PR, I'm still a bit worried about how people are going to use this new "feature".

You should also add a note in CHANGELOG.

drjayvee:
I personally don't care about this now that we are (very likely) going to implement `types` tag.

@fabpot I'd be more than happy to update the PR, but if you're not sure you want this feature in Twig, I'd rather focus on `types` for now.

fabpot:
> I personally don't care about this now that we are (very likely) going to implement `types` tag.
> 
> @fabpot I'd be more than happy to update the PR, but if you're not sure you want this feature in Twig, I'd rather focus on `types` for now.

Let's close then.

PR #4010, opened 2024-03-16
Mark implicit macro argument default values as such with an attribute in AST

drjayvee:
This change causes no difference to compiled templates or to macro argument semantics.

Consider the following macro:
```twig
{% macro marco(po, lo = null) %}{% endmacro %}
```

With this change, the `ConstantExpression` for argument `po` will have an attribute `is_implicit`, whose value will be `true`. (Note that `lo` will not have that attribute.)

This allows node visitors to distinguish between arguments that do and those that do not have explicit default values even if the value is `null`.

This is useful for [static code analysis](https://github.com/twigphp/Twig/issues/4003).

For example, a static analysis tool might consider arguments with no explicit default value as non-optional.

### DIFF
--- a/src/ExpressionParser.php
+++ b/src/ExpressionParser.php
@@ -642,7 +642,7 @@ class ExpressionParser
                 if (null === $name) {
                     $name = $value->getAttribute('name');
                     $value = new ConstantExpression(null, $this->parser->getCurrentToken()->getLine());
-                    $value->setAttribute('isImplicit', true);
+                    $value->setAttribute('is_implicit', true);
                 }
                 $args[$name] = $value;
             } else {
--- a/src/ExpressionParser.php
+++ b/src/ExpressionParser.php
@@ -591,7 +591,7 @@ class ExpressionParser
      * Parses arguments.
      *
      * @param bool $namedArguments Whether to allow named arguments or not
-     * @param bool $definition     Whether we are parsing arguments for a function definition
+     * @param bool $definition     Whether we are parsing arguments for a function (or macro) definition
      *
      * @return Node
      *
@@ -642,6 +642,7 @@ class ExpressionParser
                 if (null === $name) {
                     $name = $value->getAttribute('name');
                     $value = new ConstantExpression(null, $this->parser->getCurrentToken()->getLine());
+                    $value->setAttribute('isImplicit', true);
                 }
                 $args[$name] = $value;
             } else {
--- a/tests/ParserTest.php
+++ b/tests/ParserTest.php
@@ -178,7 +178,7 @@ EOF
 
     public function testImplicitMacroArgumentDefaultValues()
     {
-        $template = '{% macro marco (po, lo = null) %}{% endmacro %}';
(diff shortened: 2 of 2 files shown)


drjayvee:
Similar to #4009, this is a prerequisite for a desired static analysis inspection in [TwigQI](https://github.com/alisqi/twigqi/).

I can extend this feature by adding this attribute to *every* function/macro argument to make it more consistent.

src/ExpressionParser.php
```diff
@@ -642,6 +642,7 @@ public function parseArguments($namedArguments = false, $definition = false, $al
                 if (null === $name) {
                     $name = $value->getAttribute('name');
                     $value = new ConstantExpression(null, $this->parser->getCurrentToken()->getLine());
+                    $value->setAttribute('isImplicit', true);
```
fabpot:
```suggestion
                    $value->setAttribute('is_implicit', true);
```

tests/ParserTest.php
```diff
@@ -175,6 +176,37 @@ public function testGetVarName()
...
+        $argumentNodes = $this->getParser()
+            ->parse($stream)
+            ->getNode('macros')
+            ->getNode('marco')
+            ->getNode('arguments')
+        ;
+
+        self::assertTrue(
+            $argumentNodes->getNode('po')->hasAttribute('isImplicit')
+        );
```
fabpot: All assertions should be one-liners (and use `$this->assertTrue()` to be consistent with the code base).

tests/ParserTest.php
```diff
@@ -175,6 +176,37 @@ public function testGetVarName()
         $this->addToAssertionCount(1);
     }
 
+    public function testImplicitMacroArgumentDefaultValues()
+    {
+        $template = '{% macro marco (po, lo = null) %}{% endmacro %}';
```
fabpot: As the default value is `null` anyway, what about using a non-null value for `lo` so that the test is more relevant?

drjayvee: Absolutely!

drjayvee:
@fabpot thanks for the review. The latest commit includes all your suggestions.

drjayvee:
@fabpot since you're working on Twig these days, could you please have another look at this PR? 😊

tests/ParserTest.php
```diff
@@ -175,6 +176,27 @@ public function testGetVarName()
         $this->addToAssertionCount(1);
     }
 
+    public function testImplicitMacroArgumentDefaultValues()
+    {
+        $template = '{% macro marco (po, lo = true) %}{% endmacro %}';
+        $lexer = new Lexer(new Environment($this->createMock(LoaderInterface::class)));
```
fabpot, line 182:
```suggestion
        $lexer = new Lexer(new Environment(new ArrayLoader()));
```

fabpot (changes requested):
LGTM

fabpot:
Thank you @drjayvee.

fabpot:
Merged.

PR #4014, opened 2024-04-01
Fix exception when timezone is false

Seb33300:
When passing `false` in the timezone param, twig returns an exception:

> An exception has been thrown during the rendering of a template ("Twig\Extra\Intl\IntlExtension::createDateFormatter(): Argument 5 ($timezone) must be of type ?DateTimeZone, bool given

Passing `false` is supported and allows to skip the timezone conversion.

Regression introduced by https://github.com/twigphp/Twig/pull/3903

### DIFF
--- a/extra/intl-extra/IntlExtension.php
+++ b/extra/intl-extra/IntlExtension.php
@@ -372,7 +372,7 @@ final class IntlExtension extends AbstractExtension
         $date = CoreExtension::dateConverter($env, $date, $timezone);
 
         $formatterTimezone = $timezone;
-        if (null === $formatterTimezone) {
+        if (null === $formatterTimezone || false === $formatterTimezone) {
             $formatterTimezone = $date->getTimezone();
         } elseif (\is_string($formatterTimezone)) {
             $formatterTimezone = new \DateTimeZone($timezone);
--- a/extra/intl-extra/Tests/IntlExtensionTest.php
+++ b/extra/intl-extra/Tests/IntlExtensionTest.php
@@ -45,6 +45,20 @@ class IntlExtensionTest extends TestCase
         );
     }
 
+    public function testFormatterWithoutProtoSkipTimezoneConverter()
+    {
+        $ext = new IntlExtension();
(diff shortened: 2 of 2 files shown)


extra/intl-extra/IntlExtension.php
```diff
@@ -372,7 +372,7 @@ public function formatDateTime(Environment $env, $date, ?string $dateFormat = 'm
         $date = CoreExtension::dateConverter($env, $date, $timezone);
 
         $formatterTimezone = $timezone;
-        if (null === $formatterTimezone) {
+        if (null === $formatterTimezone || false === $formatterTimezone) {
```
smnandre, line 375: This seems to create a different behaviour than stated in the docblock

```php
    /**
     * @param \DateTimeInterface|string|null  $date     A date or null to use the current time
     * @param \DateTimeZone|string|false|null $timezone The target timezone, null to use the default, false to leave unchanged
*/
```

Seb33300: Can you please give more details?

In the docblock:
> false to leave unchanged

And as you can see in the test case, the original UTC date time (1pm) is still 1pm even if twig has been configured on another timezone.

BTW, I did not change the logic (inside `CoreExtension::dateConverter()`), just fixed the regression introduced by https://github.com/twigphp/Twig/pull/3903

smnandre: You're right, but i think "null" and "false" should not be handled in the same way ... at least according to the documentation

> null to use the default, false to leave unchanged

And i'm not saying this is your fault :)

Seb33300: Yes, this is handled differently inside `CoreExtension::dateConverter()` just before the `if`

smnandre: Damn.. sorry, it took me times to understand the "reversion" :(

smnandre (approved):

fabpot:
Thank you @Seb33300.

PR #4018, opened 2024-04-13
Update coding_standards.rst: "one (and only one)" => "exactly one"

ThomasLandauer:
Page: https://twig.symfony.com/doc/3.x/coding_standards.html


fabpot:

PR #4027, opened 2024-04-17
Require correct twig core version

Benedikt-Brunner:
This changes the composer version constraint of the internationalisation extension on the core twig library, to require atleast version 3.9.0. This is necessary since twig/intl-extra now depends on CoreExtension::dateConverter() which was only introduced with the release of version 3.9 (see: twigphp/Twig@54d34b9).

### DIFF
--- a/extra/intl-extra/composer.json
+++ b/extra/intl-extra/composer.json
@@ -16,7 +16,7 @@
     ],
     "require": {
         "php": ">=7.2.5",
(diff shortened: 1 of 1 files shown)


fabpot:
Thank you @Benedikt-Brunner.

shyim:
@fabpot I think the tag is missing in the repo here https://github.com/twigphp/intl-extra/tags

Issue #4033, opened 2024-04-17
[v3.9] Call to undefined method Twig\\TemplateWrapper::yieldBlock() at twig/src/Template.php:467


TLG-Gildas:
I encounter a case where the engine tries to call the `yieldBlock` method on a `TemplateWrapper` instance even though this method does not exist on the wrapper.

In my case it happens here:
https://github.com/twigphp/Twig/blob/7d2b814e6c10101a69052f31634422e6160c58fe/src/Template.php#L466-L468

And indeed the `getParent` method can return a `TemplateWrapper` in case the parent needs to be resolved via the `loadTemplate` method.

You can reproduce this issue with this test : 
```phpt
--TEST--
Call to undefined method Twig\\TemplateWrapper::yieldBlock()
--TEMPLATE--
{% extends 'parent' %}
{%- block content -%}
    {{ parent() }}
    child
{%- endblock -%}
--TEMPLATE(parent)--
{% extends ['unknowngrandparent', 'grandparent'] %}
--TEMPLATE(grandparent)--
{%- block content -%}
    grandparent
{%- endblock -%}
--DATA--
return []
--EXPECT--
    grandparent
    child
```
The key to reproduce the issue, is :
- to have an override of a block defined in a grand parent and not in parent.
- and an extends with array in parent

I can see two options to solve this issue : 
- add a call to `unwrap` everywhere a `TemplateWrapper` can be
- synchronize the contract of `TemplateWrapper` with `Template`

TLG-Gildas:
I just open a PR : https://github.com/twigphp/Twig/pull/4037

I use unwrap call strategy, instead of changing the TemplateWrapper contract, to be coherent with the rest of code base.

PR #4037, opened 2024-04-18
fix: #4033 add missing unwrap call on TemplateWrapper instance

TLG-Gildas:


fabpot:
Thank you @TLG-Gildas.

PR #4038, opened 2024-04-18
Ensure Lexer:: is always initialized

danut007ro:
I'm using Symfony Bridge `LintCommand` with `--show-deprecations` flag.

While Twig is initializing, if an deprecation error is thrown by `Lexer::getOperatorRegex()` method, the `Lexer::$regexes` will remain `null` but `$isInitialized` will be set to TRUE.

This PR ensures that `$regexes` will be initialized.

### DIFF
--- a/src/Lexer.php
+++ b/src/Lexer.php
@@ -71,8 +71,6 @@ class Lexer
             return;
         }
 
-        $this->isInitialized = true;
-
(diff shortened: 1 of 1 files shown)


stof:
to me, the simpler fix would be to move the assignment of `isInitialized` after the initialization instead of before it.

stof (approved):

xabbuh (approved):

fabpot:
Thank you @danut007ro.

PR #4044, opened 2024-04-19
Move some static extension methods to non-static

### LINKED ISSUE #4040: Twig functions becoming internal
Since 3.9 it appears all of Twigs filter/functions underlying PHP implementations are marked as internal, could this be reconsidered in some cases? I'm assuming internal means the signature might change in non backwards compatible ways, if it doesn't I guess this isn't really a problem.

The two cases that affect myself are `twig_escape_filter` and `twig_date_converter`.

For example I have a filter which generates some HTML:

```php
new TwigFilter('phone_number_link', function (Environment $env, $number) {
    $anchor = $this->libphonenumber->format($number, PhoneNumberFormat::RFC3966);

    return sprintf(
        '<a href="%s">%s</a>',
        \twig_escape_filter($env, $anchor, 'html_attr'),
        \twig_escape_filter($env, $this->formatDisplay($number), 'html'),
    );
}, ['needs_environment' => true, 'is_safe' => ['html']]),
```

I've consulted the documentation but it doesn't seem to offer any examples on how to deal with more complicated situations such as this, so perhaps this is the wrong way of doing it but it's worked since Twig 1.x.

Likewise with dates I have some custom filters to limit developers to a limited set of predefined formats, the filters look a bit like this:

```php
new TwigFilter('date_formal_format', function (Environment $env, $date, ?string $locale = null) {
    if (null === $date) {
        return null;
    }

    return DateFormat::formalDate(twig_date_converter($env, $date), $locale ?? $this->locale);
}, ['needs_environment' => true]),
```

Being able to use `twig_date_converter` means the Twig timezone setting works as expected.

I did some searching on GitHub and found numerous other cases.

https://github.com/theoboldt/juvem/blob/98af699c95c46e919dc1690ab1730754ebd65141/app/src/AppBundle/Twig/Extension/CustomFieldValue.php#L244
https://github.com/njh/twig-html-helpers/blob/d3dc45f6a9ac67512126c02a8d11c7a09889d942/lib/Twig/Extension/HTMLHelpers.php#L53
https://github.com/symfony/ux/blob/d4df61465571381ffa8692268fca6acd09b32feb/src/StimulusBundle/src/Dto/StimulusAttributes.php#L223
https://github.com/symfony/ux/blob/d4df61465571381ffa8692268fca6acd09b32feb/src/LiveComponent/src/Util/LiveAttributesCollection.php#L121
https://github.com/studiometa/twig-toolkit/blob/9e73108d84c11171ef69fd34dd8946afee8e9a5c/src/Helpers/Html.php#L209
https://github.com/mautic/mautic/blob/175ceac9098f0b9515f34820a0eafd924877864b/plugins/MauticFocusBundle/Model/FocusModel.php#L183
https://github.com/symfony/symfony/blob/a2d03c548c2e897be78504fd4a389bb7eb41e756/src/Symfony/Bundle/WebProfilerBundle/Twig/WebProfilerExtension.php#L116
https://github.com/drupal/drupal/blob/449f6fbf7c058988ba6f18a8e70cba4bf6639941/core/lib/Drupal/Core/Template/TwigExtension.php#L464
https://github.com/phpbb/phpbb/blob/4c721de243967ed4cf372caef4e54142ee6bcf6c/phpBB/phpbb/template/twig/extension.php#L215
https://github.com/IISHF/web2019/blob/0ca849abf7f247abfcf01521b6cdeee22e1d637b/src/Infrastructure/Twig/DateExtension.php#L108-L115
https://github.com/imarc/cms/blob/ef349e03202628638247a71158ff04c49da2068a/src/web/twig/Extension.php#L482
https://github.com/twigphp/intl-extra/blob/00cd46d7860cf7e814f73b290edf32801a47cafc/IntlExtension.php#L282

fabpot:
A first step to resolve #4040

### DIFF
--- a/extra/intl-extra/IntlExtension.php
+++ b/extra/intl-extra/IntlExtension.php
@@ -369,7 +369,7 @@ final class IntlExtension extends AbstractExtension
      */
     public function formatDateTime(Environment $env, $date, ?string $dateFormat = 'medium', ?string $timeFormat = 'medium', string $pattern = '', $timezone = null, string $calendar = 'gregorian', ?string $locale = null): string
     {
-        $date = CoreExtension::convertDate($env, $date, $timezone);
+        $date = $env->getExtension(CoreExtension::class)->convertDate($date, $timezone);
 
         $formatterTimezone = $timezone;
         if (null === $formatterTimezone || false === $formatterTimezone) {
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -185,11 +185,11 @@ final class CoreExtension extends AbstractExtension
     {
         return [
             // formatting filters
-            new TwigFilter('date', [self::class, 'formatDate'], ['needs_environment' => true]),
-            new TwigFilter('date_modify', [self::class, 'modifyDate'], ['needs_environment' => true]),
+            new TwigFilter('date', [$this, 'formatDate']),
+            new TwigFilter('date_modify', [$this, 'modifyDate']),
             new TwigFilter('format', [self::class, 'sprintf']),
             new TwigFilter('replace', [self::class, 'replace']),
-            new TwigFilter('number_format', [self::class, 'formatNumber'], ['needs_environment' => true]),
+            new TwigFilter('number_format', [$this, 'formatNumber']),
             new TwigFilter('abs', 'abs'),
             new TwigFilter('round', [self::class, 'round']),
 
@@ -241,7 +241,7 @@ final class CoreExtension extends AbstractExtension
             new TwigFunction('constant', [self::class, 'constant']),
             new TwigFunction('cycle', [self::class, 'cycle']),
             new TwigFunction('random', [self::class, 'random'], ['needs_charset' => true]),
-            new TwigFunction('date', [self::class, 'convertDate'], ['needs_environment' => true]),
+            new TwigFunction('date', [$this, 'convertDate']),
             new TwigFunction('include', [self::class, 'include'], ['needs_environment' => true, 'needs_context' => true, 'is_safe' => ['all']]),
             new TwigFunction('source', [self::class, 'source'], ['needs_environment' => true, 'is_safe' => ['all']]),
         ];
@@ -416,10 +416,10 @@ final class CoreExtension extends AbstractExtension
      *
      * @internal
      */
-    public static function formatDate(Environment $env, $date, $format = null, $timezone = null): string
+    public function formatDate($date, $format = null, $timezone = null): string
     {
         if (null === $format) {
-            $formats = $env->getExtension(self::class)->getDateFormat();
+            $formats = $this->getDateFormat();
             $format = $date instanceof \DateInterval ? $formats[1] : $formats[0];
         }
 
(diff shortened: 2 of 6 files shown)


src/Extension/CoreExtension.php
```diff
@@ -420,18 +420,18 @@ public static function random(Environment $env, $values = null, $max = null)
      *
      * @internal
      */
-    public static function dateFormatFilter(Environment $env, $date, $format = null, $timezone = null)
+    public function dateFormatFilter($date, $format = null, $timezone = null)
     {
         if (null === $format) {
-            $formats = $env->getExtension(self::class)->getDateFormat();
+            $formats = $this->getDateFormat();
```
fabpot, line 422: The main idea is to remove the usage of `$env` when possible so that we can extract these methods to another class that would be available for public consumption.

src/Extension/CoreExtension.php
```diff
@@ -482,12 +480,12 @@ public static function sprintf($format, ...$values)
      *
      * @internal
      */
-    public static function dateConverter(Environment $env, $date = null, $timezone = null)
+    public function dateConverter($date = null, $timezone = null)
```
stof: If the goal is to make this method public, I would suggest renaming it to `convertDate()` to have a proper verb-based naming for it.

extra/intl-extra/IntlExtension.php
```diff
@@ -369,7 +369,7 @@ public function formatNumberStyle(string $style, $number, array $attrs = [], str
      */
     public function formatDateTime(Environment $env, $date, ?string $dateFormat = 'medium', ?string $timeFormat = 'medium', string $pattern = '', $timezone = null, string $calendar = 'gregorian', ?string $locale = null): string
     {
-        $date = CoreExtension::dateConverter($env, $date, $timezone);
+        $date = $env->getExtension(CoreExtension::class)->dateConverter($date, $timezone);
```
stof: this change requires bumping the min requirement of twig/twig in this intl-extra package.

stof: And btw, we probably need to keep a deprecated `dateConverter` static method to avoid issues if using twig/intl-extra 3.9.2 with twig/twig 3.10.0.
twig/intl-extra is using an internal API of a different package, which then becomes an issue for semver-based requirements.

cs278: `twig/intl-extra` already raised it's twig requirement in the last release so it could just do that again?

stof: raising the requirement will solve my first comment. It will not solve the second one (which is about avoiding an error when using the old intl-extra with the new twig).

cs278: Very true, I guess that's already been broken once before by 3.9 not that's not a reason to do it again.

stof: 3.9 broke the case of upgrading intl-extra alone by forgetting to bump the requirement. It did not broke the case of updating twig alone (as intl-extra 3.8 was using `twig_date_converter` which has not been removed)

fabpot: Done

PR #4046, opened 2024-04-19
Optimized TextNodes have no SourceContext

### LINKED ISSUE #4045: Optimized TextNodes have no SourceContext
`TextNode` instances created by `OptimizerNodeVisitor` have no `SourceContext` set and often the wrong template line which i need for a debugging extension. `OPTIMIZE_RAW_FILTER` can be turned off, but the print node optimization is always active. Also the check in the constructor currently doesn't include `OPTIMIZE_RAW_FILTER` and would throw an `InvalidArgumentException` if only that optimization should be enabled - this can be bypassed by adding something to the bitmask to force negative values though.

Lumaraf:
Fixes twigphp/Twig#4045

### DIFF
--- a/src/NodeVisitor/OptimizerNodeVisitor.php
+++ b/src/NodeVisitor/OptimizerNodeVisitor.php
@@ -42,6 +42,7 @@ final class OptimizerNodeVisitor implements NodeVisitorInterface
 {
     public const OPTIMIZE_ALL = -1;
     public const OPTIMIZE_NONE = 0;
+    public const OPTIMIZE_PRINT = 1;
     public const OPTIMIZE_FOR = 2;
     public const OPTIMIZE_RAW_FILTER = 4;
     public const OPTIMIZE_TEXT_NODES = 8;
@@ -55,7 +56,7 @@ final class OptimizerNodeVisitor implements NodeVisitorInterface
      */
     public function __construct(int $optimizers = -1)
     {
-        if ($optimizers > (self::OPTIMIZE_FOR | self::OPTIMIZE_RAW_FILTER)) {
+        if ($optimizers !== self::OPTIMIZE_ALL && ($optimizers & ~(self::OPTIMIZE_PRINT | self::OPTIMIZE_FOR | self::OPTIMIZE_RAW_FILTER | self::OPTIMIZE_TEXT_NODES)) !== 0) {
             throw new \InvalidArgumentException(sprintf('Optimizer mode "%s" is not valid.', $optimizers));
         }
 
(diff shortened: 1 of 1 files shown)


src/NodeVisitor/OptimizerNodeVisitor.php
```diff
@@ -107,13 +110,20 @@ private function mergeTextNodeCalls(Node $node): Node
             return $node;
         }
 
+        $firstChildNode = $node->getNode($names[0]);
+        $line = $firstChildNode->getTemplateLine();
+        $sourceContext = $firstChildNode->getSourceContext();
         if (Node::class === get_class($node)) {
-            return new TextNode($text, $node->getTemplateLine());
+            $textNode = new TextNode($text, $line);
+            $textNode->setSourceContext($sourceContext);
```
stof, line 118: this will break if `$sourceContext` is null.

But maybe the right fix would be for `\Twig\Node\Node::setNode` to set the source context in the new node in case we have a source context and the node being set does not have one. This would be consistent with the fact that `setSourceContext` also sets the node on all children at the time it is set.
this way, it would work for the nodes created by OptimizerNodeVisitor but would also solve the similar issue for any other custom node visitor

src/NodeVisitor/OptimizerNodeVisitor.php
```diff
@@ -42,6 +42,7 @@ final class OptimizerNodeVisitor implements NodeVisitorInterface
 {
     public const OPTIMIZE_ALL = -1;
     public const OPTIMIZE_NONE = 0;
+    public const OPTIMIZE_PRINT = 1;
```
stof, line 45: making this optimization configurable would deserve a dedicated PR (as it deserve a changelog entry for such new feature)

stof (changes requested):

src/NodeVisitor/OptimizerNodeVisitor.php
Lumaraf: Setting the `SourceContext` in `setNode` sounds like a great idea. That solves my current problem with missing `SourceContexts` after optimization and simplifies all of my current `NodeVisitor` use cases where i have to set the `SourceContext` everywhere i create new nodes.

I created a separate PR for that small change: https://github.com/twigphp/Twig/pull/4050

src/NodeVisitor/OptimizerNodeVisitor.php
Lumaraf: Makes sense, i extracted those changes into a separate PR https://github.com/twigphp/Twig/pull/4051.

fabpot:
I think this one can be closed now, right?

PR #4051, opened 2024-04-24
allow turning all optimizations off

Lumaraf:
Currently setting the `TwigEnvironment` option `optimizers` to 0 (`OptimizerNodeVisitor::OPTIMIZE_NONE`) does not turn of the `PrintNode` optimization.

### DIFF
--- a/src/NodeVisitor/OptimizerNodeVisitor.php
+++ b/src/NodeVisitor/OptimizerNodeVisitor.php
@@ -42,6 +42,7 @@ final class OptimizerNodeVisitor implements NodeVisitorInterface
 {
     public const OPTIMIZE_ALL = -1;
     public const OPTIMIZE_NONE = 0;
+    public const OPTIMIZE_PRINT = 1;
     public const OPTIMIZE_FOR = 2;
     public const OPTIMIZE_RAW_FILTER = 4;
     public const OPTIMIZE_TEXT_NODES = 8;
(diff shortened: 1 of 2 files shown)


src/NodeVisitor/OptimizerNodeVisitor.php
```diff
@@ -81,7 +82,9 @@ public function leaveNode(Node $node, Environment $env): ?Node
             $node = $this->optimizeRawFilter($node);
         }
 
-        $node = $this->optimizePrintNode($node);
+        if (self::OPTIMIZE_PRINT === (self::OPTIMIZE_PRINT & $this->optimizers)) {
```
fabpot, line 85: Why would we want to be able to disable this optimization. I didn't provide a way to disable it on purpose.

Lumaraf: I ran into problems with one of my own extension because of missing `SourceContexts` on nodes created by the PrintNode and TextNode optimizations. The obvious solution to me would be to turn off those or all optimizers, but setting the optimizers property on `TwigEnvironment` to 0/`OptimizerNodeVisitor::OPTIMIZE_NONE` still leaves the PrintNode optimization active.

fabpot: As we are fixing the source context, would you agree that there is not need to make this optimization configurable?

Lumaraf: The source context fix is sufficient to solve my current problems. I still think this should be fixed though as the optimizers option currently doesn't behave in the way it is described in the doc comment in https://github.com/twigphp/Twig/blob/4a7de4ab5603ad8f199222cae3951fc271a8c45d/src/Environment.php#L100-L102 as setting it to `0` does not disable all optimizations.

fabpot: Let's close then.

PR #4055, opened 2024-04-28
Extract the escaping logic to a runtime

fabpot:


smnandre:
```
 If you were using the ``twig_escape_filter()`` function is your code, use
 ``$env->getRuntime(EscaperRuntime::class)->escape()`` instead.
  ```
  
For once, i'd suggest to add some "magic" because the DX seem a bit odd (to me).. 
  
Could we expose directly something like ? 

```php
$env->getEscaper()
```

With  maybe a dedicated interface ?

fabpot:
> ```
>  If you were using the ``twig_escape_filter()`` function is your code, use
>  ``$env->getRuntime(EscaperRuntime::class)->escape()`` instead.
> ```
> 
> For once, i'd suggest to add some "magic" because the DX seem a bit odd (to me)..
> 
> Could we expose directly something like ?
> 
> ```
> $env->getEscaper()
> ```
> 
> With maybe a dedicated interface ?

I don't think we need a "nicer" DX as using this directly should be the exception, not the rule.

doc/deprecated.rst
```diff
@@ -18,6 +18,20 @@ Extensions
   3.9.0, and will be removed in Twig 4.0. They have been replaced by internal
   methods on their respective extension classes.
 
+  If you were using the ``twig_escape_filter()`` function is your code, use
```
smnandre:
```suggestion
  If you were using the ``twig_escape_filter()`` function in your code, use
```

src/Extension/EscaperExtension.php
```diff
@@ -84,59 +82,72 @@ public function setDefaultStrategy($defaultStrategy): void
...
+        return $this->escaper->getDefaultStrategy($name);
     }
 
     /**
      * Defines a new escaper to be used via the escape filter.
      *
-     * @param string   $strategy The strategy name that should be used as a strategy in the escape call
-     * @param callable $callable A valid PHP callable
+     * @param string                        $strategy The strategy name that should be used as a strategy in the escape call
+     * @param callable(Environment, string) $callable A valid PHP callable
```
stof, line 99: The callable also receives the charset as a third argument.

src/Extension/EscaperExtension.php
```diff
@@ -84,59 +82,72 @@ public function setDefaultStrategy($defaultStrategy): void
...
+        };
+
+        $this->escaper->setEscaper($strategy, $callable);
     }
 
     /**
      * Gets all defined escapers.
      *
-     * @return callable[] An array of escapers
+     * @return array<callable(Environment, string)> An array of escapers
```
stof, line 122: those callables also take the charset as third argument

src/NodeVisitor/EscaperNodeVisitor.php
```diff
@@ -49,7 +50,7 @@ public function __construct()
     public function enterNode(Node $node, Environment $env): Node
     {
         if ($node instanceof ModuleNode) {
-            if ($env->hasExtension(EscaperExtension::class) && $defaultStrategy = $env->getExtension(EscaperExtension::class)->getDefaultStrategy($node->getTemplateName())) {
+            if ($env->hasExtension(EscaperExtension::class) && $defaultStrategy = $env->getRuntime(EscaperRuntime::class)->getDefaultStrategy($node->getTemplateName())) {
```
stof, line 53: should the default strategy be in the runtime ? It seems weird to me to use the runtime at compile time.

src/Extension/EscaperExtension.php
fabpot: The code and tests were wrong. Here, we should support the legacy escapers signature.

fabpot: See #4071 for the fix.

src/NodeVisitor/EscaperNodeVisitor.php
fabpot: Right. We don't even use it in the runtime, this is only needed at compilation time. See #4072

PR #4062, opened 2024-04-30
Fix capturing output from extensions that still use echo

### LINKED ISSUE #4039: [3.9] Cannot assign the output of a custom tag to a variable using `set`
After Twig 3.9.0, it's not possible to capture the output of a custom tag (that uses `echo`) and assign it to a variable using `set`.

```twig
A {% set variable %}{% trans 'value' %}{% endset %} B {{ variable }} C
```

With Twig 3.9.0+, that will print:
```
A value B  C
```

Instead of the expected:
```
A  B value C
```

It works as expected if the custom tag output is not the only content, for example:
```twig
A {% set variable %}some {% trans 'value' %}{% endset %} B {{ variable }} C
```
Outputs:
```
A  B some value C
```

nicolas-grekas:
Fix #4039

### DIFF
--- a/src/Node/CaptureNode.php
+++ b/src/Node/CaptureNode.php
@@ -43,8 +43,9 @@ class CaptureNode extends Node
         $compiler
             ->indent()
             ->subcompile($this->getNode('body'))
+            ->write("return; yield '';\n")
             ->outdent()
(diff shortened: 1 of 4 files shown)


fabpot:
Thank you @nicolas-grekas.

PR #4069, opened 2024-05-02
remove unused internal method

xabbuh:


xabbuh:
it seems to have been added accidentally in #4044

fabpot:
I've added it to avoid breaking `intl-extra`.

PR #4071, opened 2024-05-04
Fix wrong tests

fabpot:


tests/Extension/EscaperTest.php
```diff
@@ -50,24 +50,24 @@ public function testCustomEscapersOnMultipleEnvs()
...
-        $escaperExt2->setEscaper('foo', 'Twig\Tests\foo_escaper_for_test1');
+        $escaperExt2->setEscaper('foo', 'Twig\Tests\legacy_escaper_again');
 
         $this->assertSame('fooUTF-8', $env1->getRuntime(EscaperRuntime::class)->escape('foo', 'foo'));
         $this->assertSame('fooUTF-81', $env2->getRuntime(EscaperRuntime::class)->escape('foo', 'foo'));
     }
 }
 
-function foo_escaper_for_test(Environment $twig, $string, $charset)
+function legacy_escaper(Environment $twig, $string)
```
fabpot, line 65: This is the old signature that we should still support.

tests/Runtime/EscaperRuntimeTest.php
```diff
@@ -389,16 +389,11 @@ public function provideObjectsForEscaping()
     }
 }
 
-function foo_escaper_for_test_runtime($string, $charset)
+function escaper($string, $charset)
```
fabpot, line 392: This is the new signature supported by the runtime.

PR #4075, opened 2024-05-04
Make Environment::getGlobals() private

fabpot:


tests/EnvironmentTest.php
```diff
@@ -65,84 +65,87 @@ public function testGlobals()
         $loader = $this->createMock(LoaderInterface::class);
         $loader->expects($this->any())->method('getSourceContext')->willReturn(new Source('', ''));
 
+        $getGlobals = new \ReflectionMethod(Environment::class, 'getGlobals');
+        $getGlobals->setAccessible(true);
```
jdreesen: This is not necessary anymore since PHP 8.1 (see note here: https://www.php.net/manual/en/reflectionmethod.setaccessible.php).

```suggestion
```

tests/EnvironmentTest.php
```diff
@@ -280,13 +282,16 @@ public function testAddExtension()
         $twig = new Environment($this->createMock(LoaderInterface::class));
         $twig->addExtension(new EnvironmentTest_Extension());
 
+        $getGlobals = new \ReflectionMethod(Environment::class, 'getGlobals');
+        $getGlobals->setAccessible(true);
```
xabbuh: Can be removed too.

PR #4081, opened 2024-05-11
Fix blocks not available under some circumstancies

### LINKED ISSUE #4079: Warning in compiled Twig template : "Warning: Undefined variable $blocks"
With the version 3.9.3, I get this warning when I'm using variable on cache tag on my symfony project (version 6.4.7)
> An exception has been thrown during the rendering of a template ("Warning: Undefined variable $blocks").

You can find bellow a small example 
```
{% cache 'project_%s_%s'|format(project.id, project.updatedAt.timestamp) ttl(36000) %}
   {% set list %}
      <div class="list-group list-group-flush"></div>
   {% endset %}
   {{ list }}
{% endcache %}
```

Here is my composer.json file about twig packages with the current versions in my project

> "twig/cache-extra": "3.9.0",
> "twig/cssinliner-extra": "3.9.0",
> "twig/extra-bundle": "3.9.3",
> "twig/intl-extra": "3.9.2",
> "twig/string-extra": "3.9.0",
> "twig/twig": "3.9.3"

Capture d’écran 2024-05-07 à 16 40 12

Hope it can help

fabpot:
Fix #4079

### DIFF
--- a/extra/cache-extra/Node/CacheNode.php
+++ b/extra/cache-extra/Node/CacheNode.php
@@ -40,7 +40,7 @@ class CacheNode extends AbstractExpression
             ->addDebugInfo($this)
             ->raw('$this->env->getRuntime(\'Twig\Extra\Cache\CacheRuntime\')->getCache()->get(')
             ->subcompile($this->getNode('key'))
-            ->raw(", function (\Symfony\Contracts\Cache\ItemInterface \$item) use (\$context, \$macros) {\n")
+            ->raw(", function (\Symfony\Contracts\Cache\ItemInterface \$item) use (\$context, \$macros, \$blocks) {\n")
             ->indent()
         ;
 
--- a/src/Node/CaptureNode.php
+++ b/src/Node/CaptureNode.php
@@ -24,7 +24,7 @@ class CaptureNode extends Node
 {
     public function __construct(Node $body, int $lineno, ?string $tag = null)
     {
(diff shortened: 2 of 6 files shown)


src/Node/CaptureNode.php
```diff
@@ -24,7 +24,7 @@ class CaptureNode extends Node
 {
     public function __construct(Node $body, int $lineno, ?string $tag = null)
     {
-        parent::__construct(['body' => $body], ['raw' => false, 'with_blocks' => false], $lineno, $tag);
```
fabpot, line 27: `$blocks` should always be passed to the anonymous function as used by many Twig features.

extra/cache-extra/Node/CacheNode.php
```diff
@@ -40,7 +40,7 @@ public function compile(Compiler $compiler): void
             ->addDebugInfo($this)
             ->raw('$this->env->getRuntime(\'Twig\Extra\Cache\CacheRuntime\')->getCache()->get(')
             ->subcompile($this->getNode('key'))
-            ->raw(", function (\Symfony\Contracts\Cache\ItemInterface \$item) use (\$context, \$macros) {\n")
+            ->raw(", function (\Symfony\Contracts\Cache\ItemInterface \$item) use (\$context, \$macros, \$blocks) {\n")
```
fabpot, line 43: This bug is independent of the new CaptureNode feature, but it's now more obvious.

PR #4088, opened 2024-05-12
Pass charset to escaper callback again

### LINKED ISSUE #4087: Possible BC break with `EscaperExtension` in `3.10.x`
In `3.9.x` the escaper callback was called with 3 arguments:

* The environment `$env`
* The string `$string`
* The character set `$charset`

https://github.com/twigphp/Twig/blob/a842d75fed59cdbcbd3a3ad7fb9eb768fc350d58/src/Extension/EscaperExtension.php#L408

However, in `3.10.1` (and probably `3.10.0`) it is only called with 2 arguments:

* The environment `$env`
* The string `$string`

https://github.com/twigphp/Twig/blob/3af5ab2e52279e5e23dc192b1a26db3b8cffa4e7/src/Extension/EscaperExtension.php#L135

Resulting in an error like this for example:

```
ArgumentCountError:
Too few arguments to function Contao\CoreBundle\Twig\Interop\ContaoEscaper::escapeHtml(), 2 passed in vendor\twig\twig\src\Extension\EscaperExtension.php on line 135 and exactly 3 expected

  at vendor\contao\contao\core-bundle\src\Twig\Interop\ContaoEscaper.php:38
  at Contao\CoreBundle\Twig\Interop\ContaoEscaper->escapeHtml(object(Environment), 'Main navigation')
     (vendor\twig\twig\src\Extension\EscaperExtension.php:135)
  at Twig\Extension\EscaperExtension->Twig\Extension\{closure}('Main navigation', 'UTF-8')
     (vendor\twig\twig\src\Runtime\EscaperRuntime.php:311)
```

for custom escapers when updating from `3.9` to `3.10`.

fritzmg:
Fixes #4087

This restores the behaviour of `3.9.x` where the `$charset` was passed to the escaper callback.

### DIFF
--- a/src/Extension/EscaperExtension.php
+++ b/src/Extension/EscaperExtension.php
@@ -132,7 +132,7 @@ final class EscaperExtension extends AbstractExtension
 
         $this->escapers[$strategy] = $callable;
         $callable = function ($string, $charset) use ($callable) {
-            return $callable($this->environment, $string);
+            return $callable($this->environment, $string, $this->environment->getCharset());
         };
 
         $this->escaper->setEscaper($strategy, $callable);


src/Extension/EscaperExtension.php
```diff
@@ -132,7 +132,7 @@ public function setEscaper($strategy, callable $callable)
 
         $this->escapers[$strategy] = $callable;
         $callable = function ($string, $charset) use ($callable) {
-            return $callable($this->environment, $string);
+            return $callable($this->environment, $string, $this->environment->getCharset());
```
fabpot, line 135: It was indeed a mistake. I don't think we should restore this odd behavior as the callables never got the charset before (so without changing the escapers, the charset is not useful). And as adding the charset does not make migrating easier, I don't think we should support it.

fritzmg: > as the callables never got the charset before

Not sure what you mean? The callables did get the charset before:

https://github.com/twigphp/Twig/blob/a842d75fed59cdbcbd3a3ad7fb9eb768fc350d58/src/Extension/EscaperExtension.php#L408

That's the BC break.

fabpot: Sorry, I'm stupid sometimes. See #4091 for the fix. Here, we need the charset passed to the escape filter, not the default charset.

PR #4106, opened 2024-06-17
Rename hash to mapping and array to sequence

fabpot:


xabbuh:
There's also a variable named `hash` in `tests/Fixtures/regression/strings_like_numbers.test`.

doc/filters/merge.rst
```diff
@@ -1,7 +1,9 @@
 ``merge``
 =========
 
-The ``merge`` filter merges an array with another array:
+The ``merge`` filter merges sequences and mappings.
+
+The ``merge`` filter also works on sequences:
```
xabbuh: The "also" here and below reads weird. What about merging both code blocks instead?

fabpot: I've done it a bit differently, it's better now.

fabpot:
Should be good now

PR #4111, opened 2024-06-27
Add a public method clearLoadedTemplates():void to the Environment

khanhartisan:
Because when we run the application with a runner like FrankenPHP or Swoole the $loadedTemplates data will be kept in the memory. 

That will lead to a problem, the template in the cache might be changed but Twig keeps rendering old template from $loadedTemplates. 

So The best solution is to have a public method that allows developers to clear this variable.

### DIFF
--- a/src/Environment.php
+++ b/src/Environment.php
@@ -876,4 +876,17 @@ class Environment
             $this->useYield ? '1' : '0',
         ]);
     }
+
+    /**
+     * A method to clear the private variable $loadedTemplates
+     *
+     * Because when we run the application with a runner like FrankenPHP or Swoole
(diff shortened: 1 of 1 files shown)


khanhartisan:
Oh no, I just realized that even doing so, it will still render the stale template. Because after the template is loaded, it'll load the class into the memory. And there's no way to re-declare or un-declare a loaded class.
That means, when running a PHP application with a runner (FrankenPHP/Swoole), the only way to update the Twig template is to reset the runner workers.

khanhartisan:
Actually we have a solution. It's about to implement a custom Cache driver that generates the cache key dynamically based on the template version.

PR #4116, opened 2024-06-30
[Doc] Modernize Class name in examples

smnandre:
From "Project_Twig_Extension" to "CustomExtension"

Just a suggestion, feel free to close if you want to keep it like this :)

I found those namings only in "advanced"

### DIFF
--- a/doc/advanced.rst
+++ b/doc/advanced.rst
@@ -338,11 +338,11 @@ compilation. This is useful if your test can be compiled into PHP primitives.
 This is used by many of the tests built into Twig::
 
     namespace App;
-    
+
     use Twig\Environment;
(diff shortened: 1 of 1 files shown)


fabpot:
That makes a lot of sense to me. Have you change them all?

smnandre:
> That makes a lot of sense to me. Have you change them all?

I looked for other `Foo_Bar_Class` and could not find one.. it seems the remaining were mostly in this advanced page.



But i noticed something "similar-ish" a couple weeks ago (and forgot to note it...) i'll look if i can find it again

smnandre:
Got it :) https://github.com/twigphp/Twig/pull/4117

fabpot:
Thank you @smnandre.

PR #4120, opened 2024-07-05
simplify an iterable check

xabbuh:


stof (approved):

fabpot:
Thank you @xabbuh.

stof:
To fix the `--prefer-lowest` job, you can change the phpunit requirement to keep only PHPUnit 8+ (as PHPUnit 8 supports PHP 7.2)

PR #4121, opened 2024-07-05
fix the set of installed packages for TwigExtraBundle on PHP 8.0

xabbuh:
Prevents symfony/translation-contracts 3 to be installed together with symfony/string 5.4.0-RC1 (see https://github.com/twigphp/Twig/actions/runs/9804849131/job/27073398023#step:8:117). Stable releases of the String component don't suffer from this incompatibility because of symfony/symfony#44246.


fabpot:
Thank you @xabbuh.

PR #4123, opened 2024-07-06
[4.0] Support string literals

ruudk:
After posting this as an idea in https://github.com/twigphp/Twig/issues/3951#issuecomment-2211747927 I tried to implement it. It looks like it's quite doable. But maybe I'm missing something.

---

This allows to write a string literal using backticks. When using a literal, you don't need to escape backslashes. This is ideal for situation where you want to reference a fully qualified class name. For example in a `constant` function.

For example, before you had to write:
```twig
{{ constant("App\\Entity\\Class::Constant") }}
```

Now you can write:
```twig
{{ constant(`App\Entity\Class::Constant`) }}
```

Besides easier to read, there is another huge benefit: when doing a text search for a fully qualified class name, it will show up.

### DIFF
--- a/src/Lexer.php
+++ b/src/Lexer.php
@@ -57,12 +57,15 @@ class Lexer
     public const STATE_VAR = 2;
     public const STATE_STRING = 3;
     public const STATE_INTERPOLATION = 4;
+    public const STATE_STRING_LITERAL = 5;
 
     public const REGEX_NAME = '/[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*/A';
     public const REGEX_NUMBER = '/[0-9]+(?:\.[0-9]+)?([Ee][\+\-][0-9]+)?/A';
     public const REGEX_STRING = '/"([^#"\\\\]*(?:\\\\.[^#"\\\\]*)*)"|\'([^\'\\\\]*(?:\\\\.[^\'\\\\]*)*)\'/As';
     public const REGEX_DQ_STRING_DELIM = '/"/A';
     public const REGEX_DQ_STRING_PART = '/[^#"\\\\]*(?:(?:\\\\.|#(?!\{))[^#"\\\\]*)*/As';
+    public const REGEX_STRING_LITERAL_DELIM = '/`/A';
+    public const REGEX_STRING_LITERAL_PART = '/[^`]+/As';
     public const PUNCTUATION = '()[]{}?:.,|';
 
     public function __construct(Environment $env, array $options = [])
@@ -217,6 +220,10 @@ class Lexer
 
                 case self::STATE_INTERPOLATION:
                     $this->lexInterpolation();
+                    break;
+
+                case self::STATE_STRING_LITERAL:
+                    $this->lexStringLiteral();
                     break;
             }
         }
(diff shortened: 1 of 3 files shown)


smnandre:
To clarify the terminology and avoid any misunderstandings in this conversation: 

Twig **already supports** _string literals_, as  `'foo'` and `"foo"` are both already... [string literals](https://en.wikipedia.org/wiki/String_literal).  To be precise they are quoted sequence of characters using single or double-quote paired delimiters, using `\\` as escaping prefix.

What you suggest here is about introducing _another_ syntax of string literals: with \` as delimiter (and still `\` as escaping prefix, but without the need to escape... itself).

In the Javascript world, the backtick is used in [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals), offering two things harder to do without:
* multi-lines strings
* string interpolation

These two points do not offer significant advantages for Twig, as it already provides mechanisms for these needs.

Therefore, I am not entirely convinced that the advantages (facilitating PHP execution in very specific cases) outweigh the costs (significant changes to the Lexer, increased maintenance, and additional documentation).

But maybe there are other features/benefits i don't see here?

---

PhpStorm already does offer support for the `constant` function in Twig, i'm sure if it does not support Enum yet, it will be done soon. And i guess similar plugin can be found for alternative IDEs.

twig-phpstorm-enum

ruudk:
Regarding the PHPStorm screenshot: is that native or via that Symfony plugin?

smnandre:
> Regarding the PHPStorm screenshot: is that native or via that Symfony plugin?

Via the Symfony plugin it seems (found this PR that seems very related to the current PR : https://github.com/Haehnchen/idea-php-symfony2-plugin/commit/b358ff0d010493f65f78f7ca9be8b2617c385d10 )

stof:
There was a discussion in the past about adding support for unescaped slashes in existing string literals (keeping the option to escape them as well), by producing a deprecation warning for any useless escaping in strings (`\d` in a Twig string currently means `d` as unknown escape sequence is treated as resolving to the unescaped character).
This would allow to avoid escaping the `\` except when the following character has a meaning in an escape sequence, like the way double-quoted string literals work in PHP (see https://3v4l.org/A2mPT).

Implementing this involves replacing `stripcslashes` by an implementation distinguishing useless escaping in https://github.com/twigphp/Twig/blob/6b6aca52063b5c632b08286c23122372be1dc49f/src/Lexer.php#L370

If this gets implemented, we might not need to introduce a third syntax for string literals in Twig.

smnandre:
> If this gets implemented, we might not need to introduce a third syntax for string literals in Twig.

~~But to get there you'd have to change the string regexp no ?~~

Edit: nope, that works

https://3v4l.org/RfnYM

fabpot:
I agree with @stof suggestion.

ruudk:
Should I try a PR for that?

fabpot:
> Should I try a PR for that?

That would be great!

ruudk:
Created a PR to attempt to solve this: 
- https://github.com/twigphp/Twig/pull/4176

fabpot:
Can we close this one @ruudk?

ruudk:
See https://github.com/twigphp/Twig/pull/4176

PR #4125, opened 2024-07-07
[Doc] Add missing Token types

smnandre:
* Token::ARROW_TYPE
* Token::SPREAD_TYPE


fabpot:
Good catch, thanks @smnandre.

PR #4126, opened 2024-07-07
fix grammar for article

xabbuh:


smnandre:
My bad 🇫🇷

fabpot:
Good catch, thanks @xabbuh.

PR #4127, opened 2024-07-08
draft: Fix issue with accessing null attribues for active record objects

trigger202:
Do not review yet, working on unit tests

### DIFF
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -1478,7 +1478,13 @@ function twig_get_attribute(Environment $env, Source $source, $object, $item, ar
 
     // object property
     if (/* Template::METHOD_CALL */ 'method' !== $type) {
-        if (isset($object->$item) || \array_key_exists((string) $item, (array) $object)) {
+//        if (isset($object->$item) || \array_key_exists((string) $item, (array) $object)) {
+        /**
+         * NOTE: this change is added fix issue when ActiveRecord object has the property but its value is null. e.g booking.agentId.
+         * `isset($object->$item` returns false when booking.Agentid is null - as result throws an exception
+         */
+
+        if (hasAttribute($object, $item) || \array_key_exists((string) $item, (array) $object)) {
             if ($isDefinedTest) {
                 return true;
             }
@@ -1655,4 +1661,29 @@ function twig_check_arrow_in_sandbox(Environment $env, $arrow, $thing, $type)
         throw new RuntimeError(sprintf('The callable passed to the "%s" %s must be a Closure in sandbox mode.', $thing, $type));
     }
 }
+
+    /**
+     * Check if object attribute regardless of the value, it should return true
+     * It should be able to handle when object is ActiveRecord and have null value for a column e.g $user.RoleId
+     *
+     * @param $object
+     * @param $attribute
+     * @return bool
+     */
+    function hasAttribute($object, $attribute)
+    {
+        if (!is_object($object)) {
+            return false;
+        }
+
+        try {
+            $object->$attribute;
+
+            return true;
+        } catch (Exception $exception) {
+            // do nothing
+        }
+
+        return false;
+    }
 }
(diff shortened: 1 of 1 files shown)


src/Extension/CoreExtension.php
```diff
@@ -1478,7 +1478,13 @@ function twig_get_attribute(Environment $env, Source $source, $object, $item, ar
 
     // object property
     if (/* Template::METHOD_CALL */ 'method' !== $type) {
-        if (isset($object->$item) || \array_key_exists((string) $item, (array) $object)) {
+//        if (isset($object->$item) || \array_key_exists((string) $item, (array) $object)) {
+        /**
+         * NOTE: this change is added fix issue when ActiveRecord object has the property but its value is null. e.g booking.agentId.
```
stof: A comment saying `this change` is totally useless, as the comment is not describing a change (commit messages do that) but the code in the class.

And keeping the old line as a comment is also totally useless (keeping the history is handled by git in a much better way)

stof (changes requested):
This should have tests to prevent  regressions

src/Extension/CoreExtension.php
trigger202: Thanks @stof for the feedback, I have removed it.

src/Extension/CoreExtension.php
```diff
@@ -1655,4 +1655,29 @@ function twig_check_arrow_in_sandbox(Environment $env, $arrow, $thing, $type)
...
+
+    /**
+     * Check if object attribute regardless of the value, it should return true
+     * It should be able to handle when object is ActiveRecord and have null value for a column e.g $user.RoleId
+     *
+     * @param $object
+     * @param $attribute
+     * @return bool
+     */
+    function hasAttribute($object, $attribute)
```
fabpot, line 1667:
```suggestion
    private function hasAttribute($object, $attribute)
```

src/Extension/CoreExtension.php
```diff
@@ -1655,4 +1655,29 @@ function twig_check_arrow_in_sandbox(Environment $env, $arrow, $thing, $type)
...
+    {
+        if (!is_object($object)) {
```
fabpot: This check can be removed as this is always an object at this point (see line 1455).

src/Extension/CoreExtension.php
```diff
@@ -1655,4 +1655,29 @@ function twig_check_arrow_in_sandbox(Environment $env, $arrow, $thing, $type)
...
+            return false;
+        }
+
+        try {
+            $object->$attribute;
+
+            return true;
+        } catch (Exception $exception) {
```
fabpot: Does PHP throw an exception when a property does not exist? That would be `\Exception` anyway, but I think it just emits a warning, right?

fabpot (changes requested):
Without extensive tests, it's not something we will merge.

src/Extension/CoreExtension.php
trigger202: I believe you're right, accessing random attributes didn't trigger an exception, thanks.

src/Extension/CoreExtension.php
```diff
@@ -1666,18 +1666,10 @@ function twig_check_arrow_in_sandbox(Environment $env, $arrow, $thing, $type)
...
-        }
-
-        try {
-            $object->$attribute;
-
-            return true;
-        } catch (Exception $exception) {
-            // do nothing
+        if (method_exists($object, 'hasAttribute')) {
+            return $object->hasAttribute($attribute);
```
smnandre, line 1674: This seems not something generic or universal ... or is this method part of some standard ?

src/Extension/CoreExtension.php
trigger202: making the method private is making all the exists tests fail
image

trigger202:
> Without extensive tests, it's not something we will merge.

Okay, just wondering how to turn this PR into a draft so I can ask for a code review once I feel its ready. if someone can point out how to make it a draft

src/Extension/CoreExtension.php
trigger202: I think most active record patterns have this method. Yii and Laravel do at least couldnt find anything for others

https://github.com/illuminate/database/blob/master/Eloquent/Concerns/HasAttributes.php#L45

fabpot: That's not something we will add to Twig core.

PR #4131, opened 2024-07-09
Introduce a Loop object

### LINKED ISSUE #4096: loop.last with generator should throw an exception
Hello,

I painfully discovered that `loop.last` is indeed impossible to access when your looping over a generator (after a couple of hours of painful debug), but I was wondering if there's any possibility for Twig to instead throw an exception when people are accessing to `loop.last` which is currently "undefined" (obviously) to make the error clearer about WHY is `loop.last` undefined :)

fabpot:
Fixes #4096


doc/tags/for.rst
```diff
@@ -76,9 +76,9 @@ Variable              Description
 
 .. note::
 
-    The ``loop.length``, ``loop.revindex``, ``loop.revindex0``, and
-    ``loop.last`` variables are only available for PHP arrays, or objects that
-    implement the ``Countable`` interface.
+    Accessing ``loop.length``, ``loop.revindex``, ``loop.revindex0``, and
+    ``loop.last`` variables o, ...  throws a ``RuntimeException`` when
```
stof: there is an extra `o` here

src/Runtime/Loop.php
```diff
@@ -0,0 +1,126 @@
...
+namespace Twig\Runtime;
+
+use Twig\Error\RuntimeError;
+
+/**
+ * Represents a for loop variable.
+ *
+ * @author Fabien Potencier <fabien@symfony.com>
+ */
+class Loop implements \Iterator
```
stof: should this be internal ?

src/Runtime/Loop.php
```diff
@@ -0,0 +1,126 @@
...
+    private int $index0;
+    private int $revindex0;
+    private bool $first;
+    private bool $last;
+    private int $length;
+
+    public function __construct($seq, private $parent)
+    {
+        $this->seq = is_iterable($seq) ? (is_array($seq) ? new \ArrayIterator($seq) : $seq) : new \ArrayIterator([]);
+        $this->rewind();
```
stof, line 36: wouldn't this cause issues with a Generator because we would rewind the iterator twice (once in the constructor and once when the `foreach` loop calls the `rewind` method) while generators are not rewindable ?

We should have tests iterating over a Generator in Twig.

src/Runtime/Loop.php
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    public function rewind(): void
+    {
+        $this->seq->rewind();
+
+        $this->iterated = false;
+        $this->index0 = 0;
+        $this->first = true;
+        if (is_array($this->seq) || (is_object($this->seq) && $this->seq instanceof \Countable)) {
```
stof: `is_object` is useless. The `instanceof` operator will return `false` for non-objects

stof: `is_array` won't ever be true here, due to the property type.

src/Runtime/Loop.php
fabpot: Correct, same for LoopContext.

src/Runtime/Loop.php
fabpot: Right, now that I always convert to an \Iterator, I can simplify this a lot.

src/Runtime/Loop.php
fabpot: I've just added a test with a Generator.

doc/tags/for.rst
```diff
@@ -76,9 +76,9 @@ Variable              Description
 
 .. note::
 
-    The ``loop.length``, ``loop.revindex``, ``loop.revindex0``, and
-    ``loop.last`` variables are only available for PHP arrays, or objects that
-    implement the ``Countable`` interface.
+    When the underlying PHP iterator is not countable, the ``loop.length``,
+    ``loop.revindex``, ``loop.revindex0``, and ``loop.last`` variables are not
+    available and a ``RuntimeException`` is thrown if you retry to use them.
```
stof:
```suggestion
    available and a ``RuntimeException`` is thrown if you try to use them.
```

src/Runtime/Loop.php
stof: OK, I just discovered that rewinding a Generator multiple times is totally fine, as long as you do that before using `next`

stof:
I just realized that by using a Loop object, assigning it to a different variable (like `outer_loop` will now work fine, making it much easier to reference the loop context of an outer loop when using nested loops (especially when the number of nested loops can change) because it removes the need for `loop.parent.loop.parent.loop`:

```twig
{% for item in items %}
    {% set outer_loop = loop %}
    {% for property in item.properties %}
        {% for values in property.values %}
            {{ outer_loop.index0}} {{property.name}} {{loop.index0}}: {{value}}
        {% endfor %}
    {% endfor %}
{% endfor %}
```

And this makes it more similar to the way loops work in Jinja (this code snippet would also be valid Jinja code, while `loop.parent` is not)

PR #4135, opened 2024-07-12
Add `loop.changed`, `loop.previous`, `loop.next`, and `loop.cycle` variables

fabpot:


doc/tags/for.rst
```diff
@@ -66,14 +66,52 @@ Variable              Description
...
+``loop.next``         The value from the next iteration (``null`` for the first iteration)
+===================== ========================================================================
 
 .. code-block:: twig
 
     {% for user in users %}
         {{ loop.index }} - {{ user.username }}
     {% endfor %}
 
+Use ``loop.cycle`` to cycle among a list of values:
```
stof, line 81: Do we need to add this one give we already implemented the standalone cycle filter (that was added later in Jinja) ?

doc/tags/for.rst
```diff
@@ -66,14 +66,52 @@ Variable              Description
...
+        {% if not loop.first and value > loop.previous %}
+            The value just increased!
+        {% endif %}
+        {{ value }}
+        {% if not loop.last and loop.next > value %}
+            The value will increase even more!
+        {% endif %}
+    {% endfor %}
+
+Use ``loop.changed`` to check if the value has changed since the last iteration:
```
stof: `since the last call` technically (it is totally possible not to call that for each iteration, or to call it multiple times in the same iteration, even though this is not the expected usage)

doc/tags/for.rst
```diff
@@ -66,14 +66,52 @@ Variable              Description
...
+Use ``loop.cycle`` to cycle among a list of values:
+
+.. code-block:: html+twig
+
+    {% for row in rows %}
+        <li class="{{ loop.cycle('odd', 'even') }}">{{ row }}</li>
+    {% endfor %}
+
+Use ``loop.previous`` and ``loop.next`` to compare the current value with the
+previous or next values:
```
stof: I would say `to access the previous or next values`, as you might use them to do something else than comparing them to the current value.

And I would also document that `loop.previous` will be `null` when being at the first item and `loop.next` will be `null` for the last item.

src/Runtime/LoopIterator.php
```diff
@@ -35,33 +37,38 @@ public function __construct($seq)
...
+        return $this->current['key'];
     }
 
     public function next(): void
     {
-        if ($this->peek) {
-            $this->peek = false;
+        $this->previous = $this->current;
+        if ($this->next) {
+            $this->next = [];
```
stof: should we represent that case as `null` instead of an empty array ? This would allow describing the type more precisely by using an array shape, by having the type be `array{valid: bool, key: mixed, value: mixed}|null`, which would help for static analysis of the code.

doc/tags/for.rst
```diff
@@ -55,9 +55,9 @@ The ``loop`` variable
 
 Inside of a ``for`` loop block you can access some special variables:
```
stof, line 56: Describing `loop.changed` as a variable looks weird to me, as `changed` is not a getter but a method requiring an argument.

Anyway, from a strict point of view, there is only 1 special variable: `loop`. What this table describes is the API of this loop variable.

doc/tags/for.rst
fabpot: It's a shortcut. Maybe it's not needed, but I find it nice to not have to care about the second arg.

doc/tags/for.rst
fabpot: Done

src/Runtime/LoopIterator.php
fabpot: Good idea

doc/tags/for.rst
fabpot: True

stof (approved):

doc/tags/for.rst
```diff
@@ -66,14 +66,54 @@ Variable              Description
 ``loop.last``         True if last iteration
 ``loop.length``       The number of items in the sequence
 ``loop.parent``       The parent context
-===================== =============================================================
+``loop.cycle``        Cycle over a sequence of values
+``loop.changed``      True if previously called with a different value or if not called yet
+``loop.previous``     The value from the previous iteration (``null`` for the first iteration)
+``loop.next``         The value from the next iteration (``null`` for the first iteration)
```
willrowe, line 72:
```diff
-``loop.next``         The value from the next iteration (``null`` for the first iteration)
+``loop.next``         The value from the next iteration (``null`` for the last iteration)
```

fabpot: It has already been fixed in #4142 :)

PR #4137, opened 2024-07-13
User first class callable syntax

fabpot:


extra/html-extra/HtmlExtension.php
```diff
@@ -36,7 +36,7 @@ public function getFilters(): array
     public function getFunctions(): array
     {
         return [
-            new TwigFunction('html_classes', [self::class, 'htmlClasses']),
+            new TwigFunction('html_classes', self::class::htmlClasses(...)),
```
GromNaN: There are many changes like this:
```suggestion
            new TwigFunction('html_classes', self::htmlClasses(...)),
```

GromNaN (changes requested):

GromNaN (approved):
For the failing test, you can fix `Twig\Parser::subparse()` with something like this:

```php
$tokenParser = match (true) {
    \is_array($test) => $test[0],
    $test instanceof \Closure => (new \ReflectionFunction($test))->getClosureThis(),
    default => null,
};
if ($tokenParser instanceof TokenParserInterface) {
    $e->appendMessage(\sprintf(' (expecting closing tag for the "%s" tag defined near line %s).', $tokenParser->getTag(), $lineno));
}
```

fabpot:
> For the failing test, you can fix `Twig\Parser::subparse()` with something like this:
> 
> ```
> $tokenParser = match (true) {
>     \is_array($test) => $test[0],
>     $test instanceof \Closure => (new \ReflectionFunction($test))->getClosureThis(),
>     default => null,
> };
> if ($tokenParser instanceof TokenParserInterface) {
>     $e->appendMessage(\sprintf(' (expecting closing tag for the "%s" tag defined near line %s).', $tokenParser->getTag(), $lineno));
> }
> ```

I have added native support instead.

PR #4141, opened 2024-07-16
Fix support for IteratorAggregate and EmptyIterator objects in loops

fabpot:


src/Runtime/LoopIterator.php
```diff
@@ -34,7 +34,18 @@ final class LoopIterator implements \Iterator
...
-        $this->seq = is_iterable($seq) ? (is_array($seq) ? new \ArrayIterator($seq) : $seq) : new \ArrayIterator([]);
+        if (is_array($seq)) {
+            $this->seq = new \ArrayIterator($seq);
+        } elseif ($seq instanceof \IteratorAggregate) {
+            do {
+                $seq = $seq->getIterator();
+            } while ($seq instanceof \IteratorAggregate);
+            $this->seq = $seq;
+        } elseif (is_iterable($seq)) {
+            $this->seq = $seq;
```
smnandre, line 45: Here $seq is more than probably `Iterator` or `IteratorAggregate` **but** that's almost "internal"...

> Internal (built-in) classes that implement this interface can be used in a [foreach](https://www.php.net/manual/en/control-structures.foreach.php) construct and **do not need to implement** [IteratorAggregate](https://www.php.net/manual/en/class.iteratoraggregate.php) or [Iterator](https://www.php.net/manual/en/class.iterator.php).

Not sure if this could happen IRL .. but maybe a `elseif ($seq instanceof \Iterator) ` would be more precise here ?

src/Runtime/LoopIterator.php
```diff
@@ -34,7 +34,18 @@ final class LoopIterator implements \Iterator
...
     public function __construct($seq)
     {
-        $this->seq = is_iterable($seq) ? (is_array($seq) ? new \ArrayIterator($seq) : $seq) : new \ArrayIterator([]);
+        if (is_array($seq)) {
+            $this->seq = new \ArrayIterator($seq);
+        } elseif ($seq instanceof \IteratorAggregate) {
+            do {
+                $seq = $seq->getIterator();
+            } while ($seq instanceof \IteratorAggregate);
+            $this->seq = $seq;
```
smnandre, line 43: PHPStorm was not happy so i looked a bit and...

 [IteratorAggregate::getIterator()](https://www.php.net/manual/en/iteratoraggregate.getiterator.php) does not return an Iterator but only a Traversable ... 
 
 so after the loop here $this->seq could totally be a simple array
 
 
Would something like this work 'out of the box' ? 

```
if (is_array($seq))
 // ArrayIterator($seq)
elseif ($seq instanceof Traversable)
  // IteratorIterator ($seq)
else 
  // EmptyIterator

```

fabpot: The best would be to create a PR with examples of something that fails and the fix for those tests.

smnandre: Well i spend time and clearly IRL that won't ever happen. 

The last possibilities would be those classes PHP say can be used in "foreach" while not implementing "Iterator" directly, but i cannot find an example.

Sorry for the waste of time here.

src/Runtime/LoopIterator.php
```diff
@@ -34,7 +34,18 @@ final class LoopIterator implements \Iterator
 
     public function __construct($seq)
     {
-        $this->seq = is_iterable($seq) ? (is_array($seq) ? new \ArrayIterator($seq) : $seq) : new \ArrayIterator([]);
+        if (is_array($seq)) {
+            $this->seq = new \ArrayIterator($seq);
+        } elseif ($seq instanceof \IteratorAggregate) {
```
stof, line 39: There is a much simpler way: using `IteratorIterator`

src/Runtime/LoopIterator.php
```diff
@@ -63,8 +74,13 @@ public function next(): void
     public function rewind(): void
     {
         $this->seq->rewind();
-        $this->previous = ['valid' => false, 'key' => null, 'value' => null];
-        $this->current = ['valid' => $this->seq->valid(), 'key' => $this->seq->key(), 'value' => $this->seq->current()];
+        if ($this->seq->valid()) {
+            $this->previous = ['valid' => false, 'key' => null, 'value' => null];
```
stof, line 78: shouldn't we always assign `previous` ?

smnandre: That was something i wonder, but there is only one situation getPrevious() can be called without beeing set, it's if someone call "loop.previous" in the `else` part of its `for`  ... `else` .... `endfor`.

Because the node inside the `for` block can only be called if there is _something_ in the loop (so when this->seq->isValid() is true)

stof: and in such case, we still want it to be able to return `null` as value, not a PHP error

fabpot: See #4169

PR #4148, opened 2024-07-21
Deprecate NameExpression internal methods

smnandre:
Deprecate two methods from `Twig\Node\Expression\NameExpression`:
* `isSimple()`: did not find any usage in Twig (or even Symfony) repositories
* `isSpecial()`: inlined for better performance ($name is available in the method)


fabpot:
Have you checked some CMSes using Twig?

smnandre:
Only found it in repositories as hard-copies (embeded Twig source code) of Twig_NodeVisitor_Optimizer or Twig_Node_Expression_Name  (or namespaced equivalents)

Did not find a recent case (Twig 2+) of these methods beeing called, but i may have missed something there... 

Any particular CMS i could check more in depth ?

fabpot (approved):

fabpot:
Thank you @smnandre.

PR #4149, opened 2024-07-23
Remove an unused arg to FilterExpression::__construct()

fabpot:


xabbuh (approved):

src/Node/Expression/FilterExpression.php
```diff
@@ -17,9 +17,9 @@
 
 class FilterExpression extends CallExpression
 {
-    public function __construct(Node $node, ConstantExpression $filterName, Node $arguments, int $lineno, ?string $tag = null)
+    public function __construct(Node $node, ConstantExpression $filterName, Node $arguments, int $lineno)
     {
-        parent::__construct(['node' => $node, 'filter' => $filterName, 'arguments' => $arguments], [], $lineno, $tag);
+        parent::__construct(['node' => $node, 'filter' => $filterName, 'arguments' => $arguments], [], $lineno);
```
smnandre, line 22: I think you'll need to remove it from the DefaultFilter too

```
parent::__construct($node, $filterName, $arguments, $lineno, $tag);
```

https://github.com/fabpot/Twig/blob/filter-expr-unneeded-arg/src/Node/Expression/Filter/DefaultFilter.php#L45

fabpot:
Not worth it.

PR #4153, opened 2024-07-27
Add support for recursive loops

### LINKED ISSUE #3416: Feature idea: Recursive loop

### LINKED ISSUE #4144: Add tree tag to Twig 4.0
I think the tree tag for recursing data structures where the children have the same structure as the parent nodes is pretty useful.

```twig
{% tree item in menu %}
  {% if treeloop.first %}<ul>{% endif %}
    <li>
        <a href="{{ item.url }}">{{ item.name }}</a>
        {% subtree item.children %}
    </li>
  {% if treeloop.last %}</ul>{% endif %}
{% endtree %}
```

Just like a `{% for %}` loop, you can access the key of each list item:
```twig
{% tree key, item in menu %}
  <li>
    <b>Item {{ key }}</b>: {{ item.name }}
    {% subtree item.children %}
  </li>
{% endtree %}
```

Since Twig 4 is under development, would you consider adding this extension to the Twig core?

https://github.com/tacman/twig-tree-tag

I can clean up the documentation and such a bit if that helps.  I didn't write the original or even the first set of tweaks, but I did get it working with Twig 3 and the dev version of Twig 4.

fabpot:
Fixes #3416
Fixes #4144


smnandre:
Should loop(loop.parent) be forbidden / prevented ?

(or any infinite récursion if this syntax has not the effect i anticipate)

fabpot:
> Should loop(loop.parent) be forbidden / prevented ?
> 
> (or any infinite récursion if this syntax has not the effect i anticipate)

I don't think we need to do anything here. Sure, you can create an infinite loop, but there will always be ways to mess up.

fabpot:
> Should loop(loop.parent) be forbidden / prevented ?
> 
> (or any infinite récursion if this syntax has not the effect i anticipate)

My first reaction was that we should not do anything (PHP does nothing to prevent such infinite loops for instance). But as this is a templating engine that can be used by end users that you don't control, I've limited the depth to 50 (50 is kind of arbitrary, but maybe a good balance between valid use cases and malicious usage of a recursive loop).

doc/tags/for.rst
```diff
@@ -58,15 +58,17 @@ Inside of a ``for`` loop block you can access some special variables:
...
-``loop.index0``       The current iteration of the loop. (0 indexed)
+``loop.index``        The current iteration of the loop (1 indexed)
+``loop.index0``       The current iteration of the loop (0 indexed)
 ``loop.revindex``     The number of iterations from the end of the loop (1 indexed)
 ``loop.revindex0``    The number of iterations from the end of the loop (0 indexed)
 ``loop.first``        True if first iteration
 ``loop.last``         True if last iteration
 ``loop.length``       The number of items in the sequence
 ``loop.parent``       The parent context
 ``loop.cycle``        Cycle over a sequence of values
```
smnandre: Should this one be removed this list, as it's not a variable but a function ? (#nitpicking)

(the only one from this list i believe)

smnandre: (i tried to point a the `loop.cycle` line ... and failed miserably 😅 )

fabpot: See #4159

src/Runtime/LoopContext.php
```diff
@@ -25,6 +25,9 @@ final class LoopContext
     public function __construct(
         private LoopIterator $loop,
         private $parent,
+        private $blocks,
+        private $recurseFunc,
+        private $depth,
```
stof, line 30: shouldn't those be typed ?

xabbuh: see #4168

PR #4167, opened 2024-07-30
Fix various phpstan errors

fabpot:


src/Extension/ExtensionInterface.php
```diff
@@ -68,8 +67,8 @@ public function getFunctions();
      * @return array<array> First array of unary operators, second array of binary operators
      *
      * @psalm-return array{
-     *     array<string, array{precedence: int, class: class-string<AbstractUnary>}>,
-     *     array<string, array{precedence: int, class: class-string<AbstractBinary>, associativity: ExpressionParser::OPERATOR_*}>
+     *     array<string, array{precedence: int, class: class-string<AbstractExpression>}>,
+     *     array<string, array{precedence: int, class?: class-string<AbstractExpression>, associativity: ExpressionParser::OPERATOR_*}>
```
stof, line 71: why is the class optional for binary operators but mandatory for unary ones ?

And should we really widen the types there ?

src/Runtime/EscaperRuntime.php
```diff
@@ -17,6 +17,7 @@
 
 final class EscaperRuntime implements RuntimeExtensionInterface
 {
+    /** @var array<callable(string $string, string $charset): string> */
```
stof:
```suggestion
    /** @var array<string, callable(string $string, string $charset): string> */
```

This way, phpstan will be aware of the type of array keys

src/Runtime/EscaperRuntime.php
```diff
@@ -46,7 +47,7 @@ public function setEscaper($strategy, callable $callable)
     /**
      * Gets all defined escapers.
      *
-     * @return array<callable(string $string, string $charset)> An array of escapers
+     * @return array<callable(string $string, string $charset): string> An array of escapers
```
stof:
```suggestion
     * @return array<string, callable(string $string, string $charset): string> An array of escapers
```

src/Test/NodeTestCase.php
```diff
@@ -53,7 +48,7 @@ protected function getCompiler(?Environment $environment = null)
 
     protected function getEnvironment()
     {
-        return $this->currentEnv = new Environment(new ArrayLoader());
```
stof, line 56: maybe it was meant to be `??=` instead, to reuse the same instance when called multiple times in a test

src/Extension/ExtensionInterface.php
fabpot: Yes, we do want to widen the types. AbstractUnary and AbstractBinary have nothing special, the only thing we care about is the fact they are AbstractExpression.

For `??`, the class is not an AbstractBinary and that's fine:

```
'??' => ['precedence' => 300, 'class' => NullCoalesceExpression::class, 'associativity' => ExpressionParser::OPERATOR_RIGHT],
```

For `is` and `is not`, we don't have a class as it's handled by `ExpressionParser`:

```
'is' => ['precedence' => 100, 'associativity' => ExpressionParser::OPERATOR_LEFT],
'is not' => ['precedence' => 100, 'associativity' => ExpressionParser::OPERATOR_LEFT],
```

src/Test/NodeTestCase.php
fabpot: It was used when we had more methods reusing the env, but it's not that useful anymore. Maybe we can indeed keep the "cache".

PR #4169, opened 2024-07-31
Fix usage of loop in a for else clause

fabpot:


tests/Fixtures/tags/for/loop_in_else.test
```diff
@@ -0,0 +1,25 @@
...
+{% for item in [] %}
+{% else %}
+{{ loop.previous is null ? 'OK' : 'KO' }}
+{{ loop.next is null ? 'OK' : 'KO' }}
+{{ loop.first ? 'OK' : 'KO' }}
+{{ loop.last ? 'OK' : 'KO' }}
+{{ loop.index0 is same as 0 ? 'OK' : 'KO' }}
+{{ loop.revindex0 is same as 0 ? 'OK' : 'KO' }}
+{{ loop.length is same as 0 ? 'OK' : 'KO' }}
+{{ loop.depth is same as 1 ? 'OK' : 'KO' }}
```
fabpot, line 13: Should be 0 maybe?

tests/Fixtures/tags/for/loop_in_else.test
```diff
@@ -0,0 +1,29 @@
+--TEST--
+"for" tag loop behavior in an else clause
+--TEMPLATE--
+{% for iterator in iterators %}
+{% for item in iterator %}
+{% else %}
+{{ loop.previous is null ? 'OK' : 'KO' }}
+{{ loop.next is null ? 'OK' : 'KO' }}
+{{ loop.first ? 'OK' : 'KO' }}
```
fabpot, line 8: Should be false?

tests/Fixtures/tags/for/loop_in_else.test
```diff
@@ -0,0 +1,29 @@
...
+{{ loop.last ? 'OK' : 'KO' }}
```
fabpot, line 9: Should be false?

tests/Fixtures/tags/for/loop_in_else.test
stof: to me, it is fine to be 1. This is the depth of the current loop (even if it was empty)

tests/Fixtures/tags/for/loop_in_else.test
smnandre: 👍  False seems more logical  to me (it's "not even" the first iteration)

tests/Fixtures/tags/for/loop_in_else.test
smnandre: 👍  False seems more logical  to me (it's not "an" iteration, so not the last one)

stof: I would say that any usage of the `loop` variable inside the `else` block is non-sense anyway. Getting any of those boolean value would be fine IMO (a PHP error due to an uninitialized property in Twig is not)

PR #4171, opened 2024-08-04
Introduce ChainCache and ReadOnlyFilesystemCache

okhoshi:
Following the feedback in symfony/symfony#54384, I’m proposing the two new Cache implementations in Twig directly.

### DIFF
--- a/src/Cache/ChainCache.php
+++ b/src/Cache/ChainCache.php
@@ -1,0 +1,81 @@
+<?php
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Twig\Cache;
+
+/**
+ * Chains several caches together.
+ *
+ * Cached items are fetched from the first cache having them in its data store.
+ * They are saved and deleted in all adapters at once.
+ *
+ * @author Quentin Devos <quentin@devos.pm>
+ */
+final class ChainCache implements CacheInterface
+{
+    private $caches;
+
+    /**
+     * @param iterable<CacheInterface> $caches The ordered list of caches used to store and fetch cached items
+     */
+    public function __construct(iterable $caches)
+    {
+        $this->caches = $caches;
+    }
+
+    public function generateKey(string $name, string $className): string
+    {
+        return $className.'#'.$name;
+    }
+
+    public function write(string $key, string $content): void
+    {
+        $splitKey = $this->splitKey($key);
+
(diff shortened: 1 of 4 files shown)


src/Cache/ChainCache.php
```diff
@@ -0,0 +1,79 @@
...
+
+    public function generateKey(string $name, string $className): string
+    {
+        return $className.'#'.$name;
+    }
+
+    /**
+     * @return string[]
+     */
+    private function splitKey(string $key): array
```
fabpot: private methods should be move to the end of the class

src/Cache/ChainCache.php
```diff
@@ -0,0 +1,79 @@
...
+
+        foreach ($this->caches as $cache) {
+            if (0 < $timestamp = $cache->getTimestamp($cache->generateKey(...$splitKey))) {
+                return $timestamp;
+            }
+        }
+
+        return 0;
+    }
+}
```
fabpot: Missing empty line at the end of the file

src/Cache/ReadOnlyFilesystemCache.php
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * @author Quentin Devos <quentin@devos.pm>
+ */
+class ReadOnlyFilesystemCache extends FilesystemCache
+{
+    public function write(string $key, string $content): void
+    {
+        // Do nothing with the content, it's a read-only filesystem.
+    }
+}
```
fabpot: Missing empty line at the end of the file

fabpot (changes requested):
As Twig 3.x is still compatible with PHP 7.2, you need to adjust the code accordingly.

src/Cache/ChainCache.php
okhoshi: Moved

okhoshi:
> As Twig 3.x is still compatible with PHP 7.2, you need to adjust the code accordingly.

Hmm indeed, I missed that. Corrected 👍

src/Cache/ChainCache.php
```diff
@@ -0,0 +1,80 @@
...
+ * @author Quentin Devos <quentin@devos.pm>
+ */
+final class ChainCache implements CacheInterface
+{
+    private $caches;
+
+    /**
+     * @param iterable<CacheInterface> $caches The ordered list of caches used to store and fetch cached items
+     */
+    public function __construct(iterable $caches) {
```
fabpot:
```suggestion
    public function __construct(iterable $caches)
    {
```

fabpot:
Thank you @Okhoshi.

PR #4172, opened 2024-08-05
Deprecate the possibility to override a function with a macro

fabpot:


fabpot:
It means that in 4.x, an app can break when upgrading to a minor version of Twig when a new function is added and the app happens to import a macro with the same name.

Maybe we should deprecate importing a macro as a "function" or maybe we should have another way to call macros to avoid confusion.

fabpot:
And the current behavior is the same as Jinja.

stof:
I would rather keep the current behavior which avoids any breakage during upgrades (the function-like call to a macro is statically analyzable as it requires importing the macro in the same template, not in a parent or child template, so there is no risk of getting surprises for shadowing that appear during upgrades)

PR #4173, opened 2024-08-05
Add support for named arguments on special functions

fabpot:


src/ExpressionParser.php
```diff
@@ -858,4 +853,38 @@ private function checkConstantExpression(Node $node): bool
 
         return true;
     }
+
+    private function parseCalledArguments($arguments, array $names, int $required, string $callType, string $callName, int $line): array
+    {
+        if (count($arguments) > count($names)) {
+            throw new SyntaxError(sprintf('Too many arguments for %s "%s" (it accepts only %d arguments).', $callType, $callName, count($names)), $line, $this->parser->getStream()->getSourceContext());
+        } elseif (count($arguments) < $required) {
```
stof: this check is not enough when using named arguments. For instance, `attribute` has 2 required arguments and 1 optional argument. If I pass `variable` and `arguments`, I still don't have less arguments than the number of required arguments, but I don't fill all required arguments.

stof (changes requested):

src/ExpressionParser.php
```diff
@@ -858,4 +853,38 @@ private function checkConstantExpression(Node $node): bool
...
+            throw new SyntaxError(sprintf('Not enough arguments for %s "%s" (it requires at least "%s").', $callType, $callName, implode('", ', array_slice($names, 0, $required))), $line, $this->parser->getStream()->getSourceContext());
```
smnandre:
```suggestion
            throw new SyntaxError(sprintf('Not enough arguments for %s "%s" (it requires at least "%s").', $callType, $callName, implode('", "', array_slice($names, 0, $required))), $line, $this->parser->getStream()->getSourceContext());
```

src/ExpressionParser.php
```diff
@@ -858,4 +853,38 @@ private function checkConstantExpression(Node $node): bool
...
+        }
+
+        $args = [];
+        $named = false;
+        foreach ($arguments as $name => $node) {
+            if (!\is_int($name)) {
+                $named = true;
+                $name = $this->normalizeName($name);
+                if (!in_array($name, $names)) {
+                    throw new SyntaxError(\sprintf('Unexpected argument "%s" for %s "%s" (expecting one of "%s").', $name, $callType, $callName, implode('", ', $names)), $line, $this->parser->getStream()->getSourceContext());
```
smnandre:
```suggestion
                    throw new SyntaxError(\sprintf('Unexpected argument "%s" for %s "%s" (expecting one of "%s").', $name, $callType, $callName, implode('", "', $names)), $line, $this->parser->getStream()->getSourceContext());
```

tests/Fixtures/functions/attribute_with_wrong_args.test
```diff
@@ -0,0 +1,8 @@
+--TEST--
+"attribute" function
+--TEMPLATE--
+{{ attribute(var=var, template="tpl") }}
+--DATA--
+return ['var' => null]
+--EXCEPTION--
+Twig\Error\SyntaxError: Unexpected argument "var" for function "attribute" (expecting one of "variable", attribute", arguments") in "index.twig" at line 2.
```
smnandre:
```suggestion
Twig\Error\SyntaxError: Unexpected argument "var" for function "attribute" (expecting one of "variable", "attribute", "arguments") in "index.twig" at line 2.
```

src/ExpressionParser.php
fabpot: Fixed

PR #4175, opened 2024-08-07
Add a note about how macros can override existing functions

fabpot:


stof:
I think the term used generally for that is `shadowing the function` (it hides the function from being called, but does not really override the function itself)

Issue #2712, opened 2018-06-21
[Proposal] Deprecate the usage of unescaped backslashes not belonging to an escape sequence


stof:
Currently, using a backslash without escaping it and without being part of an escape sequence is silently ignored (as if any unknown escape sequence `\<char>` would produce `<char>`).
This is often confusing for PHP developers (see #2711), as this is not how PHP works in single-quoted strings (it considers it as a backslash).

I suggest triggering a (compile-time) deprecation warning for such case (and turn it into a syntax error in 3.0).
Making it produce a backslash would be a BC break (as it would change existing strings). Thus, it would make it impossible to add new escape sequences without BC break.

stof:
The implementation actually relies on PHP to handle this, thanks to https://php.net/stripcslashes so this might not be easy to implement actually.

Maybe using `$rawSource === addcslashes($strippedSource)` could allow detecting this case though for the deprecation case (this will not make it easy to implement keeping the backslash, but this is not my proposal anyway)

stof:
hmm, `addcslashes` is a pain, as it requires passing it a charlist, but `stripcslashes` has a list of builtin escape sequences, which would have to be duplicated...

stof:
and the case of `\x<code>` will also be a pain...

fabpot:
@stof Are you working on a pull request?

stof:
Well, due to the way `addcslashes` and `stripcslashes` are not symmetrical, it won't be easy to implement. https://3v4l.org/4jf9N shows me that I can detect things for most cases (while keeping the usage of the native implementation of `stripcslashes`) by hardcoding the list of built-in chars plus our own escaped chars (`\`, `'` and `"`), but the handling of `\x<code>` will be a pain (see the last case which fails while it should not).

So either we also deprecate that support, or we reimplement `stripcslashes` in userland in the lexer, allowing us to trigger the deprecation directly in the processing (instead of having to detect it after the fact)

fabpot:
We should see if reimplementing cstripcslashes in userland would make lexing much slower for typical templates.

stof:
do you have an existing benchmark for the Twig template compilation that I could reuse ?

stof:
I found https://github.com/twigphp/twig-perf but this does not seem to have "typical" templates, as it does not have much usage of Twig strings there (and so no usage of `stripcslashes` that would be replaced)

fabpot:
@stof What do we do here?

stof:
[...]

Issue #3604, opened 2021-12-19
Twig strings don't support PHP string escape sequences


GPHemsley:
I'm trying to insert a non-breaking space (NBSP) into a string in a Twig template, but all the ways I've thought to try haven't worked.

And upon further research it seems that Twig strings do not support the full gamut of escape sequences that PHP supports:
https://www.php.net/manual/en/language.types.string.php

For example, this code:
```twig
{{ "\240\xA0\u{00A0}&nbsp;" }}
```

produces:
```html
��u{00A0}&nbsp;
```

instead of:
```html
    
```

I've elaborated some here:
https://twigfiddle.com/q41fr9

As such, I have two questions:
(1) Is this expected behavior?
(2) Is there a way to do what I want to do that I'm missing?

stof:
Twig relies on [`stripcslashes`](https://php.net/stripcslashes) which indeed does not cover the PHP escape sequences but only the C ones.

PR #4176, opened 2024-08-07
Deprecate unnecessary escape characters

ruudk:
This is a first attempt at solving #4123 and #2712.

Currently it writes the deprecations to an array in the Lexer. This is probably not the way to do it. Should we directly trigger `E_USER_DEPRECATED` errors?

/cc @stof @fabpot Let me know what you think 😊

### DIFF
--- a/doc/templates.rst
+++ b/doc/templates.rst
@@ -522,10 +522,25 @@ exist:
 * ``"Hello World"``: Everything between two double or single quotes is a
   string. They are useful whenever you need a string in the template (for
   example as arguments to function calls, filters or just to extend or include
-  a template). A string can contain a delimiter if it is preceded by a
-  backslash (``\``) -- like in ``'It\'s good'``. If the string contains a
-  backslash (e.g. ``'c:\Program Files'``) escape it by doubling it
-  (e.g. ``'c:\\Program Files'``).
+  a template).
+
+  Note that certain characters require escaping:
+   * ``\f``: Form feed
+   * ``\n``: New line
+   * ``\r``: Carriage return
+   * ``\t``: Horizontal tab
+   * ``\v``: Vertical tab
+   * ``\x``: Hexadecimal escape sequence
+   * ``\0`` to ``\377``: Octal escape sequences representing characters
+   * ``\``: Backslash
+  
+   When using single-quoted strings, the single quote character (``'``) needs to be escaped with a backslash (``\'``).
+   When using double-quoted strings, the double quote character (``"``) needs to be escaped with a backslash (``\"``).
+
+   For example, a single quoted string can contain a delimiter if it is preceded by a
+   backslash (``\``) -- like in ``'It\'s good'``. If the string contains a
+   backslash (e.g. ``'c:\Program Files'``) escape it by doubling it
+   (e.g. ``'c:\\Program Files'``).
 
 * ``42`` / ``42.23``: Integers and floating point numbers are created by
   writing the number down. If a dot is present the number is a float,
--- a/src/Lexer.php
+++ b/src/Lexer.php
@@ -367,7 +367,7 @@ class Lexer
         }
         // strings
         elseif (preg_match(self::REGEX_STRING, $this->code, $match, 0, $this->cursor)) {
-            $this->pushToken(/* Token::STRING_TYPE */ 7, stripcslashes(substr($match[0], 1, -1)));
+            $this->pushToken(/* Token::STRING_TYPE */ 7, $this->stripcslashes(substr($match[0], 1, -1), substr($match[0], 0, 1)));
             $this->moveCursor($match[0]);
         }
         // opening double quoted string
@@ -380,6 +380,81 @@ class Lexer
         else {
             throw new SyntaxError(\sprintf('Unexpected character "%s".', $this->code[$this->cursor]), $this->lineno, $this->source);
         }
+    }
+
+    private function stripcslashes(string $str, string $quoteType): string
+    {
+        if (!str_contains($str, '\\')) {
+            return $str;
+        }
+
+        $result = '';
+        $length = strlen($str);
+        $specialChars = [
+            'f' => "\f",
+            'n' => "\n",
+            'r' => "\r",
+            't' => "\t",
+            'v' => "\v",
+        ];
+
+        for ($i = 0; $i < $length; $i++) {
+            if ($str[$i] !== '\\' || $i + 1 >= $length) {
+                $result .= $str[$i];
+
+                continue;
+            }
+
+            $nextChar = $str[$i + 1];
+            if (isset($specialChars[$nextChar])) {
+                $result .= $specialChars[$nextChar];
+                $i++;
+            } elseif ($nextChar === '\\') {
+                $result .= '\\';
+                $i++;
+            } elseif ($nextChar === "'" || $nextChar === '"') {
+                if ($nextChar !== $quoteType) {
+                    trigger_deprecation('twig/twig', '3.12', 'Character "%s" at position %d does not need to be escaped anymore.', $nextChar, $i + 2);
+                }
+                $result .= $nextChar;
+                $i++;
+            } elseif ($nextChar === '#' && $i + 2 < $length && $str[$i + 2] === '{') {
+                $result .= '#{';
+                $i += 2;
+            } elseif ($nextChar === 'x' && $i + 2 < $length && ctype_xdigit($str[$i + 2])) {
+                $hex = $str[$i + 2];
+                if ($i + 3 < $length && ctype_xdigit($str[$i + 3])) {
+                    $hex .= $str[$i + 3];
+                    $i++;
+                }
+                $result .= chr(hexdec($hex));
+                $i += 2;
+            } elseif (ctype_digit($nextChar) && $nextChar < '8') {
+                $octal = $nextChar;
+                for ($j = $i + 1; $j <= 3; $j++) {
+                    $o = $j + 1;
+                    if ($o >= $length) {
+                        break;
+                    }
+                    if (!ctype_digit($str[$o])) {
+                        break;
+                    }
+                    if ($str[$o] >= '8') {
+                        break;
+                    }
+                    $octal .= $str[$o];
+                    $i++;
+                }
+                $result .= chr(octdec($octal));
+                $i++;
+            } else {
+                trigger_deprecation('twig/twig', '3.12', 'Character "%s" at position %d does not need to be escaped anymore.', $nextChar, $i + 2);
+                $result .= $nextChar;
+                $i++;
+            }
+        }
+        
+        return $result;
     }
 
     private function lexRawData(): void
@@ -423,7 +498,7 @@ class Lexer
             $this->moveCursor($match[0]);
             $this->pushState(self::STATE_INTERPOLATION);
         } elseif (preg_match(self::REGEX_DQ_STRING_PART, $this->code, $match, 0, $this->cursor) && '' !== $match[0]) {
-            $this->pushToken(/* Token::STRING_TYPE */ 7, stripcslashes($match[0]));
+            $this->pushToken(/* Token::STRING_TYPE */ 7, $this->stripcslashes($match[0], '"'));
             $this->moveCursor($match[0]);
         } elseif (preg_match(self::REGEX_DQ_STRING_DELIM, $this->code, $match, 0, $this->cursor)) {
             [$expect, $lineno] = array_pop($this->brackets);
--- a/tests/LexerTest.php
+++ b/tests/LexerTest.php
@@ -12,6 +12,7 @@ namespace Twig\Tests;
  */
 
 use PHPUnit\Framework\TestCase;
+use Symfony\Bridge\PhpUnit\ExpectDeprecationTrait;
 use Twig\Environment;
 use Twig\Error\SyntaxError;
 use Twig\Lexer;
@@ -21,6 +22,8 @@ use Twig\Token;
 
 class LexerTest extends TestCase
 {
+    use ExpectDeprecationTrait;
+
     public function testNameLabelForTag()
     {
         $template = '{% § %}';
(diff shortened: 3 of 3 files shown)


src/Lexer.php
```diff
@@ -396,6 +402,64 @@ private function lexExpression(): void
...
+                    $i++;
+                } elseif ($nextChar === "'" || $nextChar === '"') {
+                    if ($nextChar !== $quoteType) {
+                        $this->deprecations[] = sprintf("Character %s at position %d does not require to be escaped.", $nextChar, $i + 2);
+                    }
+                    $result .= $nextChar;
+                    $i++;
+                } elseif ($nextChar === 'x' && $i + 2 < $length && ctype_xdigit($str[$i + 1]) && ctype_xdigit($str[$i + 2])) {
+                    $result .= chr(hexdec($str[$i + 1] . $str[$i + 2]));
+                    $i += 2;
```
ruudk, line 439: Not sure if this is correct. Does Twig even support it? https://github.com/twigphp/Twig/issues/3604

stof:
To me, this should directly trigger `E_USER_DEPRECATED` (or use the `trigger_deprecation` of the symfony/deprecation-contracts if we already use it as the way to trigger deprecations).

And this should go in the 3.x branch (so that 4.0 can ship with the new behavior instead of having to wait until 5.0)

tests/LexerTest.php
```diff
@@ -178,23 +178,97 @@ public function testBigNumbers()
...
+    {
+        yield '{{ \'App\\\\Test\' }} => App\Test' => [
+            <<<'TWIG'
+            {{ 'App\\Test' }}
+            TWIG,
+            <<<'EOF'
+            App\Test
+            EOF,
+            [
+                'Character \ at position 5 does not require to be escaped.'
```
stof: This deprecation looks weird to me. Escaping a `\` should become optional, but should stay allowed IMO (like in PHP).

tests/LexerTest.php
```diff
@@ -178,23 +178,97 @@ public function testBigNumbers()
...
+            ]
+        ];
+        yield '{{ \'App\Test\' }} => App\Test' => [
+            <<<'TWIG'
+            {{ 'App\Test' }}
+            TWIG,
+            <<<'EOF'
+            App\Test
```
stof: This cannot be the behavior in a version *deprecating* the unnecessary escaping. The result of parsing must stay the same when introducing the deprecation (we should warn about a *future* behavior change, otherwise we are already breaking BC and the deprecation is useless).

Here is a good rule of thumb: except for the assertion on the deprecation message, there should be no difference in the tests between making them pass on the current stable 3.x release and the version containing the deprecation.

ruudk:
Thanks for the review. I think it makes sense. I will change this PR to target v3 and make sure we only deprecate and not introduce new functionality.

Then for v4 we can do the other change.

tests/LexerTest.php
ruudk: Removed it.

Can we somehow notify the developer in v4 that escaping a `\` is not needed? Maybe via `E_USER_NOTICE`? Or should we just not care about that 😊

tests/LexerTest.php
ruudk: Changed it back to the old behavior. The test now reflects what happens when you don't escape `\`.

ruudk:
Oke, I updated the PR to target V3. We don't change the behavior. We only deprecate, using `trigger_deprecation`, when we detect an unnecessary escape character. 

When this is merged, v4 is rebased with v3, I can provide the PR that changes the behavior there. 

Please have another look when you have time 😊

src/Lexer.php
```diff
@@ -382,6 +382,62 @@ private function lexExpression(): void
...
+                    $i++;
+                } elseif ($nextChar === '#') {
+                    $result .= "#";
+                    $i++;
+                } elseif ($nextChar === '\\') {
+                    $result .= '\\';
+                    $i++;
+                } elseif ($nextChar === "'" || $nextChar === '"') {
+                    if ($nextChar !== $quoteType) {
+                        trigger_deprecation('twig/twig', '3.10', "Character %s at position %d does not require to be escaped.", $nextChar, $i + 2);
```
fabpot:
```suggestion
                        trigger_deprecation('twig/twig', '3.11', "Character "%s" at position %d does not need to be escaped anymore.", $nextChar, $i + 2);
```

src/Lexer.php
```diff
@@ -382,6 +382,62 @@ private function lexExpression(): void
...
+                            $octal .= $str[$i + $j + 1];
+                            $i++;
+                        } else {
+                            break;
+                        }
+                    }
+                    $result .= chr(octdec($octal));
+                    $i++;
+                } else {
+                    trigger_deprecation('twig/twig', '3.10', "Character %s at position %d does not require to be escaped.", $nextChar, $i + 2);
```
fabpot:
```suggestion
                    trigger_deprecation('twig/twig', '3.11', "Character "%s" at position %d does not need to be escaped anymore.", $nextChar, $i + 2);
```

fabpot (changes requested):
We should probably add a note in the docs explaining when one needs to escape.

ruudk:
I refactored the code a bit and added more test cases. I think good to carefully review these test cases. Let me know if you think I can add more cases. 

Also updated the docs.

doc/templates.rst
```diff
@@ -526,7 +526,18 @@ exist:
...
   (e.g. ``'c:\\Program Files'``).
-
+  
+  Note that certain characters require escaping:
+   * ``\f``: Form feed
+   * ``\n``: New line
+   * ``\r``: Carriage return
+   * ``\t``: Horizontal tab
+   * ``\v``: Vertical tab
+   * ``\#``: Hash
```
ruudk, line 536: Not sure about this one. Maybe this should only apply when `#{` is used?

src/Lexer.php
```diff
@@ -382,6 +382,58 @@ private function lexExpression(): void
...
+                    $i++;
+                } elseif ($nextChar === "'" || $nextChar === '"') {
+                    if ($nextChar !== $quoteType) {
+                        trigger_deprecation('twig/twig', '3.11', 'Character "%s" at position %d does not need to be escaped anymore.', $nextChar, $i + 2);
+                    }
+                    $result .= $nextChar;
+                    $i++;
+                } elseif ($nextChar === 'x' && $i + 2 < $length && ctype_xdigit($str[$i + 1]) && ctype_xdigit($str[$i + 2])) {
+                    $result .= chr(hexdec($str[$i + 1] . $str[$i + 2]));
+                    $i += 2;
```
ruudk: Not sure about this. If we need to handle this, we should also provide test cases for it.

tests/LexerTest.php
```diff
@@ -178,23 +181,70 @@ public function testBigNumbers()
...
+    public function getStringWithEscapedDelimiter()
+    {
+        yield '{{ \'App\\\\Test\' }} => App\Test' => [
+            '{{ \'App\\\\Test\' }}',
+            'App\\Test',
+        ];
+        yield '{{ \'App\Test\' }} => AppTest' => [
+            '{{ \'App\\Test\' }}',
+            'AppTest',
+        ];
```
ruudk: This should trigger a deprecation, as it doesn't do anything. In v4 this will become App\Test

tests/LexerTest.php
```diff
@@ -178,23 +181,70 @@ public function testBigNumbers()
...
+        yield '{{ \'A\B\C\' }} => ABC' => [
+            '{{ \'A\B\C\' }}',
+            'ABC',
+        ];
```
ruudk: This should trigger a deprecation, as it doesn't do anything. In v4 this will become A\B\C

src/Lexer.php
```diff
@@ -382,6 +382,62 @@ private function lexExpression(): void
...
+                    $i++;
+                } elseif ($nextChar === "'" || $nextChar === '"') {
+                    if ($nextChar !== $quoteType) {
+                        trigger_deprecation('twig/twig', '3.11', 'Character "%s" at position %d does not need to be escaped anymore.', $nextChar, $i + 2);
+                    }
+                    $result .= $nextChar;
+                    $i++;
+                } elseif ($nextChar === '#' && $i + 2 < $length && $str[$i + 2] === '{') {
+                    $result .= '#{';
+                    $i += 2;
```
ruudk: I added this case, to make sure that escaping string Interpolation works. 

Should we also document this?

tests/LexerTest.php
stof: we should not care (just like PHP allows escaping them). Especially given that escaping them is still required in some cases

src/Lexer.php
stof: we indeed should not add support for things without testing them. But even more important, we should not change the behavior of strings in 3.x and so any test you add must first be run on the existing 3.x version (ensuring we keep the same behavior)

tests/LexerTest.php
ruudk: Ok clear!

src/Lexer.php
ruudk: Added tests for it, and made sure it stays the same. 

Also found some PHP tests for stripcslashes https://github.com/php/php-src/blob/master/ext/standard/tests/strings/add-and-stripcslashes.phpt

ruudk:
Alright, I think I'm done. I added more documentation related to Octal and Hexadecimal escape sequences.

doc/templates.rst
```diff
@@ -527,6 +527,18 @@ exist:
...
+   * ``\n``: New line
+   * ``\r``: Carriage return
+   * ``\t``: Horizontal tab
+   * ``\v``: Vertical tab
+   * ``\x``: Hexadecimal escape sequence
+   * ``\0`` to ``\377``: Octal escape sequences representing characters
+   * ``\``: Backslash
+  
+   When using single-quoted strings, the single quote character (``'``) needs to be escaped with a backslash (``\'``).
+   When using double-quoted strings, the double quote character (``"``) needs to be escaped with a backslash (``\"``).
```
fabpot, line 538: Can you add an empty line?

ruudk: Done 😊

ruudk:
@fabpot what do you think of this? Anything we should change or add more tests?

doc/templates.rst
```diff
@@ -527,6 +527,19 @@ exist:
...
+   * ``\f``: Form feed
+   * ``\n``: New line
+   * ``\r``: Carriage return
+   * ``\t``: Horizontal tab
+   * ``\v``: Vertical tab
+   * ``\x``: Hexadecimal escape sequence
+   * ``\0`` to ``\377``: Octal escape sequences representing characters
+   * ``\``: Backslash
+  
+   When using single-quoted strings, the single quote character (``'``) needs to be escaped with a backslash (``\'``).
```
fabpot, line 537: This is already explained in the first paragraph, maybe we should remove the explanation from there and add an example here like done there :)

src/Lexer.php
```diff
@@ -382,6 +382,74 @@ private function lexExpression(): void
...
+                    }
+                    $result .= chr(octdec($octal));
+                    $i++;
+                } else {
+                    trigger_deprecation('twig/twig', '3.11', 'Character "%s" at position %d does not need to be escaped anymore.', $nextChar, $i + 2);
+                    $result .= $nextChar;
+                    $i++;
+                }
+            } else {
+                $result .= $str[$i];
```
fabpot: Can we negate the condition so that this becomes the default and use `continue` there so that we can unindent the big block?

src/Lexer.php
fabpot: As escaping string interpolation is already supported, this code is needed to avoid the deprecation in case a string interpolation is escaped, right?

Escaping a string interpolation is tested here: `LexerTest::testStringWithEscapedInterpolation()` and I've just submitted #4189 for the missing documentation.

fabpot:
I've just run a small performance test via the script below and (maybe as expected) the hit can be significant:

```php
<?php

$string = str_repeat("日本では、春になると桜の花が咲きます。多くの人々は、公園や川の近くに集まり、お花見を楽しみます。桜の花びらが風に舞い、まるで雪のように見える瞬間は、とても美しいです。", 10);
$loader = new Twig\Loader\ArrayLoader([
    'index.twig' => <<<EOF
{{ "$string" }}
EOF
    ,
]);

$twig = new Twig\Environment($loader);

for ($i = 0; $i < 10000; $i++) {
    $twig->tokenize(new Twig\Source($twig->getLoader()->getSourceContext('index.twig')->getCode(), 'index.twig'));
}
```

On my machine, ~100ms for the 3.x branch vs ~900ms for this branch.

But I suppose that this kind of long string is not very common. I would expect "hardcoded" strings to be quite short in templates.

ruudk:
Maybe we could optimize the code a bit. First check if the string contains a backslash, then run the method. Otherwise don't do anything?

fabpot:
> Maybe we could optimize the code a bit. First check if the string contains a backslash, then run the method. Otherwise don't do anything?

Sure, as we have a perf script now, we can try a few obvious strategies to see their impact. But then again, let's optimize for short strings first and make sure that long strings are not too slow to lex.

doc/templates.rst
ruudk: Done, is this better?

src/Lexer.php
ruudk: Done!

src/Lexer.php
ruudk: > this code is needed to avoid the deprecation in case a string interpolation is escaped, right

Exactly.

ruudk:
I ran the bench on my machine:

```
// v3 branch
Benchmark 1: php bench.php
  Time (mean ± σ):     110.2 ms ±   3.7 ms    [User: 98.5 ms, System: 7.4 ms]
  Range (min … max):   108.6 ms … 126.9 ms    23 runs

// this branch
Benchmark 1: php bench.php
  Time (mean ± σ):      1.018 s ±  0.010 s    [User: 0.991 s, System: 0.009 s]
  Range (min … max):    1.007 s …  1.032 s    10 runs
```

When I change the `stripcslashes` method:
```diff
 private function stripcslashes(string $str, string $quoteType): string
 {
+    if (!str_contains($str, '\\')) {
+        return $str;
+    }
+    // ...
```

it becomes this:
```
// this branch
Benchmark 1: php bench.php
  Time (mean ± σ):      95.7 ms ±   1.1 ms    [User: 84.9 ms, System: 7.5 ms]
  Range (min … max):    94.5 ms …  99.0 ms    29 runs
```

But obviously, that's because the bench does not contain any escape characters. 

So let's change the bench to:
```php
<?php

$string = str_repeat("日本では、春になると桜の花が咲きます。\n多くの人々は、公園や川の近くに集まり、お花見を楽しみます。\\n桜の花びらが風に舞い、まるで雪のように見える瞬間は、とても美しいです。", 10);
$loader = new Twig\Loader\ArrayLoader([
    'index.twig' => <<<EOF
{{ "$string" }}
EOF
    ,
]);

$twig = new Twig\Environment($loader);

for ($i = 0; $i < 10000; $i++) {
    $twig->tokenize(new Twig\Source($twig->getLoader()->getSourceContext('index.twig')->getCode(), 'index.twig'));
}
```

Now we get
```
// this branch
Benchmark 1: php bench.php
  Time (mean ± σ):      1.026 s ±  0.002 s    [User: 0.999 s, System: 0.008 s]
  Range (min … max):    1.024 s …  1.029 s    10 runs
```

So the `str_contains` check is making it a bit slower when there is a backslash found. But when there is not, it can return the raw string directly, making it even faster than v3.

src/Lexer.php
```diff
@@ -382,6 +382,75 @@ private function lexExpression(): void
...
+                    if ($str[$o] >= '8') {
+                        break;
+                    }
+                    $octal .= $str[$o];
+                    $i++;
+                }
+                $result .= chr(octdec($octal));
+                $i++;
+            } else {
+                trigger_deprecation('twig/twig', '3.11', 'Character "%s" at position %d does not need to be escaped anymore.', $nextChar, $i + 2);
```
fabpot:
```suggestion
                trigger_deprecation('twig/twig', '3.12', 'Character "%s" at position %d does not need to be escaped anymore.', $nextChar, $i + 2);
```

src/Lexer.php
```diff
@@ -382,6 +382,75 @@ private function lexExpression(): void
...
+                $result .= $nextChar;
+                $i++;
+            }
+        }
+        return $result;
```
fabpot, line 457: missing empty line above

src/Lexer.php
```diff
@@ -382,6 +382,75 @@ private function lexExpression(): void
...
+            'n' => "\n",
+            'r' => "\r",
+            't' => "\t",
+            'v' => "\v",
+        ];
+
+        for ($i = 0; $i < $length; $i++) {
+            if (!($str[$i] === '\\' && $i + 1 < $length)) {
+                $result .= $str[$i];
+                continue;
```
fabpot, line 405: missing empty line above

src/Lexer.php
```diff
@@ -382,6 +382,75 @@ private function lexExpression(): void
...
+        $specialChars = [
+            'f' => "\f",
+            'n' => "\n",
+            'r' => "\r",
+            't' => "\t",
+            'v' => "\v",
+        ];
+
+        for ($i = 0; $i < $length; $i++) {
+            if (!($str[$i] === '\\' && $i + 1 < $length)) {
```
fabpot:
```suggestion
            if ($str[$i] !== '\\' || $i + 1 >= $length) {
```

fabpot (changes requested):
You haven't added the `str_contains()` call yet, right?

ruudk:
@fabpot Applied the feedback and also added the `str_contains` optimization.

fabpot:
Thank you @ruudk.

ruudk:
Thanks! I will prep the next PR when v4 is synced.

smnandre:
Hi @ruudk ... i may have a version that reduces the runtime difference, using chunks to avoid looping on every char.

It passes the tests, but i'm not sure if this goes in the direction you wanted for it.

So if you want to check / benchmark it: 

```php
private function stripcslashes(string $str, string $quoteType): string
{
    if (!str_contains($str, '\\')) {
        return $str;
    }

    $result = '';
    $length = strlen($str);
    $specialChars = [
        'f' => "\f",
        'n' => "\n",
        'r' => "\r",
        't' => "\t",
        'v' => "\v",
    ];

    $i = 0;
    while ($i < $length) {
        if (false === $pos = strpos($str, '\\', $i)) {
            $result .= substr($str, $i);
            break;
        }

        $result .= substr($str, $i, $pos - $i);
        $i = $pos + 1;

        if ($i >= $length) {
            $result .= '\\';
            break;
        }

        $nextChar = $str[$i];

        if (isset($specialChars[$nextChar])) {
            $result .= $specialChars[$nextChar];
        } elseif ($nextChar === '\\' || $nextChar === $quoteType) {
            $result .= $nextChar;
        } elseif ($nextChar === '#' && $i + 1 < $length && $str[$i + 1] === '{') {
            $result .= '#{';
            $i++;
        } elseif ($nextChar === 'x' && $i + 1 < $length && ctype_xdigit($str[$i + 1])) {
            $hex = $str[++$i];
            if ($i + 1 < $length && ctype_xdigit($str[$i + 1])) {
                $hex .= $str[++$i];
            }
            $result .= chr(hexdec($hex));
        } elseif (ctype_digit($nextChar) && $nextChar < '8') {
            $octal = $nextChar;
            while ($i + 1 < $length && ctype_digit($str[$i + 1]) && $str[$i + 1] < '8' && strlen($octal) < 3) {
                $octal .= $str[++$i];
            }
            $result .= chr(octdec($octal));
        } else {
            trigger_deprecation('twig/twig', '3.12', sprintf('Character "%s" at position %d does not need to be escaped anymore.', $nextChar, $i + 1));
            $result .= $nextChar;
        }

        $i++;
    }

    return $result;
}
```

ruudk:
@smnandre smart! Will check it out today

ruudk:
Created the PR to optimize this further: https://github.com/twigphp/Twig/pull/4196

PR #4179, opened 2024-08-07
Move static attributes for Function|Filter|TestExpression to the constructor

fabpot:


src/Node/Expression/FilterExpression.php
```diff
@@ -19,12 +19,16 @@ class FilterExpression extends CallExpression
...
     public function __construct(Node $node, ConstantExpression $filterName, Node $arguments, int $lineno, ?string $tag = null)
     {
-        parent::__construct(['node' => $node, 'filter' => $filterName, 'arguments' => $arguments], [], $lineno, $tag);
+        parent::__construct(['node' => $node, 'filter' => $filterName, 'arguments' => $arguments], ['name' => $filterName->getAttribute('value'), 'type' => 'filter'], $lineno, $tag);
     }
 
     public function compile(Compiler $compiler): void
     {
         $name = $this->getNode('filter')->getAttribute('value');
+        if ($name !== $this->getAttribute('name')) {
```
fabpot, line 28: Drupal is doing it to override the escape filter in a NodeVisitor, they should probably override it in an Extension instead.

PR #4181, opened 2024-08-08
clarify that macros shadow other macros/functions

xabbuh:
see https://github.com/twigphp/Twig/pull/4175#issuecomment-2273646498

### DIFF
--- a/doc/tags/macro.rst
+++ b/doc/tags/macro.rst
@@ -75,7 +75,7 @@ via the ``from`` tag:
 .. caution::
 
     As macros imported via ``from`` are called like functions, be careful to
-    not override existing functions:
+    not shadow existing functions:
 
     .. code-block:: twig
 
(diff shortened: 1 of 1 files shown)


fabpot:
Thank you @xabbuh.

doc/tags/macro.rst
```diff
@@ -75,7 +75,7 @@ via the ``from`` tag:
 .. caution::
 
     As macros imported via ``from`` are called like functions, be careful to
-    not override existing functions:
+    not shadow existing functions:
```
stof, line 78: I suggest saying `be careful that they shadow existing functions` (the current wording makes it look like it is an issue to perform such shadowing, but if you don't intent to use the function, there is no issue)

fabpot: Done

PR #4182, opened 2024-08-08
Add the possibility to deprecate attributes and nodes on Node

fabpot:


stof:
How does deprecating a node interacts with node visitors ?

fabpot:
> How does deprecating a node interacts with node visitors ?

That's the main use case for me: you get deprecations as node visitors call `setAttribute()/setNode()`.

stof:
OK, NodeTraverser calls `setNode only when the visitor returns a different node. So this should be fine in most cases.

fabpot:
New implementation with the possibility to customize the package and the version in addition to a new name (all optional).

src/Node/Node.php
```diff
@@ -30,6 +30,10 @@ class Node implements \Countable, \IteratorAggregate
     protected $tag;
 
     private $sourceContext;
+    /** @var NameDeprecation[] */
```
stof: I suggest using `array<string, NameDeprecation>` to document the type of keys

fabpot: Done

src/Node/Node.php
```diff
@@ -109,14 +113,37 @@ public function getAttribute(string $name)
             throw new \LogicException(\sprintf('Attribute "%s" does not exist for Node "%s".', $name, static::class));
         }
 
+        if (isset($this->attributeNameDeprecations[$name])) {
+            $dep = $this->attributeNameDeprecations[$name];
+            if ($dep->getNewName()) {
+                trigger_deprecation($dep->getPackage(), $dep->getVersion(), 'Getting attribute "%s" on a "%s" class is deprecated, get the "%s" attribute instead.', $name, static::class, $dep->getNewName());
+            } else {
+                trigger_deprecation($dep->getPackage(), $dep->getVersion(), 'Getting attribute "%s" on a "%s" class is deprecated and will throw an exception in 4.0.', $name, static::class);
```
nicolas-grekas: do we have to hardcode 4.0?

src/Node/Node.php
```diff
@@ -109,14 +113,37 @@ public function getAttribute(string $name)
...
     }
 
     public function setAttribute(string $name, $value): void
     {
+        if (isset($this->attributeNameDeprecations[$name])) {
+            $dep = $this->attributeNameDeprecations[$name];
+            if ($dep->getNewName()) {
+                trigger_deprecation($dep->getPackage(), $dep->getVersion(), 'Setting attribute "%s" on a "%s" class is deprecated, set the "%s" attribute instead.', $name, static::class, $dep->getNewName());
+            } else {
+                trigger_deprecation($dep->getPackage(), $dep->getVersion(), 'Setting attribute "%s" on a "%s" class is deprecated and with no alternative in 4.0.', $name, static::class);
```
nicolas-grekas: 4.0?

src/Node/Node.php
fabpot: Removed

src/Node/Node.php
fabpot: Removed

nicolas-grekas (approved):

PR #4183, opened 2024-08-08
Add support for package/version on deprecations

fabpot:


src/ExpressionParser.php
```diff
@@ -774,16 +774,13 @@ private function getTestNodeClass(TwigTest $test): string
...
-                $message .= \sprintf(' since version %s', $test->getDeprecatedVersion());
-            }
             if ($test->getAlternative()) {
                 $message .= \sprintf('. Use "%s" instead', $test->getAlternative());
             }
             $src = $stream->getSourceContext();
             $message .= \sprintf(' in %s at line %d.', $src->getPath() ?: $src->getName(), $stream->getCurrent()->getLine());
 
-            @trigger_error($message, \E_USER_DEPRECATED);
+            trigger_deprecation('twig/twig', $test->getDeprecatedVersion() ? $test->getDeprecatedVersion() : '', $message);
```
stof: what if the test does not come from the `twig/twig` package ?

fabpot: I don't have a good answer unfortunately as this can come from both internally and from a third-party.
We could potentially make it configurable, but I'm not sure this is worth it.
Can we use just 'Twig' as the "package" name? Or maybe an empty package name?

fabpot: An empty package name is supported, so that's maybe the best option.

stof: We should probably add support for providing a package name alongside the version.

src/Node/DeprecatedNode.php
```diff
@@ -36,20 +36,24 @@ public function compile(Compiler $compiler): void
         $expr = $this->getNode('expr');
 
         if ($expr instanceof ConstantExpression) {
-            $compiler->write('@trigger_error(')
-                ->subcompile($expr);
+            $compiler
+                ->write('trigger_deprecation(\'twig/twig\', \'\', ')
```
stof: this should use an empty package name IMO

fabpot: I've made them configurable now.

src/Node/DeprecatedNode.php
```diff
@@ -26,30 +26,36 @@ class DeprecatedNode extends Node
 {
     public function __construct(AbstractExpression $expr, int $lineno, ?string $tag = null)
     {
-        parent::__construct(['expr' => $expr], [], $lineno, $tag);
+        parent::__construct(['expr' => $expr], ['package' => 'twig/twig', 'version' => ''], $lineno, $tag);
```
stof: defaulting to `twig/twig` looks weird to me, as templates using that node are likely to belong to a different package.

And there is no way to provide those attributes from a `{% deprecated %}` tag in a template.

fabpot: Fixed

fabpot:
I think this one is ready now (with updated docs).

src/ExpressionParser.php
```diff
@@ -826,16 +820,13 @@ private function getFilterNodeClass(string $name, int $line): string
...
-                $message .= \sprintf(' since version %s', $filter->getDeprecatedVersion());
-            }
             if ($filter->getAlternative()) {
                 $message .= \sprintf('. Use "%s" instead', $filter->getAlternative());
             }
             $src = $this->parser->getStream()->getSourceContext();
             $message .= \sprintf(' in %s at line %d.', $src->getPath() ?: $src->getName(), $line);
 
-            @trigger_error($message, \E_USER_DEPRECATED);
+            trigger_deprecation($filter->getDeprecatingPackage(), $filter->getDeprecatedVersion() ? $filter->getDeprecatedVersion() : '', $message);
```
nicolas-grekas:
```suggestion
            trigger_deprecation($filter->getDeprecatingPackage(), $filter->getDeprecatedVersion() ?: '', $message);
```

src/ExpressionParser.php
```diff
@@ -800,16 +797,13 @@ private function getFunctionNodeClass(string $name, int $line): string
...
-                $message .= \sprintf(' since version %s', $function->getDeprecatedVersion());
-            }
             if ($function->getAlternative()) {
                 $message .= \sprintf('. Use "%s" instead', $function->getAlternative());
             }
             $src = $this->parser->getStream()->getSourceContext();
             $message .= \sprintf(' in %s at line %d.', $src->getPath() ?: $src->getName(), $line);
 
-            @trigger_error($message, \E_USER_DEPRECATED);
+            trigger_deprecation($function->getDeprecatingPackage(), $function->getDeprecatedVersion() ? $function->getDeprecatedVersion() : '', $message);
```
nicolas-grekas:
```suggestion
            trigger_deprecation($function->getDeprecatingPackage(), $function->getDeprecatedVersion() ?: '', $message);
```

nicolas-grekas (approved):

PR #4191, opened 2024-08-10
Bump minimum PHP version to 8.0

fabpot:
The most recent version of PHP 7 (7.4) was deprecated more than a year and a half ago.

We won't remove code that supports PHP 7 (as this has already been done in 4.x and there is no point in redoing the work in the 3.x branch), but we will use some new PHP 8 features in the 3.x branch from now on.


derrabus:
> Drop support for 8.0

The PR title should be changed to "Drop support for PHP 7.4". PHP 8.0 is still supported after this change.

fabpot:
> > Drop support for 8.0
> 
> The PR title should be changed to "Drop support for PHP 7.4". PHP 8.0 is still supported after this change.

Good catch, fixed.

PR #4193, opened 2024-08-11
Remove escape

ruudk:
Interestingly this didn’t fail on the branch that introduced the change. Are we sure the integration tests work at all? 

See [failing job](https://github.com/twigphp/Twig/actions/runs/10341129793) on merge to main.


fabpot:
Thank you @ruudk.

PR #4194, opened 2024-08-11
Make various optimizations for dynamic Twig callables

fabpot:


tests/ExpressionParserTest.php
```diff
@@ -411,6 +415,51 @@ public function testUnknownTestWithoutSuggestions()
...
+        $env->addExtension(new class() extends AbstractExtension {
+            public function getFilters()
+            {
+                return [
+                    new TwigFilter('*_foo_*_bar', function ($foo, $bar, $a) {}),
+                ];
+            }
+        });
+
+        $this->assertStringContainsString('$this->env->getFilter(\'*_foo_*_bar\')->getCallable()("foo", "bar", "a")', $env->compile($env->parse($env->tokenize(new Source($env->getLoader()->getSourceContext('index')->getCode(), 'index')))));
```
fabpot, line 460: Having the dynamic name here (at runtime) avoids matching the original `foo_foo_bar_bar` filter in ExtensionSet (not needed anyway as we just need the callable here).

src/ExtensionSet.php
```diff
@@ -437,17 +435,26 @@ private function initExtension(ExtensionInterface $extension): void
     {
         // filters
         foreach ($extension->getFilters() as $filter) {
-            $this->filters[$filter->getName()] = $filter;
+            $this->filters[$name = $filter->getName()] = $filter;
+            if (str_contains($name, '*')) {
+                $this->dynamicFilters['#^'.str_replace('\\*', '(.*?)', preg_quote($name, '#')).'$#'] = $filter;
```
fabpot, line 440: Small optimization here to avoid doing the same work over and over again. And having just the dynamic filters in an array means looping only on them.

src/AbstractTwigCallable.php
```diff
@@ -68,8 +74,23 @@ public function needsContext(): bool
...
+        $new->dynamicName = $dynamicName;
+        $new->arguments = $arguments;
+
+        return $new;
+    }
+
+    /**
+     * @deprecated since Twig 3.12, use withDynamicArguments() instead
+     */
     public function setArguments(array $arguments): void
```
fabpot, line 90: This is the only method that makes Twig callables mutable. And as demonstrated by the new tests, this can lead to bad behaviors.

src/Node/Expression/CallExpression.php
```diff
@@ -51,7 +51,7 @@ protected function compileCallable(Compiler $compiler)
 
                 $compiler->raw(\sprintf('->%s', $callable[1]));
             } else {
-                $compiler->raw(\sprintf('$this->env->get%s(\'%s\')->getCallable()', ucfirst($this->getAttribute('type')), $this->getAttribute('name')));
+                $compiler->raw(\sprintf('$this->env->get%s(\'%s\')->getCallable()', ucfirst($this->getAttribute('type')), $this->getAttribute('dynamic_name')));
```
fabpot, line 54: At runtime, no need to parse the arguments again.

PR #4195, opened 2024-08-12
(minor) Prefer `className` over `classname`

smnandre:
For consistency with how it's written in the tested classes.


fabpot:
Thank you @smnandre.

PR #4196, opened 2024-08-12
Optimize stripcslashes

ruudk:
See https://github.com/twigphp/Twig/pull/4176#issuecomment-2283168241 by @smnandre 

```
# Old
Benchmark 1: php bench.php
  Time (mean ± σ):      1.192 s ±  0.004 s    [User: 1.163 s, System: 0.009 s]
  Range (min … max):    1.186 s …  1.198 s    10 runs

# New
Benchmark 1: php bench.php
  Time (mean ± σ):     356.5 ms ±   1.1 ms    [User: 342.0 ms, System: 7.6 ms]
  Range (min … max):   354.6 ms … 358.2 ms    10 runs
```

Using bench.php:
```php
<?php

$string = str_repeat("Lorem ipsum dolor sit amet, consectetur adipiscing elit. Ut a tincidunt turpis. \\App\\Entity\\Product \World \065 App\#{var} \'quoted\' Donec pharetra enim quis erat pharetra, dignissim molestie erat laoreet. Vivamus auctor purus sed lorem vestibulum rhoncus.", 10);
$loader = new Twig\Loader\ArrayLoader([
    'index.twig' => <<<EOF
{{ "$string" }}
EOF
    ,
]);

$twig = new Twig\Environment($loader);

for ($i = 0; $i < 10000; $i++) {
    $twig->tokenize(new Twig\Source($twig->getLoader()->getSourceContext('index.twig')->getCode(), 'index.twig'));
}

```

### DIFF
--- a/src/Lexer.php
+++ b/src/Lexer.php
@@ -50,6 +50,14 @@ class Lexer
     public const REGEX_DQ_STRING_PART = '/[^#"\\\\]*(?:(?:\\\\.|#(?!\{))[^#"\\\\]*)*/As';
     public const PUNCTUATION = '()[]{}?:.,|';
 
+    private const SPECIAL_CHARS = [
+        'f' => "\f",
+        'n' => "\n",
+        'r' => "\r",
+        't' => "\t",
+        'v' => "\v",
+    ];
+
     public function __construct(Environment $env, array $options = [])
     {
         $this->env = $env;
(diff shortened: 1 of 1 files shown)


src/Lexer.php
```diff
@@ -398,62 +398,55 @@ private function stripcslashes(string $str, string $quoteType): string
             'v' => "\v",
         ];
```
fabpot: `$specialChars` can probably be moved to a private const?

fabpot (commented):
I suppose that we can remove the `str_contains` as it's not needed anymore (the check at the beginning of the while is equivalent).

src/Lexer.php
ruudk: Done

ruudk:
> I suppose that we can remove the `str_contains` as it's not needed anymore (the check at the beginning of the while is equivalent).

@fabpot Done + updated the bench
```
# Old
Benchmark 1: php bench.php
  Time (mean ± σ):      1.204 s ±  0.012 s    [User: 1.185 s, System: 0.011 s]
  Range (min … max):    1.191 s …  1.225 s    10 runs

# New
Benchmark 1: php bench.php
  Time (mean ± σ):     368.3 ms ±   4.8 ms    [User: 354.2 ms, System: 10.1 ms]
  Range (min … max):   362.9 ms … 375.6 ms    10 runs
```

smnandre:
Nice & thank you :)

fabpot (approved):

fabpot:
Thank you @ruudk.

ruudk:
@fabpot Could you sync this back to v4 so that I can continue?

fabpot:
> @fabpot Could you sync this back to v4 so that I can continue?

I've just merged 3.x into 4.x.

PR #4198, opened 2024-08-12
Remove const optimization

### LINKED ISSUE #4197: [v4] Uncomment `/* Token::X */` references
Is this a performance optimization? Or can we move this to normal constants in v4?

fabpot:
Closes #4197


stof:
maybe we should even merge this in 3.x (which could simplify up merges between 3.x and 4.x)

fabpot:
> maybe we should even merge this in 3.x (which could simplify up merges between 3.x and 4.x)

Rebased

PR #4199, opened 2024-08-12
Do not hide unnecessary escape characters

ruudk:
Follow up of #4176, #4123 and #2712.

Now that we have deprecated unnecessary escape in v3, we can change the behavior in v4.

With this change, unnecessary will no longer be ignored, but handled like any other character.

This allows for writing fully qualified class names like this:
```twig
{{ constant('App\Entity\User::SOME_CONSTANT') }}
```

Instead of having to escape the `\` in v3:
```twig
{{ constant('App\\Entity\\User::SOME_CONSTANT') }}
```

/cc @fabpot @stof


ruudk:
I don't think the integration test failure is related to this PR.

ruudk:
@fabpot @stof Is there anything I need to change?

stof (approved):

stof:
this needs some changelog entry

ruudk:
@stof @fabpot Done 😊

src/Lexer.php
```diff
@@ -451,8 +446,7 @@ private function stripcslashes(string $str, string $quoteType): string
                 }
                 $result .= chr(octdec($octal));
             } else {
-                trigger_deprecation('twig/twig', '3.12', sprintf('Character "%s" at position %d does not need to be escaped anymore.', $nextChar, $i + 1));
-                $result .= $nextChar;
+                $result .= '\\' . $nextChar;
```
fabpot: I think there is an issue here.

I've tested with the following template:

```twig
{{ 'Foo\Bar' }}
{{ 'Foo\\Bar' }}
{{ 'Foo\\\Bar' }}
{{ 'Foo\\\\Bar' }}
```

On Twig 3.x, it results in

```twig
FooBar
Foo\Bar
Foo\Bar
Foo\\Bar
```

But with this PR, I get:

```twig
Foo\Bar
Foo\Bar
Foo\\Bar
Foo\\Bar
```

To fix it:

```suggestion
                $result .= $nextChar;
```

CHANGELOG
```diff
@@ -7,6 +7,7 @@
  * Change the compilation of `for` loops to throw an exception when a `loop.*` variable is not defined
  * Make `Environment::getGlobals()` private
  * Drop support for PHP < 8.2
+ * Unnecessary escape characters are no longer ignored
 =======
```
```suggestion

```

tests/Fixtures/functions/enum_cases/valid.test
```diff
@@ -3,13 +3,13 @@
 --CONDITION--
 \PHP_VERSION_ID >= 80100
 --TEMPLATE--
-{% for c in enum_cases('Twig\\Tests\\DummyBackedEnum') %}
+{% for c in enum_cases('Twig\Tests\DummyBackedEnum') %}
```
fabpot, line 9: Can we double the tests to also prove that double-escaping still work (for all 3 cases maybe)?

tests/LexerTest.php
```diff
@@ -229,43 +238,6 @@ public function getStringWithEscapedDelimiter()
         ];
     }
 
-    /**
-     * @group legacy
-     * @dataProvider getStringWithEscapedDelimiterProducingDeprecation
-     */
-    public function testStringWithEscapedDelimiterProducingDeprecation(string $template, string $expected, string $expectedDeprecation)
```
fabpot, line 236: I think we should keep the tests to prove that double-escaped strings still work as expected.

CHANGELOG
```diff
@@ -7,6 +7,7 @@
  * Change the compilation of `for` loops to throw an exception when a `loop.*` variable is not defined
  * Make `Environment::getGlobals()` private
  * Drop support for PHP < 8.2
+ * Unnecessary escape characters are no longer ignored
```
fabpot: I'm not sure I understand the meaning. Can you clarify?

fabpot (changes requested):

tests/LexerTest.php
ruudk: I did not remove those tests, I moved them to the `getStringWithEscapedDelimiter` above, see: https://github.com/twigphp/Twig/blob/ba4e8175cbd4c38928675ddcb0fd0f35361969d9/tests/LexerTest.php#L195-L206

CHANGELOG
ruudk: I don't know how to explain it better.

Doing `App\Test` tries to escape T which is unnecessary. In v3 this escape `\` is ignored and reported as deprecation. In v4 it's no longer ignored, and results in a literal `\`. That means that `App\Test` will show up as `App\Test` as you expect.

src/Lexer.php
ruudk: I think this is correct.

Escaping B is unnecessary. In v3 a deprecation is reported. In v4 it prints `\`.
```twig
{{ 'Foo\Bar' }}
v3: FooBar
v4: Foo\Bar
```

Escaping a backslash is ignored in v3 and still ignored in v4:
```twig
{{ 'Foo\\Bar' }}
v3: Foo\Bar
v4: Foo\Bar
```

Escaping a backslash is ignored in v3 and still ignored in v4. 
Escaping B is unnecessary. In v3 a deprecation is reported. In v4 it prints `\`.
```twig
{{ 'Foo\\\Bar' }}
v3: Foo\Bar
v4: Foo\\Bar
```

Escaping a backslash is ignored in v3 and still ignored in v4:
```twig
{{ 'Foo\\\\Bar' }}
v3: Foo\\Bar
v4: Foo\\Bar
```

ruudk: Added an integration test to better show it:
https://github.com/twigphp/Twig/pull/4199/commits/152990db5f662e3edb1bdb2248ee5436f68dc6aa

tests/Fixtures/functions/enum_cases/valid.test
ruudk: Done

CHANGELOG
smnandre: Maybe something like this, as it will be hard to explain simply with a short sentence  ?

```
 * Change tokenisation of escape sequence in strings

 Before::  
  // ... 
 {# foo equals "FooBar" #}
 
After:
 // 
 {# foo equals "Foo\Bar" #}
```

As example I would probably use a function call (instead of a print expression for instance), to support the idea it's a _lexer thing_ and thus is not related to how context vars are printed.. wdyt ?

src/Lexer.php
stof: We still need to allow escaping a backslash, otherwise writing a backslash followed by a `n` is impossible (as `\n` is a valid escape sequence.

The new behavior matches the behavior of PHP strings: escaping of backslashes is _optional_ when non-ambiguous. But if you want to write a backslash when the next char can be an escaped char (which includes the case when you want to write a backslash followed by another one), it needs to be escaped.

src/Lexer.php
```diff
@@ -428,12 +428,7 @@ private function stripcslashes(string $str, string $quoteType): string
 
             if (isset(self::SPECIAL_CHARS[$nextChar])) {
                 $result .= self::SPECIAL_CHARS[$nextChar];
-            } elseif ($nextChar === '\\') {
-                $result .= $nextChar;
-            } elseif ($nextChar === "'" || $nextChar === '"') {
-                if ($nextChar !== $quoteType) {
-                    trigger_deprecation('twig/twig', '3.12', 'Character "%s" at position %d does not need to be escaped anymore.', $nextChar, $i + 1);
```
fabpot: I think I'm confused because of the deprecation message.

Maybe it's just me, but I read the current message like this: "The character does not need to be escaped, **but that's fine**.". And of course, it's not fine as the behavior will change, so the developer must fix the code.

So, maybe we need to add some hints to the developer:

Character "%s" at position %d should not to be escaped; the "\\" character is ignored in Twig 3 but as it won't be in 4.x, please remove the extra "\\" character.

src/Lexer.php
fabpot:
```suggestion
                $result .= '\\'.$nextChar;
```

CHANGELOG
fabpot: Or maybe just "Escape characters are no longer ignored".

fabpot (changes requested):

src/Lexer.php
ruudk: You're right. The deprecation is vague. 

Fixed in https://github.com/twigphp/Twig/pull/4219

ruudk:
@fabpot Rebased the PR and applied the fixes. 

Also created https://github.com/twigphp/Twig/pull/4219 for V3.

fabpot:
Thank you @ruudk.

PR #4201, opened 2024-08-13
Disable the drupal integration tests in the 4.x branch

stof:
Drupal uses the AbstractNodeVisitor which is deprecated in 3.x and removed in 4.x, so the job fails all the time. This should be re-enabled once Drupal can be tested against Twig 4.x.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -113,6 +113,7 @@ jobs:
               run: ../../vendor/bin/simple-phpunit
 
     integration-tests:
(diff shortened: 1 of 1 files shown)


stof:
Alternatively, maybe the integration tests should be updated to use Drupal 11 instead of Drupal 10.1.x-dev. Maybe they _already_ migrated away from that base class.

fabpot:
I'm going to merge more PRs with deprecations, so Drupal will definitely not work anyway.

fabpot:
Thank you @stof.

PR #4204, opened 2024-08-15
Remove remaining const optimizations

smnandre:
After #4198, some Token:: constant optimizations remained.

This PR removes them.. **in case** this was not intended


smnandre:
(fabbot failures unrelated)

fabpot:
Thank you @smnandre.

PR #4207, opened 2024-08-15
Throw a SyntaxError exception at compile time when a Twig callable has not the minimum number of required arguments

fabpot:
When using named arguments, a compilation error is thrown when some required arguments are missing.
But when not using arguments, the error was thrown at runtime. This PR fixed this.

### DIFF
--- a/tests/ErrorTest.php
+++ b/tests/ErrorTest.php
@@ -234,28 +234,6 @@ EOHTML
         }
     }
 
-    public function testTwigArgumentCountErrorThrowsRuntimeExceptions()
-    {
-        $loader = new ArrayLoader([
-            'argument-error.html' => <<<EOHTML
-{# max requires at least one argument #}
-{{ max() }}
-EOHTML
-        ]);
-
-        $twig = new Environment($loader, ['debug' => true, 'cache' => false]);
-
-        $template = $twig->load('argument-error.html');
-        try {
-            $template->render();
-
-            $this->fail();
-        } catch (RuntimeError $e) {
-            $this->assertEquals(2, $e->getTemplateLine());
-            $this->assertEquals('argument-error.html', $e->getSourceContext()->getName());
-        }
-    }
-
     public function getErroredTemplates()
     {
         return [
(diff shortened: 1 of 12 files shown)


tests/ErrorTest.php
```diff
@@ -234,28 +234,6 @@ public function testTwigArrayReduceThrowsRuntimeExceptions()
         }
     }
 
-    public function testTwigArgumentCountErrorThrowsRuntimeExceptions()
```
fabpot, line 237: That's a compilation error now.

tests/Fixtures/functions/max_without_args.test
```diff
@@ -0,0 +1,8 @@
+--TEST--
+"max" function without an argument throws a compile time exception
+--TEMPLATE--
+{{ max() }}
+--DATA--
+return []
+--EXCEPTION--
+Twig\Error\SyntaxError: Not enough arguments for function "max" in "index.twig".
```
fabpot, line 8: This is the new compilation error.

tests/Fixtures/functions/cycle_without_enough_args.test
```diff
@@ -0,0 +1,8 @@
+--TEST--
+"cycle" function without enough args and a named argument
+--TEMPLATE--
+{{ cycle(position=2) }}
+--DATA--
+return []
+--EXCEPTION--
+Twig\Error\SyntaxError: Value for argument "values" is required for function "cycle" in "index.twig" at line 2.
```
fabpot, line 8: This was already implemented.

tests/Fixtures/functions/max_without_args.test
```diff
@@ -0,0 +1,8 @@
+--TEST--
+"max" function without an argument throws a compile time exception
+--TEMPLATE--
+{{ max() }}
+--DATA--
+return []
+--EXCEPTION--
+Twig\Error\SyntaxError: Value for argument "value" is required for function "max" in "index.twig" at line 2.
```
fabpot, line 8: This is the new compilation error (same format as the one above).

Issue #3635, opened 2022-01-20
Named argument support


kbond:
As @weaverryan and I have been working on Twig/Live components in `symfony/ux`, we've come across a few areas where named argument support would be valuable. This issue is intended to track/discuss adding/improving support for this in twig.

1. (probable bug) Filters/functions using `is_variadic` snake-cases the keys. This issue (and a possible solution) is explained fully in https://github.com/twigphp/Twig/issues/3475.
2. Support in `twig_get_attribute` for PHP8+
    ```twig
    {{ object.method(prop1='value', prop2='value') }}
    ```
3. Support in filters/functions without `is_variadic` for PHP8+
    ```twig
    {{ var|filter(prop1='value', prop2='value') }}

    {{ function(prop1='value', prop2='value') }}
    ```
4. Could `:` be used as an assignment operator instead of `=`?
    ```twig
    {{ var|filter(prop1: 'value', prop2: 'value') }}

    {{ function(prop1: 'value', prop2: 'value') }}

    {{ object.method(prop1: 'value', prop2:'value') }}
    ```

I understand there might be technical limitations that make supporting any/all of these impossible. This was the case with [named argument support in macros](https://github.com/twigphp/Twig/issues/929#issuecomment-493525160).

Bilge:
~~Your link to the macro issue just links back to this issue.~~

kbond:
Thanks @Bilge, fixed.

tacman:
@kbond Any chance of seeing this before Symfony 6.2 goes into beta?  As mentioned in a linked issue here, this format wouldn't break anything, and looks quite clean.

```twig
{{ foo(barValue: 'bar', booValue: 'boo') }}
```

kbond:
Twig has different release schedule than Symfony so this can be added to twig in a feature release (anytime). There is a plan to migrate twig to Symfony core and I believe this is issue is on the radar.

joeworkman:
I am creating a global that references a class. I would love to be able to have named parameters with my class functions. I assume that your suggested features would also cover that? This look perfect to me. I have a couple of functions that could potentially have a lot of parameters. It would be nice to be able to selectively only use parameters as I see fit.

xabbuh:
Given that Twig 4 will require PHP 8.2+ (see #3881) named arguments support for macros could also be an option then.

DenuxPlays:
It is possible now: https://twig.symfony.com/doc/3.x/templates.html#named-arguments

[...]

Issue #3475, opened 2021-01-25
Named Arguments of Filters and Functions are not passed as-is and forced to snake_case


filecage:
## Problem Description
I stumbled upon this super odd behaviour when implementing a custom function in Twig and realised that given the template
```twig
{{ foo(barValue='bar', booValue='boo') }}
```
the two named arguments, `barValue` and `booValue` will be passed to the `foo` function as `bar_value` and `boo_value`. This is due to the "name normalization" done in the [`CallExpression` on line 242](https://github.com/twigphp/Twig/blob/4a47f0f5b4bf2400ed1e3abecb486655e3970cdb/src/Node/Expression/CallExpression.php#L242) in the current 3.x branch. This has been introduced in a59dcde3, around 8 years ago and was part of the initial named arguments implementation.

As this behaviour is undocumented, technically unnecessary (named arguments are passed as array keys and PHP allows all strings to be array keys) and defnitely unexpected, I'm suggesting to classifiy this as a bug.

## Proposed Fix
To fix it without introducing backwards incompatible changes, I'm proposing a new option for functions and filters that is called `legacy_named_arguments` and that defaults to `true` in 3.x and can be set to `false` in upcoming major releases.

- Setting `legacy_named_arguments` to `true` means: keep the current behaviour. In my opinion it makes sense to trigger a `E_USER_DEPRECATED` when a variable key has been "normalized" (i. e. input key !== output key). However, this might have a very big impact on a lot of existing Symfony installations.

- Setting `legacy_named_arguments` to `false` means: leave the input argument as-is. The lexer decides whether a variable name is valid or not (which it already does implicitly [here](https://github.com/twigphp/Twig/blob/3.x/src/Lexer.php#L363)).

I am happy to contribute in a PR once you've confirmed the issue.

weaverryan:
This looks like a legit issue to me. We were trying to create a variadic Twig function, but apparently, when we pass in the named args, they are snake-cased, as you mentioned here. Your resolution sounds very reasonable, except that I'd probably skip the deprecation. The reason is just because (as you realized also) requiring everyone to set `legacy_named_arguments 
 => true` to skip the deprecation would be a huge pain. But if simply have a way to opt-intoo" normal, non-snaked-cased args, then that would be enough.

Cheers!

iquito:
[...]

PR #4209, opened 2024-08-16
Accept colons instead of equals for named arguments

fabpot:
Refs #3475
Refs #3635

### DIFF
--- a/doc/coding_standards.rst
+++ b/doc/coding_standards.rst
@@ -105,3 +105,9 @@ standards:
              true
          {% endif %}
      {% endblock %}
+
+* Use ``:`` instead of ``=`` to separate argument names and values:
+
+  .. code-block:: twig
+
+     {{ data|convert_encoding(from: 'iso-2022-jp', to: 'UTF-8') }}
--- a/doc/templates.rst
+++ b/doc/templates.rst
@@ -210,11 +210,16 @@ built-in functions.
 Named Arguments
 ---------------
 
-Named arguments are supported in functions, filters and tests.
+Named arguments are supported in functions, filters, and tests.
+
+.. versionadded:: 3.12
+
+    Twig supports both ``=`` and ``:`` as separators between argument names and
+    values, but support for ``:`` was introduced in Twig 3.12.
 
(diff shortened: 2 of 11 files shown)


smnandre:
(For information, i tested TwigComponent with this branch: all green)

doc/templates.rst
```diff
@@ -210,11 +210,16 @@ built-in functions.
 Named Arguments
 ---------------
 
-Named arguments are supported in functions, filters and tests.
+Named arguments are supported in functions, filters, and tests.
+
+.. versionadded:: 3.12
+
+    Twig supports both ``=`` and ``:`` as separators between keys and values,
```
stof:
```suggestion
    Twig supports both ``=`` and ``:`` as separators between argument names and values,
```

fabpot: Changed

doc/templates.rst
```diff
@@ -227,7 +232,7 @@ the values you pass as arguments:
 
     {# versus #}
 
-    {{ data|convert_encoding(from='iso-2022-jp', to='UTF-8') }}
+    {{ data|convert_encoding(from: 'iso-2022-jp', to: 'UTF-8') }}
```
smnandre, line 235: As you replaced the `=` with `:` in this doc, does that mean using `:` will (sooner or later) be the "recommanded" syntax ?  

(something we wondered with @VincentLanglet to chose a potential "default choice" in Twig-CS-Fixer rule)

fabpot: I think we will support both for the foreseeable future. I would recommend using `:` for a CS fixer though.

doc/coding_standards.rst
```diff
@@ -105,3 +105,9 @@ standards:
              true
          {% endif %}
      {% endblock %}
+
+* Use ``:`` instead of ``=`` to separate argument names and values:
+
+  .. code-block:: twig
+
+     {{ data|convert_encoding(from: 'iso-2022-jp', to: 'UTF-8') }}
```
fabpot, line 113: cc @smnandre @VincentLanglet

PR #4210, opened 2024-08-17
Special functions extension point

fabpot:
The `parent`, `attribute`, `block` (and `loop` in 4.0) functions are handled directly in ExpressionParser as they are converted to specific Nodes.

It means that recursive loops can only be implemented in core. By adding a new extension point, we are removing the hardcoded handling of such functions and we open up the possibility for user-land functions to do the same.

*Review is easier when hiding whitespace.*

### DIFF
--- a/src/ExpressionParser.php
+++ b/src/ExpressionParser.php
@@ -457,60 +457,37 @@ class ExpressionParser
 
     public function getFunctionNode($name, $line)
     {
-        switch ($name) {
-            case 'parent':
-                if (!\count($this->parser->getBlockStack())) {
-                    throw new SyntaxError('Calling "parent" outside a block is forbidden.', $line, $this->parser->getStream()->getSourceContext());
-                }
+        if (null !== $alias = $this->parser->getImportedSymbol('function', $name)) {
+            $arguments = new ArrayExpression([], $line);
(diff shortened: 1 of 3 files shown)


src/ExpressionParser.php
```diff
@@ -457,60 +457,38 @@ public function parsePostfixExpression($node)
...
-                if (!isset($this->readyNodes[$class = $function->getNodeClass()])) {
-                    $this->readyNodes[$class] = (bool) (new \ReflectionClass($class))->getConstructor()->getAttributes(FirstClassTwigCallableReady::class);
-                }
+            return ($function->getParserCallable())($this->parser, $fakeNode, $args, $line);
+        }
 
-                if (!$ready = $this->readyNodes[$class]) {
-                    trigger_deprecation('twig/twig', '3.12', 'Twig node "%s" is not marked as ready for passing a "TwigFunction" in the constructor instead of its name; please update your code and then add #[FirstClassTwigCallableReady] attribute to the constructor.', $class);
-                }
+        $ready = true;
```
smnandre:
```suggestion
```

You define it L487  now

fabpot: Good catch

PR #4211, opened 2024-08-17
Add missing return types in 4.0

smnandre:
Revert the removal of return types done in #3156 

Should i propagate to `extra` packages in the same PR ?

### DIFF
--- a/src/Extension/ExtensionInterface.php
+++ b/src/Extension/ExtensionInterface.php
@@ -31,35 +31,35 @@ interface ExtensionInterface
      *
      * @return TokenParserInterface[]
      */
-    public function getTokenParsers();
+    public function getTokenParsers(): array;
 
     /**
      * Returns the node visitor instances to add to the existing list.
      *
      * @return NodeVisitorInterface[]
      */
-    public function getNodeVisitors();
+    public function getNodeVisitors(): array;
 
     /**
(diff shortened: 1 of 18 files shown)


src/Extension/ExtensionInterface.php
```diff
@@ -31,35 +31,35 @@ interface ExtensionInterface
      *
      * @return TokenParserInterface[]
      */
-    public function getTokenParsers();
+    public function getTokenParsers(): array;
```
Kocal, line 34: Should we also update the PHPDoc return type to `array<TokenParserInterface>`, or keeping `TokenParserInterface[]` is fine (`array<>` is more strict)?

stof: it is not more strict. It is strictly equivalent.

Kocal: Ah, I misread/misremembered https://phpstan.org/writing-php-code/phpdoc-types#general-arrays

fabpot:
@smnandre Yes, we should do it on `extra` as well.

smnandre:
I made a pass on them and think it's good :)

fabpot:
Thank you @smnandre.

smnandre:
Arf I found others while checking sorry ....  in code runtimeloader + docs. I'll open another PR :|

PR #4212, opened 2024-08-17
ChainLoader constructor should accept iterable instead of array

### LINKED ISSUE #4200: [Twig\Loader\ChainLoader] Constructor could accept iterable instead of array.
Proposal - Instead of current constructor:

```php
    /**
     * @param LoaderInterface[] $loaders
     */
    public function __construct(array $loaders = [])
    {
        foreach ($loaders as $loader) {
            $this->addLoader($loader);
        }
    }
```

replace it with:

```php
    /**
     * @param iterable<LoaderInterface> $loaders
     */
    public function __construct(iterable $loaders = [])
    {
        foreach ($loaders as $loader) {
            $this->addLoader($loader);
        }
    }
```

No BC break, allows to use `tagged_iterator` in SF autowiring for creating our own Twig environment (beside templating delivered with SF app by default).

WDYT?

TheCelavi:
Closes #4200

### DIFF
--- a/src/Loader/ChainLoader.php
+++ b/src/Loader/ChainLoader.php
@@ -21,22 +21,33 @@ use Twig\Source;
  */
 final class ChainLoader implements LoaderInterface
 {
-    private $hasSourceCache = [];
-    private $loaders = [];
+    /**
+     * @var \Traversable<LoaderInterface>|LoaderInterface[]
+     */
+    private $loaders;
 
     /**
-     * @param LoaderInterface[] $loaders
+     * @var array<string, bool>
      */
-    public function __construct(array $loaders = [])
+    private $hasSourceCache = [];
+
+    /**
+     * @param iterable<LoaderInterface> $loaders
+     */
+    public function __construct(iterable $loaders = [])
     {
-        foreach ($loaders as $loader) {
-            $this->addLoader($loader);
-        }
+        $this->loaders = $loaders;
     }
 
     public function addLoader(LoaderInterface $loader): void
     {
-        $this->loaders[] = $loader;
+        $current = $this->loaders;
+
+        $this->loaders = (static function () use ($current, $loader): \Generator {
+            yield from $current;
+            yield $loader;
+        })();
+
         $this->hasSourceCache = [];
     }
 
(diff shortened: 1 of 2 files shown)


src/Loader/ChainLoader.php
```diff
@@ -21,21 +21,30 @@
...
-        }
+        $this->loaders = $loaders;
     }
 
     public function addLoader(LoaderInterface $loader): void
     {
+        if (!\is_array($this->loaders)) {
+            $this->loaders = \iterator_to_array($this->loaders, false);
+        }
+
```
TheCelavi, line 50: ~Iterable (generators) can not be mutated by adding a new elements, you already know that. Order of loaders must be preserved. So adding loader with `addLoader()` breaks lazy loading - assuming that loaders are provided  as generator within constructor.~

~So this is simple implementation, but it has its own flaws.~

~There is, of course, better method to this madness :) -> to add some class, per example `LoadersCollection` where it would be allowed to add loader and/or  loaders without triggering generator and by preserving order.~

~Do you want that kind of implementation, perhaps? Or you are satisfied with this one?~

stof:
ah indeed. Supporting lazy-loaded iterators would require making the ChainLoader configuration immutable instead of having this `addLoader`. So let's forget it...

TheCelavi:
There is also this method which can be used to preserve lazy loading:

```php
    public function addLoader(LoaderInterface $loader): void
    {
        $existing = $this->loaders;

        $this->loaders = static function () use ($existing, $loader): \Generator {
            yield from $existing;
            yield $loader;
        };

        $this->hasSourceCache = [];
    }
```

I don't think that it would have much of the performance impact when adding 5-10 loaders (hardly real-world scenario, people usually use 1-2 loaders and that is it).

Are you more interested in this method? @stof

xabbuh:
We could deprecate `addLoader()` in `3.x` and introduce lazy loading in Twig 4.

src/Loader/ChainLoader.php
TheCelavi: `addLoaders()` now supports preserving lazy loading of loaders provided as \Generator.

method `getLoaders()` will unpack loaders.

TheCelavi:
I have updated PR with modification which will support preserving lazy loading of loaders with `addLoaders()` method.

src/Loader/ChainLoader.php
```diff
@@ -21,22 +21,33 @@
...
-        foreach ($loaders as $loader) {
-            $this->addLoader($loader);
-        }
+        $this->loaders = $loaders;
     }
 
     public function addLoader(LoaderInterface $loader): void
     {
-        $this->loaders[] = $loader;
+        $existing = $this->loaders;
```
fabpot: Maybe

```suggestion
        $current = $this->loaders;
```

TheCelavi: Renamed and squashed.

fabpot:
Thank you @TheCelavi.

PR #4213, opened 2024-08-17
Add return types missed in #4211

smnandre:
Complete #4211 (sorry)

I forgot to update the docs, and found some others in RuntimeLoader(s)

### DIFF
--- a/doc/advanced.rst
+++ b/doc/advanced.rst
@@ -360,7 +360,7 @@ This is used by many of the tests built into Twig::
 
     class OddTestExpression extends TestExpression
     {
-        public function compile(\Twig\Compiler $compiler)
+        public function compile(\Twig\Compiler $compiler): void
         {
             $compiler
                 ->raw('(')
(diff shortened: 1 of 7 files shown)


fabpot:
Thank you @smnandre.

smnandre:
@fabpot 

Do we need or plan to do it for `all` missing return types in the Twig 4.x codebase ? 

There is still a lot of them (with or without @return tags)

Examples:

* TemplateWrapper:unwrap (`@private`)
* ExpressionParser::parseArrow (private)
* ExpressionParser::parseStringExpression (public)
* Template::renderParentBlock() (`@internal`) 
* Environment::getCache() (public)

fabpot:
I suppose we can start with everything private (arg and return types), maybe everything `@internal` as well?

smnandre:
@fabpot started this here https://github.com/twigphp/Twig/pull/4218

Most of concerned methods are in fact public or protected :|

PR #4214, opened 2024-08-18
Add test about Environment version constants.

VincentLanglet:
Hi @fabpot 

The Environment Version constants seems to be manually updated.
https://github.com/twigphp/Twig/blob/fe36e084b4e208c44e30886053c8594bb99bd78f/src/Environment.php#L46-L51

A mistake was made on the 3.11 version, the MAJOR_VERSION was set to 4 instead of 3
https://github.com/twigphp/Twig/blob/e80fb8ebba85c7341a97a9ebf825d7fd4b77708d/src/Environment.php#L48

It was fixed in another version, but I think it could be useful to have some test about the consistency of all those constants.
Such test would have fail for the 3.11 version and avoid the mistake.

### DIFF
--- a/tests/EnvironmentTest.php
+++ b/tests/EnvironmentTest.php
@@ -38,6 +38,21 @@ class EnvironmentTest extends TestCase
 {
     use ExpectDeprecationTrait;
 
+    public function testVersionConstants()
+    {
+        $version = Environment::VERSION;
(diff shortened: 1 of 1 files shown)


fabpot:
Thank you @VincentLanglet.

PR #4216, opened 2024-08-19
Swap BC layer for yield-ready and reclaim perf loss

### LINKED ISSUE #4103: Performance regression since v3.9.0
What is the performance impact of the runtime check combined with yield ?

_Originally posted by @stof in https://github.com/twigphp/Twig/pull/3999#pullrequestreview-1875847101_
            

Hello,

While investigating a performance regression in a symfony app (6.4) I found that the source was twig/twig version 3.9.0.

I made a small repro and used git bissect to find this pull request #3999 where performance is briefly discussed.

After some trial and error, performance was partially restored by passing `'use_yield' => true` to the twig Environment constructor.

Not sure if this is a bug but in certain cases the performance impact was pretty huge (from 800 ms to 2200ms).

### LINKED ISSUE #4146: yield used when use_yield = false

nicolas-grekas:
Follows #3999

Fix #4146
Fix #4103

When `use_yield` is set to false (the default), this PR reverts the implementation of the `render()` method to use a wrapping output buffer instead of hooking between each steps of generators. In this mode, the behavior of the yield method is not "pure": it triggers a mix of yield and echo. But this is fine for render and display methods.

When `use_yield` is set to `true`, we skip that wrapping output buffer. This makes twig compatible with fibers (and this also makes compilation fail if a non-YieldReady extension is found.)

That makes the name of the option not ideal, but BC rulez FTW.

### DIFF
--- a/src/Template.php
+++ b/src/Template.php
@@ -166,6 +166,17 @@ abstract class Template
      */
     public function renderParentBlock($name, array $context, array $blocks = [])
     {
+        if (!$this->useYield) {
+            if ($this->env->isDebug()) {
+                ob_start();
+            } else {
+                ob_start(function () { return ''; });
+            }
+            $this->displayParentBlock($name, $context, $blocks);
+
+            return ob_get_clean();
+        }
+
         $content = '';
         foreach ($this->yieldParentBlock($name, $context, $blocks) as $data) {
             $content .= $data;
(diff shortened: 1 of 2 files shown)


src/Template.php
```diff
@@ -166,6 +166,17 @@ public function displayBlock($name, array $context, array $blocks = [], $useBloc
      */
     public function renderParentBlock($name, array $context, array $blocks = [])
     {
+        if (!$this->useYield) {
+                if ($this->env->isDebug()) {
```
GromNaN: Indentation
```suggestion
            if ($this->env->isDebug()) {
```

nicolas-grekas: oops, fixed

fabpot:
@nicolas-grekas Can you add a note in CHANGELOG?

fabpot:
I've just tested on https://github.com/j0r1s/twig-performance-loss-reproducer, and it indeed fixes the issue with the default `use_yield => false` configuration:

Before: https://blackfire.io/profiles/a94a955c-b84d-4dfb-94f0-89300e0aa4a3/graph
After: https://blackfire.io/profiles/bde4be25-00da-432e-87a0-087ed3ab5e37/graph
Comparison: https://app.blackfire.io/profiles/compare/bde4be25-00da-432e-87a0-087ed3ab5e37...a94a955c-b84d-4dfb-94f0-89300e0aa4a3/graph?settings%5Bdimension%5D=wt&settings%5Bdisplay%5D=landscape&settings%5BtabPane%5D=nodes&selected=&callname=main()&constraintDoc=

I've also tested on a simpler test script and it works great as well.

fabpot:
Thank you @nicolas-grekas.

fabpot:
Oops, forgot about the CHANGELOG. Fixed here: 3a307cd4

smnandre:
Wow thank you @nicolas-grekas for this! 👏 🙇

PR #4218, opened 2024-08-20
Add private types on 4.x

smnandre:
(WIP but i open to gather early feedback / comments)

Objective: add missing types 

- [x] Private properties
- [x] Return type on private methods
- [ ] Argument types on private methods (?not sure)

Maybe "@internal" then

### DIFF
--- a/src/ExpressionParser.php
+++ b/src/ExpressionParser.php
@@ -96,10 +96,7 @@ class ExpressionParser
         return $expr;
     }
 
-    /**
-     * @return ArrowFunctionExpression|null
-     */
-    private function parseArrow()
+    private function parseArrow(): ?ArrowFunctionExpression
     {
         $stream = $this->parser->getStream();
 
--- a/src/NodeVisitor/EscaperNodeVisitor.php
+++ b/src/NodeVisitor/EscaperNodeVisitor.php
@@ -182,7 +182,7 @@ final class EscaperNodeVisitor implements NodeVisitorInterface
         return \in_array($type, $safe) || \in_array('all', $safe);
     }
 
-    private function needEscaping()
+    private function needEscaping(): string|false
     {
         if (\count($this->statusStack)) {
             return $this->statusStack[\count($this->statusStack) - 1];
--- a/src/ExtensionSet.php
+++ b/src/ExtensionSet.php
@@ -34,7 +34,7 @@ final class ExtensionSet
     private array $extensions;
     private bool $initialized = false;
     private bool $runtimeInitialized = false;
-    private $staging;
+    private StagingExtension $staging;
 
     /**
      * @var TokenParserInterface[]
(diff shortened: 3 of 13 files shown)


smnandre:
Remaining non-typed arguments inherit their "non-types" from protected or public methods, not sure what to do there 

Example in `Profiler\Dumper\BaseDumper`

```php

    abstract protected function formatTime(Profile $profile, $percent): string;

    private function dumpProfile(Profile $profile, $prefix = '', $sibling = false): string

```

src/ExpressionParser.php
```diff
@@ -53,7 +53,7 @@ class ExpressionParser
     private array $unaryOperators;
     /** @var array<string, array{precedence: int, class: class-string<AbstractBinary>, associativity: self::OPERATOR_*}> */
     private array $binaryOperators;
-    private $readyNodes = [];
+    private array $readyNodes = [];
```
fabpot: Removed in 2acd9e8d9fcf2fa7907db8a6e1bcdd8feffcd91f

src/NodeVisitor/EscaperNodeVisitor.php
```diff
@@ -182,7 +182,7 @@ private function isSafeFor(string $type, Node $expression, Environment $env): bo
         return \in_array($type, $safe) || \in_array('all', $safe);
     }
 
-    private function needEscaping()
+    private function needEscaping(): bool
```
fabpot:
```suggestion
    private function needEscaping(): string|false
```

fabpot (changes requested):

stof:
Actually, types on all private APIs could be added in the 3.x branch

smnandre:
So i should target 3.x for this one @stof right ? And then deal with `@internal` for 4.x ?

src/ExpressionParser.php
smnandre: Thanks, rebased and removed

smnandre:
(is there something like a "fabbot_ignore"? to prevent false-positive ? asking more generally than for this PR)

fabpot:
Let's not bother with 3.x.

smnandre:
So this seems ok for a first step.

Let's create another one for @internal function/methods if you're ok ?

```
    /**
     * Strips HTML and PHP tags from a string.
     *
     * @param string|null          $string
     * @param string[]|string|null $allowable_tags
     *
     * @internal
     */
    public static function striptags($string, $allowable_tags = null): string
    {
        return strip_tags($string ?? '', $allowable_tags);
    }
```

Because here there may be a "soft-conflict" between the signature (`string|null`) and the code (passing null to `strip_tags` is deprecated since 8.1)

fabpot:
Thank you @smnandre.

stof:
@smnandre there is no conflict in your code snippet. The code is not passing null to `strip_tags`

smnandre:
Wow... let's just attribute it to today's emotions. ;)

PR #4219, opened 2024-08-21
Improve deprecation message

ruudk:
See https://github.com/twigphp/Twig/pull/4199#discussion_r1723456834

/cc @stof @fabpot


fabpot:
Thank you @ruudk.

PR #4221, opened 2024-08-22
Deprecate OptimizerNodeVisitor::OPTIMIZE_TEXT_NODES

### LINKED ISSUE #4146: yield used when use_yield = false

fabpot:
Refs #4146

While trying to debug #4146, I realized that this optimization is not possible as we don't know how nodes are going to be used.

See the script provided in #4146 to reproduce the problem.

@brandonkelly Can you confirm that this fixes your problem? Maybe there is also an issue with yield vs echo, but the script provided in the issue runs fine for me after this PR.

### DIFF
--- a/doc/deprecated.rst
+++ b/doc/deprecated.rst
@@ -141,6 +141,11 @@ Node Visitors
 * The ``Twig\NodeVisitor\AbstractNodeVisitor`` class is deprecated, implement the
   ``Twig\NodeVisitor\NodeVisitorInterface`` interface instead.
 
+* The ``Twig\NodeVisitor\OptimizerNodeVisitor::OPTIMIZE_RAW_FILTER`` and the
+  ``Twig\NodeVisitor\OptimizerNodeVisitor::OPTIMIZE_TEXT_NODES`` options are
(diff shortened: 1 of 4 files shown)


doc/deprecated.rst
```diff
@@ -141,6 +141,11 @@ Node Visitors
 * The ``Twig\NodeVisitor\AbstractNodeVisitor`` class is deprecated, implement the
   ``Twig\NodeVisitor\NodeVisitorInterface`` interface instead.
 
+* The ``Twig\NodeVisitor\OptimizerNodeVisitor::OPTIMIZE_RAW_FILTER`` and the
+  ``Twig\NodeVisitor\OptimizerNodeVisitor::OPTIMIZE_TEXT_NODES`` options are
+  deprecated as of Twig 3.12 and will be removed in Twig 4.0; they don't
+  anything anymore.
```
jdreesen:
```suggestion
  deprecated as of Twig 3.12 and will be removed in Twig 4.0; they don't do
  anything anymore.
```

nicolas-grekas (approved):

PR #4223, opened 2024-08-23
Add support for integers in methods of `Twig\Node\Node` that take a Node name

### LINKED ISSUE #3811: getNode in Twig\Node\Node has wrong typehint
The `Twig\Node\Node::getNode` method states that the `$name` parameter should be of type `string`. However, when created using `Twig\ExpressionParser::parseAssignmentExpression` (and possibly others), the `$nodes` array is actually a regular int-indexed array. When called from code using strict types, there is no way to get at the subnodes, as the correct code is `$node->getNode(0)`, but the typehint requires the parameter to be a string.

As Twig still needs to support PHP 7, the only way to fix this is probably to remove the typehint altogether - neither `mixed` nor `string|int` are available.

I'll be happy to create a PR if the solution is OK with you.

fabpot:
Closes #3811

### DIFF
--- a/src/Node/Node.php
+++ b/src/Node/Node.php
@@ -24,6 +24,9 @@ use Twig\Source;
 #[YieldReady]
 class Node implements \Countable, \IteratorAggregate
 {
+    /**
+     * @var array<string|int, Node>
+     */
     protected $nodes;
     protected $attributes;
     protected $lineno;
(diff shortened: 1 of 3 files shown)


src/Node/Node.php
```diff
@@ -48,7 +51,13 @@ public function __construct(array $nodes = [], array $attributes = [], int $line
                 throw new \InvalidArgumentException(\sprintf('Using "%s" for the value of node "%s" of "%s" is not supported. You must pass a \Twig\Node\Node instance.', \is_object($node) ? \get_class($node) : (null === $node ? 'null' : \gettype($node)), $name, static::class));
             }
         }
-        $this->nodes = $nodes;
+        $this->nodes = [];
+        foreach ($nodes as $name => $node) {
+            if (!\is_int($name) && !is_string($name)) {
```
stof: array keys cannot be anything else that an int or a string, so this is dead code.

fabpot: Yes-ish. You can use a float even if you get a deprecation notice from PHP.

stof: you can use a float when *putting* a value in the array and it is deprecated. But you will not *read* a float as it is cast by the time you assign it

fabpot: Right

PR #4225, opened 2024-08-24
Deprecate not passing a `BodyNode` instance as the body of a `ModuleNode` or `MacroNode` constructor

fabpot:


yceruto:
FYI: Fixing the deprecation on Symfony side https://github.com/symfony/symfony/pull/58078

PR #4228, opened 2024-08-24
Use CPP in full code base

andreybolonin:


src/RuntimeLoader/ContainerRuntimeLoader.php
```diff
@@ -23,11 +23,8 @@
  */
 class ContainerRuntimeLoader implements RuntimeLoaderInterface
 {
-    private $container;
-
-    public function __construct(ContainerInterface $container)
+    public function __construct(private ContainerInterface $container)
```
fabpot:
```suggestion
    public function __construct(
        private ContainerInterface $container,
    )
```

fabpot (commented):
Would you like to update the full code base?

andreybolonin:
@fabpot ok, no problem )

smnandre:
@fabpot regarding the fabbot false positive on deprecation messages, should we add a check on "symfony/symfoy" on the "DeprecationMessageChecker" ?  Or is this something that can be configured ?

src/Extension/SandboxExtension.php
```diff
@@ -55,12 +52,7 @@ public function disableSandbox(): void
 
     public function isSandboxed(?Source $source = null): bool
     {
-        return $this->sandboxedGlobally || $this->sandboxed || $this->isSourceSandboxed($source);
-    }
-
-    public function isSandboxedGlobally(): bool
```
fabpot: Must be reverted

src/Extension/SandboxExtension.php
```diff
@@ -21,16 +21,13 @@
...
-    private $policy;
-    private $sourcePolicy;
 
-    public function __construct(SecurityPolicyInterface $policy, $sandboxed = false, ?SourcePolicyInterface $sourcePolicy = null)
+    public function __construct(
+        private SecurityPolicyInterface $policy,
+        private bool $sandboxed = false,
+        private ?SourcePolicyInterface $sourcePolicy = null,
+    )
     {
```
fabpot: CS fix for all such changes:

```suggestion
    ) {
```

src/TemplateWrapper.php
```diff
@@ -18,19 +18,17 @@
...
     /**
      * This method is for internal use only and should never be called
      * directly (use Twig\Environment::load() instead).
      *
      * @internal
      */
-    public function __construct(Environment $env, Template $template)
+    public function __construct(
+        private Environment $env,
+        private Template $template
```
fabpot:
```suggestion
        private Template $template,
```

src/Token.php
```diff
@@ -37,11 +33,12 @@ final class Token
     public const ARROW_TYPE = 12;
     public const SPREAD_TYPE = 13;
 
-    public function __construct(int $type, $value, int $lineno)
+    public function __construct(
+        private int $type,
+        private $value,
+        private int $lineno
```
fabpot:
```suggestion
        private int $lineno,
```

src/TokenStream.php
```diff
@@ -21,13 +21,13 @@
...
 final class TokenStream
 {
-    private $tokens;
     private $current = 0;
-    private $source;
 
-    public function __construct(array $tokens, ?Source $source = null)
+    public function __construct(
+        private array $tokens,
+        private ?Source $source = null
```
fabpot:
```suggestion
        private ?Source $source = null,
```

fabpot (changes requested):

src/Cache/FilesystemCache.php
```diff
@@ -20,13 +20,12 @@ class FilesystemCache implements CacheInterface
 {
     public const FORCE_BYTECODE_INVALIDATION = 1;
 
-    private $directory;
-    private $options;
-
-    public function __construct(string $directory, int $options = 0)
+    public function __construct(
+        private string $directory,
```
stof: This should not use a promoted property as the value assigned in it is not the argument directly.

src/Markup.php
```diff
@@ -18,13 +18,12 @@
  */
 class Markup implements \Countable, \JsonSerializable
 {
-    private $content;
-    private $charset;
-
-    public function __construct($content, $charset)
+    public function __construct(
+        private $content,
```
stof: Same here as the constructor does not assign the argument itself.

src/Markup.php
```diff
@@ -18,13 +18,12 @@
...
+        private $charset,
```
stof: The charset must be `?string`

src/NodeVisitor/OptimizerNodeVisitor.php
```diff
@@ -47,12 +47,13 @@ final class OptimizerNodeVisitor implements NodeVisitorInterface
...
     private $loops = [];
     private $loopsTargets = [];
-    private $optimizers;
 
     /**
      * @param int $optimizers The optimizer mode
      */
-    public function __construct(int $optimizers = -1)
+    public function __construct(
+        private int $optimizers = -1,
```
stof, line 55: Why making it a promoted properties when the constructor still assigns it separately ? This makes no sense as it means the constructor assigns it twice.

src/Extension/SandboxExtension.php
```diff
@@ -55,7 +51,7 @@ public function disableSandbox(): void
 
     public function isSandboxed(?Source $source = null): bool
     {
-        return $this->sandboxedGlobally || $this->sandboxed || $this->isSourceSandboxed($source);
```
fabpot: The two properties are still needed IIUC.

src/Extension/SandboxExtension.php
```diff
@@ -21,16 +21,12 @@
...
-    private $sourcePolicy;
 
-    public function __construct(SecurityPolicyInterface $policy, $sandboxed = false, ?SourcePolicyInterface $sourcePolicy = null)
-    {
-        $this->policy = $policy;
-        $this->sandboxedGlobally = $sandboxed;
-        $this->sourcePolicy = $sourcePolicy;
+    public function __construct(
+        private SecurityPolicyInterface $policy,
+        private bool $sandboxed = false,
```
fabpot:
```suggestion
        private bool $sandboxedGlobally = false,
```
?

stof (changes requested):

src/NodeVisitor/OptimizerNodeVisitor.php
fabpot: Line 69 (`$this->optimizers = $optimizers;`) should be removed

src/Cache/FilesystemCache.php
fabpot: I agree, let's revert cases where we need to do the assignment in the constructor.

src/Extension/SandboxExtension.php
```diff
@@ -21,16 +21,12 @@
 
 final class SandboxExtension extends AbstractExtension
 {
-    private $sandboxedGlobally;
-    private $sandboxed;
```
fabpot: This is still needed

fabpot (changes requested):

fabpot:
Thank you @andreybolonin.

PR #4230, opened 2024-08-24
Move code from ExtendsTokenParser to Parser

fabpot:


src/Parser.php
```diff
@@ -292,6 +292,10 @@ public function getParent(): ?Node
 
     public function setParent(?Node $parent): void
     {
+        if (null !== $this->parent) {
```
smnandre, line 299:
```suggestion
        if (null !== $parent && null !== $this->parent) {
```

Or maybe remove the nullable here ?

Issue #4165, opened 2024-07-29
Add tag to declare variable types


drjayvee:
See #4003 and #4009 for some context. In short, since the PHP language itself as well as its community are moving to (optional) stricter typing, it makes sense for Twig to support this as well by adding a `types` tag.

~~@fabpot said he'd like to bootstrap a spec. This is the first draft. Everything is up for debate, obviously. In addition, I've put in italics parts of the spec that I have doubts about myself.~~

Given the discussions below (thanks all for contributing), I think the spec is now crystallizing. Consider it a beta.

# Syntax
The `types` tag contains a mapping of variable name to constant type string (no expression or interpolation):

```
{% types {variable: 'Type'} %}
```
The reason for quoting the type is that several characters (e.g., `\`) are currently not supported by the Lexer. It also allows analyzer implementations to freely specify the syntax.

The variable name _must not_ be quoted. This is to ensure that only valid names can be used.

# Types
The syntax and content of the types are not part of this specification. This provides complete freedom to extensions that inspect and analyze the tags. It also keeps the implementation in Twig very simple and robust.

(In practice, it's likely that implementations will rely on widely-used specifications like [PHPDoc / PHPStan types](https://phpstan.org/writing-php-code/phpdoc-types).)

## Examples
Here are two examples:
```twig
{% types {foo: 'bool'} %}

{% types {
  bar: "list<int>",
  baz: '?\Name\Space\ClassName'
} %}
```

# Behavior
This specification does require any behavior other than validating the syntax for the tag itself. The tag does not lead to generated PHP code.

# Usage
This section describes possible implementations to demonstrate the utility of this addition.

It is up to extensions developed by the community to implement any of this.

## Development time
The types can be useful during development by enabling IDEs to auto-complete code, issue warnings, and more.

This is [already implemented](https://github.com/twigphp/Twig/pull/4009#issuecomment-2003568293) by the Symfony Support plugin for PHPStorm when using comments to declare variable types (e.g., `{# @var id int #}`).

These tags also provide precious documentation about the variables expected in a template's context.

## Compile time
During compilation, a Twig extension can:
1. Check the type and throw an appropriate Exception if it is not valid.
2. For objects, check whether properties and more methods exist, and throw an appropriate Exception if there is none.
2. Add run-time checks (see below).

## Run time
During run time, behavior may depend on `strict_variables` environment option:
* If *disabled* (likely in production environments), a PHP warning is triggered.
[...]

PR #4231, opened 2024-08-24
💡 Node `assert`❔

smnandre:
Acting as a _spinoff_ to [#4165 Add tag to declare variable types](https://github.com/twigphp/Twig/issues/4165)
(as I don't want to disrupt the main conversation too much 😅)

The idea is to introduce assertions.

I'm opening this to gather feedback and ideas. I'd like to see if there's something interesting here—or not. :)

This could enhance static analysis, define contracts in templates, and serve as a validator during includes/rendering.

```twig
{% assert name string %}
{% assert size in ['sm', 'md', 'lg'] %}
```

Open to any  feedback :)

### DIFF
--- a/src/Node/AssertNode.php
+++ b/src/Node/AssertNode.php
@@ -1,0 +1,46 @@
+<?php
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Twig\Node;
+
+use Twig\Attribute\YieldReady;
+use Twig\Compiler;
+use Twig\Node\Expression\AbstractExpression;
+
+/**
+ * Represents a assert node.
+ *
+ * @author Simon André <smn.andre@gmail.com>
+ */
+#[YieldReady]
+class AssertNode extends Node
+{
+    public function __construct(AbstractExpression $expr, int $lineno, ?string $tag = null)
+    {
+        parent::__construct(['expr' => $expr], [], $lineno, $tag);
+    }
+
+    public function compile(Compiler $compiler): void
+    {
+        $compiler->addDebugInfo($this);
+        // TODO disable in production(?)
+
+        $compiler
+            ->write(' if (!')
+            ->subcompile($this->getNode('expr'))
+            ->raw(") {\n")
+            ->indent()
+            ->write("throw new \RuntimeException('Failed');\n")
+            ->outdent()
+            ->raw("}\n")
+        ;
+    }
+}
--- a/src/Node/AssertNode.php
+++ b/src/Node/AssertNode.php
@@ -38,7 +38,7 @@ class AssertNode extends Node
             ->subcompile($this->getNode('expr'))
             ->raw(") {\n")
             ->indent()
-            ->write("throw new \RuntimeException('Failed');\n")
+            ->write("throw new RuntimeError('Failed');\n")
             ->outdent()
             ->raw("}\n")
         ;
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -69,6 +69,7 @@ use Twig\Source;
 use Twig\Template;
 use Twig\TemplateWrapper;
 use Twig\TokenParser\ApplyTokenParser;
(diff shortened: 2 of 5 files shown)


smnandre:
## Usage / syntax examples 


| Langage                                                                                  | Code example                                                 |
|------------------------------------------------------------------------------------------|-----------------------------------------------------------------|
| [**PHP**](https://www.php.net/manual/en/function.assert.php)                             | `assert(in_array($size, ['sm', 'md']));`                        |
| [**Python**](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement) | `assert size in ['sm', 'md']`                                   |
| [**Java**](https://docs.oracle.com/javase/8/docs/technotes/guides/language/assert.html)  | `assert java.util.Arrays.asList("sm", "md").contains(size);`    |
| [**C**](https://en.cppreference.com/w/c/error/assert)                                    | `assert(strcmp(size, "sm") == 0 \|\| strcmp(size, "md") == 0);` |
| [**JavaScript**](https://developer.mozilla.org/en-US/docs/Web/API/console/assert)        | `console.assert(["sm", "md"].includes(size));`                  |
| [**C++**](https://en.cppreference.com/w/cpp/error/assert)                                | `assert(size == "sm" \|\| size == "md");`                       |
| [**Ruby**](https://ruby-doc.org/core-3.1.2/Kernel.html#method-i-raise)                   | `raise 'Assertion failed' unless ['sm', 'md'].include?(size)`   |


(Table made with some gpt -  for representational purposes only)

src/Node/AssertNode.php
```diff
@@ -0,0 +1,46 @@
...
+
+        $compiler
+            ->write(' if (!')
+            ->subcompile($this->getNode('expr'))
+            ->raw(") {\n")
+            ->indent()
+            ->write("throw new RuntimeError('Failed');\n")
+            ->outdent()
+            ->raw("}\n")
+        ;
```
Kocal, line 44: What about using `assert()` function?  
With OPcache configuration, you can literally remove `assert()` calls from the code... or here you planned to not write anything if production?

smnandre: That's an idea too !  I would say a node would be easier to parse / use by node visitors but i may be wrong here 😅 

> or here you planned to not write anything if production?

I planned... _nothing_ 🙃   the "idea" just emerged earlier tonight during the "var/type" discussion

stof: The suggestion is about the compiled code, not about the Twig API

Actually, to be able to use PHP's `assert` function which is a no-op at runtime in production mode, it needs to be a Twig tag and not Twig function usable in an expression.

drjayvee:
When it comes to asserting variable types, that can be left over to extensions like [TwigQI](https://github.com/alisqi/TwigQI), once `types` is merged.

I do like the idea of using `assert` for expressions.

There's currently no way to intentionally throw an Exception of trigger a warning (aside from something silly like `{{ [].foo }}`).

I'd propose that the `assert` tag should have an optional message as part of its spec:
```
{% assert <expression> [message string] %}
```

@ruudk is working on compiling Twig templates to PHP in such a way that PHPStan can work its magic. If the `assert` tag gets a similar treatment, with properly typed variables, we can take a huge leap in template quality.

fabpot:
> @ruudk is working on compiling Twig templates to PHP in such a way that PHPStan can work its magic. If the `assert` tag gets a similar treatment, with properly typed variables, we can take a huge leap in template quality.

I'm curious to see how that would look like... but also if that's something we can include in the "default" Twig compilation (I suppose this is based on attributes/annotations with no runtime side-effects).

ruudk:
You can read more about TwigStan here: https://github.com/twigphp/Twig/discussions/4233

src/Node/AssertNode.php
smnandre: 👍

fabpot:
Now that we have the `types` tag, can we close this one?

smnandre:
> Now that we have the `types` tag, can we close this one?


In the end, i think this "assert" would be more related to assertions / tests than strict typing... 

But I do agree we should focus on the "types" tag and see  how / why it is used, the features it enables, ... 

(and we can reopen here if needed later)

PR #4235, opened 2024-08-26
Add types tag

drjayvee:
The is a draft implementation for #4165.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -31,6 +31,7 @@
  * Bump minimum PHP version to 8.0
  * Fix integration tests when a test has more than one data/expect section and deprecations
  * Add the `enum_cases` function
+ * Add the `types` tag
 
 # 3.11.0 (2024-08-08)
 
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,6 @@
 # 3.13.0 (2024-XX-XX)
 
- * n/a
+ * Add the `types` tag (experimental)
 
 # 3.12.0 (2024-08-29)
 
@@ -31,7 +31,6 @@
  * Bump minimum PHP version to 8.0
  * Fix integration tests when a test has more than one data/expect section and deprecations
  * Add the `enum_cases` function
- * Add the `types` tag
 
 # 3.11.0 (2024-08-08)
 
--- a/doc/tags/types.rst
+++ b/doc/tags/types.rst
@@ -14,6 +14,15 @@ Here is how you declare that variable ``foo`` is a boolean, while ``bar`` is an 
         bar: 'int',
     } %}
 
+You can declare variables as optional by adding the '?' suffix:
+
+.. code-block:: twig
+
+    {% types {
+        foo: 'bool',
+        bar?: 'int',
+    } %}
+
 By default, this tag does not affect the template compilation or runtime behavior.
 
 Its purpose is to enable designers and developers to document and specify the context's available
--- a/doc/tags/types.rst
+++ b/doc/tags/types.rst
@@ -1,11 +1,15 @@
 ``types``
 =========
 
-The ``types`` tag declares the types for template variables.
+.. versionadded:: 3.13
+
+    The ``types`` tag was added in Twig 3.13. This tag is **experimental** and can change based on usage and feedback.
+
+The ``types`` tag declares the types of template variables.
 
 To do this, specify a :ref:`mapping <twig-expressions>` of names to their types as strings. 
 
-Here is how you declare that variable ``foo`` is a boolean, while ``bar`` is an integer (see note below):
+Here is how to declare that ``foo`` is a boolean, while ``bar`` is an integer (see note below):
 
 .. code-block:: twig
 
@@ -14,7 +18,7 @@ Here is how you declare that variable ``foo`` is a boolean, while ``bar`` is an 
         bar: 'int',
     } %}
 
-You can declare variables as optional by adding the '?' suffix:
+You can declare variables as optional by adding the ``?`` suffix:
 
 .. code-block:: twig
 
@@ -30,7 +34,7 @@ and/or required variables. While Twig itself does not validate variables or thei
 to do this.
 
 Additionally, :ref:`Twig extensions <creating_extensions>` can analyze these tags to perform compile-time and
-runtime analysis of templates in order to increase code quality.
+runtime analysis of templates.
 
 .. note::
 
--- a/doc/tags/types.rst
+++ b/doc/tags/types.rst
@@ -1,9 +1,9 @@
 ``types``
-=======
+=========
 
-The ``types`` tag declares the types for one or more names variables.
+The ``types`` tag declares the types for template variables.
 
-To do this, specify a :ref:`mapping <twig-expressions>` of names to their types as string constants. 
+To do this, specify a :ref:`mapping <twig-expressions>` of names to their types as strings. 
 
 Here is how you declare that variable ``foo`` is a boolean, while ``bar`` is an integer (see note below):
 
@@ -14,12 +14,11 @@ Here is how you declare that variable ``foo`` is a boolean, while ``bar`` is an 
         bar: 'int',
     } %}
 
-You may put these all on one line. However, using one line per variable improves readability.
-
-This tag has no effect on the template's output or runtime behavior.
+By default, this tag does not affect the template compilation or runtime behavior.
 
 Its purpose is to enable designers and developers to document and specify the context's available
-and/or required variables.
+and/or required variables. While Twig itself does not validate variables or their types, this tag enables extensions
+to do this.
 
 Additionally, :ref:`Twig extensions <creating_extensions>` can analyze these tags to perform compile-time and
 runtime analysis of templates in order to increase code quality.
--- a/doc/tags/types.rst
+++ b/doc/tags/types.rst
@@ -1,0 +1,30 @@
+``types``
+=======
+
+The ``types`` tag declares the types for one or more names variables.
+
+To do this, specify a :ref:`mapping <twig-expressions>` of names to their types as string constants. 
+
+Here is how you declare that variable ``foo`` is a boolean, while ``bar`` is an integer (see note below):
+
+.. code-block:: twig
+
+    {% types {
+        foo: 'bool',
+        bar: 'int',
+    } %}
+
+You may put these all on one line. However, using one line per variable improves readability.
+
+This tag has no effect on the template's output or runtime behavior.
+
+Its purpose is to enable designers and developers to document and specify the context's available
+and/or required variables.
+
+Additionally, :ref:`Twig extensions <creating_extensions>` can analyze these tags to perform compile-time and
+runtime analysis of templates in order to increase code quality.
+
+.. note::
+
+    The syntax for and contents of type strings are intentionally left out of scope. 
+    
--- a/src/Node/TypesNode.php
+++ b/src/Node/TypesNode.php
@@ -12,7 +12,6 @@ use Twig\Node\Expression\NameExpression;
  * Represents a types node.
  *
  * @author Jeroen Versteeg <jeroen@alisqi.com>
- * @see    https://github.com/twigphp/Twig/issues/4165
  */
 #[YieldReady]
 class TypesNode extends Node implements NodeCaptureInterface
--- a/src/Node/TypesNode.php
+++ b/src/Node/TypesNode.php
@@ -1,0 +1,25 @@
+<?php
+
+namespace Twig\Node;
+
+use Twig\Compiler;
+use Twig\Node\Expression\ArrayExpression;
+
+/**
+ * Represents a types node.
+ *
+ * @author Jeroen Versteeg <jeroen@alisqi.com>
+ * @see https://github.com/twigphp/Twig/issues/4165
+ */
+class TypesNode extends Node implements NodeCaptureInterface
+{
+    public function __construct(ArrayExpression $typesNode, int $lineno, ?string $tag = null)
+    {
+        parent::__construct(['mapping' => $typesNode], [], $lineno, $tag);
+    }
+
+    public function compile(Compiler $compiler)
+    {
+        // Don't compile anything.
+    }
+}
--- a/src/Node/TypesNode.php
+++ b/src/Node/TypesNode.php
@@ -4,18 +4,44 @@ namespace Twig\Node;
 
 use Twig\Compiler;
 use Twig\Node\Expression\ArrayExpression;
+use Twig\Node\Expression\ConstantExpression;
+use Twig\Node\Expression\NameExpression;
 
 /**
  * Represents a types node.
  *
  * @author Jeroen Versteeg <jeroen@alisqi.com>
- * @see https://github.com/twigphp/Twig/issues/4165
+ * @see    https://github.com/twigphp/Twig/issues/4165
  */
 class TypesNode extends Node implements NodeCaptureInterface
 {
     public function __construct(ArrayExpression $typesNode, int $lineno, ?string $tag = null)
     {
+        $this->validateMapping($typesNode);
+
         parent::__construct(['mapping' => $typesNode], [], $lineno, $tag);
+    }
+
+    protected function validateMapping(ArrayExpression $typesNode): void
+    {
+        foreach ($typesNode->getKeyValuePairs() as $i => $pair) {
+            $keyExpression = $pair['key'];
+            $valueExpression = $pair['value'];
+
+            if (!$keyExpression instanceof NameExpression) {
+                throw new \InvalidArgumentException("Key at index $i is not a NameExpression");
+            }
+            $name = $keyExpression->getAttribute('name');
+
+            if (!$valueExpression instanceof ConstantExpression) {
+                throw new \InvalidArgumentException("Value for key \"$name\" is not a ConstantExpression");
+            }
+            $value = $valueExpression->getAttribute('value');
+
+            if (!is_string($value)) {
+                throw new \InvalidArgumentException("Value for key \"$name\" is not a string");
+            }
+        }
     }
 
     public function compile(Compiler $compiler)
--- a/src/Node/TypesNode.php
+++ b/src/Node/TypesNode.php
@@ -4,9 +4,6 @@ namespace Twig\Node;
 
 use Twig\Attribute\YieldReady;
 use Twig\Compiler;
-use Twig\Node\Expression\ArrayExpression;
-use Twig\Node\Expression\ConstantExpression;
-use Twig\Node\Expression\NameExpression;
 
 /**
  * Represents a types node.
@@ -16,33 +13,12 @@ use Twig\Node\Expression\NameExpression;
 #[YieldReady]
 class TypesNode extends Node implements NodeCaptureInterface
 {
-    public function __construct(ArrayExpression $typesNode, int $lineno)
+    /**
+     * @param array<string, array{type: string, optional: bool}> $types
+     */
+    public function __construct(array $types, int $lineno)
     {
-        $this->validateMapping($typesNode);
-
-        parent::__construct(['mapping' => $typesNode], [], $lineno);
-    }
-
-    protected function validateMapping(ArrayExpression $typesNode): void
-    {
-        foreach ($typesNode->getKeyValuePairs() as $i => $pair) {
-            $keyExpression = $pair['key'];
-            $valueExpression = $pair['value'];
-
-            if (!$keyExpression instanceof NameExpression) {
-                throw new \InvalidArgumentException("Key at index $i is not a NameExpression");
-            }
-            $name = $keyExpression->getAttribute('name');
-
-            if (!$valueExpression instanceof ConstantExpression) {
-                throw new \InvalidArgumentException("Value for key \"$name\" is not a ConstantExpression");
-            }
-            $value = $valueExpression->getAttribute('value');
-
-            if (!is_string($value)) {
-                throw new \InvalidArgumentException("Value for key \"$name\" is not a string");
-            }
-        }
+        parent::__construct([], ['mapping' => $types], $lineno);
     }
 
     public function compile(Compiler $compiler)
--- a/src/Node/TypesNode.php
+++ b/src/Node/TypesNode.php
@@ -17,11 +17,11 @@ use Twig\Node\Expression\NameExpression;
 #[YieldReady]
 class TypesNode extends Node implements NodeCaptureInterface
 {
-    public function __construct(ArrayExpression $typesNode, int $lineno, ?string $tag = null)
+    public function __construct(ArrayExpression $typesNode, int $lineno)
     {
         $this->validateMapping($typesNode);
 
-        parent::__construct(['mapping' => $typesNode], [], $lineno, $tag);
+        parent::__construct(['mapping' => $typesNode], [], $lineno);
     }
 
     protected function validateMapping(ArrayExpression $typesNode): void
--- a/src/Node/TypesNode.php
+++ b/src/Node/TypesNode.php
@@ -2,6 +2,7 @@
 
 namespace Twig\Node;
 
+use Twig\Attribute\YieldReady;
 use Twig\Compiler;
 use Twig\Node\Expression\ArrayExpression;
 use Twig\Node\Expression\ConstantExpression;
@@ -13,6 +14,7 @@ use Twig\Node\Expression\NameExpression;
  * @author Jeroen Versteeg <jeroen@alisqi.com>
  * @see    https://github.com/twigphp/Twig/issues/4165
  */
+#[YieldReady]
 class TypesNode extends Node implements NodeCaptureInterface
 {
     public function __construct(ArrayExpression $typesNode, int $lineno, ?string $tag = null)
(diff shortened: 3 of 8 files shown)


drjayvee:
I just force-pushed a fix to the `docs`

drjayvee:
This PR isn't complete yet. I forgot to add a `TypesTokenParser` (and tests).

doc/tags/types.rst
```diff
@@ -0,0 +1,30 @@
+``types``
+=======
```
fabpot:
```suggestion
=========
```

doc/tags/types.rst
```diff
@@ -0,0 +1,30 @@
...
+
+The ``types`` tag declares the types for one or more names variables.
```
fabpot:
```suggestion
The ``types`` tag declares the types for template variables.
```

doc/tags/types.rst
```diff
@@ -0,0 +1,30 @@
...
+
+To do this, specify a :ref:`mapping <twig-expressions>` of names to their types as string constants.
```
fabpot:
```suggestion
To do this, specify a :ref:`mapping <twig-expressions>` of names to their types as strings.
```

doc/tags/types.rst
```diff
@@ -0,0 +1,30 @@
+``types``
+=======
+
+The ``types`` tag declares the types for one or more names variables.
+
+To do this, specify a :ref:`mapping <twig-expressions>` of names to their types as string constants. 
+
+Here is how you declare that variable ``foo`` is a boolean, while ``bar`` is an integer (see note below):
```
fabpot:
```suggestion
Here is how you declare that variable ``foo`` is a ``boolean``, while ``bar`` is an ``integer``:
```

doc/tags/types.rst
```diff
@@ -0,0 +1,30 @@
...
+
+.. code-block:: twig
+
+    {% types {
+        foo: 'bool',
+        bar: 'int',
+    } %}
+
+You may put these all on one line. However, using one line per variable improves readability.
```
fabpot: I think we can remove this sentence as this is true for any mapping.

doc/tags/types.rst
```diff
@@ -0,0 +1,30 @@
...
+
+This tag has no effect on the template's output or runtime behavior.
```
fabpot:
```suggestion
By default, this tag does not affect the template compilation or runtime behavior.
```

doc/tags/types.rst
```diff
@@ -0,0 +1,30 @@
...
+
+Its purpose is to enable designers and developers to document and specify the context's available
+and/or required variables.
+
+Additionally, :ref:`Twig extensions <creating_extensions>` can analyze these tags to perform compile-time and
+runtime analysis of templates in order to increase code quality.
+
+.. note::
+
+    The syntax for and contents of type strings are intentionally left out of scope.
```
fabpot, line 41: Thinking out loud here: maybe we can document a default subset that should always be supported by extensions, the basic types used by Twig: boolean, number (maybe integer, float), mapping, sequence, string/stringable, object, iterable?

doc/tags/types.rst
```diff
@@ -0,0 +1,30 @@
...
+
+You may put these all on one line. However, using one line per variable improves readability.
+
+This tag has no effect on the template's output or runtime behavior.
+
+Its purpose is to enable designers and developers to document and specify the context's available
+and/or required variables.
+
+Additionally, :ref:`Twig extensions <creating_extensions>` can analyze these tags to perform compile-time and
+runtime analysis of templates in order to increase code quality.
```
fabpot:
```suggestion
runtime analysis of templates.
```

IMHO, it's not just about code quality. Let's people be creative :)

doc/tags/types.rst
```diff
@@ -0,0 +1,30 @@
...
+        foo: 'bool',
+        bar: 'int',
+    } %}
+
+You may put these all on one line. However, using one line per variable improves readability.
+
+This tag has no effect on the template's output or runtime behavior.
+
+Its purpose is to enable designers and developers to document and specify the context's available
+and/or required variables.
```
fabpot: I would remove this sentence (see my comment below about creativity).

doc/tags/types.rst
drjayvee: I wanted to make clear that interpolated strings (let alone concatenated ones) are not supported. But perhaps that's a bit pedantic. Users will run into a `SyntaxError` if they try anything other than constant strings.

doc/tags/types.rst
drjayvee: I'm not sure using back ticks for "boolean" and "integer" is a good idea here, since the actual code below uses `bool` (not `boolean`) and `int` (not `integer`). Doesn't that create confusion?

doc/tags/types.rst
drjayvee: Let's discuss this in #4165. I'll quote your point and respond there.

doc/tags/types.rst
drjayvee: See my comment below.

doc/tags/types.rst
drjayvee: I do want to point out the raison d'être for this to hopefully push adoption.

If the tag doesn't affect templates and seemingly has no point (if you're just looking at the official documentation), then won't many developers ignore and forget about this tag entirely?

I'm fine with removing this to keep the docs clean and to the point. I just wanted to bring up why I included this. Let me know what you prefer.

doc/tags/types.rst
fabpot: A concatenation of strings is an expression, not a string. So, "string" is very clear IMHO.

doc/tags/types.rst
fabpot: See my other comment below about defining a minimal set of types that must be supported.

doc/tags/types.rst
fabpot: I see. Maybe we could rephrase to suggest that this is one possible use case?

drjayvee: How about

> While Twig itself does not validate variables or their types, this tag enables extensions to do this.

fabpot: That works for me

drjayvee:
To quote my last commit message:
> Move parsing of `{name: 'string'}` mappings to ExpressionParser 
> 
> This also adds support for optional keys as discussed in https://github.com/twigphp/Twig/issues/4165.
> 
> I really don't like using a static method, but I couldn't figure out how
else to call the method in tests. See the notes on the method itself.
> 
> Any help is appreciated!

src/TokenParser/TypesTokenParser.php
```diff
@@ -0,0 +1,58 @@
...
+ * @see    https://github.com/twigphp/Twig/issues/4165
+ * @internal
+ */
+final class TypesTokenParser extends AbstractTokenParser
+{
+    public function parse(Token $token): Node
+    {
+        $stream = $this->parser->getStream();
+
+        $expression = ExpressionParser::parseNameToStringMapping($stream, true);
```
ruudk:
```suggestion
        $expression = $this->parser->getExpressionParser()->parseNameToStringMapping($stream, true);
```

drjayvee: That would work, but the `TokenStream` argument could be completely different than `$this->parser`'s instance variable.

It's also not possible to create a `new ExpressionParser($this->parser, $env)` in `TypesTokenParser` because the latter doesn't have the required `Twig\Environment`. (`Parser->env` is private and there's no `public` getter.)

Is there a more elegant solution?

ruudk: But this is how it's done everywhere. I would follow that.

drjayvee: Thanks for the suggestion.

Unfortunately, this doesn't work, because `Parser` only sets its `$expressionParser` in `parse()`. Once that method has returned, the stream has already been parsed. (There's no way to rewind the `TokenStream`.)

I did remove `static` by simply creating a `new ExpressionParser()`.

drjayvee:
I don't understand the CI error:
```
Remaining self deprecation notices (2)

  2x: Since twig/twig 3.12: The "tag" constructor argument of the "Twig\Node\TypesNode" class is deprecated and ignored (check which TokenParser class set it to "null"), the tag is now automatically set by the Parser when needed.
    1x in TypesTest::getTests from Twig\Tests\Node
    1x in TypesTest::testConstructor from Twig\Tests\Node
```

[That test](https://github.com/twigphp/Twig/blob/917cdd5d457e6ba4769d350983457221803ac24a/tests/Node/TypesTest.php) _does not_ pass the `tag` argument.

Update: I also can't reproduce this on my local dev stack.

I think the problem is that the `TypesNode` constructor passes `tag` to `parent::__construct()`. I'll remove that and push again.

src/TokenParser/TypesTokenParser.php
stof: why wouldn't it work ? token parsers are called from inside the Parser parse method, so the ExpressionParser will be set.

src/Node/TypesNode.php
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * @author Jeroen Versteeg <jeroen@alisqi.com>
+ * @see    https://github.com/twigphp/Twig/issues/4165
+ */
+#[YieldReady]
+class TypesNode extends Node implements NodeCaptureInterface
+{
+    public function __construct(ArrayExpression $typesNode, int $lineno)
+    {
+        $this->validateMapping($typesNode);
```
ruudk: I don't think this needs to be validated anymore, as parseNameToStringMapping already handles this.

src/ExpressionParser.php
```diff
@@ -432,6 +432,52 @@ public function parseMappingExpression()
         return $node;
     }
 
+    /*
+     * Unlike its sibling methods above, method requires a TokenStream parameter.
```
stof: I vote for making this method work like the others instead.

src/TokenParser/TypesTokenParser.php
ruudk: This works:
```patch
diff --git a/src/ExpressionParser.php b/src/ExpressionParser.php
index 6b18c535..4166505e 100644
--- a/src/ExpressionParser.php
+++ b/src/ExpressionParser.php
@@ -437,7 +437,7 @@ class ExpressionParser
      * Calling `$this->parser->getStream()` as done in `parseMappingExpression()` only works if
      * `parser->parse()` has been called, but that won't ever call this method.
      */
-    public static function parseNameToStringMapping(TokenStream $stream, bool $allowOptionalKeys): ArrayExpression
+    public function parseNameToStringMapping(TokenStream $stream, bool $allowOptionalKeys): ArrayExpression
     {
         $stream->expect(Token::PUNCTUATION_TYPE, '{', 'A mapping element was expected');
 
diff --git a/src/Parser.php b/src/Parser.php
index 28cc8a0e..92d3f686 100644
--- a/src/Parser.php
+++ b/src/Parser.php
@@ -49,6 +49,7 @@ class Parser
     public function __construct(Environment $env)
     {
         $this->env = $env;
+        $this->expressionParser = new ExpressionParser($this, $this->env);
     }
 
@@ -67,10 +68,6 @@ class Parser
             $this->visitors = $this->env->getNodeVisitors();
         }
 
-        if (null === $this->expressionParser) {
-            $this->expressionParser = new ExpressionParser($this, $this->env);
-        }
-
         $this->stream = $stream;
         $this->parent = null;
         $this->blocks = [];
diff --git a/src/TokenParser/TypesTokenParser.php b/src/TokenParser/TypesTokenParser.php
index 8e5f65ae..cb9a5c00 100644
--- a/src/TokenParser/TypesTokenParser.php
+++ b/src/TokenParser/TypesTokenParser.php
@@ -44,7 +44,7 @@ final class TypesTokenParser extends AbstractTokenParser
     {
 
-        $expression = ExpressionParser::parseNameToStringMapping($stream, true);
+        $expression = $this->parser->getExpressionParser()->parseNameToStringMapping($stream, true);
 
diff --git a/tests/ExpressionParserTest.php b/tests/ExpressionParserTest.php
index b6a4fdea..2b1d6d26 100644
--- a/tests/ExpressionParserTest.php
+++ b/tests/ExpressionParserTest.php
@@ -213,10 +213,12 @@ class ExpressionParserTest extends TestCase
     public function testNameToStringMapping(string $template, ArrayExpression $expected, bool $allowOptionalKeys): void
     {
         $env = new Environment(new ArrayLoader(), ['cache' => false, 'autoescape' => false]);
+        $parser = new Parser($env);
+
         $stream = $env->tokenize(new Source($template, ''));
         $stream->next();
-        
-        $this->assertEquals($expected, ExpressionParser::parseNameToStringMapping($stream, $allowOptionalKeys));
+
+        $this->assertEquals($expected, $parser->getExpressionParser()->parseNameToStringMapping($stream, $allowOptionalKeys));
     }
 
     public function getTestsForNameToStringMapping(): array
@@ -276,9 +278,12 @@ class ExpressionParserTest extends TestCase
         $this->expectException(SyntaxError::class);
         
         $env = new Environment(new ArrayLoader(), ['cache' => false, 'autoescape' => false]);
+        $parser = new Parser($env);
+
         $stream = $env->tokenize(new Source($template, ''));
-        
-        ExpressionParser::parseNameToStringMapping($stream, $allowOptionalKeys);
+        $stream->next();
+
+        $parser->getExpressionParser()->parseNameToStringMapping($stream, $allowOptionalKeys);
     }
 
     public function getFailingTestsForNameToStringMapping(): array
```

src/ExpressionParser.php
```diff
@@ -432,6 +432,52 @@ public function parseMappingExpression()
...
+     * 
+     * The reason is that calling `$this->parser->getStream()` (as in `parseMappingExpression()`)
+     * only works _while_ `parser->parse()` is running, because that method sets and unsets the $stream property.
+     * 
+     * The upside is that this method is fully pure, since it doesn't depend on the state of $this.
+     */
+    public function parseNameToStringMapping(TokenStream $stream, bool $allowOptionalKeys): ArrayExpression
+    {
+        $stream->expect(Token::PUNCTUATION_TYPE, '{', 'A mapping element was expected');
```
ruudk: I think this should be like this, to follow the rest of the ExpressionParser

```suggestion
    public function parseNameToStringMapping(bool $allowOptionalKeys): ArrayExpression
    {
        $stream = $this->parser->getStream();
        $stream->expect(Token::PUNCTUATION_TYPE, '{', 'A mapping element was expected');
```

src/ExpressionParser.php
```diff
@@ -432,6 +432,52 @@ public function parseMappingExpression()
...
 
+    /*
+     * Unlike its sibling methods above, method requires a TokenStream parameter.
+     * 
+     * The reason is that calling `$this->parser->getStream()` (as in `parseMappingExpression()`)
+     * only works _while_ `parser->parse()` is running, because that method sets and unsets the $stream property.
+     * 
+     * The upside is that this method is fully pure, since it doesn't depend on the state of $this.
+     */
+    public function parseNameToStringMapping(TokenStream $stream, bool $allowOptionalKeys): ArrayExpression
```
stof: This `$allowOptionalKeys` parameter should be removed, as it is always used as `true`

tests/ExpressionParserTest.php
```diff
@@ -208,6 +209,142 @@ public function getTestsForSequence()
...
+
+        $this->assertEquals($expected, $expressionParser->parseNameToStringMapping($stream, $allowOptionalKeys));
+    }
+
+    public function getTestsForNameToStringMapping(): array
+    {
+        return [
+            // empty map
+            [
+                '{{ {} }}',
```
stof: Those tests don't make sense as they don't parse the source that is visible here (btw, those hacky tests seem to be the primary reason why the design of `ExpressionParser::parseNameToStringMapping` does not match the design of other parts of the Twig parser).

Your tests should be parsing a `{% types %}` tag and asserting its structure instead.

src/Node/TypesNode.php
```diff
@@ -0,0 +1,53 @@
...
+
+        parent::__construct(['mapping' => $typesNode], [], $lineno);
+    }
+
+    protected function validateMapping(ArrayExpression $typesNode): void
```
stof: this validation looks useless as the TypesTokenParser already guarantees that this is respected.

src/ExpressionParser.php
stof: I would even suggest removing that method entirely by moving this logic to the TypesTokenParser instead. This would be cleaner by keeping the ExpressionParser responsible for parsing expressions (while the TypesTokenParser explicitly expects a different syntax than a Twig expression)

src/ExpressionParser.php
```diff
@@ -432,6 +432,52 @@ public function parseMappingExpression()
...
+
+            $nameToken = $stream->expect(Token::NAME_TYPE);
+            $nameExpression = new NameExpression($nameToken->getValue(), $nameToken->getLine());
+
+            if ($allowOptionalKeys) {
+                $isOptional = $stream->nextIf(Token::PUNCTUATION_TYPE, '?') !== null;
+                $nameExpression->setAttribute('is_optional', $isOptional);
+            }
+
+            $stream->expect(Token::PUNCTUATION_TYPE, ':', 'A mapping key must be followed by a colon (:)');
```
stof: This is not a mapping key

src/ExpressionParser.php
```diff
@@ -432,6 +432,52 @@ public function parseMappingExpression()
...
+    public function parseNameToStringMapping(TokenStream $stream, bool $allowOptionalKeys): ArrayExpression
+    {
+        $stream->expect(Token::PUNCTUATION_TYPE, '{', 'A mapping element was expected');
+
+        $node = new ArrayExpression([], $stream->getCurrent()->getLine());
+
+        $first = true;
+        while (!$stream->test(Token::PUNCTUATION_TYPE, '}')) {
+            if (!$first) {
+                $stream->expect(Token::PUNCTUATION_TYPE, ',', 'A mapping value must be followed by a comma');
```
stof: this is not a mapping value

src/TokenParser/TypesTokenParser.php
```diff
@@ -0,0 +1,58 @@
...
+ * 
+ * Names may be suffixed with `?` to mark them as optional (i.e., the context may not
+ * have that variable defined).
+ * 
+ * Example:
+ * ```
+ *    {% types {foo: 'int', bar?: 'string'} %}
+ * ```
+ * @author Jeroen Versteeg <jeroen@alisqi.com>
+ * @see    https://github.com/twigphp/Twig/issues/4165
```
stof: Those `@see` linking to the Twig issue requesting the feature should be removed.

stof (changes requested):

src/ExpressionParser.php
drjayvee: First off, thanks for the review! 🙇

> I vote for making this method work like the others instead.

I tried that at first, but I couldn't write tests for it like the others (e.g., `testSequenceExpression`) because the `ExpressionParser` never calls this new method directly.

> I would even suggest removing that method entirely by moving this logic to the TypesTokenParser instead. 

Yeah, that's what I had initially. @ruudk also agrees that its better (for now) to move this (back) to `TypesTokenParser`.

I'll `rebase -i` my branch accordingly.

tests/ExpressionParserTest.php
drjayvee: You're spot on re: test design.

> Your tests should be parsing a {% types %} tag and asserting its structure instead.

Yep, will do!

src/Node/TypesNode.php
drjayvee: But the constructor takes a (generic) `ArrayExpression`.

I considered creating a specialized `NameToStringMappingExpression` to lock this down, but thought that would be overkill. What would you advise?

I _do_ want to nail down the structure of the expression to prevent errors down the line (post-construction). It doesn't hurt, right?

drjayvee: @ruudk suggested taking in an `array` in the constructor and store it as an attribute instead.

stof: I also think that this does not need to be an ArrayExpression (especially as we want to be a lot more strict)

ruudk: What if we make it an array typed as `array<string, string>`. Then store it in the attributes of the TypesNode? 

Optional variables will have an `?` at the end of the string key.

stof: this would indeed be a lot simpler as it would not expect type-checking extensions to support expressions (or to hardcode the restrictions on the possible expressions)

Another solution is `array<string, array{type: string, optional: bool}>`

ruudk: > array<string, array{type: string, optional: bool}>

Also fine!

drjayvee: Done!

ruudk (approved):
Looks great 👏

CHANGELOG
```diff
@@ -16,6 +16,7 @@
  * Bump minimum PHP version to 8.0
  * Fix integration tests when a test has more than one data/expect section and deprecations
  * Add the `enum_cases` function
+ * Add the `types` tag
```
```suggestion
 * Add the `types` tag (experimental)
```

doc/tags/types.rst
```diff
@@ -0,0 +1,38 @@
+``types``
+=========
+
```
fabpot, line 3:
```suggestion
.. versionadded:: 3.13

    The ``types`` tag was added in Twig 3.13. This tag is **experimental** and can change based on usage and feedback.
```

CHANGELOG
fabpot: I've just bumped the version to 3.13 so that we can merge this PR. Can you rebase and move this line to the 3.13 section?

doc/tags/types.rst
```diff
@@ -0,0 +1,38 @@
...
+The ``types`` tag declares the types for template variables.
```
fabpot:
```suggestion
The ``types`` tag declares the types of template variables.
```

doc/tags/types.rst
```diff
@@ -0,0 +1,38 @@
...
+
+To do this, specify a :ref:`mapping <twig-expressions>` of names to their types as strings. 
+
+Here is how you declare that variable ``foo`` is a boolean, while ``bar`` is an integer (see note below):
```
fabpot:
```suggestion
Here is how to declare that ``foo`` is a boolean, while ``bar`` is an integer (see note below):
```

doc/tags/types.rst
```diff
@@ -0,0 +1,38 @@
...
+
+.. code-block:: twig
+
+    {% types {
+        foo: 'bool',
+        bar: 'int',
+    } %}
+
+You can declare variables as optional by adding the '?' suffix:
```
fabpot:
```suggestion
You can declare variables as optional by adding the ``?`` suffix:
```

doc/tags/types.rst
```diff
@@ -0,0 +1,38 @@
...
+    } %}
+
+By default, this tag does not affect the template compilation or runtime behavior.
+
+Its purpose is to enable designers and developers to document and specify the context's available
+and/or required variables. While Twig itself does not validate variables or their types, this tag enables extensions
+to do this.
+
+Additionally, :ref:`Twig extensions <creating_extensions>` can analyze these tags to perform compile-time and
+runtime analysis of templates in order to increase code quality.
```
fabpot:
```suggestion
runtime analysis of templates.
```

Could be more than just code quality (performance optimization comes to mind for instance).

src/TokenParser/TypesTokenParser.php
```diff
@@ -0,0 +1,84 @@
...
+
+        $types = $this->parseSimpleMappingExpression($stream);
+
+        $stream->expect(Token::BLOCK_END_TYPE);
+
+        return new TypesNode($types, $token->getLine());
+    }
+
+    /**
+     * @throws SyntaxError
```
fabpot: Should be after the `@return` one with one empty line in between.

src/TokenParser/TypesTokenParser.php
```diff
@@ -0,0 +1,84 @@
...
+                if ($stream->test(Token::PUNCTUATION_TYPE, '}')) {
+                    break;
+                }
+            }
+            $first = false;
+
+            $nameToken = $stream->expect(Token::NAME_TYPE);
+            $isOptional = $stream->nextIf(Token::PUNCTUATION_TYPE, '?') !== null;
+
+            $stream->expect(Token::PUNCTUATION_TYPE, ':', 'A name must be followed by a colon (:)');
```
fabpot:
```suggestion
            $stream->expect(Token::PUNCTUATION_TYPE, ':', 'A type name must be followed by a colon (:)');
```

fabpot (changes requested):
LGTM. I've added some minor comments.

fabpot:
Thank you @drjayvee.

fabpot:
Not sure why Github didn't work as expected here. I'm closing but it's merged.

VincentLanglet:
Hi @drjayvee,

Is there a reason to not support
```
{% types {
        'foo': 'bool',
        'bar'?: 'int',
    } %}
```
too ? Like it's done for classic objects `{% set bar = {'foo': 42} %}`.

ruudk:
I guess we did not consider this as it looks a bit ugly for optionally. But sure, we could still accept this. Especially as it's better in line with mappings. 

I do hope we can turn of the quoting for the `types` tag in Twig-CS-Fixer though 😉

stof:
@VincentLanglet variable names need to be valid identifiers anyway to be usable, so I don't see the benefit of supporting string literals as keys there instead of only identifiers.
This syntax is not an actual hash anyway (the parser does not allow arbitrary expressions)

PR #4236, opened 2024-08-26
Deprecate the spaceless filter

### LINKED ISSUE #3576: [BUG] Extension::spaceless: Algorithm removes spaces in the text and causes the meaning to be broken.
The current algorithm causes the meaning of a sentence to be broken because it concatenates multiple tags together, which can be, for example, a link and a formatting tag.

In case of template:

```html
{% block content %}
{% apply spaceless %}
<p>
    Open <a href="{{ url }}">Google</a> <strong>for more</strong> info.
</p>
{% endapply %}
{% endblock %}
```

Current output is:

```html
<p>
    Open <a href="https://google.cz">Google</a><strong>for more</strong> info.
</p>
```

But expected output **must be**:

```html
<p>
    Open <a href="https://google.cz">Google</a> <strong>for more</strong> info.
</p>
```

Snímek obrazovky 2021-10-14 v 15 31 06

I think more better algorithm should be inspired from Latte engine: https://github.com/nette/latte/blob/53e5da9aa2ec20e6e7e4149448e5d0ad2af282ba/src/Latte/Runtime/Filters.php#L309

Possible solution
------

The algorithm could replace spaces for some tag types with a single separate space, not necessarily remove everything. Formatting tags can be like `p`, `a`, `strong`, `b`, `i`, `u`, ...

Thanks.

Rel #3574

fabpot:
Closes #3576

Deprecate the `spaceless` filter for the following reasons:

* The performance is bad (as the work is done at runtime via a regexp)
* Optimizing the size of an HTML doc server side is "almost never" a good idea (compression is better and enough)
* There are some edge cases where you want to keep some spaces (see #3576)
* Controlling whitespace is possible and fine-grained via the dedicated Twig modifiers on `{{ }}`

If someone find it useful, re-creating it is trivial (`return trim(preg_replace('/>\s+</', '><', $content ?? ''));`),
but with so many caveats and not so many use cases, I think it does not belong to core.


src/Extension/CoreExtension.php
```diff
@@ -1144,9 +1144,13 @@ public static function nl2br($string): string
      * @param string|null $content
      *
      * @internal
+     *
+     * @deprecated since Twig 3.12
      */
     public static function spaceless($content): string
     {
+        trigger_deprecation('twig/twig', '3.12', 'The "spaceless" filter is deprecated.');
```
stof: Do we need a runtime deprecation for each usage of the filter ? can't we rely only on the compile-time deprecation instead ?

stof (changes requested):
This filter should be marked as deprecated in the filter definition, so that using the filter reports the deprecation at compile time (and so that the linter also reports it)

src/Extension/CoreExtension.php
stof: OK, actually, the compile-time deprecation is not implemented.

fabpot: Indeed, as the function is `@internal`, let's use the built-in deprecation feature. Code updated accordingly.

stof (approved):

ttk:
I didn't find any good examples of how to replace `spaceless` with the existing whitespace controls.  So here's what I did:

```twig
{% apply spaceless %}
  <span>1</span>
  <span>2</span>
  <span>3</span>
{% endapply %}
```

To rewrite this without spaceless, I used `{{ }}` with a null condition and the appropriate whitespace controls:
```twig
  <span>1</span>{{ null -}}
  <span>2</span>{{ null -}}
  <span>3</span>
```

I experimented with other "null" conditions, and the following also work: `{{ '' -}}` , `{{ "" -}}` , and `{{ false -}}`, but I find `{{ null -}}` to be my personal preference.

Not sure if there is a better syntax to handle this?  It might be worth adding this example to the docs?

stof:
@ttk the shortest solution (and most efficient by disappearing entirely at runtime) is to use a comment to hold the whitespace control: `{# -#}`

ttk:
@stof Thanks, that is exactly what I was looking for!   I honestly didn't know that whitespace control worked on comments as well but after re-reviewing the docs, I see a single example shown [here](https://twig.symfony.com/doc/3.x/templates.html#whitespace-control).

PR #4237, opened 2024-08-26
Fix two-word tests precedence over one-word tests

### LINKED ISSUE #3282: Unable to register a test with spaces in name
When registering a test named "empty element" I am getting the following error.

> Unexpected token "name" of value "element" ("end of statement block" expected).

That's likely because the Twig core has a test named [empty](https://twig.symfony.com/doc/3.x/tests/empty.htmll).

fabpot:
Closes #3282

### DIFF
--- a/src/ExpressionParser.php
+++ b/src/ExpressionParser.php
@@ -750,15 +750,15 @@ class ExpressionParser
         $stream = $this->parser->getStream();
         $name = $stream->expect(Token::NAME_TYPE)->getValue();
 
-        if (!$test = $this->env->getTest($name)) {
-            if ($stream->test(Token::NAME_TYPE)) {
-                // try 2-words tests
-                $name = $name.' '.$this->parser->getCurrentToken()->getValue();
+        if ($stream->test(Token::NAME_TYPE)) {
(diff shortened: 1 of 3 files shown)


src/ExpressionParser.php
```diff
@@ -750,15 +750,15 @@ private function getTest(int $line): TwigTest
...
-                // try 2-words tests
-                $name = $name.' '.$this->parser->getCurrentToken()->getValue();
+        if ($stream->test(Token::NAME_TYPE)) {
+            // try 2-words tests
+            $name = $name.' '.$this->parser->getCurrentToken()->getValue();
 
-                if ($test = $this->env->getTest($name)) {
-                    $stream->next();
-                }
+            if ($test = $this->env->getTest($name)) {
```
stof, line 757: shouldn't this have a `else` to use the 1-word test ?

fabpot: Having something after the test is not valid anyway. So, I prefer to have an exception saying `Unknown "empty element" test.` instead of an exception saying `Unexpected token "name" of value "element" ("end of statement block" expected).`

Issue #4158, opened 2024-07-28
Fix behaviour of `cycle` function & mappings ?


smnandre:
I have a couple questions about the "[cycle](https://github.com/twigphp/Twig/blob/71021834cc0f8694e0d858830799137e1714b0ee/src/Extension/CoreExtension.php#L318-L339C6)" function... 

I'm not sure what is really expected in some scenarios (mainly "what to do with `mappings`"), and the [documentation](https://twig.symfony.com/doc/3.x/functions/cycle.html) uses "mapping" "sequence" and "list"  terms.

I'll gladly open a PR, but i'm not sure what to do here :) 

---

<details open><summary>Current code of the <code>cycle</code> function here</summary>
<p>

```php
/**
 * Cycles over a value.
 *
 * @param \ArrayAccess|array $values
 * @param int                $position The cycle position
 *
 * @return string The next value in the cycle
 *
 * @internal
 */
public static function cycle($values, $position): string
{
    if (!\is_array($values) && !$values instanceof \ArrayAccess) {
        return $values;
    }

    if (!\count($values)) {
        throw new RuntimeError('The "cycle" function does not work on empty sequences/mappings.');
    }

    return $values[$position % \count($values)];
}
```

</p>
</details> 

---

### ArrayAccess ?

If $values implements `ArrayAccess` but not `Countable`, depending on the PHP version the effects can be a simple warning or ... a `DivisionByZeroError`

* Genuine cursiosity: what is the use case with `ArrayAccess` here ? 
* Should we check is_countable first ? 
* Or should this be added in the docblock ? 
* Should we deprecate passing a not countable `ArrayAccess` ? 

### Sequences / Mappings 

There is nothing explicit in the docblock, and the exception is about "_sequences/mappings_" but .. in reality this method does only works with lists (sequences) of strings.

* Should this be checked before accessing the result / counting items ? 
* Or inversely, if $values is a mapping, should we convert it to a sequence first ? 
* Should the "position" value be tested ?  (a negative value will not work here)
* Should the 'string' return type be enforced / checked  somewhere ?

stof:
I would suggest removing the return type, as there is no reason to restrict it to strings (we could want to cycle over a list of booleans).

I would suggest documenting that this function expects a sequence and a non-negative position (not sure whether we need runtime checks for those, I will let @fabpot decide).

We should indeed forbid passing a non-countable ArrayAccess (with a deprecation first).

[...]

PR #4241, opened 2024-08-26
Adjust `cycle` implementation

smnandre:
Update code following [#4158](https://github.com/twigphp/Twig/issues/4158)
* return mixed
* trigger when non countable
* type the position argument

Add doc precisions

### DIFF
--- a/doc/functions/cycle.rst
+++ b/doc/functions/cycle.rst
@@ -1,7 +1,7 @@
 ``cycle``
 =========
 
-The ``cycle`` function cycles on a sequence or mapping:
+The ``cycle`` function cycles on a sequence:
 
 .. code-block:: twig
 
@@ -23,7 +23,9 @@ The ``cycle`` function cycles on a sequence or mapping:
         
     #}
 
-The array can contain any number of values:
+The ``cycle`` function takes two arguments: the ``sequence`` to cycle through and the ``position`` in the sequence.
+
+The ``sequence`` must be non-empty and can contain any number of values:
 
 .. code-block:: twig
 
@@ -52,5 +54,5 @@ The array can contain any number of values:
 Arguments
 ---------
 
-* ``values``: The list of values to cycle on
-* ``position``: The cycle position
+* ``values``: The sequence to cycle on
+* ``position``: The position in the sequence
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -330,26 +330,39 @@ final class CoreExtension extends AbstractExtension
     }
 
     /**
-     * Cycles over a value.
+     * Cycles over a sequence.
      *
-     * @param \ArrayAccess|array $values
-     * @param int                $position The cycle position
+     * @param array|\ArrayAccess $values   A non-empty sequence of values
+     * @param positive-int       $position The position of the value to return in the cycle
      *
-     * @return string The next value in the cycle
+     * @return mixed The value at the given position in the sequence, wrapping around as needed
      *
      * @internal
      */
-    public static function cycle($values, $position): string
+    public static function cycle($values, $position): mixed
     {
-        if (!\is_array($values) && !$values instanceof \ArrayAccess) {
-            return $values;
+        if (!\is_array($values)) {
+            if (!$values instanceof \ArrayAccess) {
+                throw new RuntimeError('The "cycle" function expects an array or "ArrayAccess" as first argument.');
+            }
+
+            if (!\is_countable($values)) {
+                // To be uncommented in 4.0
+                // throw new RuntimeError('The "cycle" function expects a countable sequence as first argument.');
+
+                trigger_deprecation('twig/twig', '3.12', 'Passing a non-countable sequence of values to "%s()" is deprecated.', __METHOD__);
+
+                return $values;
+            }
+
+            $values = self::toArray($values, false);
         }
 
-        if (!\count($values)) {
-            throw new RuntimeError('The "cycle" function does not work on empty sequences/mappings.');
+        if (!$count = \count($values)) {
+            throw new RuntimeError('The "cycle" function does not work on empty sequences.');
         }
 
-        return $values[$position % \count($values)];
+        return $values[$position % $count];
     }
 
     /**
--- a/tests/Extension/CoreTest.php
+++ b/tests/Extension/CoreTest.php
@@ -18,6 +18,46 @@ use Twig\Extension\CoreExtension;
 class CoreTest extends TestCase
 {
     /**
+     * @dataProvider provideCycleCases
+     */
+    public function testCycleFunction($values, $position, $expected)
+    {
+        $this->assertSame($expected, CoreExtension::cycle($values, $position));
+    }
+
+    public static function provideCycleCases()
+    {
+        return [
+            [[1, 2, 3], 0, 1],
+            [[1, 2, 3], 1, 2],
+            [[1, 2, 3], 2, 3],
+            [[1, 2, 3], 3, 1],
+            [[false, 0, null], 0, false],
+            [[false, 0, null], 1, 0],
+            [[false, 0, null], 2, null],
+
+            [[['a', 'b'], ['c', 'd']], 3, ['c', 'd']],
+        ];
+    }
+
+    /**
+     * @dataProvider provideCycleInvalidCases
+     */
+    public function testCycleFunctionThrowRuntimeError($values,  mixed $position = null)
+    {
+        $this->expectException(RuntimeError::class);
+        CoreExtension::cycle($values, $position ?? 0);
+    }
+
+    public static function provideCycleInvalidCases()
+    {
+        return [
+            'empty' => [[]],
+            'non-countable' => [new class extends \ArrayObject{}],
+        ];
+    }
+
+    /**
      * @dataProvider getRandomFunctionTestData
      */
     public function testRandomFunction(array $expectedInArray, $value1, $value2 = null)
--- a/tests/Fixtures/functions/cycle.test
+++ b/tests/Fixtures/functions/cycle.test
@@ -2,15 +2,19 @@
 "cycle" function
 --TEMPLATE--
 {% for i in 0..6 %}
-{{ cycle(array1, i) }}-{{ cycle(array2, i) }}
+{{ cycle(array1, i) }}-{{ cycle(array2, i) }}-{{ cycle(array3, i) }}
 {% endfor %}
 --DATA--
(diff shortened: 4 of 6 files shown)


doc/functions/cycle.rst
```diff
@@ -23,34 +23,33 @@ The ``cycle`` function cycles on a sequence or mapping:
         
     #}
 
-The array can contain any number of values:
+The `cycle` function takes two arguments: the `sequence` of values to cycle through and the `position` in the sequence.
```
fabpot:
```suggestion
The ``cycle`` function takes two arguments: the ``sequence`` of values to cycle through and the ``position`` in the sequence.
```

Also, we cut at 80 chars per line.

doc/functions/cycle.rst
```diff
@@ -23,34 +23,33 @@ The ``cycle`` function cycles on a sequence or mapping:
...
+
+The `sequence` must not be empty, the provided values can be of any type:
```
fabpot:
```suggestion
The ``sequence`` must not be empty, the provided values can be of any type:
```

src/Extension/CoreExtension.php
```diff
@@ -11,6 +11,8 @@
 
 namespace Twig\Extension;
 
+use ArrayAccess;
```
fabpot: We don't use statements for built-in classes.

src/Extension/CoreExtension.php
```diff
@@ -330,26 +332,36 @@ public function getOperators(): array
...
+     * @param array|ArrayAccess $values   A non-empty sequence of values.
+     * @param positive-int      $position The position of the value to return in the cycle.
      *
-     * @return string The next value in the cycle
+     * @return mixed The value at the given position in the sequence, wrapping around as needed.
      *
      * @internal
      */
-    public static function cycle($values, $position): string
+    public static function cycle(array|ArrayAccess $values, int $position): mixed
```
fabpot: That's a BC break.

fabpot (changes requested):
Just CS review for now.

doc/functions/cycle.rst
```diff
@@ -1,7 +1,7 @@
 ``cycle``
 =========
 
-The ``cycle`` function cycles on a sequence or mapping:
+The ``cycle`` function cycles on a sequence of values:
```
fabpot:
```suggestion
The ``cycle`` function cycles on a sequence:
```

doc/functions/cycle.rst
```diff
@@ -23,34 +23,33 @@ The ``cycle`` function cycles on a sequence or mapping:
         
     #}
 
-The array can contain any number of values:
+The ``cycle`` function takes two arguments: the ``sequence`` of values to cycle through and the ``position`` in the sequence.
```
fabpot:
```suggestion
The ``cycle`` function takes two arguments: the ``sequence`` to cycle through and the ``position`` in the sequence.
```

doc/functions/cycle.rst
```diff
@@ -23,34 +23,33 @@ The ``cycle`` function cycles on a sequence or mapping:
...
+
+The ``sequence`` must not be empty, the provided values can be of any type:
 
 .. code-block:: twig
 
-    {% set fruits = ['apple', 'orange', 'citrus'] %}
```
fabpot: I like concrete examples instead of random ones. Let's revert to the `fruits` example.

doc/functions/cycle.rst
```diff
@@ -23,34 +23,33 @@ The ``cycle`` function cycles on a sequence or mapping:
...
+        c B 1
+        
     #}
 
 Arguments
 ---------
 
-* ``values``: The list of values to cycle on
-* ``position``: The cycle position
+* ``values``: The sequence of values to cycle on
```
fabpot:
```suggestion
* ``values``: The sequence to cycle on
```

src/Extension/CoreExtension.php
```diff
@@ -330,26 +330,36 @@ public function getOperators(): array
     }
 
     /**
-     * Cycles over a value.
+     * Cycles over of sequence of values.
```
```suggestion
     * Cycles over a sequence.
```

src/Extension/CoreExtension.php
```diff
@@ -330,26 +330,36 @@ public function getOperators(): array
...
+     * @param array|\ArrayAccess $values   A non-empty sequence of values.
+     * @param positive-int      $position The position of the value to return in the cycle.
      *
-     * @return string The next value in the cycle
+     * @return mixed The value at the given position in the sequence, wrapping around as needed.
      *
      * @internal
      */
-    public static function cycle($values, $position): string
+    public static function cycle(mixed $values, int $position): mixed
```
```suggestion
    public static function cycle($values, $position): mixed
```

fabpot (changes requested):

smnandre:
~~All fixed~~

Documentation states it "cycles over" .. and we see it behaves more as a "static modulo-accessor". 

I was wonderning if this function could return (or act as) an iterator ?

doc/functions/cycle.rst
smnandre: I keep this open: i'll rework it later today.

stof:
@smnandre the common use case is to use this `cycle` inside the body of a loop, using `loop.index` as position. The cycle function is not the one controlling the iteration behavior (if you want to iterate over the array passed as input, you don't need a function for that as the for loop can work with the array directly)

smnandre:
> @smnandre the common use case is to use this `cycle` inside the body of a loop, using `loop.index` as position. The cycle function is not the one controlling the iteration behavior (if you want to iterate over the array passed as input, you don't need a function for that as the for loop can work with the array directly)

This is what i was thinking of ..  would it be possible to _not_ pass the "position" ? 

```twig
{% for turtle in turtles %}

    🐢  {{ cycle([ '🔵', '🔴', '🟠', '🟣']) }}   

{% endfor %}

{# would render #}
🐢 🔵  🐢 🔴  🐢 🟠  🐢 🟣
```

fabpot:
> > @smnandre the common use case is to use this `cycle` inside the body of a loop, using `loop.index` as position. The cycle function is not the one controlling the iteration behavior (if you want to iterate over the array passed as input, you don't need a function for that as the for loop can work with the array directly)
> 
> This is what i was thinking of .. would it be possible to _not_ pass the "position" ?
> 
> ```
> {% for turtle in turtles %}
> 
>     🐢  {{ cycle([ '🔵', '🔴', '🟠', '🟣']) }}   
> 
> {% endfor %}
> 
> {# would render #}
> 🐢 🔵  🐢 🔴  🐢 🟠  🐢 🟣
> ```

It already works:

```
{% for turtle in turtles %}
    🐢  {{ loop.cycle('🔵', '🔴', '🟠', '🟣') }}
{% endfor %}
```

doc/functions/cycle.rst
smnandre: No time nor inspiration these last day so ... back to fruits :)

smnandre:
With  `loop.` yep ! And i really like this DX :) 

Maybe this is something to highlight in the doc ? Or even deprecate the cycle function ?

fabpot:
> With `loop.` yep ! And i really like this DX :)
> 
> Maybe this is something to highlight in the doc ? Or even deprecate the cycle function ?

I think both serve different purposes. I forgot to say that this is available as of 4.0. Docs are available: https://github.com/twigphp/Twig/blob/4.x/doc/tags/for.rst

src/Extension/CoreExtension.php
```diff
@@ -330,26 +330,36 @@ public function getOperators(): array
     }
 
     /**
-     * Cycles over a value.
+     * Cycles over a sequence.
      *
-     * @param \ArrayAccess|array $values
-     * @param int                $position The cycle position
+     * @param array|\ArrayAccess $values A non-empty sequence of values.
+     * @param positive-int $position     The position of the value to return in the cycle.
```
fabpot: `$position` should be aligned with `$values` above

src/Extension/CoreExtension.php
```diff
@@ -330,26 +330,36 @@ public function getOperators(): array
     }
 
     /**
-     * Cycles over a value.
+     * Cycles over a sequence.
      *
-     * @param \ArrayAccess|array $values
-     * @param int                $position The cycle position
+     * @param array|\ArrayAccess $values A non-empty sequence of values.
```
```suggestion
     * @param array|\ArrayAccess $values A non-empty sequence of values
```

src/Extension/CoreExtension.php
```suggestion
     * @param positive-int $position     The position of the value to return in the cycle
```

src/Extension/CoreExtension.php
```diff
@@ -330,26 +330,36 @@ public function getOperators(): array
...
+     * @param positive-int $position     The position of the value to return in the cycle.
      *
-     * @return string The next value in the cycle
+     * @return mixed The value at the given position in the sequence, wrapping around as needed.
```
```suggestion
     * @return mixed The value at the given position in the sequence, wrapping around as needed
```

src/Extension/CoreExtension.php
```diff
@@ -330,26 +330,36 @@ public function getOperators(): array
...
-            return $values;
+        if (!\is_array($values)) {
+            if (!$values instanceof \ArrayAccess) {
+                throw new RuntimeError('The "cycle" function expects an array or "ArrayAccess" as first argument.');
+            }
+
+            if (!\is_countable($values)) {
+                trigger_deprecation('twig/twig', '3.12', 'Passing a non-countable sequence of values to "%s()" is deprecated.', __METHOD__);
+
+                return $values;
```
fabpot, line 355: We are going to throw an Exception is 4.0? If yes, can you add it commented here?

fabpot (changes requested):

doc/functions/cycle.rst
```diff
@@ -52,5 +54,5 @@ The array can contain any number of values:
 Arguments
 ---------
 
-* ``values``: The list of values to cycle on
-* ``position``: The cycle position
+* ``values``: The sequence to cycle on
+* ``position``: The position in the sequence
```
fabpot:
```suggestion
* ``position``: The position in the sequence
```

fabpot (changes requested):

src/Extension/CoreExtension.php
smnandre: Added  "will not be possible in Twig 4.0."  

Not sure here.. do you prefer to mention explicitely "will throw an exception" ?

fabpot: I meant adding the code that throws the exception as commented. So that when merging, we dont forget about it.

smnandre: Done (sorry)

fabpot:
Thank you @smnandre.

PR #4242, opened 2024-08-27
Fix CoreExtension::captureOutput

### LINKED ISSUE #4146: yield used when use_yield = false

nicolas-grekas:
Fix #4146

I forgot these lines in #4216


fabpot:
Thank you @nicolas-grekas.

PR #4247, opened 2024-08-28
deprecate passing a tag to ImportNode

xabbuh:
following #4244


fabpot:
Thank you @xabbuh.

jdreesen:
Do we need an entry in the changelog?

stof:
I don't think it is needed. Creating ImportNode outside Twig core should be very rare (and we could consider it covered by the changelog entry of the previous PR)

PR #4248, opened 2024-08-28
Add conditional return types to `ensureTraversable`

ruudk:
This way, static analyzers can understand what the output of this call will be.

### DIFF
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -1280,6 +1280,9 @@ final class CoreExtension extends AbstractExtension
 
     /**
      * @internal
+     * @template TSequence
+     * @param TSequence $seq
+     * @return ($seq is iterable ? TSequence : array{})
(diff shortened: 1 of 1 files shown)


src/Extension/CoreExtension.php
```diff
@@ -1280,6 +1280,9 @@ public static function callMacro(Template $template, string $method, array $args
 
     /**
      * @internal
+     * @template TSequence of mixed
```
stof: `of mixed` is useless (as `mixed` is the top type)

ruudk: You mean it's not needed?
So we can do this:
```suggestion
     * @template TSequence
```

ruudk: I like to be explicit, but fine. I will remove it 😊

fabpot:
Thank you @ruudk.

PR #4249, opened 2024-08-28
Only unset loop when with_loop

ruudk:
Even though PHP does not complain, PHPStan does.
> Cannot unset offset 'loop' on array{ ... the context values ... }

Since this is compiled code, we can easily produce a bit more valid code in the eyes of PHPStan.

### DIFF
--- a/src/Node/ForNode.php
+++ b/src/Node/ForNode.php
@@ -101,7 +101,11 @@ class ForNode extends Node
         $compiler->write("\$_parent = \$context['_parent'];\n");
 
         // remove some "private" loop variables (needed for nested loops)
-        $compiler->write('unset($context[\'_seq\'], $context[\'_iterated\'], $context[\''.$this->getNode('key_target')->getAttribute('name').'\'], $context[\''.$this->getNode('value_target')->getAttribute('name').'\'], $context[\'_parent\'], $context[\'loop\']);'."\n");
+        $compiler->write('unset($context[\'_seq\'], $context[\'_iterated\'], $context[\''.$this->getNode('key_target')->getAttribute('name').'\'], $context[\''.$this->getNode('value_target')->getAttribute('name').'\'], $context[\'_parent\']');
+        if ($this->getAttribute('with_loop')) {
+            $compiler->raw(', $context[\'loop\']');
+        }
+        $compiler->raw(");\n");
 
         // keep the values set in the inner context for variables defined in the outer context
         $compiler->write("\$context = array_intersect_key(\$context, \$_parent) + \$_parent;\n");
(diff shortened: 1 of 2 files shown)


src/Node/ForNode.php
```diff
@@ -101,7 +101,11 @@ public function compile(Compiler $compiler): void
         $compiler->write("\$_parent = \$context['_parent'];\n");
 
         // remove some "private" loop variables (needed for nested loops)
-        $compiler->write('unset($context[\'_seq\'], $context[\'_iterated\'], $context[\''.$this->getNode('key_target')->getAttribute('name').'\'], $context[\''.$this->getNode('value_target')->getAttribute('name').'\'], $context[\'_parent\'], $context[\'loop\']);'."\n");
+        $compiler->write('unset($context[\'_seq\'], $context[\'_iterated\'], $context[\''.$this->getNode('key_target')->getAttribute('name').'\'], $context[\''.$this->getNode('value_target')->getAttribute('name').'\'], $context[\'_parent\']);'."\n");
+
+        if ($this->getAttribute('with_loop')) {
+            $compiler->write('unset($context[\'loop\']);'."\n");
```
fabpot: Can we use a single `unset()` call in this case?

ruudk: Sure, done!

src/Node/ForNode.php
```diff
@@ -101,7 +101,11 @@ public function compile(Compiler $compiler): void
         $compiler->write("\$_parent = \$context['_parent'];\n");
 
         // remove some "private" loop variables (needed for nested loops)
-        $compiler->write('unset($context[\'_seq\'], $context[\'_iterated\'], $context[\''.$this->getNode('key_target')->getAttribute('name').'\'], $context[\''.$this->getNode('value_target')->getAttribute('name').'\'], $context[\'_parent\'], $context[\'loop\']);'."\n");
+        $compiler->write('unset($context[\'_seq\'], $context[\'_iterated\'], $context[\''.$this->getNode('key_target')->getAttribute('name').'\'], $context[\''.$this->getNode('value_target')->getAttribute('name').'\'], $context[\'_parent\']');
+        if ($this->getAttribute('with_loop')) {
+            $compiler->write(', $context[\'loop\']');
```
stof: this needs to use `->raw` so that it does not add indentation again (as it is not meant to write new lines but to continue the content of the current line).
Same below for the end of the line

stof (changes requested):

src/Node/ForNode.php
ruudk: Good to know, changed!

fabpot:
Thank you @ruudk.

PR #4250, opened 2024-08-28
Add return type `isTraitable`

ruudk:
I'd like to do this for more methods that are exposed in the compiled code. This will ease static analysis as there are less errors that are coming from Twig. 

This is just a first small PR to see if this is acceptable 😊

### DIFF
--- a/src/Node/ModuleNode.php
+++ b/src/Node/ModuleNode.php
@@ -115,7 +115,7 @@ final class ModuleNode extends Node
         $parent = $this->getNode('parent');
 
         $compiler
-            ->write("protected function doGetParent(array \$context)\n", "{\n")
+            ->write("protected function doGetParent(array \$context): bool|string|Template|TemplateWrapper\n", "{\n")
             ->indent()
             ->addDebugInfo($parent)
             ->write('return ')
@@ -160,7 +160,9 @@ final class ModuleNode extends Node
                 ->write("use Twig\Sandbox\SecurityNotAllowedFilterError;\n")
                 ->write("use Twig\Sandbox\SecurityNotAllowedFunctionError;\n")
                 ->write("use Twig\Source;\n")
-                ->write("use Twig\Template;\n\n")
+                ->write("use Twig\Template;\n")
+                ->write("use Twig\TemplateWrapper;\n")
+                ->write("\n")
             ;
         }
         $compiler
@@ -170,8 +172,11 @@ final class ModuleNode extends Node
             ->raw(" extends Template\n")
             ->write("{\n")
             ->indent()
-            ->write("private \$source;\n")
-            ->write("private \$macros = [];\n\n")
+            ->write("private Source \$source;\n")
+            ->write("/**\n")
+            ->write(" * @var array<string, Template>\n")
+            ->write(" */\n")
+            ->write("private array \$macros = [];\n\n")
         ;
     }
 
@@ -377,7 +382,7 @@ final class ModuleNode extends Node
             ->write("/**\n")
             ->write(" * @codeCoverageIgnore\n")
             ->write(" */\n")
-            ->write("public function getTemplateName()\n", "{\n")
+            ->write("public function getTemplateName(): string\n", "{\n")
             ->indent()
             ->write('return ')
             ->repr($this->getSourceContext()->getName())
@@ -434,7 +439,7 @@ final class ModuleNode extends Node
             ->write("/**\n")
             ->write(" * @codeCoverageIgnore\n")
             ->write(" */\n")
-            ->write("public function isTraitable()\n", "{\n")
+            ->write("public function isTraitable(): bool\n", "{\n")
             ->indent()
             ->write("return false;\n")
             ->outdent()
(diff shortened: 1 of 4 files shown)


stof:
I suggest also adding the `@return` phpdoc in 3.x

ruudk:
I can change the PR, target v3, and then start applying `@return` types. Then in v4 we can (or not) change them to native types.

stof:
@ruudk I would change them to native type in v4 (as done for the Twig public API)

ruudk:
Alright, then we start with this PR for v4. When it's merged, I'll continue for v3 with `@return` tags, and then later the native types in v4 again.

fabpot:
> Alright, then we start with this PR for v4. When it's merged, I'll continue for v3 with `@return` tags, and then later the native types in v4 again.

Let's start with v3, it makes things easier for me and more logical.

ruudk:
Alright done. Targeting v3 and rebased.

~After this is merged, I will do a slightly bigger PR that changes a few more places in one go.~

ruudk:
Made some changes. Turns out that we need to add the missing type hints to the `abstract Template`. 

Next to that, we now also type the 2 properties in the compiled templates.

src/Node/ModuleNode.php
```diff
@@ -170,7 +171,13 @@ protected function compileClassHeader(Compiler $compiler)
             ->raw(" extends Template\n")
             ->write("{\n")
             ->indent()
+            ->write("/**\n")
+            ->write(" * @var Source\n")
+            ->write(" */\n")
             ->write("private \$source;\n")
```
fabpot: What about using a proper type hint (this class is final anyway and the variable private)?

```suggestion
            ->write("private Source \$source;\n")
```

src/Node/ModuleNode.php
```diff
@@ -170,7 +171,13 @@ protected function compileClassHeader(Compiler $compiler)
...
+            ->write("/**\n")
+            ->write(" * @var array<string, Template>\n")
+            ->write(" */\n")
             ->write("private \$macros = [];\n\n")
```
fabpot:
```suggestion
            ->write("private array \$macros = [];\n\n")
```

src/Template.php
```diff
@@ -107,11 +107,17 @@ public function getParent(array $context)
...
+     */
     protected function doGetParent(array $context)
     {
         return false;
     }
 
+    /**
+     * @return bool
+     */
     public function isTraitable()
```
fabpot: As this class is internal, I think we can use a proper tye hint:

```suggestion
    public function isTraitable(): bool
```

fabpot (changes requested):

ruudk:
I applied the feedback. I added more return types, since the `Template` class is `@internal` anyway. 

What do you think?

src/Node/ModuleNode.php
```diff
@@ -115,7 +115,7 @@ protected function compileGetParent(Compiler $compiler)
         $parent = $this->getNode('parent');
 
         $compiler
-            ->write("protected function doGetParent(array \$context)\n", "{\n")
+            ->write("protected function doGetParent(array \$context): bool | string | Template | TemplateWrapper\n", "{\n")
```
fabpot: I think we don't add whitespace between the pipes (in Symfony I mean)

```suggestion
            ->write("protected function doGetParent(array \$context): bool|string|Template|TemplateWrapper\n", "{\n")
```

src/Template.php
```diff
@@ -107,12 +103,12 @@ public function getParent(array $context)
         return $this->parents[$parent];
     }
 
-    protected function doGetParent(array $context)
+    protected function doGetParent(array $context) : bool | string | self | TemplateWrapper
```
fabpot:
```suggestion
    protected function doGetParent(array $context): bool|string|self|TemplateWrapper
```

src/Template.php
```diff
@@ -107,12 +103,12 @@ public function getParent(array $context)
...
     {
         return false;
     }
 
-    public function isTraitable()
+    public function isTraitable() : bool
```
fabpot:
```suggestion
    public function isTraitable(): bool
```

src/Template.php
```diff
@@ -52,24 +52,20 @@ public function __construct(Environment $env)
...
+    abstract public function getTemplateName() : string;
 
     /**
      * Returns debug information about the template.
      *
-     * @return array Debug information
+     * @return array<int, int> Debug information
      */
-    abstract public function getDebugInfo();
+    abstract public function getDebugInfo() : array;
```
fabpot:
```suggestion
    abstract public function getDebugInfo(): array;
```

src/Template.php
```diff
@@ -52,24 +52,20 @@ public function __construct(Environment $env)
 
     /**
      * Returns the template name.
-     *
-     * @return string The template name
      */
-    abstract public function getTemplateName();
+    abstract public function getTemplateName() : string;
```
fabpot:
```suggestion
    abstract public function getTemplateName(): string;
```

fabpot (changes requested):

src/Template.php
ruudk: How come this is not automatically enforced by PHP CS Fixer or fabbot? I feel your valuable time can be used for other things 😁

ruudk:
Applied the changes.

fabpot:
Thank you @ruudk.

PR #4252, opened 2024-08-29
Unify error messages when a filter expects a mapping/sequence

fabpot:
Besides unifying the error messages, I've also removed the mention of `\Traversable` as this does not help and is an implementation detail (mapping and sequence are Twig terms, Traversable is not).

Or maybe we should introduce the notion of a "iterable"?

### DIFF
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -366,7 +366,7 @@ final class CoreExtension extends AbstractExtension
         }
 
         if (!$count = \count($values)) {
-            throw new RuntimeError('The "cycle" function does not work on empty sequences.');
+            throw new RuntimeError('The "cycle" function expects a non-empty sequence.');
         }
 
         return $values[$position % $count];
@@ -436,7 +436,7 @@ final class CoreExtension extends AbstractExtension
         $values = self::toArray($values);
 
         if (0 === \count($values)) {
-            throw new RuntimeError('The "random" function cannot pick from an empty sequence/mapping.');
+            throw new RuntimeError('The "random" function cannot pick from an empty sequence or mapping.');
         }
 
         return $values[array_rand($values, 1)];
@@ -573,7 +573,7 @@ final class CoreExtension extends AbstractExtension
     public static function replace($str, $from): string
     {
         if (!is_iterable($from)) {
-            throw new RuntimeError(\sprintf('The "replace" filter expects a sequence/mapping or "Traversable" as replace values, got "%s".', get_debug_type($from)));
+            throw new RuntimeError(\sprintf('The "replace" filter expects a sequence or a mapping, got "%s".', get_debug_type($from)));
         }
 
         return strtr($str ?? '', self::toArray($from));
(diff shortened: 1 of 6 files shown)


src/Extension/CoreExtension.php
```diff
@@ -346,7 +346,7 @@ public static function cycle($values, $position): string
         }
 
         if (!\count($values)) {
-            throw new RuntimeError('The "cycle" function does not work on empty sequences/mappings.');
+            throw new RuntimeError('The "cycle" function expects a non-empty sequence or mapping.');
```
stof: it actually only expects a sequence in this case

src/Extension/CoreExtension.php
```diff
@@ -1763,10 +1763,12 @@ public static function getAttribute(Environment $env, Source $source, $object, $
      */
     public static function column($array, $name, $index = null): array
     {
+        if (!is_iterable($array)) {
+            throw new RuntimeError(\sprintf('The "column" filter expects a sequence/mapping or "Traversable", got "%s".', \is_object($array) ? \get_class($array) : \gettype($array)));
```
```suggestion
            throw new RuntimeError(\sprintf('The "column" filter expects a sequence/mapping or "Traversable", got "%s".', get_debug_type($array)));
```

src/Extension/CoreExtension.php
```diff
@@ -1796,6 +1798,10 @@ public static function filter(Environment $env, $array, $arrow)
      */
     public static function find(Environment $env, $array, $arrow)
     {
+        if (!is_iterable($array)) {
+            throw new RuntimeError(\sprintf('The "find" filter expects a sequence or a mapping, got "%s".', \gettype($array)));
```
```suggestion
            throw new RuntimeError(\sprintf('The "find" filter expects a sequence or a mapping, got "%s".', \get_debug_type($array)));
```

stof: same below

stof (commented):
I would suggest using the concept of "iterable" in those error messages

fabpot:
> I would suggest using the concept of "iterable" in those error messages

I've just started to add some docs about Twig types. @drjayvee @ruudk

src/Extension/CoreExtension.php
```diff
@@ -553,7 +553,7 @@ public function convertDate($date = null, $timezone = null)
     public static function replace($str, $from): string
     {
         if (!is_iterable($from)) {
-            throw new RuntimeError(\sprintf('The "replace" filter expects a sequence/mapping or "Traversable" as replace values, got "%s".', \is_object($from) ? \get_class($from) : \gettype($from)));
+            throw new RuntimeError(\sprintf('The "replace" filter expects a sequence or a mapping, got "%s".', \is_object($from) ? \get_class($from) : \gettype($from)));
```
derrabus: While we're at it, shall we also pick a consistent way of outputting the type of the invalid parameter?

```suggestion
            throw new RuntimeError(\sprintf('The "replace" filter expects a sequence or a mapping, got "%s".', get_debug_type($from)));
```

doc/templates.rst
```diff
@@ -76,8 +76,18 @@ Twig templates have access to variables provided by the PHP application and
 variables created in templates via the :doc:`set <tags/set>` tag. These
 variables can be manipulated and displayed in the template.
 
-Use a dot (``.``) to access attributes of a variable (methods or properties of a
-PHP object, or items of a PHP array):
+Twig tries to abstract PHP types as much as possible and works with a few basic
+types, supported by ``filters``, ``functions``, and ``tests`` among others:
+
+* ``string``: A sequence of characters;
+* ``number``: An integer or a float;
```
stof: should we distinguish integer and floats (or maybe integers and numbers, with integers being a subtype) ? We have a bunch of places expecting an integer (`cycle` for instance)

fabpot: I would just cast to an integer when needed

PR #4253, opened 2024-08-29
Document how to iterate over a string

fabpot:


doc/tags/for.rst
```diff
@@ -1,8 +1,8 @@
 ``for``
 =======
 
-Loop over each item in a sequence or a mapping. For example, to display a list
-of users provided in a variable called ``users``:
+Loop over each item in a sequence, a mapping, or a string. For example, to
```
stof: This needs more info about what iterator over a string means. Is it iterating bytes, unicode codepoints, unicode grapheme clusters, words ?
Maybe this is best handled by a filter turning a string in an iterable, allowing to have several filters with different behaviors (and a name making the behavior clear)

fabpot: I've changed the PR to document how to use the `split` filter to achieve the same.

stof (approved):

PR #4255, opened 2024-08-29
Add more types to Template class and compiled templates

ruudk:
Similar to https://github.com/twigphp/Twig/pull/4250.

### DIFF
--- a/src/Template.php
+++ b/src/Template.php
@@ -74,7 +74,7 @@ abstract class Template
      *
      * @return self|TemplateWrapper|false The parent template or false if there is no parent
      */
-    public function getParent(array $context)
+    public function getParent(array $context): self|TemplateWrapper|false
     {
         if (null !== $this->parent) {
             return $this->parent;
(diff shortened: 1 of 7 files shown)


ruudk:
@fabpot @stof This is ready for review. Whenever you have time 😊

src/Template.php
```diff
@@ -494,6 +490,7 @@ public function yieldParentBlock($name, array $context, array $blocks = [])
      *
      * @param array $context An array of parameters to pass to the template
      * @param array $blocks  An array of blocks to pass to the template
+     * @return iterable<string>
```
fabpot, line 494: missing blank line above

fabpot (changes requested):

src/Template.php
ruudk, line 494:
```suggestion

     * @return iterable<string>
```

ruudk: Done

fabpot:
Thank you @ruudk.

PR #4257, opened 2024-09-01
Only unset `_iterated` when there is an `else` node

ruudk:
Same as #4249

Even though PHP does not complain, PHPStan does.
> Cannot unset offset '_iterated' on array{ ... the context values ... }

Since this is compiled code, we can easily produce a bit more valid code in the eyes of PHPStan.


fabpot:
Thank you @ruudk.

PR #4259, opened 2024-09-02
Replace `return; yield` with `yield from []`

ruudk:
This has the same effect, but looks less hacky and makes PHPStan happy.
https://phpstan.org/r/df7fcc88-1df8-428e-b675-5dc6965c34d6

Previously this was needed to work properly with output capturing.

/cc @fabpot @nicolas-grekas 

https://3v4l.org/KXste

### DIFF
--- a/src/Node/CaptureNode.php
+++ b/src/Node/CaptureNode.php
@@ -39,7 +39,7 @@ class CaptureNode extends Node
             ->raw("(function () use (&\$context, \$macros, \$blocks) {\n")
             ->indent()
             ->subcompile($this->getNode('body'))
-            ->write("return; yield '';\n")
+            ->write("yield from [];\n")
             ->outdent()
             ->write('})()')
         ;
--- a/src/Node/BlockNode.php
+++ b/src/Node/BlockNode.php
@@ -42,7 +42,7 @@ class BlockNode extends Node
 
         $compiler
             ->subcompile($this->getNode('body'))
-            ->write("return; yield '';\n") // needed when body doesn't yield anything
+            ->write("yield from [];\n")
             ->outdent()
             ->write("}\n\n")
         ;
--- a/src/Node/ModuleNode.php
+++ b/src/Node/ModuleNode.php
@@ -353,7 +353,7 @@ final class ModuleNode extends Node
         $compiler->subcompile($this->getNode('display_end'));
 
         if (!$this->hasNode('parent')) {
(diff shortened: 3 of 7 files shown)


derrabus:
I think, the goal is to have a generator function which is not executed until we iterate over the result. This does not happen if you remove all `yield` instructions and return an empty array. But we could do `yield from [];` instead.

ruudk:
> This does not happen if you remove a `yield` instructions and return an empty array

Are you sure? https://3v4l.org/GrX71

ruudk:
> we could do `yield from [];` instead

I didn't think of this. I like it as well, as this will make sure the return type is always a Generator.

ruudk:
Currently, PHPStan does not complain about this 👏 
https://phpstan.org/r/df7fcc88-1df8-428e-b675-5dc6965c34d6

ruudk:
I rebased the PR and targeted v3 and applied the `yield from []` trick suggested by @derrabus.

src/Node/CaptureNode.php
```diff
@@ -36,10 +36,10 @@ public function compile(Compiler $compiler): void
         }
         $compiler
             ->raw($useYield ? "implode('', iterator_to_array(" : '\\Twig\\Extension\\CoreExtension::captureOutput(')
-            ->raw("(function () use (&\$context, \$macros, \$blocks) {\n")
+            ->raw("(function () use (&\$context, \$macros, \$blocks): iterable {\n")
```
nicolas-grekas: this is not needed, generated code don't need extra runtime checks, let's remove them all from this PR

nicolas-grekas (commented):
looks nicer indeed

derrabus:
> > This does not happen if you remove a `yield` instructions and return an empty array
> 
> Are you sure? https://3v4l.org/GrX71

This reproducer misses the point because the function call and the start of the iteration are on the same line.

see https://3v4l.org/78LhS

nicolas-grekas (approved):

src/Node/CaptureNode.php
ruudk: Done!

ruudk:
@nicolas-grekas You approved and I requested review at roughly the same time 😅

derrabus (approved):

xabbuh (approved):

stof (approved):

fabpot:
Thank you @ruudk.

PR #4260, opened 2024-09-02
don't read current key and value when end of iterator is reached

xabbuh:
see failures in symfony/symfony#58145


stof (approved):

fabpot:
Is it possible to add a test?

xabbuh:
@fabpot test added

fabpot:
Thank you @xabbuh.

PR #4263, opened 2024-09-02
Fix testExtensionsAreNotInitializedWhenRenderingACompiledTemplate

derrabus:
I stumbled over this test while checking for PHPUnit deprecations. In this case, we need to do something about the `setMethods()` call which has been removed in PHPUnit 11.

However, we add an expectation for a method `initExtensions()` which has been removed a long time ago. So, if that method was actually called, we'd get an error anyway. Expecting this method not to be called feels a little redundant. ~~This basically means, we can run this test against an unmocked `Environment`. But if we do so, the test does not really test what it's supposed to test which is why I decided to remove the whole test.~~

Update: Instead, I'm now registering an extension that throws if we try to initialize it.

### DIFF
--- a/tests/EnvironmentTest.php
+++ b/tests/EnvironmentTest.php
@@ -178,19 +178,26 @@ class EnvironmentTest extends TestCase
 
         // force compilation
         $twig = new Environment($loader = new ArrayLoader(['index' => '{{ foo }}']), $options);
+        $twig->addExtension($extension = new class extends AbstractExtension {
+            public bool $throw = false;
+
+            public function getFilters(): array
+            {
+                if ($this->throw) {
+                    throw new \RuntimeException('Extension are not supposed to be initialized.');
+                }
+
+                return parent::getFilters();
+            }
+        });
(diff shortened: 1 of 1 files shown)


stof:
you could check that `getFunctions` or `getFilters` is not called on a registered extension as a way to detect that the initialization is not performed (which also does not require mocking the Environment AFAIK).

derrabus:
I tried your suggestion:

```php
$twig = new Environment($loader, $options);
$extension = $this->createMock(ExtensionInterface::class);
$extension->expects($this->never())->method($this->anything());
$twig->addExtension($extension);
```

With this, the test fails. My assumption is that the test guesses the cache key wrong.

derrabus:
Adding the extension changes the cache key. 🤦🏻‍♂️  I'll try to work around that.

derrabus:
I think, I've fixed the test. Thank you for your help, @stof.

fabpot:
Thank you @derrabus.

PR #4264, opened 2024-09-02
Fix MacroTest

derrabus:
Because of a typo in a variable name, `MacroTest` was never executed. This PR fixes the typo and adjusts the expected compiled PHP code, assuming that the current behavior is correct.

### DIFF
--- a/tests/Node/MacroTest.php
+++ b/tests/Node/MacroTest.php
@@ -36,8 +36,6 @@ class MacroTest extends NodeTestCase
 
     public function getTests()
     {
-        $tests = [];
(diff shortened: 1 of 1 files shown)


xabbuh:
The failure looks related.

stof:
The failure is because the compiled code is different depending on the state of `use_yield` (and we have CI jobs running for both modes)

derrabus:
Indeed. I'm now testing both variants and set that parameter explicitly.

fabpot:
Thank you @derrabus.

PR #4265, opened 2024-09-02
Migrate NodeTestCase to static data providers

derrabus:
PHPUnit 11 requires data providers to be static. This PR prepares our abstract `NodeTestCase` by deprecating the non-static `getTests()` and adding a static `provideTests()` as a replacement. I've also added PHPUnit attributes which newer PHPUnit releases prefer over PHPDoc annotations.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,10 @@
 # 3.13.0 (2024-XX-XX)
 
  * Add the `types` tag (experimental)
+ * Deprecate the `Twig\Test\NodeTestCase::getTests()` data provider, override `provideTests()` instead.
+ * Mark `Twig\Test\NodeTestCase::getEnvironment()` as final, override `createEnvironment()` instead.
+ * Deprecate `Twig\Test\NodeTestCase::getVariableGetter()`, call `createVariableGetter()` instead.
+ * Deprecate `Twig\Test\NodeTestCase::getAttributeGetter()`, call `createAttributeGetter()` instead.
 
 # 3.12.0 (2024-08-29)
 
--- a/src/Test/NodeTestCase.php
+++ b/src/Test/NodeTestCase.php
@@ -11,6 +11,8 @@
 
 namespace Twig\Test;
 
+use PHPUnit\Framework\Attributes\BeforeClass;
+use PHPUnit\Framework\Attributes\DataProvider;
 use PHPUnit\Framework\TestCase;
 use Twig\Compiler;
 use Twig\Environment;
@@ -24,11 +26,26 @@ abstract class NodeTestCase extends TestCase
      */
     private $currentEnv;
 
-    abstract public function getTests();
+    public function getTests()
+    {
+        return [];
+    }
+
+    /**
+     * @return iterable<array{0: Node, 1: string, 2?: Environment|null, 3?: bool}>
+     */
+    public static function provideTests(): iterable
+    {
+        trigger_deprecation('twig/twig', '3.13', 'Not implementing "%s()" in "%s" is deprecated. This method will be abstract in 4.0.', __METHOD__, static::class);
+
+        return [];
+    }
 
     /**
      * @dataProvider getTests
+     * @dataProvider provideTests
      */
+    #[DataProvider('getTests'), DataProvider('provideTests')]
     public function testCompile($node, $source, $environment = null, $isPattern = false)
     {
         $this->assertNodeCompilation($source, $node, $environment, $isPattern);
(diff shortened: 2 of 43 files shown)


src/Test/NodeTestCase.php
```diff
@@ -51,24 +65,53 @@ protected function getCompiler(?Environment $environment = null)
...
+     * @deprecated since Twig 3.13, use createVariableGetter() instead.
+     */
     protected function getVariableGetter($name, $line = false)
+    {
+        trigger_deprecation('twig/twig', '3.13', 'Method "%s()" is deprecated, use "createVariableGetter()" instead.', __METHOD__);
+
+        return self::createVariableGetter($name, $line);
+    }
+
+    final protected static function createVariableGetter($name, $line = false): string
```
stof: the new methods should have parameter types

src/Test/NodeTestCase.php
```diff
@@ -51,24 +65,53 @@ protected function getCompiler(?Environment $environment = null)
...
 
         return \sprintf('%s($context["%s"] ?? null)', $line, $name);
     }
 
+    /**
+     * @deprecated since Twig 3.13
+     */
     protected function getAttributeGetter()
     {
+        trigger_deprecation('twig/twig', '3.13', 'Method "%s()" is deprecated.', __METHOD__);
```
stof: the deprecation message (and the `@deprecated` tag) should mention the replacement

src/Test/NodeTestCase.php
```diff
@@ -24,11 +26,23 @@ abstract class NodeTestCase extends TestCase
      */
     private $currentEnv;
 
-    abstract public function getTests();
+    public function getTests()
+    {
+        return [];
+    }
+
+    public static function provideTests(): iterable
```
stof, line 37: I suggest adding a phpdoc `@return iterable<array{0: Node, 1: string, 2?: Environment|null, 3?: bool}>` to document the expected signature of the data provider (based on the test signature)

src/Test/NodeTestCase.php
derrabus: There is none. This method was only used by one subclass which does not really justify its presence on an abstract test case class.

stof: This is also used in other projects (like Symfony).

The whole point is to make it easier to write tests that support multiple Twig versions by making the NodeTestCase shipped in Twig itself responsible for knowing how a variable access gets compiled.
This was very important during the migration from Twig 2 to Twig 3 where the compilation output changed (Twig 3 to Twig 4 changes the compilation output for the strict variable getter, for which we don't have a similar method in the test case, but maybe we should add it to avoid making downstream projects add it, like in the TransNodeTest of Symfony)

stof: well, this one is about the `.` operator, not variable access, but the reason is the same (btw, without that method, the migration from `twig_get_attribute` to `CoreExtension::getAttribute` would have been painful for the ecosystem, as it would mean their tests would have to rely on the internal implementation details)

derrabus: All right, I'll add a static replacement.

src/Test/NodeTestCase.php
derrabus: Done, thanks.

CHANGELOG
```diff
@@ -1,6 +1,10 @@
 # 3.13.0 (2024-XX-XX)
 
  * Add the `types` tag (experimental)
+ * Deprecate the `Twig\Test\NodeTestCase::getTests()` data provider, override `provideTests()` instead.
+ * Mark `Twig\Test\NodeTestCase::getEnvironment()` as final, override `createEnvironment()` instead.
+ * Deprecate `Twig\Test\NodeTestCase::getVariableGetter()`, call `createVariableGetter()` instead.
+ * Deprecate `Twig\Test\NodeTestCase::getAttributeGetter()`, call `createAttributeGetter()` instead.
```
fabpot, line 7: Can you add the same information in `doc/deprecated.rst`?

src/Test/NodeTestCase.php
```diff
@@ -51,24 +68,58 @@ protected function getCompiler(?Environment $environment = null)
...
         return 'CoreExtension::getAttribute($this->env, $this->source, ';
     }
+
+    /** @beforeClass */
+    #[BeforeClass]
+    final public static function checkDataProvider(): void
+    {
+        $r = new \ReflectionMethod(static::class, 'getTests');
+        if ($r->getDeclaringClass()->getName() !== self::class) {
+            trigger_deprecation('twig/twig', '3.13', 'Overriding "%s::getTests()" in "%s" is deprecated, override "provideTests()" instead.', self::class, static::class);
```
fabpot: As those methods are abstract, should we say "implementing" instead?

```suggestion
            trigger_deprecation('twig/twig', '3.13', 'Implementing "%s::getTests()" in "%s" is deprecated, implement "provideTests()" instead.', self::class, static::class);
```

fabpot (commented):
LGTM

CHANGELOG
derrabus: Done.

src/Test/NodeTestCase.php
derrabus: Good idea. I've adjusted the wording.

fabpot:
Thank you @derrabus.

PR #4266, opened 2024-09-03
Prepare IntegrationTestCase for static data providers

derrabus:
Data providers need to be static in PHPUnit 11. Because of this, I'd like to declare the two methods we use as data providers in `IntegrationTestCase` as static in 4.0. This PR prepares that change:

* The non-static `getFixturesDir()` method is replaced with a static `getFixturesDirectory()`.
* Both methods `getTests()` and `getLegacyTests()` are marked as final, so we can declare them static in the next major.

This however means that we're delaying PHPUnit 11 compatibility of integration tests to Twig 4.0. If that's too late for us, we could deprecate the whole `IntegrationTestCase` in favor of a compatible replacement. That's a bigger change, but I would work on it if you think it's worth it.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -5,6 +5,8 @@
  * Mark `Twig\Test\NodeTestCase::getEnvironment()` as final, override `createEnvironment()` instead.
  * Deprecate `Twig\Test\NodeTestCase::getVariableGetter()`, call `createVariableGetter()` instead.
  * Deprecate `Twig\Test\NodeTestCase::getAttributeGetter()`, call `createAttributeGetter()` instead.
+ * Deprecate not overriding `Twig\Test\IntegrationTestCase::getFixturesDirectory()`, this method will be abstract in 4.0
+ * Marked `Twig\Test\IntegrationTestCase::getTests()` and `getLegacyTests()` as final
 
 # 3.12.0 (2024-08-29)
 
--- a/doc/deprecated.rst
+++ b/doc/deprecated.rst
@@ -216,3 +216,10 @@ Testing Utilities
 * The method ``Twig\Test\NodeTestCase::getEnvironment()`` is considered final
   as of Twig 3.13. If you want to override how the Twig environment is
   constructed, override ``createEnvironment()`` instead.
+
+* The method ``getFixturesDir()`` on ``Twig\Test\IntegrationTestCase`` is
+  deprecated, implement the new static method ``getFixturesDirectory()``
+  instead, which will be abstract in 4.0.
+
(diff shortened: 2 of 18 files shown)


derrabus:
The failures in the extra packages happen because we pull a version of Twig as dependency that does not include my changes. How do we deal with this kind of failure on this repo?

stof:
@derrabus accept it for now, as it will be resolved after the merge-up in the 4.x branch.

CHANGELOG
```diff
@@ -1,6 +1,8 @@
 # 3.13.0 (2024-XX-XX)
 
  * Add the `types` tag (experimental)
+ * Deprecate not overriding `Twig\Test\IntegrationTestCase::getFixturesDirectory()`, this method will be final in 4.0
```
stof:
```suggestion
 * Deprecate not overriding `Twig\Test\IntegrationTestCase::getFixturesDirectory()`, this method will be abstract in 4.0
```

stof (changes requested):

CHANGELOG
derrabus: Good catch, thank you.

fabpot:
Having PHPUnit 11 support only in 4.0 sounds good to me.

fabpot:
Thank you @derrabus.

williamdes:
The documentation example is missing an update https://github.com/twigphp/Twig/blob/4.x/doc/advanced.rst#testing-an-extension

doc/deprecated.rst
```diff
@@ -216,3 +216,10 @@ Testing Utilities
 * The method ``Twig\Test\NodeTestCase::getEnvironment()`` is considered final
   as of Twig 3.13. If you want to override how the Twig environment is
   constructed, override ``createEnvironment()`` instead.
+
+* The method ``getFixturesDir()`` on ``Twig\Test\IntegrationTestCase`` is
+  deprecated, implement the new static method ``getFixturesDirectory()``
+  instead, which will be abstract in 4.0.
+
+* The data providers ``getTests()`` and ``getLegacyTests()`` on
+  ``Twig\Test\IntegrationTestCase`` are considered final als of Twig 3.13.
```
williamdes, line 225: typo

GromNaN:
@williamdes Would you like to open a PR to improve documentation?

williamdes:
> @williamdes Would you like to open a PR to improve documentation?

🇫🇷 I will try soon :)

williamdes:
> @williamdes Would you like to open a PR to improve documentation?

Update: I am trying to cleanup my TODO lists. I will not be the one making the doc update :/

PR #4267, opened 2024-09-03
Fix Xdebug detection

derrabus:
The ini setting `xdebug.overload_var_dump` has been removed in Xdebug 3. The `var_dump()` method is now always overloaded if the Xdebug runs in develop mode. I've adjusted our detection logic accordingly.

Since our minimum PHP version is 8.0 now, we can stop caring about Xdebug 2 which has never received a PHP 8 compatible release.

### DIFF
--- a/src/Extension/DebugExtension.php
+++ b/src/Extension/DebugExtension.php
@@ -22,10 +22,8 @@ final class DebugExtension extends AbstractExtension
     {
         // dump is safe if var_dump is overridden by xdebug
         $isDumpOutputHtmlSafe = \extension_loaded('xdebug')
-            // false means that it was not set (and the default is on) or it explicitly enabled
-            && (false === \ini_get('xdebug.overload_var_dump') || \ini_get('xdebug.overload_var_dump'))
-            // false means that it was not set (and the default is on) or it explicitly enabled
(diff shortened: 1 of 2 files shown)


src/Extension/DebugExtension.php
```diff
@@ -22,10 +22,8 @@ public function getFunctions(): array
     {
         // dump is safe if var_dump is overridden by xdebug
         $isDumpOutputHtmlSafe = \extension_loaded('xdebug')
-            // false means that it was not set (and the default is on) or it explicitly enabled
-            && (false === \ini_get('xdebug.overload_var_dump') || \ini_get('xdebug.overload_var_dump'))
-            // false means that it was not set (and the default is on) or it explicitly enabled
-            // xdebug.overload_var_dump produces HTML only when html_errors is also enabled
+            // Xdebug overloads var_dump in develop mode when html_errors is enabled
+            && str_contains(\ini_get('xdebug.mode'), 'develop')
```
derrabus, line 26: Note: the safer check would be:

```suggestion
            && in_array('develop', xdebug_info('mode'), true)
```

But that does not work with Xdebug 3.0 which could technically be installed in environments with PHP 8.0.

fabpot:
Thank you @derrabus.

PR #4268, opened 2024-09-03
Remove obsolete "requires PHP" annotation

derrabus:


xabbuh (approved):

fabpot:
As the work has already been done in 4.x, I don't want us to do it again in 3.x now that we've bumped the min PHP version to 8.0.

derrabus:
This PR removes two annotations. One of them is still present on 4.x, while the other one is gone. So my change actually contains one line removal "done again" on 3.x. Well… 😓 

Shall I rebase the PR onto 4.x?

derrabus:
Rebased. 🙂

fabpot:
Thank you @derrabus.

PR #4262, opened 2024-09-02
Switch to expectUserDeprecationMessage()

derrabus:
This PR leverages symfony/symfony#54593. This will make it easier to switch to a vanilla PHPUnit 11 in the future.

PR #4269, opened 2024-09-03
PHPUnit 11

derrabus:
This PR switches our CI to PHPUnit 11.

TODO:

- [x] #4262
- [x] #4263
- [x] #4264
- [x] #4265
- [x] #4268
- [x] #4266
- [x] #4270
- [x] #4271
- [x] #4272
- [x] #4274

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,20 +35,15 @@ jobs:
                   coverage: "none"
                   php-version: ${{ matrix.php-version }}
                   ini-values: memory_limit=-1
+                  tools: phpunit:11.3
 
             - name: "Add PHPUnit matcher"
               run: echo "::add-matcher::${{ runner.tool_cache }}/phpunit.json"
 
             - run: composer install
 
-            - name: "Install PHPUnit"
-              run: vendor/bin/simple-phpunit install
-
-            - name: "PHPUnit version"
-              run: vendor/bin/simple-phpunit --version
-
             - name: "Run tests"
-              run: vendor/bin/simple-phpunit
+              run: phpunit
 
     extension-tests:
         needs:
@@ -86,18 +81,10 @@ jobs:
                   coverage: "none"
                   php-version: ${{ matrix.php-version }}
                   ini-values: memory_limit=-1
+                  tools: phpunit:11.3
 
             - name: "Add PHPUnit matcher"
               run: echo "::add-matcher::${{ runner.tool_cache }}/phpunit.json"
-
-            - name: "Composer install Twig"
-              run: composer install
-
-            - name: "Install PHPUnit"
-              run: vendor/bin/simple-phpunit install
-
-            - name: "PHPUnit version"
-              run: vendor/bin/simple-phpunit --version
 
             - name: "Prevent installing symfony/translation-contract 3.0"
               if: "matrix.extension == 'twig-extra-bundle'"
@@ -110,7 +97,9 @@ jobs:
 
             - name: "Run tests for ${{ matrix.extension }}"
               working-directory: extra/${{ matrix.extension }}
-              run: ../../vendor/bin/simple-phpunit
+              env:
+                  LANG: 'en_US.UTF-8'
+              run: phpunit
 
     integration-tests:
         if: "false" # TODO re-enable that job when Drupal does not use features deprecated in 3.x anymore
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 /doc/_build/vendor
 /doc/_build/output
 /composer.lock
+/phpunit.phar
 /phpunit.xml
 /vendor
-.phpunit.result.cache
+.phpunit.cache
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -1,6 +1,17 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" backupGlobals="false" backupStaticAttributes="false" colors="true" convertErrorsToExceptions="true" convertNoticesToExceptions="true" convertWarningsToExceptions="true" processIsolation="false" stopOnFailure="false" bootstrap="vendor/autoload.php" xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/9.6/phpunit.xsd">
-  <coverage/>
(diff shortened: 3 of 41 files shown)


.gitignore
```diff
@@ -1,6 +1,7 @@
 /doc/_build/vendor
 /doc/_build/output
 /composer.lock
+/phpunit.phar
```
derrabus, line 4: This will allow developer to place a PHAR distribution of PHPUnit in the main directory which is picked up and indexed by PHPStorm for code completion.

tests/ErrorTest.php
```diff
@@ -199,8 +198,15 @@ public function testTwigArrayMapThrowsRuntimeExceptions()
         $out = $template->render(['variable' => [1, 2, 3, 4]]);
         $this->assertEquals('36912', trim($out));
 
+        set_error_handler(function ($errno, $errstr, $errfile, $errline): never {
+            throw new \ErrorException($errstr, $errno, $errfile, $errline);
+        });
```
derrabus: It is my understanding that this test only ever worked because PHPUnit 9 registered an error handler that throws on PHP warnings. Not sure if that was actually intended.

.github/workflows/ci.yml
```diff
@@ -35,20 +35,18 @@ jobs:
                   coverage: "none"
                   php-version: ${{ matrix.php-version }}
                   ini-values: memory_limit=-1
+                  tools: phpunit
```
stof: I would suggest to use a dev requirement instead, which would have 2 benefits:

- the CI would use a known PHPUnit version instead of whatever `setup-php` decides to use (the current setup is essentially a [death star version constraint](https://thephp.cc/articles/the-death-star-version-constraint)
- it will make it easy for contributors to use the appropriate PHPUnit version locally

phpunit.xml.dist
```diff
@@ -1,17 +1,17 @@
...
   <php>
     <ini name="error_reporting" value="-1" />
-    <ini name="xdebug.mode" value="off" />
-    <ini name="xdebug.overload_var_dump" value="0" />
+    <ini name="xdebug.mode" value="debug" />
   </php>
-  <listeners>
-    <listener class="Symfony\Bridge\PhpUnit\SymfonyTestsListener" />
-  </listeners>
+  <source ignoreSuppressionOfDeprecations="true">
```
stof, line 24: shouldn't this be done also in all extra packages ?

phpunit.xml.dist
```diff
@@ -1,17 +1,17 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" backupGlobals="false" backupStaticAttributes="false" colors="true" convertErrorsToExceptions="true" convertNoticesToExceptions="true" convertWarningsToExceptions="true" processIsolation="false" stopOnFailure="false" bootstrap="vendor/autoload.php" xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/9.6/phpunit.xsd">
-  <coverage/>
+<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" backupGlobals="false" colors="true" processIsolation="false" stopOnFailure="false" bootstrap="vendor/autoload.php" xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/11.3/phpunit.xsd" cacheDirectory=".phpunit.cache" backupStaticProperties="false">
```
stof: this misses the `failOnWarning`, `failOnNotice` and so on, to reproduce the current configuration.

tests/ErrorTest.php
stof: To me, the right fix is to fix the implementation of `map` to check the type of variable as done in `reduce`: https://github.com/twigphp/Twig/blob/3.x/src/Extension/CoreExtension.php#L1834

.github/workflows/ci.yml
derrabus: We could tell setup-php exactly which PHPUnit version to use if we wanted. Also, setup-php tells us which version it installed, so there's no secret here.

My understanding is that we don't want to have dev tooling in our deps, which is e.g. also why we don't install CS Fixer or Psalm this way. If that does not apply to PHPUnit, I can add to to `require-dev`.

tests/ErrorTest.php
derrabus: We could do that. Shall I submit that as a bugfix for 3.x?

phpunit.xml.dist
derrabus: Done!

.github/workflows/ci.yml
stof: the issue is not that it is a secret. The issue is that it can upgrade to a new major version of PHPUnit without any work done in Twig (making the CI fail on an unrelated contribution in Twig after such upgrade of PHPUnit is released externally).
As explained in the blog post I linked, upgrading to a new major version must be a conscious choice (made by the Twig team, not by the setup-php maintainer).

stof: and Psalm is not installed through require-dev because Psalm is not used at all (if you thought about the Symfony repo, Psalm is used in a very special way there, so we explicitly don't want contributors to have it locally)

derrabus: > upgrading to a new major version must be a conscious choice

That makes total sense. I'm pinning PHPUnit to 11.3 now.

phpunit.xml.dist
derrabus: Done.

tests/ErrorTest.php
derrabus: see #4272

derrabus:
I don't understand the intl-extra failure. They look like they're failing because of data that could differ between ICU versions, but I'm not aware I have changed anything that could lead to different ICU data being loaded.

Also, I get a _different_ failure localy, but that one is at least consistent between 4.x and this branch. 😓

derrabus:
Apart from that, the test suite is green. 🎉

stof:
Could it be that the `LANG` env variable [set in the phpunit configuration](https://github.com/twigphp/Twig/blob/fb2663eb69f43251ea3399e9a72101f9bb94ae5f/extra/intl-extra/phpunit.xml.dist#L13) does not work anymore to define the default ext-intl locale ?

fabpot:
Thank you @derrabus.

PR #4270, opened 2024-09-03
Make data providers static

derrabus:
Data providers must be static in PHPUnit 11. This PR fixes all the low hanging fruit.


stof (approved):

fabpot:
Thank you @derrabus.

PR #4271, opened 2024-09-03
Rename AbstractTest to ProfilerTestCase

derrabus:
PHPUnit 11 doesn't allow abstract test classes to have the suffix "Test". This PR renames a class that violates that rule.


stof (approved):

fabpot:
Thank you @derrabus.

PR #4272, opened 2024-09-03
Validate the input of CoreExtension::map()

derrabus:
Fixes https://github.com/twigphp/Twig/pull/4269#discussion_r1742182828

As discovered in #4269, we have a test that checks if the map filter fails if we pass a non-iterable value. However, this test relies on an active error handler that throws on PHP warnings.

This PR adds input validation the the `map()` function like we can already find it in similar functions.


stof (approved):

fabpot:
Thank you @derrabus.

PR #4273, opened 2024-09-03
Fix tests that don't perform assertions

derrabus:
The method `doesNotPerformAssertions()` is a getter that tells us if the current test is expected to perform assertions. I think the method that the author _wanted_ to call is `expectNotToPerformAssertions()` which tells PHPUnit not to complain if the current test does not perform any assertions.


fabpot:
Thank you @derrabus.

PR #4274, opened 2024-09-03
Switch to static data provides for integration tests

derrabus:
This change should make our integration test case compatible with PHPUnit 10 and 11. Merging the PR will hopefully give us a green build on #4269. 🤞🏻


fabpot:
Thank you @derrabus.

PR #4275, opened 2024-09-04
Fix method visibility

derrabus:
The method's visibility was accidentally changed to private in 6f885353f3eba94decad73d31aeaf674046a45ed. Since it's a utility function for child classes, it should remain protected.


xabbuh (approved):

fabpot:
Thank you @derrabus.

PR #4276, opened 2024-09-04
Make FilesystemLoader::findTemplate public

ruudk:
For [TwigStan](https://github.com/twigstan/twigstan), I want to be able to find a template by the namespaced path. 

Currently, the only way to do this is to use getSourceContext() but this also loads the template into memory.

/cc @fabpot

### DIFF
--- a/src/Loader/FilesystemLoader.php
+++ b/src/Loader/FilesystemLoader.php
@@ -166,7 +166,7 @@ class FilesystemLoader implements LoaderInterface
     /**
      * @return string|null
      */
-    protected function findTemplate(string $name, bool $throw = true)
+    public function findTemplate(string $name, bool $throw = true)
     {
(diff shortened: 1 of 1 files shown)


stof:
I'm -1 on that change. This would encourage writing code that works only for the FilesystemLoader instead of supporting any configured loader (note that in a Symfony 7.2 project, the twig loader registered in the environment is *not* a FilesystemLoader but a ChainLoader).

ruudk:
Any suggestions on how to resolve it to a file on disk then? This only applies to FilesystemLoader as Loader can load from anywhere.

stof:
why do you need to resolve to a file path ? this will conflict with support for other loaders

ruudk:
I came up with a different approach that does not require me to resolve this afterwards.

PR #4277, opened 2024-09-04
Improve exception expectations reliability

alexandre-daubois:
This PR moves exception expectations the closest possible to the line intended to throw. It improves reliability of tests ensuring that the right code is throwing the expected exception and not something else (resulting in false-positive). This has been done many times across Symfony code base.


stof (approved):

yceruto (approved):

fabpot:
Thank you @alexandre-daubois.

PR #4279, opened 2024-09-05
Fix iterable return type

ruudk:
Currently, the PHPDoc states that a string should be returned, but that's not correct.

It can be anything that is echo-able.

That means any scalar, Stringable and even null.

/cc @stof @fabpot


fabpot:
Thank you @ruudk.

PR #4280, opened 2024-09-06
Fix isset in ForLoopNode

ruudk:
Even though the code works perfectly fine, PHPStan doesn't understand it.

> Offset 'revindex0' does not exist on array{parent: array{index0: int<1, max>, index: int<2, max>, first: false, revindex0?: int, revindex?: int, length: int<0, max>, last?: bool}.
> Offset 'revindex' does not exist on array{parent: array{index0: int<1, max>, index: int<2, max>, first: false, revindex0: int, revindex?: int, length: int<0, max>, last?: bool}.

Since we want to work with `revindex` and `revindex0`, we can better check for their existence, instead of `length`.

/cc @stof @fabpot


stof (approved):

xabbuh (approved):

fabpot:
Thank you @ruudk.

PR #4281, opened 2024-09-06
Fix wrong format of `Environment::VERSION_ID` constant

emodric:
It seems that format of `Environment::VERSION_ID` constant was (erroneously?) changed with Twig 3.10.0.

### DIFF
--- a/tests/EnvironmentTest.php
+++ b/tests/EnvironmentTest.php
@@ -50,7 +50,7 @@ class EnvironmentTest extends TestCase
         $this->assertEquals(Environment::MINOR_VERSION, $exploded[1]);
         $this->assertEquals(Environment::RELEASE_VERSION, $exploded[2]);
 
-        $this->assertEquals(Environment::VERSION_ID, \sprintf('%s0%s0%s', $exploded[0], $exploded[1], $exploded[2]));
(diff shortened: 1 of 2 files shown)


stof:
The test encoding the wrong assertion needs to be updated too.

emodric:
Whoops, sorry!

emodric:
@stof How about now?

tests/EnvironmentTest.php
```diff
@@ -50,7 +50,7 @@ public function testVersionConstants()
         $this->assertEquals(Environment::MINOR_VERSION, $exploded[1]);
         $this->assertEquals(Environment::RELEASE_VERSION, $exploded[2]);
 
-        $this->assertEquals(Environment::VERSION_ID, \sprintf('%s0%s0%s', $exploded[0], $exploded[1], $exploded[2]));
+        $this->assertEquals(Environment::VERSION_ID, \sprintf('%d%02d%02d', $exploded[0], $exploded[1], $exploded[2]));
```
stof: I suggest rewriting the test so that it does not depend on the number of digits in each number:

```php
$this->assertEquals(Environment::VERSION_ID, Environment::MAJOR_VERSION * 10000 + Environment::MINOR_VERSION * 100 + Environment::PATCH_VERSION);
```

This will work without having to change the test when reaching version `10` and without relying on generating a string to validate an int (relying on the fact that the 3 constants are already validated by the test)

emodric:
@stof Done!

stof (approved):

fabpot:
Thank you @emodric.

PR #4283, opened 2024-09-07
Deprecate Environment::mergeGlobals()

fabpot:


stof (approved):

doc/deprecated.rst
```diff
@@ -223,3 +223,17 @@ Testing Utilities
...
+* The ``Twig\Environment::mergeGlobals()`` method is deprecated as of Twig 3.13
+  and will be removed in Twig 4.0:
+
+  Before::
+
+      $context = $twig->mergeGlobals($context);
+
+  After::
+
+      $context += $twig->getGlobals();
```
derrabus, line 239: `getGlobals()` is flagged as `@internal`. Shall we remove that annotation?

ruudk:
@fabpot It is a bit unfortunate, that this was merged while Symfony TwigBridge has not yet been tagged.

Given this is included in the security release 3.14, we cannot upgrade because of deprecated usage in TwigBridge:
> Remaining indirect deprecation notices (109)
>  109x: Since twig/twig 3.14: The "Twig\Environment::mergeGlobals" method is deprecated.

Would it be an idea to get this tagged? https://github.com/symfony/symfony/pull/58207

I know one solution could be to turn off detecting of indirect deprecations, but it's super valuable to detect deprecations in vendor and help to get those fixed.

derrabus:
@ruudk There's no shame in temporarily installing a dev snapshot of the Twig Bridge if that solves a problem for you.

ruudk:
True. But I'm raising it, as there might be more people getting this "problem".

doc/deprecated.rst
davidkmenta: To anyone curious: it seems that `getGlobals` is `private` in v4, so the `@internal` flag makes sense.

derrabus: But then it doesn't make sense to document that method as a replacement for a deprecated public method.

@fabpot Can you please clarify how the deprecation should be resolved downstream?

xabbuh: Making `getGlobals()` `private` has been reverted.

davidkmenta: @xabbuh I’m sorry, you’re right! I didn’t notice the direct commit in v4. But then @derrabus question is still valid: shouldn’t be the flag removed?

stof: The `@internal` should indeed be removed in the 3.x branch to account for the revert of the private visibility in 4.x

fabpot: Done in 194cced47b

derrabus: Thanks!

PR #4286, opened 2024-09-07
Add the possibility to reset globals

### LINKED ISSUE #4007: Twig doesn't support alternative PHP servers like RoadRunner
Alternative PHP servers like RoadRunner don't unload PHP processes and process multiple requests per process allowing to get a significant performance increase. Unfortunately, Twig is not fully compatible with such servers because it relies on stateful classes that store some data. The intention is good - to cache data during the request. However, if the PHP process is not unloaded between HTTP requests, the internal state becomes an issue. On the second and further requests, Twig gets data from the first one which easily can be broken and make the request fail.

From my investigation, I understood that the issue is mostly in the global variables. They're cached in the `\Twig\Environment::$resolvedGlobals` property. Some extensions like EasyAdmin may add global variables conditionally. For example, if the first request doesn't meet EasyAdmin's criteria, the `ea` global variable is not loaded and not added to `$resolvedGlobals`. On the next request, Twig sees the non-null `$resolvedGlobals` and doesn't try to reload global variables. So any subsequent request that tries to render EasyAdmin's stuff will fail because the `ea` variable is not loaded.

Currently, to fix this issue, I'm using the following workaround in my project:

Define a custom Twig class:
```php
class Twig extends Environment implements EventSubscriberInterface
{
    public function onKernelTerminate(TerminateEvent $event): void
    {
        // Clear the resolvedGlobals cache to reinitialize global variables on the next RR request
        $this->resolvedGlobals = null;
    }

    #[\Override]
    public static function getSubscribedEvents(): array
    {
        return [
            KernelEvents::TERMINATE => 'onKernelTerminate',
        ];
    }
}
```

Add a compiler pass to replace the Twig object with the custom one:

```php
class ReplaceTwigCompilerPass implements CompilerPassInterface
{

    #[\Override]
    public function process(ContainerBuilder $container): void
    {
        $twigDefinition = $container->getDefinition('twig');
        $twigDefinition->setClass(Twig::class);
        $twigDefinition->addTag('kernel.event_subscriber');
    }
}
```

The custom class subscribes to Symfony's `kernel.terminate` event and cleans up globals cache. So on the next request, Twig properly reinitializes it with variables for the current request.

And the last point - to allow the Environment's `$resolvedGlobals` field modification, I apply the following patch with `cweagans/composer-patches`:
```diff
--- a/src/Environment.php
+++ b/src/Environment.php
@@ -57,7 +57,7 @@
     private $compiler;
     /** @var array<string, mixed> */
     private $globals = [];
-    private $resolvedGlobals;
+    protected $resolvedGlobals;
     private $loadedTemplates;
     private $strictVariables;
     private $templateClassPrefix = '__TwigTemplate_';
```

For my project, this works well. But I would be more than happy to see native support to not apply custom workarounds. The fix doesn't look complex so I believe it can be implemented quickly with no breaking changes.

fabpot:
Closes #4007

### DIFF
--- a/doc/advanced.rst
+++ b/doc/advanced.rst
@@ -104,8 +104,8 @@ What?      Implementation difficulty? How often? When?
 Globals
 -------
 
-A global variable is like any other template variable, except that it's
-available in all templates and macros::
+Global variables are available in all templates and macros. Use ``addGlobal()``
+to add a global variable to a Twig environment::
 
     $twig = new \Twig\Environment($loader);
     $twig->addGlobal('text', new Text());
@@ -679,6 +679,16 @@ method::
 
         // ...
     }
+
+.. caution::
+
+    Globals are fetched once from extensions and then cached for the lifetime
+    of the Twig environment. It means that globals should not be used to store
+    values that can change during the lifetime of the Twig environment. For
+    instance, if you're using an application server like RoadRunner or
+    FrakenPHP, you should not store values related to the current context (like
+    the HTTP request). If you do so, don't forget to reset the cache between
+    requests by calling ``Environment::resetGlobals()``.
 
 Functions
 ~~~~~~~~~
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
 # 3.13.1 (2024-XX-XX)
 
+ * Add the possibility to reset globals via `Environment::resetGlobals()`
  * Deprecate `Environment::mergeGlobals()`
 
 # 3.13.0 (2024-09-07)
--- a/src/Environment.php
+++ b/src/Environment.php
@@ -830,6 +830,12 @@ class Environment
         return array_merge($this->extensionSet->getGlobals(), $this->globals);
     }
 
+    public function resetGlobals(): void
+    {
+        $this->resolvedGlobals = null;
+        $this->extensionSet->resetGlobals();
+    }
+
     /**
      * @deprecated since Twig 3.13
      */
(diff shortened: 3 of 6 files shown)


doc/advanced.rst
```diff
@@ -680,6 +680,16 @@ method::
         // ...
     }
 
+.. caution::
+
+    Globals are fetched once from extensions and then cached for the lifetime
+    of the Twig environment. It means that globals should not be used to store
+    values that can change during the lifetime of the Twig environment. For
+    instance, if you're using an application server like RoadRunner of
```
javiereguiluz: `RoadRunner of FrakenPHP`

->

`RoadRunner or FrankenPHP`

doc/advanced.rst
```diff
@@ -680,6 +680,16 @@ method::
...
+    FrakenPHP, you should not store values related to the current context (like
+    the HTTP request). If you do so, don't forget to reset the cache between
+    requests by calling ``Environment::reset()``.
```
stof:
```suggestion
    requests by calling ``Environment::resetGlobals()``.
```

stof (changes requested):

doc/advanced.rst
fabpot: Fixed

doc/advanced.rst
fabpot: Fixed

JJCLane:
Unfortunately this doesn't seem to work for me when using Workerman, as the `$environment->globals` array doesn't get reset after calling `$environment->resetGlobals()`, if I update the method to include `$this->globals = [];` and call the function at the end of the request, it does seem to clear all global state. Do you have any thoughts?

stof:
@JJCLane please open a dedicated issue with the full details about reproducing the case that breaks.

PR #4289, opened 2024-09-09
Remove unused private methods

fabpot:


src/Node/ModuleNode.php
```diff
@@ -418,14 +418,6 @@ protected function compileIsTraitable(Compiler $compiler)
                     continue;
                 }
 
-                if ($node instanceof TextNode && ctype_space($node->getAttribute('data'))) {
```
fabpot, line 421: For a child template, these nodes are removed before getting here anyway.

src/Node/ModuleNode.php
```diff
@@ -418,14 +418,6 @@ protected function compileIsTraitable(Compiler $compiler)
...
-                    continue;
-                }
-
-                if ($node instanceof BlockReferenceNode) {
```
fabpot, line 425: For a child template, these nodes are removed before getting here anyway.

src/Node/ModuleNode.php
fabpot: In `Parser::filterBodyNodes()`

src/Node/ModuleNode.php
fabpot: In `Parser::filterBodyNodes()`

PR #4291, opened 2024-09-09
Improve the way one can deprecate a Twig callable

fabpot:


nicolas-grekas (approved):

src/DeprecatedCallableInfo.php
```diff
@@ -0,0 +1,62 @@
...
+                $message .= \sprintf(' from the "%s" package', $this->altPackage);
+            }
+            if ($this->altVersion) {
+                $message .= \sprintf(' (available since "%s")', $this->altVersion);
+            }
+            $message .= ' instead';
+        }
+
+        if ($file) {
+            $message .= \sprintf(' in %s at line %d.', $file, $line);
```
stof: What if `$line` is `null` when providing `$file` ?

src/DeprecatedCallableInfo.php
```diff
@@ -0,0 +1,62 @@
...
+{
+    private string $type;
+    private string $name;
+
+    public function __construct(
+        private string $package,
+        private string $version,
+        private ?string $altPackage = null,
+        private ?string $altName = null,
+        private ?string $altVersion = null,
```
stof, line 27: Shouldn't we put `altName` before `altPackage` to keep the package name and package version together ?

src/DeprecatedCallableInfo.php
fabpot: Fixed, test added

src/DeprecatedCallableInfo.php
fabpot: Good idea, done.

src/DeprecatedCallableInfo.php
```diff
@@ -0,0 +1,67 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Twig;
+
+/**
+ * @author Fabien Potencier <fabien@symfony.com>
+ */
+final class DeprecatedCallableInfo
```
Kocal, line 17: I'm not really sure about that, but is it really a callable? 
What about naming it `DeprecatedInfo`, which would match `deprecated_info` option?

Kocal (commented):
Thanks for working on this! 🫶

src/DeprecatedCallableInfo.php
stof: It is the info about a deprecated callable.

fabpot: Recently, I've created an abstraction on top of filters/functions/tests that I called a Twig callable :)
Filters/functions/tests are an abstraction on top of a PHP callable.

fabpot: See the AbstractTwigCallable class

doc/advanced.rst
```diff
@@ -277,29 +277,53 @@ filter: ``('a', 'b', 'foo')``.
 Deprecated Filters
 ~~~~~~~~~~~~~~~~~~
 
-You can mark a filter as being deprecated by setting the ``deprecated`` option
-to ``true``. You can also give an alternative filter that replaces the
-deprecated one when that makes sense::
+.. versionadded:: 3.14
+
+    The ``deprecated_info`` option was added in Twig 3.14.
```
stof: maybe `deprecation_info` would be a better name for that, as `deprecated_info` reads like the info itself is deprecated (while it is the info about a deprecation)

src/DeprecatedCallableInfo.php
Kocal: Oh okay, that makes sense then, thanks :)

doc/advanced.rst
fabpot: Makes sense, renamed.

src/AbstractTwigCallable.php
```diff
@@ -33,10 +33,35 @@ public function __construct(string $name, $callable = null, array $options = [])
...
             'alternative' => null,
         ], $options);
+
+        if ($this->options['deprecation_info'] && !$this->options['deprecation_info'] instanceof DeprecatedCallableInfo) {
+            throw new \LogicException(\sprintf('The "deprecation_info" option must be an instance of "%s".', DeprecatedCallableInfo::class));
+        }
+
+        if ($this->options['deprecated']) {
+            if ($this->options['deprecation_info']) {
+                throw new \LogicException('When setting the "deprecation_info" option, you need to remove the obsolete deprecated options.');
```
GromNaN, line 48: The option `deprecated` could accept an instance of `DeprecatedCallableInfo` so that you don't have to introduce a new option. The object being truthy, that is compatible with older versions.

fabpot: Not sure, I like the new name, it's more descriptive.

Issue #3926, opened 2023-12-03
Error "A block definition cannot be nested under non-capturing nodes" not emitted when there is no parent


ericmorand:
When rendering the following template, I get the error `A block definition cannot be nested under non-capturing nodes`:

```twig
{% extends "layout.twig" %}

{% if false %}
    {% block content %}FOO{% endblock %}
{% endif %}
```

(note that there is no need for the `layout.twig` template to exist for the error to be thrown)

But when rendering the following one, I don't get any error:

```twig
{% if false %}
    {% block content %}FOO{% endblock %}
{% endif %}
```

In _both_ cases, there is a block definition nested under a non-capturing node (`if`), so why is the first template failing and why is the second one not failing?

stof:
A `{% block %}` has 2 effects:

- it creates a block definition
- if the block is not top-level in a child template, it creates a block reference to render that block

That's why the second case is not forbidden: block references are perfectly fine inside non-capturing nodes.

This is related to the way the `extends` feature works: https://twig.symfony.com/doc/3.x/tags/extends.html
If there were top-level non-capturing nodes in a child template, we would not know where this output should go in the final output.

ericmorand:
In the first example, it is a block definition? That sounds counter intuitive to me. Isn't the block definition in the parent template - `layout.twig`?

I mean the first example is a template that inherits from another template and that _references_ a block that was _defined_ in the parent template.

In the second example however there is no parent so the block tag _defines_ a new block. It doesn't reference anything since there is no previous definition of the block.

I have the feeling that your explanation says exactly the opposite.

Anyway, my original question was more about user land: I feel that the error message doesn't convey anything that can help understand and fix the issue.

ericmorand:
@stof , I've been testing this capturing node rule and it is very confusing. Look at this template:

https://twigfiddle.com/j1gr3s

### child

```twig
{% extends "layout.twig" %}

{% set foo %}
    {% block content %}child{% endblock %}
{% endset %}
```

### parent

```twig
I am the

{% block content %}
parent
{% endblock %}
```

Even though the `content` block is captured in the child template - i.e. is not expected by the template designer to be displayed -, rendering the child template outputs this with Twig 3:

```text
I am the

child
```

And it is not different from...

### child

```twig
{% extends "layout.twig" %}

{% if false %}
    {% block content %}child{% endblock %}
{% endif %}
```

### parent

```twig
I am the

{% block content %}
parent
{% endblock %}
```

[...]

PR #4292, opened 2024-09-09
Introduce a CorrectnessNodeVisitor to validate that templates are semantically correct

### LINKED ISSUE #3698: Conditional {% use %} not working
**Version context:**

Twig version: v3.3.10
Twig Bridge version: 6.0.8
Symfony version: 6.0.8
PHP version: 8.1.5

**What's happening:**

I just tried to use a different twig trait depending on whether a form variable exists or not.

Below is a simplified version of my template structure. 

1. I'm building a "Create" form that can run inside a modal window driven by Turbo Frames, but the same page also works outside of the modal window, with different surrounding HTML markup, when loaded directly, not via the modal. 
2. My Symfony controller is aware of whether or not the request came from a modal window, and sets a `modal` variable to either true, or false. 

**base.html.twig**
```twig
<html>
<head> ... something shared by a lot of scripts - no markup, just metadata ...</head>
<body>
{%block shell %}{%endblock %}
</body>
</html>
```

**form-base.html.twig**
```twig
{% extends 'base.html.twig' %}

{% if modal is defined %}
    {% dump('FIRST') %}
    {% use 'standard-form.html.twig' %}
{% else %}
    {% dump('SECOND') %}
    {% use 'modal-form.html.twig' %}
{% endif %}
```

**standard-form.html.twig**
```twig
{% block shell %}
    {% block body %}
        {{ form(form) }}
    {% endblock body %}
{% endblock shell %}
```

**modal-form.html.twig**
```
{% block shell %}
<turbo-frame id="modal-content">
    {{ form_start(form) }}
    <div class="header">some header contents</div>
    {% block body %}
        By default we just render the form, but people can override the block body, if they need something more specialised
        {{ form_rest(form) }}
    {% endblock body %}
    <button type="submit">
    {{ form_end(form) }}
</turbo-frame>
    ... some additional modal markup ...
{% endblock shell %}
```

**create.html.twig**
```twig
{% extends 'form-base.html.twig' %}

{% block body %}
 ... some custom form body markup - not important. 
{% endblock %}
```

---

Now, let's not get into whether I'm doing it wrong or not - I know I could extend different base templates etc., my point here is that I tried to use the {% use %} trait and it behaved differently than expected:

Consider the base template:

```twig
{% if modal is defined %}
    {% dump('FIRST') %}
    {% use 'standard-form.html.twig' %}
{% else %}
    {% dump('SECOND') %}
    {% use 'modal-form.html.twig' %}
{% endif %}
```

The core of what's wrong: Even though my debug toolbar dumped FIRST and not the SECOND,  twig used the {% block shell %} present in `modal-form.html.twig`, while it should have used `standard-form.html.twig`.

Docs (https://twig.symfony.com/doc/3.x/tags/use.html) say that if multiple traits define the same block, the last one wins. When I commented the `{% use 'modal-form.html.twig' %}` out, the correct template showed up. 

**Conclusion in what I think is wrong:**

When Twig parses the templates, it ignores the if/else construct around the `use` tag.

Is that expected behaviour? I think it should be at least documented?

I also tried `{% use (form is defined ? '_stimulus/shell/modal-form.html.twig' : '_stimulus/shell/modal.html.twig') %}`, which would work for example with `{% include %}`, but I got an error `The template references in a "use" statement must be a string.`, suggesting the template path at that point was not resolved, further confirming my suspition.

fabpot:
This PR addresses several issues around the correctness of templates.
Being able to parse and compile a template does not mean that it is semantically correct. To enforce correctness, we currently have several places where we deal with it:

 * `Parser::filterBodyNodes()`: This method is a mix of ensuring the correctness of a template, but it also changes the body node of a child template (something that is always needed and not part of the correctness checks)
 * `ExtendsTokenParser`: It checks that an `extend` tag is not embedded into a block or a macro. The `extend` tag is not the only one that must be at the root of a template

This PR introduces a new `CorrectnessNodeVisitor` that has the responsibility to check that a template is semantically correct. It's the continuation of work that started a long time ago in #2687 (where I mentioned the weirdness of some supported templates like those mentioned in #3926 and deprecated by this PR).

Closes #3698: Having a `use` tag embedded in another tag (like `if` in the mentioned PR) is deprecated and will not be possible in 4.0.


src/NodeVisitor/CorrectnessNodeVisitor.php
```diff
@@ -0,0 +1,131 @@
...
+
+            // allows to identify when we enter/leave the block nodes
+            $this->blockNodes = new \SplObjectStorage();
+            foreach ($node->getNode('blocks') as $n) {
+                $this->blockNodes->attach($n);
+            }
+
+            $body = $node->getNode('body')->getNode('0');
+            // see Parser::subparse() which does not wrap the parsed Nodes if there is only one node
+            foreach (count($body) ? $body : new Node([$body]) as $k => $n) {
```
stof: what if there is a single parsed node with children ? `count($body)` would be truthy so it would iterate over the children of that node instead of checking the body

src/NodeVisitor/CorrectnessNodeVisitor.php
```diff
@@ -0,0 +1,131 @@
...
+
+        if ($this->hasParent && $node->getNodeTag() && !$node instanceof BlockReferenceNode) {
+            $this->currentTagNode = $node;
+        }
+
+        if ($node instanceof ConfigNode && !$this->rootNodes->contains($node)) {
+            throw new SyntaxError(sprintf('The "%s" tag must always be at the root of the body of a template.', $node->getNodeTag()), $node->getTemplateLine(), $node->getSourceContext());
+        }
+
+        if ($this->currentTagNode && $node instanceof BlockReferenceNode) {
```
stof: shouldn't this check `$this->currentBlockNodeLevel` ? Once we are inside a block, we can have block definitions anywhere as those are not top-level blocks anymore.

src/NodeVisitor/CorrectnessNodeVisitor.php
```diff
@@ -0,0 +1,131 @@
...
+            if ($this->currentTagNode instanceof NodeCaptureInterface || count($this->blockNodes) > 1) {
```
stof: what is the goal of this `count($this->blockNodes)` condition ? why would we have a behavior that depends on the number of blocks defined in the template ?

src/NodeVisitor/CorrectnessNodeVisitor.php
stof: maybe we need to check for `Node::class === \get_class($node)` instead, as done in the old `filterBodyNodes`

src/Parser.php
```diff
@@ -334,50 +332,16 @@ public function getCurrentToken(): Token
...
     {
-        // check that the body does not contain non-empty output nodes
-        if (
-            ($node instanceof TextNode && !ctype_space($node->getAttribute('data')))
-            || (!$node instanceof TextNode && !$node instanceof BlockReferenceNode && $node instanceof NodeOutputInterface)
-        ) {
-            if (str_contains((string) $node, \chr(0xEF).\chr(0xBB).\chr(0xBF))) {
-                $t = substr($node->getAttribute('data'), 3);
-                if ('' === $t || ctype_space($t)) {
-                    // bypass empty nodes starting with a BOM
```
stof, line 563: Is the filtering of the BOM gone ?

src/Parser.php
```diff
@@ -334,50 +332,16 @@ public function getCurrentToken(): Token
...
-            return null;
-        }
-
-        // here, $nested means "being at the root level of a child template"
-        // we need to discard the wrapping "Node" for the "body" node
-        $nested = $nested || Node::class !== \get_class($node);
-        foreach ($node as $k => $n) {
-            if (null !== $n && null === $this->filterBodyNodes($n, $nested)) {
-                $node->removeNode($k);
+        foreach ($body as $k => $node) {
```
stof, line 561: What happens when `$body` is not a `Nodes` but another `Node` because `subparse` did not wrap it ? AFAICT, this logic would fail to remove the node in such case.

src/NodeVisitor/CorrectnessNodeVisitor.php
stof: maybe we should introduce a `StatementListNode` (name to be bikeshed)  to represent such list of other nodes to allow identifying this case more easily and deprecate instantiating the `Node` class itself (making it abstract in Twig 4)

tests/Fixtures/tags/inheritance/conditional_block_nested.legacy.test
```diff
@@ -0,0 +1,40 @@
+--TEST--
+conditional "block" tag with "extends" tag (nested)
+--DEPRECATION--
+Since twig/twig 3.14: Having a "block" tag under a "if" tag (line 7) is deprecated in layout.twig at line 8.
```
stof: This is a tag inside a non-top-level condition, so I don't think we need to deprecate this. Such tag is useful when you want to reuse part of the rendering between the `if` and `else` clause by defining the block in the `if` and referencing it in the `else`.
Having to move it to a top-level block (that does not override a parent block) makes the intent a lot less clear as it is easy to miss that this block is related to the rendering inside the condition (and so that it can be removed if the condition body does not need it anymore).

For the same reason, it can make sense to define a block inside the body of a loop, which would probably also trigger such a deprecation.

stof: To me, once entering a top-level block of a child template, we should not have any restriction on blocks anymore, just like there is no such restriction in a standalone template.

src/NodeVisitor/CorrectnessNodeVisitor.php
```diff
@@ -0,0 +1,131 @@
...
+    public function leaveNode(Node $node, Environment $env): Node
+    {
+        if ($node instanceof ModuleNode) {
+            $this->rootNodes = null;
+            $this->hasParent = false;
+            $this->blockNodes = null;
+            $this->currentBlockNodeLevel = 0;
+        }
+        if ($this->hasParent && $node->getNodeTag() && !$node instanceof BlockReferenceNode) {
+            $this->currentTagNode = null;
```
stof: This looks wrong to me, as it will reset the current tag node to `null` rather than resetting it to its state before entering that node (try putting a block inside a nested `if` in a template followed by a block after the inner `if` but inside the outer one, and your current code would not report the second block as being incorrect)

stof: or even just some *other* nested node inside the `if` followed by a block after the `if`

doc/deprecated.rst
```diff
@@ -188,6 +188,29 @@ Templates
   ``Template::loadTemplate()``); pass instances of ``Twig\TemplateWrapper``
   instead.
 
+* Having a "block" definition nested in another node that captures the output
```
smnandre: Not answering on the "why", i have'nt tried this PR locally yet... but i feel the before/after examples do not represent the same usage/need.

For instance (please don't be too harsh on this code 🙏 ) this is a pattern i used from time to time, and i _feel_ this is often how the set/block was used. 

https://github.com/symfony/ux/blob/8a66e74d5e7070e5cbf60042cae535fae28d3e7a/ux.symfony.com/templates/components/Code/CodeLine.html.twig

I don't mind we cannot in 4.0 and will find a better way, but the "after" here would not help in this situation, right ?

stof:
@fabpot what is the state of this PR ? Do you plan to fix it or to abandon it ?

fabpot:
> @fabpot what is the state of this PR ? Do you plan to fix it or to abandon it ?

I plan to still work on this (soon).

src/Parser.php
fabpot: The BOM logic now lives in `TextNode::isBlank()`, used by both the visitor and `cleanupBodyForChildTemplates`.

doc/deprecated.rst
```diff
@@ -291,6 +291,34 @@ Templates
   in ``Environment::resolveTemplate()`` and ``Environment::load()``); pass
   instances of ``Twig\TemplateWrapper`` instead.
 
+* Having a "block" definition nested in another node that captures the output
```
stof: is this deprecated only in child templates ?

tests/Fixtures/tags/inheritance/conditional_block_nested.legacy.test
stof: @fabpot what about this case ?

fabpot:
I've just pushed a new version which hopefully addresses most of you comments.

stof:
I'm wondering whether we should really deprecate the definition of blocks inside a capturing node. My [argument about block definitions nested inside a block](https://github.com/twigphp/Twig/pull/4292#discussion_r1751659218) could easily apply to block definitions inside a capturing node.

In term of implementation, the CorrectnessNodeVisitor could maintain a `capturingNodeDepth` parameter similar to the `blockDepth` and use it in the guard clause of the method.

src/Parser.php
stof: @fabpot it looks like you haven't answered that comment.

fabpot: Additional commit to cover this.

stof: shouldn't we also return `$body` as is when it is not a `Nodes` (or `Node` for BC reasons) instead of iterating over it ?

PR #4298, opened 2024-09-11
Replace all empty() calls

fabpot:


src/Extension/CoreExtension.php
```diff
@@ -545,7 +545,7 @@ public function convertDate($date = null, $timezone = null)
         }
 
         $asString = (string) $date;
-        if (ctype_digit($asString) || (!empty($asString) && '-' === $asString[0] && ctype_digit(substr($asString, 1)))) {
+        if (ctype_digit($asString) || (isset($asString[0]) && '-' === $asString[0] && ctype_digit(substr($asString, 1)))) {
```
stof, line 548: I would even suggest `'' !== $asString` to make the code more readable.

fabpot: 1a20e961742de10bca43d4d02b7f92b005e3e2a2

PR #4300, opened 2024-09-11
Add support for argument unpacking

### LINKED ISSUE #3291: Is there a way to destructuring an array ?
Something like :
```twig
{{ '%s %s %s'|format(...params) }}
```

fabpot:
Closes #3291

### DIFF
--- a/src/ExpressionParser.php
+++ b/src/ExpressionParser.php
@@ -30,6 +30,7 @@ use Twig\Node\Expression\Unary\AbstractUnary;
 use Twig\Node\Expression\Unary\NegUnary;
 use Twig\Node\Expression\Unary\NotUnary;
 use Twig\Node\Expression\Unary\PosUnary;
+use Twig\Node\Expression\Unary\SpreadUnary;
 use Twig\Node\Node;
 
 /**
@@ -618,6 +619,7 @@ class ExpressionParser
         $stream = $this->parser->getStream();
 
         $stream->expect(Token::PUNCTUATION_TYPE, '(', 'A list of arguments must begin with an opening parenthesis');
+        $hasSpread = false;
         while (!$stream->test(Token::PUNCTUATION_TYPE, ')')) {
             if ($args) {
                 $stream->expect(Token::PUNCTUATION_TYPE, ',', 'Arguments must be separated by a comma');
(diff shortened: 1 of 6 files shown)


stof:
The parsing seems to support spreading any expression while the compilation has such support only for ArrayExpression. This seems inconsistent. Shouldn't we have a SpreadExpression node in the AST instead ?

stof:
Btw, now that we require PHP 8 (and so we can rely on the PHP spread operator in the compiled output), we could simplify the implementation of spreading in sequences/mapping by using the same SpreadExpression node using PHP's spread operator instead of the custom logic compiling the ArrayExpression to a `merge` call.

fabpot:
> The parsing seems to support spreading any expression while the compilation has such support only for ArrayExpression. This seems inconsistent. Shouldn't we have a SpreadExpression node in the AST instead ?

Done, I've added another test using a variable to prove it fixes that use case.

fabpot:
> Btw, now that we require PHP 8 (and so we can rely on the PHP spread operator in the compiled output), we could simplify the implementation of spreading in sequences/mapping by using the same SpreadExpression node using PHP's spread operator instead of the custom logic compiling the ArrayExpression to a `merge` call.

The code has been simplified in 4.x. Once this PR is merged, I will do another one on 4.x.

src/ExpressionParser.php
```diff
@@ -632,7 +634,14 @@ public function parseArguments($namedArguments = false, $definition = false, $al
                 $token = $stream->expect(Token::NAME_TYPE, null, 'An argument must be a name');
                 $value = new NameExpression($token->getValue(), $this->parser->getCurrentToken()->getLine());
             } else {
-                $value = $this->parseExpression(0, $allowArrow);
+                if ($stream->nextIf(Token::SPREAD_TYPE)) {
+                    $hasSpread = true;
+                    $value = new Node([new SpreadExpression(), $this->parseExpression(0, $allowArrow)]);
```
stof: Shouldn't this be `$value = new SpreadExpression($this->parseExpression(0, $allowArrow))`, ensuring that the value is an expression node instead of using a Node containing 2 expressions ? And this would also prevent invalid usages of SpreadExpression (if not followed by another node) as the SpreadExpression itself would enfoce being spreading an expression (similar to how unary operators are represented in the AST)

fabpot: Of course 😊

src/Node/Expression/SpreadExpression.php
```diff
@@ -0,0 +1,30 @@
...
+{
+    public function __construct(AbstractExpression $node, int $lineno)
+    {
+        parent::__construct(['node' => $node], [], $lineno);
+    }
+
+    public function compile(Compiler $compiler): void
+    {
+        $compiler
+            ->raw('...')
```
stof: shouldn't we enforce putting a space before the `...` to avoid issues in compiled code about where it is associated (as done in AbstractUnary) ?
Btw, maybe this could extend `AbstractUnary`

fabpot: Done

PR #4308, opened 2024-09-12
[Doc] Provide an alternative for the deprecated spaceless filter

javiereguiluz:
When a feature is deprecated, it's recommended to provide an alternative for it (or mention that there's no alternative).

So, let's mention the whitespace control features in the deprecation message of the `spaceless` filter.

### DIFF
--- a/doc/filters/spaceless.rst
+++ b/doc/filters/spaceless.rst
@@ -3,7 +3,8 @@
 
 .. warning::
 
-    The ``spaceless`` filter is deprecated as of Twig 3.12.
+    The ``spaceless`` filter is deprecated as of Twig 3.12. Instead, use the
+    :ref:`whitespace control features <templates-whitespace-control>`.
 
 Use the ``spaceless`` filter to remove whitespace *between HTML tags*, not
 whitespace within HTML tags or whitespace in plain text:
--- a/doc/filters/spaceless.rst
+++ b/doc/filters/spaceless.rst
@@ -3,8 +3,8 @@
 
 .. warning::
 
-    The ``spaceless`` filter is deprecated as of Twig 3.12. Instead, use the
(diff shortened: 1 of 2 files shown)


stof:
The mention should be expanded IMO to make it clear that whitespace control is not a drop-in replacement of `spaceless`. It is a *different* feature (that might satisfy the need of the project if their intent was to remove spaces coming from indentation in the Twig template)

javiereguiluz:
@stof in Symfony Docs, in similar situations, we don't do that because it's like duplicating some doc content. Also, the admonition messages look bad when they are too long.

I'd say it's better to just link to the new docs so visitors can quickly learn about it ... but if Twig Docs does differently, tell me and I'll update this. Thanks.

stof:
Well, the message as is can make people think that this is a direct replacement, while it is a totally different feature.

I'm not asking to duplicate the documentation of whitespace control in the admonition message. I'm only asking to make it explicit that whitespace control is not a direct replacement.

doc/filters/spaceless.rst
```diff
@@ -3,7 +3,8 @@
 
 .. warning::
 
-    The ``spaceless`` filter is deprecated as of Twig 3.12.
+    The ``spaceless`` filter is deprecated as of Twig 3.12. Instead, use the
+    :ref:`whitespace control features <templates-whitespace-control>`.
```
mbabker:
```suggestion
    The ``spaceless`` filter is deprecated as of Twig 3.12. If needed, use the
    :ref:`whitespace control features <templates-whitespace-control>`.
```

This is the best suggestion I can come up with which doesn't imply that those controls are a direct replacement without making the message too wordy.

fabpot:
Thank you @javiereguiluz.

PR #4310, opened 2024-09-12
Allow to use a dynamic attribute on the . operator via ()

### LINKED ISSUE #3598: attribute() not looking for camelCase method

fabpot:
Closes #3598

As one can think that the `attribute` function has some special power compared to the `.` operator (like using Symfony PropertyAccess), I think it's better to use the `.` operator for dynamic attributes as well.

Using `()` makes it clear that the behavior is the same as the `.` operator (using `()` is similar to the way one can use dynamic mapping names (like in `{ (name): value }`).


CHANGELOG
```diff
@@ -1,5 +1,6 @@
 # 3.15.0 (2024-XX-XX)
 
+ * Deprecate the `attribute` function; use the `.` notation and wrap the name with parenthesis instead
```
stof, line 3: Deprecating the `attribute` function would make it hard to write Twig templates supporting older Twig versions (we cannot do a `{% if %}` to switch between 2 cases, we have to switch between 2 templates instead as the deprecation is at compile time).

As the `attribute` function is free to maintain (the implementation uses the same PHP function than the `.` operator), I suggest delaying the deprecation until some future 4.x release (we have some maintained versions of Symfony that still support Twig 2.x and so could not upgrade).

fabpot: That works for me. Code updated accordingly.

fabpot: See 8d1af0fc03314a2786decd6457447a583f7dd146

doc/functions/attribute.rst
```diff
@@ -1,6 +1,13 @@
 ``attribute``
 =============
 
+.. warning::
+
+    The ``attribute`` filter is deprecated as of Twig 3.15. Use the ``.``
```
stof: should this be a link to the dot operator documentation ?

fabpot: Good idea, done

mishoo:
I just noticed this in the documentation, and I was wondering why the weird syntax? Instead of `{{ user.('get' ~ name) }}`, wouldn't this work: `user['get' ~ name]`?

Asking because I maintain a Twig→JS compiler, and I'm pondering whether to implement the new syntax.. it seems superfluous to me.

stof:
@mishoo both work to access an array index. But `[ ]` does not allow calling a method (in JS, calling a method is about reading a property containing a function and calling it, but that's absolutely not the case in PHP)

Issue #4113, opened 2024-06-28
EscaperRuntime::setSafeClasses() implementation


joeworkman:
My Twig extension implements a global that is an adapter for my CMS. 

```php
public function getGlobals(): array
{
    return [
            'cms' => $this->adapter,
    ];
}
```

I would like to set all functions for this class to be safe. I thought that I could use `EscaperRuntime::setSafeClasses()` for this. 

```php
$this->twig->getRuntime(EscaperRuntime::class)->setSafeClasses([
	TotalForm::class           => ['html'],
	TotalCMSTwigAdapter::class => ['html'],
]);
```

It's not working. Am I close?

fabpot:
Can you tell us more about what does not work? I've just tried and it works well. I realize that `setSafeClasses()` might be a confusing method name. This is **only** to avoid escaping the return value for `__toString()` methods.

joeworkman:
Let me explain what I am trying to accomplish. The TotalForm class is essentially a form builder for my CMS. Its still a work in progress, but the twig currently looks like this... 

```twig
{% set form  = cms.objectFormBuilder('blog', {}) %}

{{ form.addField('text', 'mytext', {}) }}
{{ form.build() }}
```

The problem that I was trying to overcome was the need to add `| raw` to the build statement since it outputs HTML. I thought setting the TotalForm class (the `form` variable is an instance of TotalForm) to be "safe" would fix that.

stof:
What you are displaying is not the `form` variable but the output of `form.build()`

joeworkman:
Yes. That is correct. For now, I have globally disabled the need to use raw. Not ideal but I have to admit that it's nice since so much of my implementation displays HTML from my CMS.

stof:
Make your build method return a stringable object from a class configured as safe class (which could be the `\Twig\Markup` object).

When you return a string, you have no way to mark it as safe.

An alternative method is to create a Twig function that would do this build (as Twig functions can tell Twig whether they output a safe value or no), making it look like `{{ render_form(form) }}`

stof:
I'm closing this issue as it was based on a misunderstanding of what safe classes are about.
The phpdoc type of the method has been improved to make it more likely for such mistake to be detected.

PR #4315, opened 2024-09-13
Add more precise types for the registration of safe classes

stof:
Safe classes only impact the auto-escaping of the string casting of instance of those classes.
By defining that this method only expects names of stringable classes, it allows static analysis tools to report non-working configurations (for instance for people thinking this would impact the auto-escaping for values returned by getters of a safe class, as in https://github.com/twigphp/Twig/issues/4113 for instance).

The method technically support prefixing the class name with a `\`. Static analyzers would reject such string when checking for `class-string`. However, I think this is fine. Projects wanting to use a SA tool are probably already using the `::class` constant anyway to refer to class names (and otherwise, removing the leading `\` in their string literal is easy)

### DIFF
--- a/src/Runtime/EscaperRuntime.php
+++ b/src/Runtime/EscaperRuntime.php
@@ -52,6 +52,9 @@ final class EscaperRuntime implements RuntimeExtensionInterface
         return $this->escapers;
     }
 
+    /**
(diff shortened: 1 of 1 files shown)


fabpot:
Thank you @stof.

PR #4316, opened 2024-09-13
Set IteratorAggregate generics for Node

ruudk:
This will help PHPStan and Psalm to understand what happens when iterating over a Node.

/cc @fabpot


stof (approved):

fabpot:
Thank you @ruudk.

PR #4317, opened 2024-09-13
Add support for PHPUnit 11

MauricioFauth:
- Related to https://github.com/twigphp/Twig/pull/4313

The test cases `Twig\Test\NodeTestCase` and `Twig\Test\IntegrationTestCase` are currently incompatible with PHPUnit 11.

This pull request adds an option to the user to add support to run the tests with PHPUnit 11, if they implement the `Twig\Test\IntegrationTestCase::getFixturesDirectory()` method. Currently the only solution to the user is to fork `IntegrationTestCase` and `NodeTestCase` test cases, and run them only for PHPUnit 11.

The difference from the previous pull request is that this time, if `getFixturesDirectory()` is not implemented, it does not run the tests and trigger a deprecation, instead of causing an error.

### DIFF
--- a/src/Test/IntegrationTestCase.php
+++ b/src/Test/IntegrationTestCase.php
@@ -108,12 +108,24 @@ abstract class IntegrationTestCase extends TestCase
 
     final public static function provideTests(): iterable
     {
-        return self::assembleTests(static::getFixturesDirectory(), false);
+        try {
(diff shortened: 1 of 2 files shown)


fabpot:
I'm closing here as the work to support PHPUnit 11 has been done in Twig 4.0.

PR #4318, opened 2024-09-14
Allow Twig callable arguments to use camel or snake names

### LINKED ISSUE #3475: Named Arguments of Filters and Functions are not passed as-is and forced to snake_case

fabpot:
Closes #3475

Twig callables (functions/filters/tests) now accept snake and camel argument names independently of the name of the underlying PHP callable parameter names.

### DIFF
--- a/src/Util/CallableArgumentsExtractor.php
+++ b/src/Util/CallableArgumentsExtractor.php
@@ -40,22 +40,25 @@ final class CallableArgumentsExtractor
     public function extractArguments(Node $arguments): array
     {
         $extractedArguments = [];
+        $extractedArgumentNameMap = [];
         $named = false;
         foreach ($arguments as $name => $node) {
             if (!\is_int($name)) {
                 $named = true;
-                $name = $this->normalizeName($name);
             } elseif ($named) {
                 throw new SyntaxError(\sprintf('Positional arguments cannot be used after named arguments for %s "%s".', $this->twigCallable->getType(), $this->twigCallable->getName()), $this->node->getTemplateLine(), $this->node->getSourceContext());
             }
 
-            $extractedArguments[$name] = $node;
+            $extractedArguments[$normalizedName = $this->normalizeName($name)] = $node;
+            $extractedArgumentNameMap[$normalizedName] = $name;
         }
 
         if (!$named && !$this->twigCallable->isVariadic()) {
             $min = $this->twigCallable->getMinimalNumberOfRequiredArguments();
             if (\count($extractedArguments) < $this->rc->getReflector()->getNumberOfRequiredParameters() - $min) {
-                throw new SyntaxError(\sprintf('Value for argument "%s" is required for %s "%s".', $this->rc->getReflector()->getParameters()[$min + \count($extractedArguments)]->getName(), $this->twigCallable->getType(), $this->twigCallable->getName()), $this->node->getTemplateLine(), $this->node->getSourceContext());
+                $argName = $this->toSnakeCase($this->rc->getReflector()->getParameters()[$min + \count($extractedArguments)]->getName());
+
+                throw new SyntaxError(\sprintf('Value for argument "%s" is required for %s "%s".', $argName, $this->twigCallable->getType(), $this->twigCallable->getName()), $this->node->getTemplateLine(), $this->node->getSourceContext());
             }
 
             return $extractedArguments;
@@ -76,7 +79,7 @@ final class CallableArgumentsExtractor
         $optionalArguments = [];
         $pos = 0;
         foreach ($callableParameters as $callableParameter) {
-            $callableParameterName = $this->normalizeName($callableParameter->name);
+            $callableParameterName = $callableParameter->name;
             if (\PHP_VERSION_ID >= 80000 && 'range' === $callable) {
                 if ('start' === $callableParameterName) {
                     $callableParameterName = 'low';
@@ -86,8 +89,9 @@ final class CallableArgumentsExtractor
             }
 
             $callableParameterNames[] = $callableParameterName;
+            $normalizedCallableParameterName = $this->normalizeName($callableParameterName);
 
-            if (\array_key_exists($callableParameterName, $extractedArguments)) {
+            if (\array_key_exists($normalizedCallableParameterName, $extractedArguments)) {
                 if (\array_key_exists($pos, $extractedArguments)) {
                     throw new SyntaxError(\sprintf('Argument "%s" is defined twice for %s "%s".', $callableParameterName, $this->twigCallable->getType(), $this->twigCallable->getName()), $this->node->getTemplateLine(), $this->node->getSourceContext());
                 }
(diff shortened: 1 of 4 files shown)


smnandre:
I love how this implementation _dances_ around most of the differences between the  previous `normalize($str)` and the String  implementation.

I encountered many problems when the `snake` implementation changed in String...  so i tried to double-check many cases here :) 

> [!NOTE]
> i only looked at the code of the changed class, i am **not 100% sure** this concerns _real_ use cases (especially after the recent naming convention updates)
> 

## Results

A  ❌  in the `Same` column indicate a change of behaviour **before / after** this PR (but only some of them are "regressions" .... and again not sure about..

| Node Attribute | Callable Parameter | Same | norm(attr)   | norm(call)   | Valid Before? | node attr    | res = param  | camel(param) | snake(camel(param)) | Valid After? | 
| - | - | - | - | - | - | - | - | - | - | - |
 no_svg         |  noSVG          |  ✅  |  no_svg         |  no_svg         |  ✅  |  no_svg         |  noSVG          |  noSVG          |  no_svg         |  ✅  | 
 noSVG          |  no_svg         |  ❌  |  no_svg         |  no_svg         |  ✅  |  noSVG          |  no_svg         |  noSvg          |  no_svg         |  ❌  | 
 error_404      |  error404       |  ✅  |  error_404      |  error404       |  ❌  |  error_404      |  error404       |  error404       |  error404       |  ❌  | 
 error404       |  error_404      |  ❌  |  error404       |  error_404      |  ❌  |  error404       |  error_404      |  error404       |  error404       |  ✅  | 
 errCode_404    |  err_code_404   |  ❌  |  err_code_404   |  err_code_404   |  ✅  |  errCode_404    |  err_code_404   |  errCode404     |  err_code404    |  ❌  | 
 errCode404     |  err_code_404   |  ❌  |  err_code404    |  err_code_404   |  ❌  |  errCode404     |  err_code_404   |  errCode404     |  err_code404    |  ✅  | 
 err_code_404   |  errCode404     |  ✅  |  err_code_404   |  err_code404    |  ❌  |  err_code_404   |  errCode404     |  errCode404     |  err_code404    |  ❌  | 
 err_code_404   |  errCode_404    |  ❌  |  err_code_404   |  err_code_404   |  ✅  |  err_code_404   |  errCode_404    |  errCode404     |  err_code404    |  ❌  | 
 aBc            |  a_b_c          |  ✅  |  a_bc           |  a_b_c          |  ❌  |  aBc            |  a_b_c          |  aBC            |  a_bc           |  ❌  | 
 a_b_c          |  aBc            |  ✅  |  a_b_c          |  a_bc           |  ❌  |  a_b_c          |  aBc            |  aBc            |  a_bc           |  ❌  | 
 a_b_c          |  a_b_c          |  ✅  |  a_b_c          |  a_b_c          |  ✅  |  a_b_c          |  a_b_c          |  aBC            |  a_bc           |  ✅  | 
 aBC            |  a_b_c          |  ❌  |  a_bc           |  a_b_c          |  ❌  |  aBC            |  a_b_c          |  aBC            |  a_bc           |  ✅  | 
 a_b_c          |  aBC            |  ✅  |  a_b_c          |  a_bc           |  ❌  |  a_b_c          |  aBC            |  aBC            |  a_bc           |  ❌  | 


## Reproducer

I took the code from the current [normalize](https://github.com/twigphp/Twig/blob/3.x/src/Util/CallableArgumentsExtractor.php) method, the two new toSnake and toCamel from [your changes](https://github.com/twigphp/Twig/blob/3149ab346f29fa67f959e38d7628152dd33babfe/src/Util/CallableArgumentsExtractor.php) and the [snake and camel](https://github.com/symfony/symfony/blob/7.2/src/Symfony/Component/String/AbstractUnicodeString.php#L155) methods from Symfony

You can find the small script i made to test on [this gist](https://gist.github.com/smnandre/09cb10e5ce4e95fe45ce87ab7f67b14c).

fabpot:
Your comment @smnandre made me realize that there is nothing forcing people to use proper snake or camel cases.
So, I've taken a simpler and more robust approach. I now compare the argument name and the callable parameter name after removing the `_`s and converting to lowercase. I've added all your tests to confirm that they are all behaving correctly now.

There are of course some edge cases where several parameters might resolve to the same name, but having such similar names in a signature would not be very smart :)

smnandre:
I feared I was gonna make you lose time here .. so i'm really glad if in the end the implementation is more robust and more "tolerant" at the same time. 

🎩

src/Util/CallableArgumentsExtractor.php
```diff
@@ -163,7 +166,17 @@ public function extractArguments(Node $arguments): array
...
+    }
+
+    // logic from Symfony\Component\String\AbstractUnicodeString::snake()
+    private function toSnakeCase(string $str): string
+    {
+        $camel = str_replace(' ', '', preg_replace_callback('/\b.(?!\p{Lu})/u', static function ($m) use (&$i) {
+            return 1 === ++$i ? ('İ' === $m[0] ? 'i̇' : mb_strtolower($m[0], 'UTF-8')) : mb_convert_case($m[0], \MB_CASE_TITLE, 'UTF-8');
+        }, preg_replace('/[^\pL0-9]++/u', ' ', $str)));
+
+        return mb_strtolower(preg_replace(['/(\p{Lu}+)(\p{Lu}\p{Ll})/u', '/([\p{Ll}0-9])(\p{Lu})/u'], '\1_\2', $camel), 'UTF-8');
```
nicolas-grekas: since normalizeName isn't unicode-aware, I suggest simplifying here:
```suggestion
    private function toSnakeCase(string $str): string
    {
        return strtolower(preg_replace(['/([A-Z]+)([A-Z][a-z])/', '/([a-z0-9])([A-Z])/'], '\1_\2', $str));
```

stof:
what happens for named arguments passed to variadic functions ? AFAIU, the issue you referenced was about that case, not about the case where Twig is resolving named arguments for its callables.

fabpot:
> what happens for named arguments passed to variadic functions ? AFAIU, the issue you referenced was about that case, not about the case where Twig is resolving named arguments for its callables.

Variadic is just one problem, which I "fixed" in the last commit. For variadics, I keep the attribute name as is.

stof:
Not normalizing variadic arguments to snake case anymore will fix #3475 but would be a BC break for functions/templates that currently rely on this behavior (because the function implementation expects snake case but the template does not provide it).

fabpot:
> Not normalizing variadic arguments to snake case anymore will fix #3475 but would be a BC break for functions/templates that currently rely on this behavior (because the function implementation expects snake case but the template does not provide it).

I know :)

fabpot:
> > Not normalizing variadic arguments to snake case anymore will fix #3475 but would be a BC break for functions/templates that currently rely on this behavior (because the function implementation expects snake case but the template does not provide it).
> 
> I know :)

I should elaborate a bit more here. Even if it is a BC break, is it a problem in practice? The current behavior is to always convert to snake_case. So, for any app that runs today, developers must have used snake case names as well for their variadic names. And I bet developers also changed their templates to use snake_case as well to be consistent. Is it a fair assumption?

fabpot:
Now, we trigger a deprecation when a name for a variadic arg is not snake-cased. That way, we're sure that the upgrade is smooth.

src/Util/CallableArgumentsExtractor.php
nicolas-grekas: if we want to keep unicode casing, we should likely do it only when the charset is utf-?8?

tacman:
Since twig 4 will allow BC breaks, perhaps this issue can be fixed for variadic functions then.

PR #4321, opened 2024-09-16
use static data providers in the IntegrationTestCase class

### LINKED ISSUE #3810: IntegrationTestCase: Compatibility with PHPUnit 10

xabbuh:
fixes #3810

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,7 @@
 # 3.15.0 (2024-XX-XX)
 
+ * Deprecate the `Twig\Test\IntegrationTestCase::getTests()` data provider, override `provideTests()` instead.
+ * Deprecate the `Twig\Test\IntegrationTestCase::getLegacyTests()` data provider, override `provideLegacyTests()` instead.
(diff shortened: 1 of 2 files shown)


stof:
This does not make any sense as IntegrationTestCase does not expect users of the test case to implement the data provider.

Support for PHPUnit 11+ (and PHPUnit 10 without warning) will be provided by Twig 4.0 (it is already available in the 4.x branch)

xabbuh:
ah indeed, let's close here

PR #4325, opened 2024-09-19
Use a more precise return type in AbstractTwigCallable

stof:
This method is not expected to change the type of callable.

As using it is the only way to configure arguments for a Twig callable (there is no constructor argument for them), it avoids getting static analysis errors in `Extension::getFunctions()` when using it.


xabbuh (approved):

fabpot:
Thank you @stof.

PR #4328, opened 2024-09-20
Improve formatDate methods phpdoc

VincentLanglet:
`convertDate`
- Supports null https://github.com/twigphp/Twig/blob/3.x/src/Extension/CoreExtension.php#L539
- Supports timestamp passed as integer https://github.com/twigphp/Twig/blob/3.x/src/Extension/CoreExtension.php#L549

So the phpdoc should not be
`\DateTimeInterface|string|null` but `\DateTimeInterface|string|int|null`

And all the date methods using `convertDate` should also accepts int and null.

This is important since some tools are developing static analysis on twig.


fabpot:
Thank you @VincentLanglet.

PR #4330, opened 2024-09-21
Add compile-time checks for the "matches" operator

fabpot:


src/Node/Expression/Binary/MatchesBinary.php
```diff
@@ -12,9 +12,27 @@
...
+use Twig\Node\Node;
+use Twig\Node\Expression\ConstantExpression;
 
 class MatchesBinary extends AbstractBinary
 {
+    public function __construct(Node $left, Node $right, int $lineno)
+    {
+        if ($right instanceof ConstantExpression) {
+            $regexp = $right->getAttribute('value');
+            set_error_handler(static fn ($t, $m) => throw new SyntaxError(\sprintf('Regexp "%s" passed to "matches" is not valid: %s.', $regexp, substr($m, 14)), $lineno));
```
Kocal, line 25: The number `14` feels like a bit magical to me, can we add a comment or use `strlen('...')` (which will be inlined to `14`) instead?

jdreesen: This also somehow doesn't look right:
> Twig\Error\RuntimeError: Regexp "3" passed to "matches" is not valid: Delimiter must not be alphanumeric%sbackslash%sin "index.twig" at line 2

It seems like there's at least a space missing before `in "index.twig" at line 2`.

fabpot: The message is correct, here, this is only the format to cover differences between PHP versions.

PR #4331, opened 2024-09-23
Add return types in FileystemLoader

ruudk:


stof:
If you improve the types in that class, I suggest also replacing `@param string|array $paths` with `@param string|string[] $paths` everywhere (otherwise, we cannot actually guarantee that the returned list is a list of strings)

ruudk:
You're right, done!

src/Loader/FilesystemLoader.php
```diff
@@ -58,14 +63,16 @@ public function getPaths(string $namespace = self::MAIN_NAMESPACE): array
...
+     * @return list<string>
      */
     public function getNamespaces(): array
     {
         return array_keys($this->paths);
     }
 
     /**
-     * @param string|array $paths A path or an array of paths where to look for templates
+     * @param string|list<string> $paths A path or an array of paths where to look for templates
```
stof: The argument actually accepts `string[]`. It does not care whether the array is a list (as it only iterates over it ignoring the keys). Forcing it to be a list will only make it more annoying for projects using a SA tool supporting the `list` type (i.e. phpstan with bleeding-edge mode enabled or Psalm)

ruudk: Fine, changed to `string[]` everywhere.

stof: My comment applied only to arguments.

You should not have changed it for return types where we actually always return a list (and where having a more precise type will be *less* annoying for projects using SA tools)

ruudk: But then I have to do array_values on the input side, oké ?

stof: The input is never assigned directly. As said in my previous comment, it is only used in a `foreach` loop ignoring the keys.

ruudk: You're right, I commented without seeing the actual code. Fixed it now. Thanks.

stof (approved):

fabpot:
Thank you @ruudk.

PR #4333, opened 2024-09-24
Optimize compiled code for "set" tag

fabpot:


src/Node/SetNode.php
```diff
@@ -78,11 +78,19 @@ public function compile(Compiler $compiler): void
...
                 if ($this->getAttribute('safe')) {
-                    $compiler
-                        ->raw("('' === \$tmp = ")
-                        ->subcompile($this->getNode('values'))
-                        ->raw(") ? '' : new Markup(\$tmp, \$this->env->getCharset())")
-                    ;
+                    if ($this->getNode('values') instanceof ConstantExpression) {
+                        $compiler
+                            ->raw('new Markup(')
+                            ->subcompile($this->getNode('values'))
```
stof: isn't this missing the case of a ConstantExpression being the empty string ?

tests/Node/SetTest.php
```diff
@@ -77,7 +77,7 @@ public static function provideTests(): iterable
         $node = new SetNode(true, $names, $values, 1);
         $tests[] = [$node, <<<EOF
 // line 1
-\$context["foo"] = ('' === \$tmp = "foo") ? '' : new Markup(\$tmp, \$this->env->getCharset());
+\$context["foo"] = new Markup("foo", \$this->env->getCharset());
```
stof, line 80: shouldn't we also have a test covering the case of non-constant expressions ?

fabpot: We already have such tests above, right?

src/Node/SetNode.php
fabpot: Fixed an covered with a test now.

src/Node/SetNode.php
```diff
@@ -33,9 +33,15 @@ public function __construct(bool $capture, Node $names, Node $values, int $linen
...
             $safe = true;
-            if ($values instanceof TextNode) {
+            if (Node::class === get_class($values) && !count($values)) {
+                $values = new ConstantExpression('', $values->getTemplateLine());
+                $capture = false;
+            } elseif ($values instanceof TextNode) {
                 $values = new ConstantExpression($values->getAttribute('data'), $values->getTemplateLine());
                 $capture = false;
+            } elseif ($values instanceof PrintNode && $values->getNode('expr') instanceof ConstantExpression) {
+                $values = new ConstantExpression($values->getNode('expr')->getAttribute('value'), $values->getTemplateLine());
```
stof: why creating a new ConstantExpression here instead of using `$values->getNode('expr')` directly ?

fabpot: You're right, updated.

src/Node/SetNode.php
```diff
@@ -33,9 +33,15 @@ public function __construct(bool $capture, Node $names, Node $values, int $linen
         $safe = false;
         if ($capture) {
             $safe = true;
-            if ($values instanceof TextNode) {
+            if (Node::class === get_class($values) && !count($values)) {
```
stof, line 36: See this makes me think we should really investigate having a dedicated node class for representing a statement list in 4.x (as suggested in https://github.com/twigphp/Twig/pull/4292#discussion_r1751616965), as we have this special case of `Node::class === get_class($values)` in more and more places

stof (approved):

PR #4335, opened 2024-09-24
Deprecate instantiating Node directly, introduce EmptyNode and Nodes

fabpot:
Based on some comments from @stof:

See https://github.com/twigphp/Twig/pull/4292#discussion_r1751651242
See https://github.com/twigphp/Twig/pull/4333#discussion_r1773531733

First interesting usage here: 65ee72a71f60c3680c2aceef7b9932cee7f84e89

### DIFF
--- a/src/Node/Nodes.php
+++ b/src/Node/Nodes.php
@@ -1,0 +1,28 @@
+<?php
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Twig\Node;
+
+use Twig\Attribute\YieldReady;
+
+/**
+ * Represents a list of nodes.
+ *
+ * @author Fabien Potencier <fabien@symfony.com>
+ */
+#[YieldReady]
+final class Nodes extends Node
+{
+    public function __construct(array $nodes = [], int $lineno = 0)
+    {
+        parent::__construct($nodes, [], $lineno);
+    }
+}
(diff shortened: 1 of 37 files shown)


src/Node/SetNode.php
```diff
@@ -33,7 +33,7 @@ public function __construct(bool $capture, Node $names, Node $values, int $linen
         $safe = false;
         if ($capture) {
             $safe = true;
-            if (Node::class === get_class($values) && !count($values)) {
+            if ($values instanceof Nodes && !count($values)) {
```
stof: this should probably keep a `|| Node::class === get_class($values)` for BC reasons

src/Parser.php
```diff
@@ -371,7 +373,7 @@ private function filterBodyNodes(Node $node, bool $nested = false): ?Node
 
         // here, $nested means "being at the root level of a child template"
         // we need to discard the wrapping "Node" for the "body" node
-        $nested = $nested || Node::class !== \get_class($node);
+        $nested = $nested || Nodes::class !== \get_class($node);
```
stof:
```suggestion
        $nested = $nested || $node instance Nodes || Node::class !== \get_class($node);
```

src/Node/Nodes.php
```diff
@@ -0,0 +1,28 @@
...
+
+use Twig\Attribute\YieldReady;
+
+/**
+ * Represents a list of nodes.
+ *
+ * @author Fabien Potencier <fabien@symfony.com>
+ */
+#[YieldReady]
+class Nodes extends Node
```
stof: I think this class should be final, as custom nodes should not extend it (otherwise, we would have the same issue than with the usage of `Node` today)

src/Node/SetNode.php
fabpot: Indeed, added

src/Parser.php
fabpot: Done

src/Node/Nodes.php
fabpot: You're right, done here and for `EmptyNode` as well.

PR #4336, opened 2024-09-24
Fix template name in error for an unknown dynamic extends called from an include

### LINKED ISSUE #3265: Wrong error location when failing to load a dynamic parent
When an included template has a dynamic parent and the computed parent name is invalid, the error gets reported with the location of the `include` tag rather than the location of the broken `extends` tag.

See these 2 fiddles to compare the reported location:

- an invalid static parent reports the error properly: https://twigfiddle.com/hmgy4z
- an invalid dynamic parent reports the error in the wrong location (`main.twig:3`): https://twigfiddle.com/tu86dd

The loading of the parent template in the compiled code already provides the appropriate location: https://github.com/twigphp/Twig/blob/19f98c7298359f62a65ccf670b123e8836582394/src/Node/ModuleNode.php#L118-L125

But `getParent` catches the exception and resets that info and runs the guessing (which then finds the wrong template apparently): https://github.com/twigphp/Twig/blob/19f98c7298359f62a65ccf670b123e8836582394/src/Template.php#L101-L105

This affects Twig 1.x, 2.x and 3.x in similar ways.

The reason things work fine for static parents is that the compile template bypasses `getParent()` in this case, and so bypasses the catch block.

fabpot:
Closes #3265

The code was introduced in #1529, but removing it nowadays does not break any existing test.

Review ignoring whitespace.

### DIFF
--- a/src/Template.php
+++ b/src/Template.php
@@ -80,23 +80,16 @@ abstract class Template
             return $this->parent;
         }
 
-        try {
-            if (!$parent = $this->doGetParent($context)) {
-                return false;
-            }
+        if (!$parent = $this->doGetParent($context)) {
+            return false;
+        }
 
-            if ($parent instanceof self || $parent instanceof TemplateWrapper) {
-                return $this->parents[$parent->getSourceContext()->getName()] = $parent;
-            }
+        if ($parent instanceof self || $parent instanceof TemplateWrapper) {
+            return $this->parents[$parent->getSourceContext()->getName()] = $parent;
+        }
 
-            if (!isset($this->parents[$parent])) {
-                $this->parents[$parent] = $this->loadTemplate($parent);
-            }
-        } catch (LoaderError $e) {
-            $e->setSourceContext(null);
-            $e->guess();
-
-            throw $e;
+        if (!isset($this->parents[$parent])) {
+            $this->parents[$parent] = $this->loadTemplate($parent);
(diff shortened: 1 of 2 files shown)


src/Template.php
```diff
@@ -80,23 +80,16 @@ public function getParent(array $context): self|TemplateWrapper|false
...
-            }
+        if (!$parent = $this->doGetParent($context)) {
+            return false;
+        }
 
-            if (!isset($this->parents[$parent])) {
-                $this->parents[$parent] = $this->loadTemplate($parent);
-            }
-        } catch (LoaderError $e) {
-            $e->setSourceContext(null);
```
fabpot, line 96: This is the "fix".

src/Template.php
```diff
@@ -80,23 +80,16 @@ public function getParent(array $context): self|TemplateWrapper|false
...
-                return $this->parents[$parent->getSourceContext()->getName()] = $parent;
-            }
+        if (!$parent = $this->doGetParent($context)) {
+            return false;
+        }
 
-            if (!isset($this->parents[$parent])) {
-                $this->parents[$parent] = $this->loadTemplate($parent);
-            }
-        } catch (LoaderError $e) {
```
fabpot, line 95: This try/catch is not needed as it's already taken care of in `Template::loadTemplate()`.

PR #4338, opened 2024-09-25
Add template cache hot reload

### LINKED ISSUE #3880: Simplify refreshing a template’s cache
https://twig.symfony.com/doc/3.x/recipes.html#using-a-database-to-store-templates shows how to load templates from a database, but following this recipe will make templates’ source update inoperant by default when debug mode is disabled.

You can enable `auto_reload` to work around the issue, but this will cause every loader to check for cache freshness every time a template is rendered.

I think it would be more efficient to refresh a template’s cache after its source is updated, but the corresponding code is somewhat complex and needs the `@internal` `Environment::getTemplateClass` method:

```php
$twigCache = $twig->getCache(false);

$twigCache->write(
    $twigCache->generateKey($templateName, $twig->getTemplateClass($templateName)),
    $twig->compileSource($twig->getLoader()->getSourceContext($templateName))
);
```

Would it make sense for `Environment` to expose something like a `refreshCache(string $templateName)` method?

fabpot:
Closes #3880

This PR achieves 2 things in the minimum amount of code (in 2 commits):

* It adds a new `RemovableCacheInterface` that allows cache to be removed.
* It adds a memory cache to allow updating existing templates in the same process (think FrakenPHP for instance) without polluting the permanent cache.

### DIFF
--- a/src/Cache/FilesystemCache.php
+++ b/src/Cache/FilesystemCache.php
@@ -16,7 +16,7 @@ namespace Twig\Cache;
  *
  * @author Andrew Tch <andrew@noop.lv>
  */
-class FilesystemCache implements CacheInterface
+class FilesystemCache implements CacheInterface, RemovableCacheInterface
 {
     public const FORCE_BYTECODE_INVALIDATION = 1;
 
@@ -74,6 +74,14 @@ class FilesystemCache implements CacheInterface
         }
 
         throw new \RuntimeException(\sprintf('Failed to write cache file "%s".', $key));
+    }
+
+    public function remove(string $name, string $cls): void
+    {
+        $key = $this->generateKey($name, $cls);
+        if (!@unlink($key) && file_exists($key)) {
+            throw new \RuntimeException(\sprintf('Failed to delete cache file "%s".', $key));
+        }
     }
 
     public function getTimestamp(string $key): int
--- a/src/Environment.php
+++ b/src/Environment.php
@@ -14,6 +14,7 @@ namespace Twig;
 use Twig\Cache\CacheInterface;
 use Twig\Cache\FilesystemCache;
 use Twig\Cache\NullCache;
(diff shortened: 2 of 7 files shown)


src/Environment.php
```diff
@@ -233,6 +235,18 @@ public function isStrictVariables()
         return $this->strictVariables;
     }
 
+    public function removeCache(string $name): void
+    {
+        $cls = $this->getTemplateClass($name);
+        $this->hotCache[$name] = $cls.'_'.bin2hex(random_bytes(16));
```
fabpot, line 241: Make sure that the template class name changes.

src/Environment.php
```diff
@@ -379,8 +393,10 @@ public function loadTemplate(string $cls, string $name, ?int $index = null): Tem
             if (!class_exists($cls, false)) {
                 $source = $this->getLoader()->getSourceContext($name);
                 $content = $this->compileSource($source);
-                $this->cache->write($key, $content);
-                $this->cache->load($key);
+                if (!isset($this->hotCache[$name])) {
```
fabpot, line 396: Do not pollute the cache with a temporary class that will never be reused in another/new PHP process.

src/Cache/FilesystemCache.php
```diff
@@ -76,6 +76,14 @@ public function write(string $key, string $content): void
         throw new \RuntimeException(\sprintf('Failed to write cache file "%s".', $key));
     }
 
+    public function remove(string $name, string $cls): void
+    {
+        $key = $this->generateKey($name, $cls);
+        if (is_file($key) && !@unlink($key)) {
```
nicolas-grekas: guards against concurrent unlinks?
```suggestion
        if (!@unlink($key) && file_exists($key)) {
```

fabpot: 👍 Done

nicolas-grekas (approved):

PR #4340, opened 2024-09-26
Fix 'ignore missing' when used on an 'embed' tag

### LINKED ISSUE #3085: 'ignore missing' does not ignore missing templates
This appears like it might well be a Twig bug, on my local, using the version on my branch (v1.38.4 ) adding a made up template to the homepage:

{% set x = 'bad' %}
{% embed x ~ 'ger.twig' ignore missing %}
{% endembed %}

Does not error, removing 'ignore missing' causes it to error…

Upgrading twig/twig (v1.38.4 => v1.42.1)

At which point the ignore missing directive appears to no longer work and the above test fails 

Twig\Error\LoaderError thrown with message "Unable to find template

fabpot:
Closes #3085

### DIFF
--- a/tests/Node/IncludeTest.php
+++ b/tests/Node/IncludeTest.php
@@ -80,9 +80,10 @@ EOF
 // line 1
 \$__internal_%s = null;
 try {
-    \$__internal_%s =     \$this->loadTemplate("foo.twig", null, 1);
+    \$__internal_%s = \$this->loadTemplate("foo.twig", null, 1);
 } catch (LoaderError \$e) {
     // ignore missing template
+    \$__internal_%s = null;
 }
 if (\$__internal_%s) {
     yield from \$__internal_%s->unwrap()->yield(CoreExtension::toArray(["foo" => true]));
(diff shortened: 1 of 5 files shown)


tests/Node/IncludeTest.php
```diff
@@ -80,9 +80,10 @@ public static function provideTests(): iterable
 // line 1
 \$__internal_%s = null;
 try {
-    \$__internal_%s =     \$this->loadTemplate("foo.twig", null, 1);
+    \$__internal_%s = \$this->loadTemplate("foo.twig", null, 1);
 } catch (LoaderError \$e) {
     // ignore missing template
+    \$__internal_%s = null;
```
stof, line 86: this seems useless as it is already assigned before the try/catch. Why do we need it ?

fabpot: The EmbedNode extends this one and there the template loading doesn’t fail. It fails when trying to load the parent. So we need to reset the variable here for this use case.

ruudk: Removed the assign before the try/catch here: 
- https://github.com/twigphp/Twig/pull/4346

Issue #4343, opened 2024-09-26
the "\" character is ignored in Twig v3 but will not be in v4


ianmclaughlin:
This deprecation message isn't actually true? 

For Twig 3.14
```
<br>double_slash: {{ date | date('Ymd\\THis') }}
<br>single_slash: {{ date | date('Ymd\THis') }}
```
results in 
double_slash: 20240926T103021
single_slash: 20240926BST103021

How do I fix the deprecation?

stof:
are you sure your reported results are not mixed up ?

In Twig 3.14, `\T` is parsed as being `T` (i.e. it considers the `\` as being part of an escape sequence `\T` even though `T` does not need any escaping)

ianmclaughlin:
feel free to run the code yourself...

I suspect the deprecation message is incorrectly showing for `\\` cases when it should only show for `\`cases

stof:
This is the output I get when running this (adding a case `no_slash`:
```
<br>double_slash: 20240926T103031\n
<br>single_slash: 20240926UTC103031\n
<br>no_slash:     20240926UTC103031
```

As you can see, the `single_slash` and `no_slash` cases produce the same output in Twig 3.x. they format the date using `YmdTHis` as format (where `T` will show the timezone name, which is `UTC` in my case).

and I get a single deprecation being reported.

@fabpot something is still wrong with that deprecation. It says `User Deprecated: Since twig/twig 3.12: Character "T" at position 5 should not be escaped; the "\" character is ignored in Twig v3 but will not be in v4. Please remove the extra "\" character.`. However, this `position 5` seems to be the position inside the string literal, not the position inside the template, making it totally useless to identify where the deprecation actually comes from.

ianmclaughlin:
The example I provided showed a difference between double-slash case and a single slash case, and yours does too.

The deprecation is shown for the double-slash case, but when I change the code to take out a slash the deprecation is not shown. This made me think the deprecation was fixed, when I had instead altered the result.

ruudk:
I'll try to tweak the deprecation message, so that it's more clear.

stof:
@ianmclaughlin of course it does in 3.x. That's precisely why we have a deprecation warning (in 4.0, the single_slash case will be changed so that `\T` is parsed as a `\` followed by a `T` instead as `\T` is not an actual escape sequence).

The deprecation is triggered for the single_slash case, not for the double_slash case

ruudk:
Created a PR to improve a few things:
[...]

PR #4344, opened 2024-09-26
Improve escape deprecation message

ruudk:
This was introduced in https://github.com/twigphp/Twig/pull/4176

According to #4343 the message can be improved.

I made a few changes to the test to make it easier to work with. Best to review this PR **commit by commit**.

/cc @stof

### DIFF
--- a/src/Lexer.php
+++ b/src/Lexer.php
@@ -418,7 +418,7 @@ class Lexer
                 $result .= $nextChar;
             } elseif ("'" === $nextChar || '"' === $nextChar) {
                 if ($nextChar !== $quoteType) {
-                    trigger_deprecation('twig/twig', '3.12', 'Character "%s" at position %d should not be escaped; the "\" character is ignored in Twig v3 but will not be in v4. Please remove the extra "\" character.', $nextChar, $i + 1);
+                    trigger_deprecation('twig/twig', '3.12', 'Character "%s" at position %d in string on line %d should not be escaped; the "\" character is ignored in Twig v3 but will not be in v4. Please remove the extra "\" character.', $nextChar, $i + 1, $this->lineno);
                 }
                 $result .= $nextChar;
             } elseif ('#' === $nextChar && $i + 1 < $length && '{' === $str[$i + 1]) {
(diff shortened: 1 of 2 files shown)


src/Lexer.php
```diff
@@ -418,7 +418,7 @@ private function stripcslashes(string $str, string $quoteType): string
                 $result .= $nextChar;
             } elseif ("'" === $nextChar || '"' === $nextChar) {
                 if ($nextChar !== $quoteType) {
-                    trigger_deprecation('twig/twig', '3.12', 'Character "%s" at position %d in string on line %d should not be escaped; the "\" character is ignored in Twig v3 but will not be in v4. Please remove the extra "\" character.', $nextChar, $i + 1, $this->lineno);
+                    trigger_deprecation('twig/twig', '3.12', 'Character "%s" at position %d in string on line %d should not be escaped; the "\" character is ignored in Twig v3 but will not be in v4. Please remove the extra "\" character or add another one if you want to print the escape character too.', $nextChar, $i + 1, $this->lineno);
```
stof: I'm not sure we need this alternative, as the deprecation generally explains how to keep the same behavior.

tests/LexerTest.php
```diff
@@ -268,6 +276,15 @@ public function testStringWithEscapedDelimiterProducingDeprecation(string $templ
 
     public static function getStringWithEscapedDelimiterProducingDeprecation()
     {
+        yield [
+            <<<'EOF'
+            {{ 'Ymd\THis' }}
```
stof: this case is exactly the same one that the next case, so no need to add it.

src/Lexer.php
ruudk: Alright, removed it.

tests/LexerTest.php
ruudk: Removed it.

stof (approved):

fabpot:
Thank you @ruudk.

smnandre:
Would it be possible to add the source path (if it exists) ?   `$this->source?->getPath()` ? 

It took me time to figure where one came from in my code (spoiler alert: a date format 😅)

PR #4345, opened 2024-09-26
Add fixture filename as key

ruudk:
This way, it's easier to see which test failed and what the file was.

In PHPStorm this is visualized very nice.
Screenshot 2024-09-26 at 14 19 49@2x


fabpot:
Thank you @ruudk.

PR #4346, opened 2024-09-26
Remove useless assign in compiled code

ruudk:
It's already unset / initialized on catch. 

Having it this way, is a bit easier for me when analyzing this in TwigStan.


fabpot:
Thank you @ruudk.

PR #4347, opened 2024-09-26
Remove useless Closure::bind in for loop

ruudk:
This has no effect. The returned value is not used.

https://www.php.net/manual/en/closure.bind.php


fabpot:
Thank you @ruudk.

PR #4349, opened 2024-09-26
Add support for inline comments

### LINKED ISSUE #4001: Allow comments within components / function calls
It is so natural to comment out HTML attributes that it's easy to think that Twig supports this. So easy that even the Symfony UX documentation has an example.

```twig
// ...

<div class="toggle-password-container"> // Add "toggle-password-container" or a class that applies position: relative to this container.
    <label for="password">Password</label>
    <input
        id="password"
        name="password"
        type="password"
        {{ stimulus_controller('symfony/ux-toggle-password/toggle-password', {
                {# visibleLabel: 'Show password', // If you want to modify this label. #}
                {# visibleIcon: 'Some svg icon', // If you want to modify this icon. #}
                {# hiddenLabel: 'Hide password', // If you want to modify this label. #}
                {# hiddenIcon: 'Some svg icon', // If you want to modify this icon. #}
                buttonClasses: ['toggle-password-button'], // Add as many classes as you wish. "toggle-password-button" is needed to activate the default CSS.
        }) }}
    >
</div>

// ...
```

https://symfony.com/bundles/ux-toggle-password/current/index.html#usage-without-symfony-forms

I started to make a PR to fix the docs, but it's not easy to decide where to move it.  It'd be awesome if twig allowed this, as I frequently use it for in-line comments as well as disabling attributes.

fabpot:
Closes #4001

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
 # 3.15.0 (2024-XX-XX)
 
+ * Add support for inline comments
  * Add support for accessing class constants with the dot operator
  * Add `Profile::getStartTime()` and `Profile::getEndTime()`
  * Fix "ignore missing" when used on an "embed" tag
--- a/doc/templates.rst
+++ b/doc/templates.rst
@@ -259,9 +259,9 @@ Go to the :doc:`tags<tags/index>` page to learn more about the built-in tags.
 Comments
 --------
 
-To comment-out part of a line in a template, use the comment syntax ``{# ...
-#}``. This is useful for debugging or to add information for other template
-designers or yourself:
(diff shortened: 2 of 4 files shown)


stof:
I suggest adding tests covering that a `#` inside strings is not triggering the parsing as inline comments

fabpot:
> I suggest adding tests covering that a `#` inside strings is not triggering the parsing as inline comments

Added

VincentLanglet:
Hi, I dunno if it was wanted but I feel like inline comment also works inside interpolation
```
{{ "#{1 # This is a comment
         ~ 1}" }}
```

PR #4350, opened 2024-09-26
Replace strtr() by strtolower()

nicolas-grekas:
The performance of `strtolower()` has been fixed in PHP 8.

### DIFF
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -1686,7 +1686,7 @@ final class CoreExtension extends AbstractExtension
         if (!isset($cache[$class])) {
             $methods = get_class_methods($object);
             sort($methods);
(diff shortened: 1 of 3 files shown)


src/Extension/CoreExtension.php
```diff
@@ -1686,7 +1686,7 @@ public static function getAttribute(Environment $env, Source $source, $object, $
         if (!isset($cache[$class])) {
             $methods = get_class_methods($object);
             sort($methods);
-            $lcMethods = array_map(function ($value) { return strtr($value, 'ABCDEFGHIJKLMNOPQRSTUVWXYZ', 'abcdefghijklmnopqrstuvwxyz'); }, $methods);
+            $lcMethods = array_map('strtrolower', $methods);
```
```suggestion
            $lcMethods = array_map('strtolower', $methods);
```

stof (changes requested):

src/Extension/CoreExtension.php
nicolas-grekas: fixed :)

fabpot:
Thank you @nicolas-grekas.

PR #4351, opened 2024-09-26
Fix getting a property on an object casted to an array

fabpot:
Closes #3371


nicolas-grekas:
FTR, I propose another approach in #4425, which is to make the case for DateTime[Immutable] objects special, because those objects are special.

PR #4352, opened 2024-09-26
Implement the `enum` function

### LINKED ISSUE #3681: Support reading enums from PHP 8.1

nicolas-grekas:
Fixes #3681
Requires #4353

### DIFF
--- a/doc/functions/enum.rst
+++ b/doc/functions/enum.rst
@@ -1,0 +1,26 @@
+``enum_cases``
+==============
+
+.. versionadded:: 3.15
+
+    The ``enum`` function was added in Twig 3.15.
+
+``enum`` gives access to enums:
+
+.. code-block:: twig
+
+    {# display one specific case of a backed enum #}
+    {{ enum('App\\MyEnum').SomeCase.value }}
+
+    {# get all cases of an enum #}
+    {% enum('App\\MyEnum').cases() %}
+
+    {# call any methods of the enum class #}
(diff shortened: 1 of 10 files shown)


fabpot:
@nicolas-grekas Can you rebase?

doc/functions/enum.rst
```diff
@@ -0,0 +1,26 @@
...
+
+``enum`` gives access to enums:
+
+.. code-block:: twig
+
+    {# display one specific case of a backed enum #}
+    {{ enum('App\\MyEnum').SomeCase.value }}
+
+    {# get all cases of an enum #}
+    {% enum('App\\MyEnum').cases() %}
```
fabpot, line 16: Should we deprecate `enum_cases()`?

stof: I would not deprecated it. `enum_cases` compiles to optimized code when the enum class name is a string literal, avoiding any dot operator resolution at runtime.

fabpot: 👍

fabpot:
Thank you @nicolas-grekas.

PR #4353, opened 2024-09-26
Add support for accessing class constants with the dot operator

nicolas-grekas:
This PR allows accessing class constants from objects: `{{ foo.SOME_CONSTANT }}`. This should nicely replace the `constant()` function when an object is at hand, and is actually needed to make the `enum` function work in #4352.

### DIFF
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -1675,6 +1675,18 @@ final class CoreExtension extends AbstractExtension
 
                 return $object->$item;
             }
+
+            if (\defined($object::class.'::'.$item)) {
+                if ($isDefinedTest) {
+                    return true;
+                }
+
+                if ($sandboxed) {
+                    $env->getExtension(SandboxExtension::class)->checkPropertyAllowed($object, $item, $lineno, $source);
+                }
+
+                return \constant($object::class.'::'.$item);
+            }
         }
 
         static $cache = [];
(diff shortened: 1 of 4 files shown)


fabpot:
It cannot replace `constant()` for all use cases, right? Here, it's only when you want to get a constant on an object, correct?

nicolas-grekas:
Correct, we need an object so this cannot replace `constant('FQCN::SOME_CONST')` calls.

fabpot:
Thank you @nicolas-grekas.

stof:
isn't this a BC break as a constant can now be accessed instead of calling a getter ?

nicolas-grekas:
> isn't this a BC break as a constant can now be accessed instead of calling a getter ?

It can yes, but given the typical conventions for consts (all uppercase + underscores), I'd rate this as low proba. Let's see :)

src/Extension/CoreExtension.php
```diff
@@ -1675,6 +1675,18 @@ public static function getAttribute(Environment $env, Source $source, $object, $
...
                 return $object->$item;
             }
+
+            if (\defined($object::class.'::'.$item)) {
+                if ($isDefinedTest) {
+                    return true;
+                }
+
+                if ($sandboxed) {
+                    $env->getExtension(SandboxExtension::class)->checkPropertyAllowed($object, $item, $lineno, $source);
```
stof, line 1685: this should probably be a new sandbox method as it is not a property access

PR #4354, opened 2024-09-26
Do not pass closure as reference

ruudk:
It's not needed, as the tests pass.


ruudk:
Handled in https://github.com/twigphp/Twig/pull/4355

PR #4355, opened 2024-09-26
Simplify compiled `for` loop code

ruudk:
**Do not pass closure as reference**
It's not needed.

**Do not use same var names in loop function** 
These variables are passed in when the loop function is invoked when yielding.
Giving them the same internal var names complicates things for no good reason.

### DIFF
--- a/src/Node/ForNode.php
+++ b/src/Node/ForNode.php
@@ -43,25 +43,24 @@ class ForNode extends Node
     {
         $iteratorVar = $compiler->getVarName();
         $functionVar = $compiler->getVarName();
-        $parentVar = $compiler->getVarName();
 
         $compiler
             ->addDebugInfo($this)
             ->write("\$$iteratorVar = new \Twig\Runtime\LoopIterator(")
             ->subcompile($this->getNode('seq'))
             ->raw(");\n")
-            ->write("\$$functionVar = function (\$$iteratorVar, &\$context, \$blocks, \$$functionVar, \$depth) {\n")
+            ->write("\$$functionVar = function (\$iterator, &\$context, \$blocks, \$recurseFunc, \$depth) {\n")
             ->indent()
             ->write("\$macros = \$this->macros;\n")
-            ->write("\$$parentVar = \$context;\n")
+            ->write("\$parent = \$context;\n")
         ;
 
         if ($this->getAttribute('with_loop')) {
-            $compiler->write("\$context['loop'] = new \Twig\Runtime\LoopContext(\$$iteratorVar, \$$parentVar, \$blocks, \$$functionVar, \$depth);\n");
+            $compiler->write("\$context['loop'] = new \Twig\Runtime\LoopContext(\$iterator, \$parent, \$blocks, \$recurseFunc, \$depth);\n");
         }
 
         $compiler
-            ->write("foreach (\$$iteratorVar as ")
+            ->write("foreach (\$iterator as ")
             ->subcompile($this->getNode('key_target'))
             ->raw(' => ')
             ->subcompile($this->getNode('value_target'))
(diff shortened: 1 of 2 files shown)


fabpot:
Thank you @ruudk.

src/Node/ForNode.php
```diff
@@ -43,25 +43,24 @@ public function compile(Compiler $compiler): void
...
             ->addDebugInfo($this)
             ->write("\$$iteratorVar = new \Twig\Runtime\LoopIterator(")
             ->subcompile($this->getNode('seq'))
             ->raw(");\n")
-            ->write("\$$functionVar = function (\$$iteratorVar, &\$context, \$blocks, &\$$functionVar, \$depth) {\n")
+            ->write("\$$functionVar = function (\$iterator, &\$context, \$blocks, \$recurseFunc, \$depth) {\n")
             ->indent()
             ->write("\$macros = \$this->macros;\n")
-            ->write("\$$parentVar = \$context;\n")
+            ->write("\$parent = \$context;\n")
```
stof, line 55: wouldn't naming it `$parent` all the time cause issue with nested `{% for %}` tags ?

stof: same for the iterator btw

ruudk: Good question. Since it uses an anonymous function (no short arrow), it's protected.
See compiled code for a nested loop:

```php
protected function doDisplay(array $context, array $blocks = []): iterable
    {
        $macros = $this->macros;
        // line 1
        yield "
";
        // line 2
        $__internal_compile_0 = new \Twig\Runtime\LoopIterator((array_key_exists("items", $context) ? $context["items"] : throw new RuntimeError('Variable "items" does not exist.', 2, $this->source)));
        $__internal_compile_1 = function ($iterator, &$context, $blocks, $recurseFunc, $depth) {
            $macros = $this->macros;
            $parent = $context;
            $context['loop'] = new \Twig\Runtime\LoopContext($iterator, $parent, $blocks, $recurseFunc, $depth);
            foreach ($iterator as $context["key"] => $context["item"]) {
                // line 3
                yield "* ";
                yield $this->env->getRuntime('Twig\Runtime\EscaperRuntime')->escape($this->sandbox->ensureToStringAllowed($context["key"], 3, $this->source), "html", null, true);
                yield " (";
                yield $this->env->getRuntime('Twig\Runtime\EscaperRuntime')->escape($this->sandbox->ensureToStringAllowed(CoreExtension::getAttribute($this->env, $this->source, $context["loop"], "length", arguments: [], sandboxed: true, lineno: 3), 3, $this->source), "html", null, true);
                yield "):
";
                // line 4
                $__internal_compile_2 = new \Twig\Runtime\LoopIterator($context["item"]);
                $__internal_compile_3 = function ($iterator, &$context, $blocks, $recurseFunc, $depth) {
                    $macros = $this->macros;
                    $parent = $context;
                    $context['loop'] = new \Twig\Runtime\LoopContext($iterator, $parent, $blocks, $recurseFunc, $depth);
                    foreach ($iterator as $context["_key"] => $context["value"]) {
                        // line 5
                        yield "  * ";
                        yield $this->env->getRuntime('Twig\Runtime\EscaperRuntime')->escape($this->sandbox->ensureToStringAllowed($context["value"], 5, $this->source), "html", null, true);
                        yield " (";
                        yield $this->env->getRuntime('Twig\Runtime\EscaperRuntime')->escape($this->sandbox->ensureToStringAllowed(CoreExtension::getAttribute($this->env, $this->source, $context["loop"], "length", arguments: [], sandboxed: true, lineno: 5), 5, $this->source), "html", null, true);
                        yield ")
";
                    }
                    unset($context['_key'], $context['value'], $context['loop']);
                    $context = array_intersect_key($context, $parent) + $parent;
                    yield from [];
                };
                yield from $__internal_compile_3($__internal_compile_2, $context, $blocks, $__internal_compile_3, 0);
            }
            unset($context['key'], $context['item'], $context['loop']);
            $context = array_intersect_key($context, $parent) + $parent;
            yield from [];
        };
        yield from $__internal_compile_1($__internal_compile_0, $context, $blocks, $__internal_compile_1, 0);
        yield from [];
    }
```

PR #4357, opened 2024-09-26
Update plural.rst: Fixing(?) broken links

ThomasLandauer:
More important question: Is the filter's name `plural` or `pluralize`??


fabpot:

fabpot:
`plural` is the name of the filter in Twig, `pluralize` is the name of the method in Symfony.

PR #4358, opened 2024-09-27
Update plural.rst: Fixing filter name

ThomasLandauer:
Info is taken from https://github.com/twigphp/Twig/pull/4357#issuecomment-2378435199

Besides, I changed to a more illustrative example word. One of the rare cases where ChatGPT was indeed useful... ;-)


fabpot:

PR #4361, opened 2024-09-29
Use get_debug_type() everywhere

fabpot:


smnandre:
Do you plan (or would like) to replace the few remaining `get_class($foo)` with `$foo::class` ?

fabpot:
> Do you plan (or would like) to replace the few remaining `get_class($foo)` with `$foo::class` ?

Not in 3.x as this was done in 4.x already. See #4380 as we had a few remaining calls.

Issue #4256, opened 2024-08-30
Determine which types to support for `types` tag


drjayvee:
Following discussions in #4165, we want to determine which types (if any), Twig wants to suggest or even mandate extensions support.

There seem to be four options:

1. Remain completely agnostic
2. Use PHP's types: `string`, `bool`, etc
3. Use more generic types such as `stringable`, `iterable`
4. Use [PHPStan/Psalm](https://phpstan.org/writing-php-code/phpdoc-types) types

fabpot:
It's also related to the work done here: https://github.com/twigphp/Twig/pull/4252/files#diff-6fd6db196be65abd4dcc8132c2bc70a2e0b232c4a5732f29ea26d0a09676ff74R79-R90

Haehnchen:
the PhpStorm Symfony integration now supports `types`: https://github.com/Haehnchen/idea-php-symfony2-plugin/issues/2396

Basically most projects needs a class type hint with array. Thats what is now supported here for know:

```
{% types {
    foobar: '\App\Entity\ErrorReport',
    foobars: '\App\Entity\ErrorReport[]',
} %}
```
I am also not sure about the type value itself. imho should be as open as possible but guidlines would be helpful.
In the end we are talking about templates and to supports all the "syntax sugar" from the start looks too heavy:

- `array<\App\Entity\ErrorReport>`
- `array<int|\App\Entity\ErrorReport>`
- `\App\Entity\ErrorReport|\App\Entity\ErrorReportFoo`

fabpot:
Can we start small with all the types from https://github.com/twigphp/Twig/pull/4362/files + some more precise ones like the above example?

drjayvee:
I'm working on a `types` validation inspection for [TwigQI](https://github.com/alisqi/twigqi), so allow me to add my thoughts here.

First off, I've come around to preferring the higher-level Twig types `number` and `iterable` over the actual PHP ones. Maybe `integer`, `float` and `array` should be allowed, but deprecated straight away?

However, I really being able to tell the distinction between lists (e.g., `[13, 37]`) and maps (e.g., `{foo: 13, bar: 37}`). I'd be in favor of deprecating the ambiguous `iterable` in favor of ~~list~~ `sequence` and `mapping`. Just my two cents.

💡 What if we support:
* `iterable<valueType>` for ~~lists~~ sequences
* `iterable<keyType, valueType>` for ~~maps~~ mappings
* `iterable` for *either*

[...]

PR #4362, opened 2024-09-29
Document Twig vs PHP types

fabpot:
Refs #4256

### DIFF
--- a/doc/templates.rst
+++ b/doc/templates.rst
@@ -76,8 +76,24 @@ Twig templates have access to variables provided by the PHP application and
 variables created in templates via the :doc:`set <tags/set>` tag. These
 variables can be manipulated and displayed in the template.
 
-Use a dot (``.``) to access attributes of a variable (methods, properties
-or constants of a PHP object, or items of a PHP array):
+Twig tries to abstract PHP types as much as possible and works with a few basic
+types, supported by ``filters``, ``functions``, and ``tests`` among others:
+
+===================  ===============================
+Twig Type            PHP Type
+===================  ===============================
+string               A string or a Stringable object
+number               An integer or a float
+boolean              ``true`` or ``false``
+null                 ``null``
+iterable (mapping)   An array
+iterable (sequence)  An array
+iterable (object)    An iterable object
+object               An object
+===================  ===============================
+
+The ``iterable`` and ``object`` types expose attributes you can access via the
+dot (``.``) operator:
 
 .. code-block:: twig
 


fabpot:
I think this one is ready, comments are welcomed.
Note that this is about documenting "native" Twig types and not all types supported by the `@types` tag.

fabpot:
@drjayvee @ruudk I'd love to have your thoughts here.

drjayvee:
Thanks for the heads up, Fabian. I'll have a look at this tomorrow or Wednesday and will provide feedback.

javiereguiluz (approved):

ruudk (approved):
Looks good! :+1:

doc/templates.rst
```diff
@@ -76,8 +76,22 @@ Twig templates have access to variables provided by the PHP application and
...
+types, supported by ``filters``, ``functions``, and ``tests`` among others:
+
+=========  ===============================================
+Twig Type  PHP Type
+=========  ===============================================
+string     A sequence of characters or a Stringable object
+number     An integer or a float
+boolean    ``true`` or ``false``
+null       ``null``
+iterable   A sequence, a mapping, or an iterable object
```
xabbuh: Doesn't Twig also distinguish sequences and mappings?

xabbuh: On the PHP side `sequence` and `mapping` should probably be replaced with `array`?

fabpot: You're right, updated.

xabbuh (approved):

drjayvee:
LGTM as well 👍

doc/templates.rst
```diff
@@ -76,8 +76,24 @@ Twig templates have access to variables provided by the PHP application and
...
-Use a dot (``.``) to access attributes of a variable (methods, properties
-or constants of a PHP object, or items of a PHP array):
+Twig tries to abstract PHP types as much as possible and works with a few basic
+types, supported by ``filters``, ``functions``, and ``tests`` among others:
+
+===================  ===============================
+Twig Type            PHP Type
+===================  ===============================
+string               A string or a Stringable object
+number               An integer or a float
```
alexander-schranz, line 86: In the docs `int` is used here: https://twig.symfony.com/doc/3.x/tags/types.html#types
Do we specify for the new feature PHP types instead of Twig types?

drjayvee: Good catch!

While the documentation states "The syntax for and contents of type strings are intentionally left out of scope", I do agree that "number" and "boolean" are better choices to align with the documentation on types.

I'll create a PR for updating the `types` documentation.

drjayvee: Voila: #4403

PR #4363, opened 2024-09-29
Deprecate using ~ with + or - in an expression without using parentheses to clarify precedence

fabpot:
See https://wiki.php.net/rfc/concatenation_precedence for the related change in PHP 7.4.


stof:
Can we also have tests using parenthesis on the other branch to check that we trigger deprecation only for the ambiguous case ?

fabpot:
> Can we also have tests using parenthesis on the other branch to check that we trigger deprecation only for the ambiguous case ?

Done :)

doc/deprecated.rst
```diff
@@ -296,3 +296,23 @@ Node
...
+
+Operators
+---------
+
+* Using ``~`` with ``+`` or ``-`` in an expression without using parentheses to
+  clarify precedence is deprecated as of Twig 3.15 (in Twig 4.0, parentheses
+  won't be needed anymore as ``+`` / ``-`` will have a higher precedence than
+  ``~``):
+
+  Before::
```
javiereguiluz: This is a bit confusing:

```rst
  Before::

    {{ '42' ~ 1 + 41 }}

    {# which is equivalent to #}

    {{ ('42' ~ 1) + 41 }}

  After::

    {{ '42' ~ (1 + 41) }}
```

Maybe use something like this instead?:

```twig
  {# the following expression: #}
  {{ '42' ~ 1 + 41 }}

  {# in Twig 3.x is equivalent to #}
  {{ ('42' ~ 1) + 41 }}

  {# in Twig 4.x is equivalent to #}
  {{ '42' ~ (1 + 41) }}
```

fabpot: See how I changed the example to be clearer.

PR #4367, opened 2024-09-30
Trigger deprecations when using "??" and "not" without explicit parentheses when precedence will change in 4.0

### LINKED ISSUE #3387: Precedence of ?? operator
Hi!

This might be a bug:
The precedence of ?? is not the same in twig as it is in php.

> php:
$x = 'a';
var_dump(null ?? $x.'_1'); // a_1
var_dump('notnull' ?? $x.'_1'); // notnull

> twig:
{{ null ?? x ~ '_1' }} // a_1
{{ 'notnull' ?? x ~ '_1' }} // notnull_1 expected: nutnull
{{ 'notnull' ?? (x ~ '_1') }} // notnull

### LINKED ISSUE #3642: Wrong precedence of NOT operator
In most programming languages (such as C++, Java, or PHP) the boolean unary NOT operator has a higher precedence than e.g. the multiplication, division, remainder, or relational operators. It usually has a precedence same as the unary '+' or unary '-'.

But why does Twig deviate from this practice and puts the unary 'NOT' operator precedence between the ones below:

....
'+' (binary add)
'-' (binary minus)
'~' (binary concat)
**'NOT' (unary not)**
'*' (binary multiply)
'/' (binary division)
....

Not even Django (Python) has such a strange precedence placement.

There are several other bad habits in the Twig expression syntax, but I think above mentioned issue could be fixed, and if need be, add a runtime switch to ensure backward compatibility with older Twig versions, but have future Twig releases use a proper operator precedence.

fabpot:
Closes #3387
Closes #3642

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
 # 3.15.0 (2024-XX-XX)
 
+ * Deprecate using `??` without explicit parentheses
  * Deprecate using `~` with `+` or `-` in an expression without using parentheses to clarify precedence
  * Deprecate not passing `AbstractExpression` args to most constructor arguments for classes extending `AbstractExpression`
  * Fix `power` expressions with a negative number in parenthesis (`(-1) ** 2`)
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,7 +1,9 @@
 # 3.15.0 (2024-XX-XX)
 
- * Deprecate using `??` without explicit parentheses
- * Deprecate using `~` with `+` or `-` in an expression without using parentheses to clarify precedence
+ * Add support for triggering deprecations for future operator precedence changes
+ * Deprecate using the `not` unary operator in an expression with ``*``, ``/``, ``//``, or ``%`` without using explicit parentheses to clarify precedence
+ * Deprecate using the `??` binary operator without explicit parentheses
+ * Deprecate using the `~` binary operator in an expression with `+` or `-` without using parentheses to clarify precedence
  * Deprecate not passing `AbstractExpression` args to most constructor arguments for classes extending `AbstractExpression`
  * Fix `power` expressions with a negative number in parenthesis (`(-1) ** 2`)
  * Deprecate instantiating `Node` directly. Use `EmptyNode` or `Nodes` instead.
--- a/src/ExpressionParser.php
+++ b/src/ExpressionParser.php
@@ -64,29 +64,36 @@ class ExpressionParser
         $this->unaryOperators = $env->getUnaryOperators();
         $this->binaryOperators = $env->getBinaryOperators();
 
-        foreach ($this->binaryOperators as $name => $config) {
-            if (!isset($config['future_precedence'])) {
+        $ops = [];
+        foreach ($this->unaryOperators as $n => $c) {
+            $ops[] = $c + ['name' => $n, 'type' => 'unary'];
+        }
+        foreach ($this->binaryOperators as $n => $c) {
+            $ops[] = $c + ['name' => $n, 'type' => 'binary'];
+        }
+        foreach ($ops as $config) {
+            if (!isset($config['precedence_change'])) {
                 continue;
             }
-
-            $min = min($config['future_precedence'], $config['precedence']);
-            $max = max($config['future_precedence'], $config['precedence']);
-            foreach ($this->binaryOperators as $n => $c) {
+            $name = $config['type'].'_'.$config['name'];
+            $min = min($config['precedence_change']->getNewPrecedence(), $config['precedence']);
+            $max = max($config['precedence_change']->getNewPrecedence(), $config['precedence']);
+            foreach ($ops as $c) {
                 if ($c['precedence'] > $min && $c['precedence'] < $max) {
-                    $this->precedenceChanges[$n][] = $name;
+                    $this->precedenceChanges[$c['type'].'_'.$c['name']][] = $name;
                 }
             }
         }
     }
 
-    public function parseExpression($precedence = 0, $allowArrow = false)
+    public function parseExpression($precedence = 0, $allowArrow = false, bool $deprecationCheck = true)
     {
         if ($allowArrow && $arrow = $this->parseArrow()) {
             return $arrow;
         }
 
         $expr = $this->getPrimary();
-        $token = $this->parser->getCurrentToken();
+        $previousToken = $token = $this->parser->getCurrentToken();
         while ($this->isBinary($token) && $this->binaryOperators[$token->getValue()]['precedence'] >= $precedence) {
             $op = $this->binaryOperators[$token->getValue()];
             $this->parser->getStream()->next();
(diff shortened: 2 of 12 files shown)


tests/Fixtures/tests/null_coalesce.test
```diff
@@ -10,9 +10,9 @@ Twig supports the ?? operator
 {{ foo.bar.baz.missing ?? 'OK' }}
 {{ foo['bar'] ?? 'KO' }}
 {{ foo['missing'] ?? 'OK' }}
-{{ nope ?? nada ?? 'OK' }}
-{{ 1 + nope ?? nada ?? 2 }}
-{{ 1 + nope ?? 3 + nada ?? 2 }}
+{{ nope ?? (nada ?? 'OK') }}
```
stof, line 13: should we really report this case ? It will be parsed the same in 4.0

src/ExpressionParser.php
```diff
@@ -117,6 +118,25 @@ private function triggerPrecedenceDeprecations(AbstractExpression $expr, Token $
...
+        // Precedence of the ?? operator will be lower than everything else in Twig 4.0
+        if ($expr instanceof NullCoalesceExpression) {
+            /** @var AbstractExpression $expr2 */
+            $expr2 = $expr->getNode('expr2');
+            /** @var AbstractExpression $expr3 */
+            $expr3 = $expr->getNode('expr3');
+            if (
+                (($expr2 instanceof AddBinary || $expr2 instanceof SubBinary || $expr2 instanceof ConditionalExpression) && !$expr2->hasExplicitParentheses())
+                ||
+                (($expr3 instanceof AddBinary || $expr3 instanceof SubBinary || $expr3 instanceof ConditionalExpression) && !$expr3->hasExplicitParentheses())
```
stof: Don't we have a lot more binary nodes that can be impacted (basically all of them as we go from the highest precedence among binary nodes to the lowest one).

stof (commented):
I'm wondering whether this could be implemented in the ExpressionParser while resolving precedence (by providing a `future_precedence` or something like that) instead of inspecting the AST after the fact. This could make it work in combination with third-party operators as well (which might have a precedence between the old and new ones and so be impacted).

fabpot:
@stof I've rewritten how this all works. It's now all automated in the sense that it's done in a generic way where I just need the current and future precedence. It's a quick prototype I've done in a few minutes, so I might have forgotten something. Can you have a look?

fabpot:
The implementation is now much better. I'm done for this PR.

src/Extension/CoreExtension.php
```diff
@@ -296,7 +296,7 @@ public function getOperators(): array
     {
         return [
             [
-                'not' => ['precedence' => 50, 'class' => NotUnary::class],
+                'not' => ['precedence' => 50, 'future_precedence' => 70, 'class' => NotUnary::class],
```
stof: Please update the phpdoc type in ExtensionInterface to document the optional `future_precedence` key here, as the implementation actually makes it possible for third-party extensions to use it for their own operators.

src/ExpressionParser.php
```diff
@@ -102,20 +127,36 @@ public function parseExpression($precedence = 0, $allowArrow = false)
...
+                if ($node->hasAttribute('operator') && $operatorName === $node->getAttribute('operator')) {
+                    trigger_deprecation('twig/twig', '3.15', \sprintf('Add explicit parentheses around the "%s" unary operator to avoid behavior change in Twig 4.0 as its precedence will change in "%s" at line %d.', $target, $this->parser->getStream()->getSourceContext()->getName(), $token->getLine()));
+                }
+            }
+        } else {
+            foreach ($this->precedenceChanges[$expr->getAttribute('operator')] as $operatorName) {
+                foreach ($expr as $node) {
+                    /** @var AbstractExpression $node */
+                    if ($node->hasAttribute('operator') && $operatorName === $node->getAttribute('operator') && !$node->hasExplicitParentheses()) {
+                        trigger_deprecation('twig/twig', '3.15', \sprintf('Add explicit parentheses around the "%s" binary operator to avoid behavior change in Twig 4.0 as its precedence will change in "%s" at line %d.', explode('_', $operatorName)[1], $this->parser->getStream()->getSourceContext()->getName(), $token->getLine()));
```
stof: Can we make the operator provide the package and version for the deprecation instead ? Maybe Twig will perform precedence deprecations in other minor versions for other operators, and maybe other packages might want to benefit from this deprecation for their custom operators.

src/Extension/CoreExtension.php
fabpot: Done

src/ExpressionParser.php

src/Extension/CoreExtension.php
```diff
@@ -296,7 +296,7 @@ public function getOperators(): array
     {
         return [
             [
-                'not' => ['precedence' => 50, 'class' => NotUnary::class],
+                'not' => ['precedence' => 50, 'future_precedence' => 70, 'class' => NotUnary::class, 'package' => 'twig/twig', 'version' => '3.15'],
```
stof: should we have `'future_precedence' => new PrecedenceChangeInfo(70, 'twig/twig', '3.15')` instead, for consistency with the new way to provide deprecations for callables ? The current naming does not make it clear that `version` applies to the precedence deprecation.

src/Extension/ExtensionInterface.php
```diff
@@ -67,8 +67,8 @@ public function getFunctions();
      * @return array<array> First array of unary operators, second array of binary operators
      *
      * @psalm-return array{
-     *     array<string, array{precedence: int, class: class-string<AbstractExpression>}>,
-     *     array<string, array{precedence: int, class?: class-string<AbstractExpression>, associativity: ExpressionParser::OPERATOR_*}>
+     *     array<string, array{precedence: int, future_precedence: int, package: string, version: string, class: class-string<AbstractExpression>}>,
+     *     array<string, array{precedence: int, future_precedence: int, package: string, version: string, class?: class-string<AbstractExpression>, associativity: ExpressionParser::OPERATOR_*}>
```
stof: the keys need to be optional:

```suggestion
     *     array<string, array{precedence: int, future_precedence?: int, package?: string, version?: string, class?: class-string<AbstractExpression>, associativity: ExpressionParser::OPERATOR_*}>
```

src/ExpressionParser.php
```diff
@@ -104,18 +129,37 @@ public function parseExpression($precedence = 0, $allowArrow = false)
...
+                    trigger_deprecation($targetPackage, $targetVersion, \sprintf('Add explicit parentheses around the "%s" unary operator to avoid behavior change in Twig 4.0 as its precedence will change in "%s" at line %d.', $target, $this->parser->getStream()->getSourceContext()->getName(), $token->getLine()));
+                }
+            }
+        } else {
+            foreach ($this->precedenceChanges[$expr->getAttribute('operator')] as $operatorName) {
+                foreach ($expr as $node) {
+                    /** @var AbstractExpression $node */
+                    if ($node->hasAttribute('operator') && $operatorName === $node->getAttribute('operator') && !$node->hasExplicitParentheses()) {
+                        $op = explode('_', $operatorName)[1];
+                        trigger_deprecation($this->binaryOperators[$op]['package'], $this->binaryOperators[$op]['version'], \sprintf('Add explicit parentheses around the "%s" binary operator to avoid behavior change in Twig 4.0 as its precedence will change in "%s" at line %d.', $op, $this->parser->getStream()->getSourceContext()->getName(), $token->getLine()));
```
stof: this should say `in the next major version` as the package might not be Twig

src/Extension/ExtensionInterface.php
fabpot: Done

src/ExpressionParser.php
fabpot: Indeed

src/Extension/CoreExtension.php
fabpot: Good idea

stof (approved):

CHANGELOG
```diff
@@ -1,6 +1,8 @@
 # 3.15.0 (2024-XX-XX)
 
- * Deprecate using `~` with `+` or `-` in an expression without using parentheses to clarify precedence
+ * Deprecate using the `not` unary operator in an expression with ``*``, ``/``, ``//``, or ``%`` without using explicit parentheses to clarify precedence
+ * Deprecate using the `??` binary operator without explicit parentheses
+ * Deprecate using the `~` binary operator in an expression with `+` or `-` without using parentheses to clarify precedence
```
stof, line 6: this should also mention `Add support for triggering deprecations for future operator precedence changes` (and maybe this should have docymentation as well)

PR #4370, opened 2024-10-01
Don't wrap `throw` in closure

HypeMC:
I believe this was done because in versions of PHP prior to 8, the `throw` keyword was a statement and not an expression. Since Twig now requires PHP 8+, this should no longer be necessary.


fabpot:
That's already done in 4.x and I don't want to redo all the work that has been done in 4.x to remove support/upgrade the code base for PHP8+.

PR #4371, opened 2024-10-01
Allow running CI on all branches

HypeMC:
This allows contributors to verify if the CI passes on their forks before opening a PR.


fabpot:
Thank you @HypeMC.

PR #4372, opened 2024-10-02
Clarify docs for some operators

### LINKED ISSUE #3922: The documentation for the `<=>` operator is missing
Except if I miss something, there is no documentation for the `<=>` operator even though it is mentioned in both Twig 2 and 3 documentation:

https://twig.symfony.com/doc/2.x/templates.html#expressions
https://twig.symfony.com/doc/3.x/templates.html#expressions

fabpot:
Closes #3922

### DIFF
--- a/doc/templates.rst
+++ b/doc/templates.rst
@@ -727,35 +727,26 @@ You can combine multiple expressions with the following operators:
 Comparisons
 ~~~~~~~~~~~
 
-The following comparison operators are supported in any expression: ``==``,
-``!=``, ``<``, ``>``, ``>=``, and ``<=``.
+The following mathematical comparison operators are supported in any
+expression: ``==``, ``!=``, ``<``, ``>``, ``>=``, and ``<=``.
 
-Check if a string ``starts with`` or ``ends with`` another string:
+Spaceship Operator
+~~~~~~~~~~~~~~~~~~
 
-.. code-block:: twig
-
-    {% if 'Fabien' starts with 'F' %}
-    {% endif %}
-
-    {% if 'Fabien' ends with 'n' %}
-    {% endif %}
-
-Check that a string contains another string via the containment operator (see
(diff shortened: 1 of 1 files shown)


doc/templates.rst
```diff
@@ -727,31 +727,15 @@ You can combine multiple expressions with the following operators:
...
-    `regular expressions`_:
-
-    .. code-block:: twig
-
-        {% if phone matches '/^[\\d\\.]+$/' %}
-        {% endif %}
+Iterator Operators
+~~~~~~~~~~~~~~~~~~
 
 Check that a sequence or a mapping ``has every`` or ``has some`` of its
```
stof: Maybe we should document that `has every` returns `true` for an empty iterable and `has some` returns `false`

stof: should this talk about iterable instead of `sequence or mapping` ?

doc/templates.rst
```diff
@@ -727,31 +727,15 @@ You can combine multiple expressions with the following operators:
 Comparisons
 ~~~~~~~~~~~
 
-The following comparison operators are supported in any expression: ``==``,
-``!=``, ``<``, ``>``, ``>=``, and ``<=``.
+The following mathematical comparison operators are supported in any
+expression: ``==``, ``!=``, ``<``, ``>``, ``>=``, ``<=``, and ``<=>``.
```
stof: should the spaceship operator be considered a mathematical operator ?

doc/templates.rst
fabpot: Done

fabpot: Done

doc/templates.rst
fabpot: I've created a specific section for the spaceship operator.

stof (approved):

PR #4376, opened 2024-10-04
Move sandbox docs to its own chapter in the docs

fabpot:
I've also added more information about potential security issues when exposing something that takes a callable as an argument.

### DIFF
--- a/doc/sandbox.rst
+++ b/doc/sandbox.rst
@@ -1,0 +1,95 @@
+Twig Sandbox
+============
+
+The ``sandbox`` extension can be used to evaluate untrusted code.
+
+Registering the Sandbox
+-----------------------
+
+Register the ``SandboxExtension`` extension via the ``addExtension()`` method::
+
+    $twig->addExtension(new \Twig\Extension\SandboxExtension($policy));
+
+Configuring the Sandbox Policy
+------------------------------
+
+The sandbox security is managed by a policy instance, which must be passed to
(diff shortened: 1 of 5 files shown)


doc/sandbox.rst
```diff
@@ -0,0 +1,87 @@
...
+    $twig->addExtension(new \Twig\Extension\SandboxExtension($policy));
+
+Configuring the Sandbox Policy
+------------------------------
+
+The sandbox security is managed by a policy instance, which must be passed to
+the ``SandboxExtension`` constructor.
+
+By default, Twig comes with one policy class: ``\Twig\Sandbox\SecurityPolicy``.
+This class allows you to white-list some tags, filters, functions, but also
```
stof:
```suggestion
This class allows you to allow-list some tags, filters, functions, but also
```

doc/sandbox.rst
```diff
@@ -0,0 +1,87 @@
...
+You can sandbox all templates by passing ``true`` as the second argument of
+the extension constructor::
+
+    $sandbox = new \Twig\Extension\SandboxExtension($policy, true);
+
+Accepting Callables Arguments
+-----------------------------
+
+The Twig sandbox allows you to configure which functions, filters, tests and
+dot operations are allowed. Many these calls can accept arguments. As these
```
stof:
```suggestion
dot operations are allowed. Many of these calls can accept arguments. As these
```

PR #4377, opened 2024-10-04
Deprecate passing a string or an array to Twig callable arguments accepting arrow functions (pass a Closure)

fabpot:
Instead of restricting callable arguments only in sandbox mode, let's deprecate using functions/arrays as callables.


stof (approved):
This is good:

- Twig already uses a Closure when using an arrow function
- If passing a callable coming from PHP code, using the first-class callable syntax is an easy way to get a Closure for any callable.

PR #4378, opened 2024-10-04
Allow arrow functions everywhere

### LINKED ISSUE #3192: Arrow function are not allowed outside filters
Arrow functions currently can only be used in a filter as per ExpressionParser.
This prevents multiple functionalities such as
- passing arrows to an include context so it can forward it to an actual filter;
- passing arrows to a macro or a function;
- defining and setting a callable so it can be used in subsequent expressions;

### LINKED ISSUE #3402: How to use arrow functions as callable?
I'd be really great if Twig arrow functions can be passed as arguments.

e.g. [Laravel collections unique method](https://laravel.com/docs/8.x/collections#method-unique) accepts a callable.

I'd like to do something like this:

```twig
{{ mycollection.unique(item => item.id) }}
```

Is that possible? Or an RFC?

fabpot:
Closes #3192
Closes #3402

Arrow functions are now supported for all Twig callables (filters, functions, and tests), macro arguments, and method call arguments.

Arrow functions can also be stored in variables.

### DIFF
--- a/doc/templates.rst
+++ b/doc/templates.rst
@@ -960,6 +960,32 @@ The following operators don't fit into any of the other categories:
     Support for expanding the arguments of a function call was introduced in
     Twig 3.15.
 
+* ``=>``: The arrow operator allows the creation of functions. A function is
+  made of arguments (use parentheses for multiple arguments) and an arrow
+  (``=>``) followed by an expression to execute. The expression has access to
+  all passed arguments. Arrow functions are supported as arguments for filters,
+  functions, tests, macros, and method calls.
+
+  For instance, the built-in ``map``, ``reduce``, ``sort``, ``filter``, and
+  ``find`` filters accept arrow functions as arguments:
+
+  .. code-block:: twig
+
+      {{ people|map(p => p.first_name)|join(', ') }}
+
+  Arrow functions can be stored in variables:
+
+  .. code-block:: twig
+
+      {% set first_name_fn = (p) => p.first_name %}
+
+      {{ people|map(first_name_fn)|join(', ') }}
+
+  .. versionadded:: 3.15
+
+    Arrow function support for functions, macros, and method calls was added in
+    Twig 3.15 (filters and tests were already supported).
+
 Operators
 ~~~~~~~~~
 
--- a/src/ExpressionParser.php
+++ b/src/ExpressionParser.php
@@ -87,9 +87,13 @@ class ExpressionParser
         }
     }
 
-    public function parseExpression($precedence = 0, $allowArrow = false)
+    public function parseExpression($precedence = 0)
     {
-        if ($allowArrow && $arrow = $this->parseArrow()) {
+        if (func_num_args() > 1) {
+            trigger_deprecation('twig/twig', '3.15', 'Passing a second argument ($allowArrow) to "%s()" is deprecated.', __METHOD__);
+        }
+
+        if ($arrow = $this->parseArrow()) {
             return $arrow;
         }
 
@@ -108,7 +112,7 @@ class ExpressionParser
             } else {
                 $previous = $this->setDeprecationCheck(true);
                 try {
-                    $expr1 = $this->parseExpression(self::OPERATOR_LEFT === $op['associativity'] ? $op['precedence'] + 1 : $op['precedence'], true);
+                    $expr1 = $this->parseExpression(self::OPERATOR_LEFT === $op['associativity'] ? $op['precedence'] + 1 : $op['precedence']);
                 } finally {
                     $this->setDeprecationCheck($previous);
                 }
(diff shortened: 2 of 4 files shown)


src/ExpressionParser.php
```diff
@@ -238,7 +243,12 @@ private function getPrimary(): AbstractExpression
             return $this->parsePostfixExpression($expr);
         } elseif ($token->test(Token::PUNCTUATION_TYPE, '(')) {
             $this->parser->getStream()->next();
-            $expr = $this->parseExpression(deprecationCheck: false)->setExplicitParentheses();
+            $this->disableDeprecationCheck();
+            try {
+                $expr = $this->parseExpression()->setExplicitParentheses();
+            } finally {
+                $this->enableDeprecationCheck();
```
stof: shouldn't this rather restore the previous state of the check instead ?

src/ExpressionParser.php
```diff
@@ -105,7 +110,7 @@ public function parseExpression($precedence = 0, $allowArrow = false, bool $depr
             } elseif (isset($op['callable'])) {
                 $expr = $op['callable']($this->parser, $expr);
             } else {
-                $expr1 = $this->parseExpression(self::OPERATOR_LEFT === $op['associativity'] ? $op['precedence'] + 1 : $op['precedence'], true);
+                $expr1 = $this->parseExpression(self::OPERATOR_LEFT === $op['associativity'] ? $op['precedence'] + 1 : $op['precedence']);
```
stof: what if this is called from a place where `$this->deprecationCheck` was set to `false` ? Should this be a place where we turn it on again ? Wouldn't this miss deprecation checks for operators used in an parenthesized expression in a larger expression ?

src/ExpressionParser.php
fabpot: Correct

src/ExpressionParser.php
fabpot: Right, I've also updated this part.

fabpot: I found more problems, fixed in #4382

doc/templates.rst
```diff
@@ -960,6 +960,32 @@ The following operators don't fit into any of the other categories:
...
+
+  .. code-block:: twig
+
+      {% set first_name_fn = (p) => p.first_name %}
+
+      {{ people|map(first_name_fn)|join(', ') }}
+
+  .. versionadded:: 3.15
+
+    Arrow function support for functions, tests, macros, and method calls was
```
xabbuh:
```suggestion
    Arrow function support for filters, functions, tests, macros, and method calls was
```

xabbuh:
```suggestion
    Arrow function support for filters, functions, tests, macros, and method calls was
```

doc/templates.rst
```diff
@@ -960,6 +960,32 @@ The following operators don't fit into any of the other categories:
     Support for expanding the arguments of a function call was introduced in
     Twig 3.15.
 
+* ``=>``: The arrow operator allows the creation of functions. A function is
+  made of arguments (use parentheses for multiple arguments) and an arrow
+  (``=>``) followed by an expression to execute. The expression has access to
+  all passed arguments. Arrow functions are supported as arguments for Twig
+  callables, macros, and method calls.
```
xabbuh:
```suggestion
    filters, functions, tests, macros, and method calls.
```

doc/templates.rst
fabpot: Support for filters (and tests) was already there. I've updated the comment to make it clear.

fabpot:
Rebased

xabbuh (approved):

Lorenzschaef:
I updated twig for this feature. However, I found that there is one piece missing, to what I was actually looking for. I can store an arrow function in a variable and pass it around, but I can't _call_ it.

My use case:

I have a reusable component/partial, that displays a list of documents as links. Depending on the context, I want the link's URL to be different, so I need to pass a function that generates the link for each item.

What I would expect to work:

`<a href="{{ buildUrl(document) }}">`

My workaround that does work:

`<a href="{{ [document]|map(buildUrl)|first }}>`

Are there any plans to support this?

stof:
The way to call it (in 3.17.1 and newer) is `buildUrl.__invoke(document)`

Twig does not support using `buildUrl(document)` as this would be ambiguous whether `buildUrl` refers to a Twig function or a Twig variable (forcing to remove all validation of function existence from the compile time of templates)

Lorenzschaef:
Ok, it makes sense to not allow the normal function call syntax then. But isn't the fact that arrow functions have an __invoke() mehtod an (undocumented) implementation detail? I wouldn't really feel comfortable relying on that in my code. I think an `invoke` filter would be a good solution to this.

fabpot:
> Ok, it makes sense to not allow the normal function call syntax then. But isn't the fact that arrow functions have an __invoke() mehtod an (undocumented) implementation detail? I wouldn't really feel comfortable relying on that in my code. I think an `invoke` filter would be a good solution to this.

Having a dedicated filter makes sense to me. Would you like to work on this @Lorenzschaef ?

Lorenzschaef:
@fabpot I can do that, adding the filter shouldn't be hard.

PR #4379, opened 2024-10-04
Replace `var_` with `print_` in Lexer / Token ?

smnandre:
While playing with the Lexer, using `Token::typeToEnglish()` and reading the docs, this change _felt_ more coherent to me. By keeping the old constants, I believe it will be BC-safe.

Please let me know what you think (especially if I'm missing something here..).

### DIFF
--- a/doc/internals.rst
+++ b/doc/internals.rst
@@ -32,8 +32,8 @@ The lexer tokenizes a template source code into a token stream (each token is
 an instance of ``\Twig\Token``, and the stream is an instance of
 ``\Twig\TokenStream``). The default lexer recognizes 15 different token types:
 
-* ``\Twig\Token::BLOCK_START_TYPE``, ``\Twig\Token::BLOCK_END_TYPE``: Delimiters for blocks (``{% %}``)
-* ``\Twig\Token::VAR_START_TYPE``, ``\Twig\Token::VAR_END_TYPE``: Delimiters for variables (``{{ }}``)
+* ``\Twig\Token::BLOCK_START_TYPE``, ``\Twig\Token::BLOCK_END_TYPE``: Delimiters for blocks statements (``{% %}``)
+* ``\Twig\Token::PRINT_START_TYPE``, ``\Twig\Token::PRINT_END_TYPE``: Delimiters for print statements (``{{ }}``)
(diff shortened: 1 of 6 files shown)


doc/internals.rst
```diff
@@ -32,8 +32,8 @@ The lexer tokenizes a template source code into a token stream (each token is
 an instance of ``\Twig\Token``, and the stream is an instance of
 ``\Twig\TokenStream``). The default lexer recognizes 15 different token types:
 
-* ``\Twig\Token::BLOCK_START_TYPE``, ``\Twig\Token::BLOCK_END_TYPE``: Delimiters for blocks (``{% %}``)
-* ``\Twig\Token::VAR_START_TYPE``, ``\Twig\Token::VAR_END_TYPE``: Delimiters for variables (``{{ }}``)
+* ``\Twig\Token::BLOCK_START_TYPE``, ``\Twig\Token::BLOCK_END_TYPE``: Delimiters for blocks statements (``{% %}``)
```
jdreesen:
```suggestion
* ``\Twig\Token::BLOCK_START_TYPE``, ``\Twig\Token::BLOCK_END_TYPE``: Delimiters for block statements (``{% %}``)
```

smnandre: Good catch, thanks!

src/Lexer.php
```diff
@@ -63,10 +68,15 @@ public function __construct(Environment $env, array $options = [])
     {
         $this->env = $env;
 
+        if (isset($options['tag_variable'])) {
+            trigger_deprecation('twig/twig', '3.XX', 'The "tag_variable" option is deprecated, use "tag_print" instead.');
```
fabpot, line 72:
```suggestion
            trigger_deprecation('twig/twig', '3.15', 'The "tag_variable" option is deprecated, use "tag_print" instead.');
```

fabpot (commented):
About the name: as we're using `yield` now, I'm wondering if it should be `OUTPUT_START_TYPE` or maybe `OUT_START_TYPE`?

fabpot:
Jinja2 uses VARIABLE like us: https://github.com/pallets/jinja/blob/main/src/jinja2/lexer.py#L176

smnandre:
Well let's close then

PR #4382, opened 2024-10-06
Fix deprecation dep feature

fabpot:


src/ExpressionParser.php
```diff
@@ -147,7 +150,7 @@ private function triggerPrecedenceDeprecations(AbstractExpression $expr, Token $
                     continue;
                 }
                 if ($node->hasAttribute('operator') && $operatorName === $node->getAttribute('operator')) {
-                    trigger_deprecation($change->getPackage(), $change->getVersion(), \sprintf('Add explicit parentheses around the "%s" unary operator to avoid behavior change in the next major version as its precedence will change in "%s" at line %d.', $target, $this->parser->getStream()->getSourceContext()->getName(), $token->getLine()));
+                    trigger_deprecation($change->getPackage(), $change->getVersion(), \sprintf('Add explicit parentheses around the "%s" unary operator to avoid behavior change in the next major version as its precedence will change in "%s" at line %d.', $target, $this->parser->getStream()->getSourceContext()->getName(), $node->getTemplateLine()));
```
fabpot, line 153: I added some tests to make sure we report the correct line.

src/ExpressionParser.php
```diff
@@ -105,19 +106,21 @@ public function parseExpression($precedence = 0, $allowArrow = false, bool $depr
             } elseif (isset($op['callable'])) {
                 $expr = $op['callable']($this->parser, $expr);
             } else {
-                $expr1 = $this->parseExpression(self::OPERATOR_LEFT === $op['associativity'] ? $op['precedence'] + 1 : $op['precedence'], true);
+                $previous = $this->setDeprecationCheck(true);
```
fabpot, line 109: I added a test that would fail without this.

src/ExpressionParser.php
```diff
@@ -105,19 +106,21 @@ public function parseExpression($precedence = 0, $allowArrow = false, bool $depr
...
             }
 
             $expr->setAttribute('operator', 'binary_'.$token->getValue());
 
-            $previousToken = $token;
-            $token = $this->parser->getCurrentToken();
-        }
+            $this->triggerPrecedenceDeprecations($expr, $token);
 
-        if ($deprecationCheck) {
```
fabpot, line 119: Not needed as the `setExplicitParentheses()` is set before the check here. The `deprecationCheck` is only useful for unary operators.

PR #4392, opened 2024-10-11
Improve deprecation messages

fabpot:
Refs [in string should](https://github.com/twigphp/Twig/pull/4344#issuecomment-2406110007)
cc @smnandre


ruudk (approved):
Much better!

PR #4397, opened 2024-10-14
Check reserved names in TempNameExpression

fabpot:


src/ExpressionParser.php
```diff
@@ -689,20 +689,24 @@ public function parseFilterExpressionRaw($node)
...
-     * @param bool $definition     Whether we are parsing arguments for a function (or macro) definition
-     *
      * @return Node
      *
      * @throws SyntaxError
      */
-    public function parseArguments($namedArguments = false, $definition = false)
+    public function parseArguments()
     {
+        $namedArguments = true;
```
stof: this changes the default value, which could be unexpected if some places where reusing this without expecting named arguments.

fabpot: Fixed

src/ExpressionParser.php
```diff
@@ -945,10 +950,55 @@ private function setDeprecationCheck(bool $deprecationCheck): bool
...
         $arguments = new ArrayExpression([], $line);
-        foreach ($this->parseArguments(true) as $k => $n) {
+        foreach ($this->parseOnlyArguments() as $k => $n) {
             $arguments->addElement($n, new TempNameExpression($k, $line));
         }
 
         return $arguments;
     }
+
+    public function parseOnlyArguments()
```
fabpot, line 960: This method will be updated again in another PR (with a specialized return Node).

PR #4398, opened 2024-10-16
Rename Node classes related to variables

fabpot:
Whenever I work on Twig internals, it's always complicated to reason about variable names, probably because the class names are confusing. This PR is an attempt to find "better" and more explicit names.

This PR does the following renaming:

 * `NameExpression` to `Variable\ContextVariable`
    Represents the value of a context variable like `$context[VAR] ?? null`

 * `AssignNameExpression` to `Variable\AssignContextVariable`
    Represents a context variable assignment like in `$context[VAR] = `

 * `TempNameExpression` to `Variable\LocalVariable`
    Represents a "private" local variable like `$_l111`

### DIFF
--- a/src/Node/Expression/TempNameExpression.php
+++ b/src/Node/Expression/TempNameExpression.php
@@ -18,14 +18,18 @@ class TempNameExpression extends AbstractExpression
 {
     public const RESERVED_NAMES = ['varargs', 'context', 'macros', 'blocks', 'this'];
 
-    public function __construct(string|int $name, int $lineno)
+    public function __construct(string|int|null $name, int $lineno)
     {
         // All names supported by ExpressionParser::parsePrimaryExpression() should be excluded
-        if (\in_array(strtolower($name), ['true', 'false', 'none', 'null'])) {
+        if ($name && \in_array(strtolower($name), ['true', 'false', 'none', 'null'])) {
             throw new SyntaxError(\sprintf('You cannot assign a value to "%s".', $name), $lineno);
         }
 
-        if (is_int($name) || ctype_digit($name)) {
+        if (self::class === static::class) {
+            trigger_deprecation('twig/twig', '3.15', 'The "%s" class is deprecated.', self::class);
+        }
+
+        if (null !== $name && (is_int($name) || ctype_digit($name))) {
             $name = (int) $name;
         } elseif (in_array($name, self::RESERVED_NAMES)) {
             $name = '_'.$name.'_';
@@ -36,6 +40,10 @@ class TempNameExpression extends AbstractExpression
 
     public function compile(Compiler $compiler): void
     {
+        if (null === $this->getAttribute('name')) {
+            $this->setAttribute('name', \sprintf('_l%d', $compiler->getVarName()));
+        }
+
         $compiler->raw('$'.$this->getAttribute('name'));
     }
 }
--- a/src/Node/Expression/Variable/AssignContextVariable.php
+++ b/src/Node/Expression/Variable/AssignContextVariable.php
@@ -1,0 +1,18 @@
+<?php
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Twig\Node\Expression\Variable;
+
+use Twig\Node\Expression\AssignNameExpression;
+
+final class AssignContextVariable extends AssignNameExpression
+{
+}
--- a/src/Parser.php
+++ b/src/Parser.php
@@ -18,6 +18,7 @@ use Twig\Node\BlockReferenceNode;
 use Twig\Node\BodyNode;
 use Twig\Node\EmptyNode;
 use Twig\Node\Expression\AbstractExpression;
+use Twig\Node\Expression\Variable\TemplateVariable;
 use Twig\Node\MacroNode;
 use Twig\Node\ModuleNode;
 use Twig\Node\Node;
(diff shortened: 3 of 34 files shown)


src/Node/Expression/Variable/AssignContextVariable.php
```diff
@@ -0,0 +1,21 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Twig\Node\Expression\Variable;
+
+use Twig\Node\Expression\AssignNameExpression;
+
+/**
+ * @final
```
stof: The new classes could be final directly instead of `@final`

src/Node/Expression/TempNameExpression.php
```diff
@@ -13,19 +13,28 @@
...
             throw new SyntaxError(\sprintf('You cannot assign a value to "%s".', $name), $lineno);
         }
 
-        if (is_int($name) || ctype_digit($name)) {
+        if (self::class === static::class) {
+            trigger_deprecation('twig/twig', '3.15', 'The "%s" class is deprecated, use "%s" instead.', self::class, LocalVariable::class);
+        }
+
+        if (null === $name) {
+            $name = \sprintf('_l%d', self::$salt++);
```
stof: would such static state still allow to have reproducible template compilation ? I doubt it as compiling multiple templates in the same process will produce different local variable names.
The parser `varNameSalt` was introduced as a member variable precisely when working on such reproducible compilation, because that state is reset for each parsed template.

fabpot: I've moved the creation of the name at compilation time.

src/Parser.php
```diff
@@ -60,7 +60,9 @@ public function getEnvironment(): Environment
 
     public function getVarName(): string
     {
-        return \sprintf('__internal_parse_%d', $this->varNameSalt++);
+        trigger_deprecation('twig/twig', '3.15', 'The "%s()" method is deprecated.', __METHOD__);
+
+        return \sprintf('_n%d', $this->varNameSalt++);
```
stof: why changing the name of the variable being generated in this deprecated method ?

src/TokenParser/ApplyTokenParser.php
```diff
@@ -32,11 +32,7 @@ final class ApplyTokenParser extends AbstractTokenParser
     public function parse(Token $token): Node
     {
         $lineno = $token->getLine();
-        $name = $this->parser->getVarName();
-
-        $ref = new TempNameExpression($name, $lineno);
-        $ref->setAttribute('always_defined', true);
```
stof, line 38: was this unnecessary ?

src/TokenParser/FromTokenParser.php
```diff
@@ -50,7 +51,7 @@ public function parse(Token $token): Node
 
         $stream->expect(Token::BLOCK_END_TYPE);
 
-        $var = new AssignNameExpression($this->parser->getVarName(), $token->getLine());
+        $var = new LocalVariable(null, $token->getLine());
```
stof: is it actually safe to migrate from a context variable to a local variable ? What happens for usages that are inside nested blocks, which won't have access to the same PHP local variables ?

src/Node/Expression/Variable/AssignContextVariable.php
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Twig\Node\Expression\Variable;
+
+use Twig\Node\Expression\AssignNameExpression;
+
+final class AssignContextVariable extends AssignNameExpression
```
stof, line 16: should AssignContextVariable actually be an expression ? AFAIK, we don't support assigning context variables in arbitrary expressions (while the AST allows representing such case)

src/Parser.php
fabpot: It makes reading the output code easier to read. But you're right, it's not needed.

src/TokenParser/ApplyTokenParser.php
fabpot: TempNameExpression does not support this attribute anyway.

src/TokenParser/FromTokenParser.php
fabpot: The `ImportNode` class only uses the `name` attribute from `$var` (the node is never compiled). So, here, we're only interested in the name, which is then going to be used as `$macros[NAME]`, which is propagated in nested blocks. See #4402 for a fix where we remove the wrong usage of the `AssignNameExpression` class.

src/Node/Expression/Variable/AssignContextVariable.php
fabpot: We might support `{% if foo = some_expression %}...{% endif %}` at some point (I'm working on a new expression parser that would, at least, make it easy to support it).

fabpot:
Rebased

src/Node/Expression/Variable/TemplateVariable.php
```diff
@@ -12,12 +12,19 @@
 namespace Twig\Node\Expression\Variable;
 
 use Twig\Compiler;
-use Twig\Node\Expression\NameExpression;
+use Twig\Node\Expression\TempNameExpression;
 
-final class TemplateVariable extends NameExpression
+/**
+ * @final
```
xabbuh, line 18: Is this intended? The added `GlobalTemplateVariable` class extends this class. This currently makes the Symfony tests fail:

> ```
> Remaining indirect deprecation notices (1)
> 
>   1x: The "Twig\Node\Expression\Variable\TemplateVariable" class is considered final. It may change without further notice as of its next major version. You should not extend it from "Twig\Node\Expression\Variable\GlobalTemplateVariable".
>     1x in ProfilerControllerTest::testPanelActionWithLatestToken from Symfony\Bundle\WebProfilerBundle\Tests\Controller
> ```

fabpot: Fixed in #4408

PR #4399, opened 2024-10-17
Call For recurseFunc inline

ruudk:
After https://github.com/phpstan/phpstan-src/pull/3554 this now makes it easier for PHPStan to understands what's going on.

This will allow TwigStan to handle for loops in v4.


ruudk:
@fabpot This one is ready for review.

fabpot:
Thank you @ruudk.

PR #4400, opened 2024-10-17
fix function is already exist

jimmydupre:
added function_exists on multiple files because if twig is stocked in include_path on php config and another place. We risk a conflict.
Thank you

### DIFF
--- a/src/Resources/core.php
+++ b/src/Resources/core.php
@@ -16,12 +16,14 @@ use Twig\Extension\CoreExtension;
  * @internal
  *
  * @deprecated since Twig 3.9
- */
-function twig_cycle($values, $position)
-{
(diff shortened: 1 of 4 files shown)


fabpot:
If these functions already exist, I suppose that this should be fixed instead of hiding the problem by checking for existence (which would not guarantee that the implementation is the same).

jimmydupre:
In my case, twig is incorporated into the include_path for general use. This is an unusual usage. However, in the virtualhost of the main domain there is a subproject (CMS) that uses twig. This causes a conflict because the functions are already declared.

fabpot:
I understand your (unusual) setup. But this is not something that we will support for the reason explained above.

PR #4401, opened 2024-10-18
reduce the deprecation noise when more than one argument is passed

xabbuh:
When more than one argument is passed to parseArguments(), one deprecation is triggered for each argument. Since passing any argument is deprecated the number of triggered deprecations can be reduced for more than one argument being passed.

### DIFF
--- a/src/ExpressionParser.php
+++ b/src/ExpressionParser.php
@@ -697,15 +697,11 @@ class ExpressionParser
     {
         $namedArguments = false;
         $definition = false;
-        if (func_num_args() > 2) {
-            trigger_deprecation('twig/twig', '3.15', 'Passing a third argument ($allowArrow) to "%s()" is deprecated.', __METHOD__);
-        }
         if (func_num_args() > 1) {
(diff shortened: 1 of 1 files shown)


src/ExpressionParser.php
```diff
@@ -697,15 +697,8 @@ public function parseArguments()
     {
         $namedArguments = false;
         $definition = false;
-        if (func_num_args() > 2) {
-            trigger_deprecation('twig/twig', '3.15', 'Passing a third argument ($allowArrow) to "%s()" is deprecated.', __METHOD__);
-        }
-        if (func_num_args() > 1) {
-            trigger_deprecation('twig/twig', '3.15', 'Passing a second argument ($definition) to "%s()" is deprecated.', __METHOD__);
-            $definition = func_get_arg(1);
```
stof, line 705: this handling of `$definition` is lost

xabbuh: oops, fixed

fabpot:
Good catch, thanks @xabbuh.

PR #4402, opened 2024-10-20
Rework macros handling

fabpot:


tests/Fixtures/tests/defined_for_macros.test
```diff
@@ -1,41 +1,84 @@
...
+    {{~ baz is not defined ? 'OK' : 'KO' }}
+    {{~ baz() is not defined ? 'OK' : 'KO' }}
 
-{% if foo is not defined -%}
-    OK
-{% endif %}
+    {{~ _self.hello is defined ? 'OK' : 'KO' }}
+    {{~ _self.hello() is defined ? 'OK' : 'KO' }}
+
+    {{~ _self.bar is not defined ? 'OK' : 'KO' }}
```
fabpot, line 36: This is currently not supported.

tests/Fixtures/tests/defined_for_macros.test
```diff
@@ -1,41 +1,80 @@
...
-{% if hello is defined -%}
-    OK
-{% endif %}
+    {{~ _self.bar is not defined ? 'OK' : 'KO' }}
+    {{~ _self.bar() is not defined ? 'OK' : 'KO' }}
 
-{% if bar is not defined -%}
-    OK
-{% endif %}
+    {{~ _self.lol is defined ? 'OK' : 'KO' }}
```
fabpot, line 39: This is currently not supported

tests/Fixtures/tests/defined_for_macros.test
```diff
@@ -1,41 +1,80 @@
...
+    {{~ lol is defined ? 'OK' : 'KO' }}
+    {{~ lol() is defined ? 'OK' : 'KO' }}
+
+    {{~ baz is not defined ? 'OK' : 'KO' }}
+    {{~ baz() is not defined ? 'OK' : 'KO' }}
 
-{% if macros.foo is not defined -%}
-    OK
-{% endif %}
+    {{~ _self.hello is defined ? 'OK' : 'KO' }}
```
fabpot, line 33: This is currently not supported.

src/NodeVisitor/MacroAutoImportNodeVisitor.php
```diff
@@ -28,41 +21,18 @@
...
     public function leaveNode(Node $node, Environment $env): Node
     {
-        if ($node instanceof ModuleNode) {
-            $this->inAModule = false;
-            if ($this->hasMacroCalls) {
-                $node->getNode('constructor_end')->setNode('_auto_macro_import', new ImportNode(new NameExpression('_self', 0), new AssignNameExpression('_self', 0), 0));
-            }
-        } elseif ($this->inAModule) {
-            if (
-                $node instanceof GetAttrExpression
```
fabpot, line 53: This code has been moved to `ExpressionParser` where it belongs.

src/Template.php
```diff
@@ -478,6 +478,35 @@ public function yieldParentBlock($name, array $context, array $blocks = []): ite
         }
     }
 
+    protected function hasMacro(string $name, array $context): bool
+    {
+        if (method_exists($this, $name)) {
+            return true;
+        }
+
+        if (!$parent = $this->getParent($context)) {
```
fabpot, line 487: The current code didn't have a look at parent templates for the "is defined" test.

src/TokenParser/FromTokenParser.php
```diff
@@ -50,11 +50,11 @@ public function parse(Token $token): Node
 
         $stream->expect(Token::BLOCK_END_TYPE);
 
-        $var = new AssignNameExpression($this->parser->getVarName(), $token->getLine());
```
fabpot, line 53: Using `AssignNameExpression` here is wrong as we're not storing macros in `$context`, but more importantly, we never compile this node, it's only used for its `name` attribute.

src/ExpressionParser.php
```diff
@@ -531,11 +532,7 @@ public function parsePostfixExpression($node)
     public function getFunctionNode($name, $line)
     {
         if (null !== $alias = $this->parser->getImportedSymbol('function', $name)) {
-            $arguments = $this->createArguments($line);
-            $node = new MethodCallExpression($alias['node'], $alias['name'], $arguments, $line);
-            $node->setAttribute('safe', true);
```
fabpot, line 536: The `safe` attribute is always set to `true` in all usages, so I've removed the `safe` attribute (see `SafeAnalysisNodeVisitor`).

src/Node/Expression/Variable/TemplateVariable.php
```diff
@@ -0,0 +1,31 @@
...
+
+final class TemplateVariable extends NameExpression
+{
+    public function compile(Compiler $compiler): void
+    {
+        if ('_self' === $this->getAttribute('name')) {
+            $compiler->raw('$this');
+        } else {
+            $compiler
+                ->raw('$macros[')
```
fabpot, line 25: Maybe we can rename this variable as it stores templates, not macros, maybe `$templates` or `$templateAliases`?

src/Extension/CoreExtension.php
```diff
@@ -1286,6 +1285,8 @@ public static function capitalize(string $charset, $string): string
 
     /**
      * @internal
+     *
+     * to be removed in 4.0
```
stof, line 1289: As it is internal, can't it be removed immediately ?

src/Node/Expression/Variable/TemplateVariable.php
stof: If those templates are only the ones used for macros, maybe `macroTemplates` ?

src/NodeVisitor/MacroAutoImportNodeVisitor.php
```diff
@@ -28,41 +21,18 @@
  */
 final class MacroAutoImportNodeVisitor implements NodeVisitorInterface
 {
-    private $inAModule = false;
-    private $hasMacroCalls = false;
+    public function __construct()
+    {
+        trigger_deprecation('twig/twig', '3.15', 'The "%s" class is deprecated.', __CLASS__);
```
stof: As the class is internal, do we need to go through a deprecation ?

src/NodeVisitor/MacroAutoImportNodeVisitor.php
```diff
@@ -28,41 +21,18 @@
...
-
         return $node;
     }
 
     public function leaveNode(Node $node, Environment $env): Node
     {
-        if ($node instanceof ModuleNode) {
-            $this->inAModule = false;
-            if ($this->hasMacroCalls) {
-                $node->getNode('constructor_end')->setNode('_auto_macro_import', new ImportNode(new NameExpression('_self', 0), new AssignNameExpression('_self', 0), 0));
```
stof, line 49: Where is this going ?

src/Template.php
```diff
@@ -478,6 +478,35 @@ public function yieldParentBlock($name, array $context, array $blocks = []): ite
...
+    }
+
+    protected function getTemplateForMacro(string $name, array $context, int $line, Source $source): Template
+    {
+        if (method_exists($this, $name)) {
+            return $this;
+        }
+
+        $parent = $this;
+        while ($parent = $parent->getParent($context)) {
```
stof, line 501: Was the current code inheriting macros ?

src/Extension/CoreExtension.php
fabpot: It's used by the deprecated `MethodCallExpression` class, which is public.

src/NodeVisitor/MacroAutoImportNodeVisitor.php
fabpot: Sure, I was just trying to be nice :) But I suppose you're right.

src/NodeVisitor/MacroAutoImportNodeVisitor.php
fabpot: It's here: https://github.com/twigphp/Twig/pull/4402/files#diff-ad87623d49894167ed5e4ecd1814164bdbe5c4c250e6b2ad9da415edd22e8e85R965

src/Template.php
fabpot: Yes, see `CoreExtension::callMacro()` code https://github.com/twigphp/Twig/pull/4402/files#diff-29e85e483c6ec4a9c2fd144820b6722c86df60d54175b355d85e806253313c1aR1293-R1302

src/NodeVisitor/MacroAutoImportNodeVisitor.php
stof: but this does not add anymore the auto-import at the end of the constructor AFAICT.

fabpot: Because it's not needed anymore

src/Node/Expression/Variable/TemplateVariable.php
fabpot: I've just had a closer look and it can break extensions, like cache, so let's keep it like it is today for now.

PR #4403, opened 2024-10-21
Documentation for types tag uses Twig types in examples instead of PHP

drjayvee:
Specifically, "bool" => "boolean" and "int" => "number".

This aligns the `types` documentation with templates.rst. See #4362

Thanks, @alexander-schranz!

### DIFF
--- a/doc/tags/types.rst
+++ b/doc/tags/types.rst
@@ -9,14 +9,13 @@ The ``types`` tag declares the types of template variables.
 
 To do this, specify a :ref:`mapping <twig-expressions>` of names to their types as strings.
 
-Here is how to declare that ``is_correct`` is a boolean, while ``score`` is an
-integer (see note below):
+Here is how to declare that ``is_correct`` is a boolean, while ``score`` is a number (see note below):
 
 .. code-block:: twig
 
     {% types {
-        is_correct: 'bool',
-        score: 'int',
+        is_correct: 'boolean',
+        score: 'number',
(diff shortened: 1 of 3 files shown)


ruudk:
`boolean` is a valid PHPStan type, but `number` is not. 
See https://phpstan.org/writing-php-code/phpdoc-types

Can we maybe update the example, to not use a `number` instead? 

[TwigStan](https://github.com/twigstan/twigstan) will only accept PHPStan types, as it's a PHPStan analyzer.

Making this `types` tag too strict up front, only limits its usage in the wild.

ruudk:
The documentation is not up to date. PHPStan _does_ properly understand `number` 🎉 
I'll fix the docs.


See https://phpstan.org/r/24895f82-975a-4e08-b7b1-5da29d058ddd

ruudk:
Done:
- https://github.com/phpstan/phpstan/pull/11882

smnandre:
Do you think it's worth it to update TypesTest.php and TypesTokenParser.php ?

fabpot:
Thank you @drjayvee.

fabpot:
Not sure why Github doesn't understand this one was merged 🤷

stof:
@fabpot this happens when we squash commits but we are not allowed to push the squashed commits back to the branch from which the PR is submitted.
There is no way to mark a PR as merged when performing a squash or rebase entirely locally.

PR #4404, opened 2024-10-22
Unset `_parent` when `for` loop is done

ruudk:
By not doing this, PHPStan has a really hard time dealing with loops in loops.

See https://github.com/phpstan/phpstan/issues/11890

This feels like a breaking change for v3. But since `_parent` variable is not documented, it might be OK?

/cc @stof @fabpot

### DIFF
--- a/src/Node/ForNode.php
+++ b/src/Node/ForNode.php
@@ -101,7 +101,7 @@ class ForNode extends Node
         $compiler->write("\$_parent = \$context['_parent'];\n");
 
         // remove some "private" loop variables (needed for nested loops)
-        $compiler->write('unset($context[\'_seq\'], $context[\''.$this->getNode('key_target')->getAttribute('name').'\'], $context[\''.$this->getNode('value_target')->getAttribute('name').'\'], $context[\'_parent\']');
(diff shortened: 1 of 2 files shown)


ruudk:
I handle this manually in TwigStan for now:
- https://github.com/twigstan/twigstan/pull/72

src/Node/ForNode.php
```diff
@@ -101,7 +101,7 @@ public function compile(Compiler $compiler): void
         $compiler->write("\$_parent = \$context['_parent'];\n");
 
         // remove some "private" loop variables (needed for nested loops)
-        $compiler->write('unset($context[\'_seq\'], $context[\''.$this->getNode('key_target')->getAttribute('name').'\'], $context[\''.$this->getNode('value_target')->getAttribute('name').'\'], $context[\'_parent\']');
+        $compiler->write('unset($context[\'_parent\'], $context[\'_seq\'], $context[\''.$this->getNode('key_target')->getAttribute('name').'\'], $context[\''.$this->getNode('value_target')->getAttribute('name').'\'], $context[\'_parent\']');
```
smnandre, line 104: This is done at the end of the line :)

smnandre (commented):
Should be removed (in all changes i think) from the end of the line, or reverted if order has no importance for you

ruudk:
The problem is not in Twig but in TwigStan 🥲 I found the problem there.

src/Node/ForNode.php
ruudk: Thanks for spotting this. I didn't realize it and was too quick with the PR.

PR #3930, opened 2023-12-06
Add an `html_attr` function to make outputting HTML attributes easier

mpdude:
**Updated:** This description has been updated to reflect changes from the discussion up to https://github.com/twigphp/Twig/pull/3930#issuecomment-3870445987.

This PR suggests adding an `html_attr` function and two filters `html_attr_merge` and `html_attr_type`. Together, they are intended to make it easier to collect HTML attributes in arrays, in order to pass them in Twig to included templates or macros, and to ultimately print such attribute sets as HTML strings.

`html_attr_merge` can be used to either merge such arrays over default values, or to override (say, inside a macro) particular values in a given attribute array. As described in #3907, it favors overwriting simple (scalar) attribute values and appending to multi-valued attributes over all the other operations one could conceive (like, for example, replacing a list of two CSS `class` names with two other ones). This is a design decision to keep the API simple and optimized for the primary use case that I see. But, since we're mostly dealing with arrays after all, users are free to do in parallel any other kind of array wrangling they see fit.

So, this PR is _not_ trying to design a full-fledged, object-oriented API with all the necessary methods to add, replace, remove attributes; to add, change or toggle elements in "list" style attributes like `class`; to provide extension points for custom (arbitrary) attributes or to provide a fluent API to do all that from within PHP code. See https://github.com/symfony/ux/issues/3269 for a Symfony UX component RFC that does that.

A little bit of special case handling is present for `aria-*`, `data-*` and inline CSS `style` attributes. But apart from that, there is no special knowledge about the attributes defined in HTML, ARIA or other standards, nor about the structure and sementic of attribute values. The approach works in a generic way, so it should be possible to use it for many custom attributes as well.

In order to support "list" style attributes like `class`, `aria-labelledby` or `srcset` that may come in different flavors, users have to be disciplined and consistently use iterables (arrays) to represent such attribute values, possibly assisted by the `html_attr_type` filter (see below).

When printing attributes, names and values are escaped. For names, the `html_attr_relaxed` strategy (#4743) is used.

#### Motivation and practical examples

I have seen repeating patterns when dealing with HTML attributes in Twig templates and macros. Typical examples can be found in Symfony's form theme, where an `attr` variable is present in various blocks.

https://github.com/symfony/symfony/blob/4a5d8cf03e1e31d1a7591921c6fa1fe7ec1c2015/src/Symfony/Bridge/Twig/Resources/views/Form/form_div_layout.html.twig#L453-L458

```twig
{%- block widget_attributes -%}
[...]

Issue #3907, opened 2023-11-21
Helper function to make dealing with HTML attributes easier


mpdude:
I'd like to discuss adding a new function, let's say `attr()`, to make juggling with HTML attributes easier.

I'm willing to work on the implementation given that the general idea is backed here and we can sketch out some details.

#### Motivation 

One pattern I see coming up in Twig templates again and again is along the lines of...

```twig
{% set attr = attr ?? {} %}
{% set attr = attr|merge({ id, type, href, class }) %}
{% set attr = attr|merge({class: (attr.class|default('') ~ ' form-horizontal')|trim}) %}
{% set attr = attr|merge({ 'aria-disabled': 'true', 'onclick': 'event.preventDefault()' }) %}
{% set attr = attr|merge({ 'style': {
    'color': groupFontColor ? groupFontColor : null,
    'background-color': groupBackgroundColor ? groupBackgroundColor : null
}}) %}
```

Then a custom macro along the lines of `<div {{ printAttributes.print(attr) }}>...</div>` follows. This macro basically puts all the key-value pairs into a `key="value" other-key="other-value"` sequence. 

#### Suggested syntax, thoughts

Be able to pass multiple hashes into the function to avoid `|merge` calls:

`<a {{ attr({ id, rel, href, class }, disabled ? { 'aria-disabled': 'true' }) }}>`

Here, the short-hand ternary for the second argument resolves to the `''` (empty string) value. Maybe this should be interpreted as an empty array to allow this notation.

Then, maybe know about special attributes like `class` that might work better if values are converted to array first and then a deep merge is performed:

`<div {{ attr({ class: ['zero', 'first'] }, { class: 'second' }, third ? { class: 'third' }) }}>`

In the case of `class`, all values should be concatenated with spaces and the result being trimmed.

For `style`, the attribute value might be a hash by itself. It would be turned into a string by concatenating all values with a pattern like `{key}: {value};`:

`<div {{ attr({ style: { color: 'red' }}, { style: { 'background-color': 'green' }}) }}>` gives `<div style="color: red; background-color: green;">`

In this case, maybe latter `attr.style` keys should override former ones?

`data` key/value pairs might be converted to `data-firstkey="firstvalue" data-secondkey="secondvalue"`.

In general, if that should matter, optimize for the use case of adding attributes or adding to styles, CSS classes and the like instead of unsetting/overriding or completely replacing them with latter values.

#### tbd

[...]

PR #4405, opened 2024-10-22
RFC / WIP - html_attributes function

leevigraham:
An alternative implementation to #3930

Addresses:

* https://github.com/twigphp/Twig/issues/3907
* https://github.com/twigphp/html-extra/pull/4
* https://github.com/twigphp/Twig/pull/3760

This WIP pr demonstrates a html attribute merging strategy for html attributes, aria-attributes with special handling for `class`, `style` `data` and `aria` attributes.

I've currently focussed on the `HtmlAttributes::merge` function which returns the merged attributes array.

## Examples

See [`./Tests/HtmlAttributesTest.php`](https://github.com/twigphp/Twig/pull/4405/files#diff-210291f849102679e6c0a1050d5bcd3076cf55b3cc9677c20fab26dcd15f543c) for usage examples

```php
// Basic attribute merging
HtmlAttributes::merge(
    ['id' => 'a', 'disabled' => true], 
    ['hidden' => true]
) 
=== 
['id' => 'a', 'disabled' => true, 'hidden' => true]
``` 

```php
// Attribute overriding
HtmlAttributes::merge(
    ['id' => 'a'], 
    ['id' => 'b']
) 
===
['id' => 'b']
``` 

```php
// class merging
// class attributes produce an array of key / values where the value is true, false, null.
// multiple classnames are split on spaces
HtmlAttributes::merge(
    ['class' => 'a'], 
    ['class' => 'b'], 
    ['class' => 'c'],
    ['class' => 'd e']
) 
===
['class' => ['a' => true, 'b' => true, 'c' => true, 'd' => true, 'e' => true]]
``` 

```php
// Style merging
// values provided as string, array of strings or key / value array
// styles are split on ":" and converted to key / value pairs
HtmlAttributes::merge(
    ['style' => 'color: red'], 
    ['style' => ['color: green']], 
    ['style' => ['background-color' => 'blue']],
    ['style' => ['display: block' => false]]
) 
===
['style' => ['color: green;' => 'true', 'background-color: blue;' => 'true', 'display: block;' => false]]
```

```php
// data expansion:
HtmlAttributes::merge(
    [data' => ['count' => '1']]
) 
===
['data-count' => '1']
``` 

```php
// aria expansion:
HtmlAttributes::merge(
    ['aria' => ['hidden' => true]]
) 
===
['aria-hidden' => true]
``` 

The return value of `HtmlAttributes::merge` should also be able to be used as an input for `HtmlAttributes::merge` as demonstrated here: 

https://github.com/twigphp/Twig/pull/4405/files#diff-210291f849102679e6c0a1050d5bcd3076cf55b3cc9677c20fab26dcd15f543cR27-R75

## Rendering Attributes

The `HtmlAttributes::renderAttributes` method that takes the output of `HtmlAttributes::merge` and creates the attribute string.

This method:

1. skips `null` attribute values
2. filters and implodes `class`, `style` and `data-controller` attribute values
3. coerces `aria-*` boolean attribute values to `'true'` `'false'` strings.
4. json encodes `data-*` array values
5. skips remaining `false` values (see `aria-*` coercion above)
6. returns attribute name for `true` boolean values
7. returns attribute name and encoded value for everything else

[`./Tests/HtmlAttributesTest.php`](https://github.com/twigphp/Twig/pull/4405/files#diff-210291f849102679e6c0a1050d5bcd3076cf55b3cc9677c20fab26dcd15f543c) demonstrate the return value of `HtmlAttributes::merge` and `HtmlAttributes::renderAttributes`

There was some consideration made to coerce `data-*` boolean values to string `'true'` and `'false'` but this was decided against. StimulusJs uses the following conditional to determine if a `data-*-value` attribute is `true` or `false` when internally coercing to a javascript `Boolean`.

```js
// https://stimulus.hotwired.dev/reference/values#types
!(value == "0" || value == "false")
```

Illustrated here: https://codepen.io/leevigraham/pen/MWNOyLr

## Challenges

The challenge with a `html_attributes` like function is that the merging strategy is arbitrary.

`class` and `style` attributes are usually merged together. Other attributes override the previous values.

Symfony [UX twig components](https://symfony.com/bundles/ux-twig-component/current/index.html) recommend [Stimulus](https://stimulus.hotwired.dev) for [interaction](https://symfony.com/bundles/ux-live-component/current/index.html#working-with-the-component-in-javascript). Stimulus uses `data-controller` attributes for functionality. The `data-controller` value can be a space delimited list of strings. In this case should the multiple `data-controller` values be merged or overridden? For StimulusJs also applies to `data-target` and `data-action`

### Merge strategy options

Given the challenges with `data-controller` above maybe the method should take 2 arguments:

1. The attributes to be merged
2. An array which defines the merging strategy

```php
HtmlAttributes::merge(
    attributes: ['style' => ['background-color' => 'blue', 'color' => 'red']],
    options: ['merge' => ['class', 'data-controller']]
)
```

In the example above the `$options` argument would be used to determine which values to merge. Other values would replace.

Given the return value of `HtmlAttributes::merge` can also be used as the `$attributes` argument of `HtmlAttributes::merge` the developer could call `HtmlAttributes::merge` multiple times which would be the equivalent of multiple attribute arrays / argument unpacking.

## References in other platforms / frameworks:

Yii2 has a similar function: https://github.com/yiisoft/yii2/blob/master/framework/helpers/BaseHtml.php#L1966-L2046

The [renderTagAttributes](https://www.yiiframework.com/doc/api/2.0/yii-helpers-basehtml#renderTagAttributes()-detail) method has the following rules:

* Attributes whose values are of boolean type will be treated as [boolean attributes](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attributes).
* Attributes whose values are null will not be rendered.
* aria and data attributes get special handling when they are set to an array value. In these cases, the array will be "expanded" and a list of ARIA/data attributes will be rendered. For example, `'aria' => ['role' => 'checkbox', 'value' => 'true']` would be rendered as `aria-role="checkbox" aria-value="true`".
* If a nested data value is set to an array, it will be JSON-encoded. For example, `'data' => ['params' => ['id' => 1, 'name' => 'yii']]` would be rendered as `data-params='{"id":1,"name":"yii"}`'.

CraftCMS uses twig and provides an attr() twig method that implements renderTagAttributes. I've used this helper many times and the rules above are great. Especially the "Attributes whose values are null will not be rendered".

Vuejs v2 -> v3 also went through some changes for false values https://v3-migration.vuejs.org/breaking-changes/attribute-coercion.html. This aligns with "Attributes whose values are null will not be rendered." above.

## TODO

- [ ] add twig filter function
- [ ] cleanup code
- [ ] write docs

### DIFF
--- a/extra/html-extra/HtmlAttributes.php
+++ b/extra/html-extra/HtmlAttributes.php
@@ -1,0 +1,221 @@
+<?php
+
+namespace Twig\Extra\Html;
+
+use Twig\Error\RuntimeError;
+
+final class HtmlAttributes
+{
+    /**
+     * Merges multiple attribute group arrays into a single array.
+     *
+     * `HtmlAttributes::merge(['id' => 'a', 'disabled' => true], ['hidden' => true])` becomes
+     * `['id' => 'a', 'disabled' => true, 'hidden' => true]`
+     *
+     * attributes override each other in the order they are provided.
+     *
+     * `HtmlAttributes::merge(['id' => 'a'], ['id' => 'b'])` becomes `['id' => 'b']`.
+     *
+     * However, `class` and `style` attributes are merged into an array so they can be concatenated in later processing.
+     *
+     * `HtmlAttributes::merge(['class' => 'a'], ['class' => 'b'], ['class' => 'c'])` becomes
+     * `['class' => ['a' => true, 'b' => true, 'c' => true]]`.
+     *
+     * style attributes are also merged into an array so they can be concatenated in later processing.
+     *
+     * `HtmlAttributes::merge(['style' => 'color: red'], ['style' => ['background-color' => 'blue']])` becomes
+     * `['style' => ['color: red;' => true, 'background-color: blue;' => true]]`.
+     *
+     * style attributes which are arrays with false and null values are also processed
+     *
+     * `HtmlAttributes::merge(['style' => ['color: red' => true]], ['style' => ['display: block' => false]]) becomes
+     * `['style' => ['color: red;' => true, 'display: block;' => false]]`.
+     *
+     * attributes can be provided as an array of key, value where the value can be true, false or null.
+     *
+     * Example:
+     * `HtmlAttributes::merge(['class' => ['a' => true, 'b' => false], ['class' => ['c' => null']])` becomes
+     * `['class' => ['a' => true, 'b' => false, 'c' => null]]`.
+     *
+     * `aria` and `data` arrays are expanded into `aria-*` and `data-*` attributes before further processing.
+     *
+     * Example:
+     *
+     * `HtmlAttributes::merge([data' => ['count' => '1']])` becomes `['data-count' => '1']`.
+     * `HtmlAttributes::merge(['aria' => ['hidden' => true]])` becomes `['aria-hidden' => true]`.
+     *
+     * @param ...$attributeGroup
+     * @return array
+     * @throws RuntimeError
+     * @see ./Tests/HtmlAttributesTest.php for usage examples
+     *
+     */
+    public static function merge(...$attributeGroup): array
+    {
+        $result = [];
+
+        $attributeGroupCount = 0;
+
+        foreach ($attributeGroup as $attributes) {
+
+            $attributeGroupCount++;
+
+            // Skip empty attributes
+            // Return early if no attributes are provided
+            // This could be false or null when using the twig ternary operator
+            if (!$attributes) {
+                continue;
+            }
+
+            if (!is_iterable($attributes)) {
+                throw new RuntimeError(sprintf('"%s" only works with mappings or "Traversable", got "%s" for argument %d.', self::class, \gettype($attributes), $attributeGroupCount));
+            }
+
+            // Alternative to is_iterable check above, cast the attributes to an array
+            // This would produce weird results but would not throw an error
+//            $attributes = (array)$attributes;
+
+            // data and aria arrays are expanded into data-* and aria-* attributes
+            $expanded = [];
+            foreach ($attributes as $key => $value) {
+                if (in_array($key, ['data', 'aria'])) {
+                    $value = (array)$value;
+                    foreach ($value as $k => $v) {
+                        $k = $key . '-' . $k;
+                        $expanded[$k] = $v;
+                    }
+                    continue;
+                }
+                $expanded[$key] = $value;
+            }
+
+            // Reset the attributes array to the flattened version
+            $attributes = $expanded;
+
+            foreach ($attributes as $key => $value) {
+
+                // Treat class and data-controller attributes as arrays
+                if (in_array($key, [
+                    'class',
+                    'data-controller',
+                    'data-action',
+                    'data-targets',
+                ])) {
+                    if (!array_key_exists($key, $result)) {
+                        $result[$key] = [];
+                    }
+                    $value = (array)$value;
+                    foreach ($value as $k => $v) {
+                        if (is_int($k)) {
+                            $classes = explode(' ', $v);
+                            foreach ($classes as $class) {
+                                $result[$key][$class] = true;
+                            }
+                        } else {
+                            $classes = explode(' ', $k);
+                            foreach ($classes as $class) {
+                                $result[$key][$class] = $v;
+                            }
+                        }
+                    }
+                    continue;
+                }
+
+                if ($key === 'style') {
+                    if (!array_key_exists('style', $result)) {
+                        $result['style'] = [];
+                    }
+                    $value = (array)$value;
+                    foreach ($value as $k => $v) {
+                        if (is_int($k)) {
+                            $styles = array_filter(explode(';', $v));
+                            foreach ($styles as $style) {
+                                $style = explode(':', $style);
+                                $sKey = trim($style[0]);
+                                $sValue = trim($style[1]);
+                                $result['style']["$sKey: $sValue;"] = true;
+                            }
+                        } elseif (is_bool($v) || is_null($v)) {
+                            $styles = array_filter(explode(';', $k));
+                            foreach ($styles as $style) {
+                                $style = explode(':', $style);
+                                $sKey = trim($style[0]);
+                                $sValue = trim($style[1]);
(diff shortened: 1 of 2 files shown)


leevigraham:
* Attribute rendering added to the class and tests: https://github.com/twigphp/Twig/pull/4405/files#diff-210291f849102679e6c0a1050d5bcd3076cf55b3cc9677c20fab26dcd15f543cR100
* Changed style merging so that the array keys contain the case attribute and value
* Updated the original PR description.

fabpot:
I suppose that the implementation of `html_classes` would use this new function? https://github.com/twigphp/Twig/blob/3.x/extra/html-extra/HtmlExtension.php#L91-L113

leevigraham:
@fabpot Yep it could do.

I still haven't fully considered the twig methods yet. Given the existing static methods the `htmlClasses` method could look like:

```php
public static function htmlClasses(...$args): string
{
    $attributes = HtmlAttributes::merge(['class' => $args]);
    return HtmlAttributes::renderAttributes($attributes);
}
```

I'm also interested in your (and @stof) opinion on merging vs replacing on some attributes. `class` and `style` use merging (seems to be standard across other frameworks)  but there are use cases for merging `data-controller`, `data-action`, some `aria-*` properties.

I thought using the `CVA` pattern here might work. ie… add a `HtmlAttributes::__construct()` method which takes a config argument (or null), and add a `apply` function which takes the object or attributes and returns the attribute string.

image

extra/html-extra/HtmlAttributes.php
```diff
@@ -0,0 +1,221 @@
...
+                if ($key === 'style') {
+                    if (!array_key_exists('style', $result)) {
+                        $result['style'] = [];
+                    }
+                    $value = (array)$value;
+                    foreach ($value as $k => $v) {
+                        if (is_int($k)) {
+                            $styles = array_filter(explode(';', $v));
+                            foreach ($styles as $style) {
+                                $style = explode(':', $style);
```
stof, line 133: this is broken in case CSS values contains `:` (which can happen, for instance in the `if` function or in strings)

extra/html-extra/HtmlAttributes.php
```diff
@@ -0,0 +1,221 @@
...
+                }
+
+                if ($key === 'style') {
+                    if (!array_key_exists('style', $result)) {
+                        $result['style'] = [];
+                    }
+                    $value = (array)$value;
+                    foreach ($value as $k => $v) {
+                        if (is_int($k)) {
+                            $styles = array_filter(explode(';', $v));
```
stof, line 131: this is broken if a CSS string contains a `;`. You cannot perform CSS parsing based on naive `explode` calls. The CSS syntax is not meant to allow that.

extra/html-extra/HtmlAttributes.php
```diff
@@ -0,0 +1,221 @@
...
+            }
+
+            // Alternative to is_iterable check above, cast the attributes to an array
+            // This would produce weird results but would not throw an error
+//            $attributes = (array)$attributes;
+
+            // data and aria arrays are expanded into data-* and aria-* attributes
+            $expanded = [];
+            foreach ($attributes as $key => $value) {
+                if (in_array($key, ['data', 'aria'])) {
```
stof, line 81: should we really add such support ? Right now, those `data` and `aria` keys will be merged with the explicit `aria-*` and `data-*` keys, which can be a confusing behavior. I'm not sure this special case is worth it.

extra/html-extra/HtmlAttributes.php
```diff
@@ -0,0 +1,221 @@
...
+
+            // Skip null values regardless of attribute key
+            if ($value === null) {
+                continue;
+            }
+
+            // Handle class, style, data-controller value coercion
+            // array[] -> concatenate string
+            if (in_array($key, ['class', 'style', 'data-controller'])) {
+                $value = array_filter($value);
```
stof, line 176: what if the value is not an array there ?

extra/html-extra/HtmlAttributes.php
```diff
@@ -0,0 +1,221 @@
...
+            // true -> 'true'
+            // false -> 'false,
+            // array[] ->  concatenate string
+            if (str_starts_with($key, 'aria-')) {
+                if ($value === true) {
+                    $value = 'true';
+                } elseif ($value === false) {
+                    $value = 'false';
+                } elseif(is_array($value)) {
+                    $value = join(" ", array_filter($value));
```
stof, line 191:
```suggestion
                    $value = implode(' ', array_filter($value));
```

extra/html-extra/HtmlAttributes.php
```diff
@@ -0,0 +1,221 @@
...
+            }
+
+            // Boolean attribute doesn't have a value
+             if ($value === true) {
+                $return[] = $key;
+                continue;
+            }
+
+            // Everything else gets added as an encoded value
+            $return[] = $key . '="' . htmlspecialchars($value) . '"';
```
stof, line 216: keys must be escaped as well (and with the `html_attr` strategy)

extra/html-extra/HtmlAttributes.php
```diff
@@ -0,0 +1,221 @@
...
+     * `HtmlAttributes::merge([data' => ['count' => '1']])` becomes `['data-count' => '1']`.
+     * `HtmlAttributes::merge(['aria' => ['hidden' => true]])` becomes `['aria-hidden' => true]`.
+     *
+     * @param ...$attributeGroup
+     * @return array
+     * @throws RuntimeError
+     * @see ./Tests/HtmlAttributesTest.php for usage examples
+     *
+     */
+    public static function merge(...$attributeGroup): array
```
stof, line 53: this signature does not correspond to the PR description.

stof:
> I still haven't fully considered the twig methods yet. Given the existing static methods the `htmlClasses` method could look like:

no it cannot. `html_classes` returns only the value of the class attribute. It does not render the attribute name (which is totally different both in term of usage and of auto-escaping)

Kocal:
https://github.com/twigphp/Twig/pull/3930 has been merged, I think this one can be closed

PR #4409, opened 2024-10-24
Deprecate the range function

fabpot:


doc/functions/range.rst
```diff
@@ -1,6 +1,11 @@
 ``range``
 =========
 
+.. warning::
+
+    The ``attribute`` filter is deprecated as of Twig 3.15. Use the ``..``
```
xabbuh:
```suggestion
    The ``range()`` function is deprecated as of Twig 3.15. Use the ``..``
```

fabpot: oops :)

doc/functions/range.rst
```diff
@@ -1,6 +1,11 @@
 ``range``
 =========
 
+.. warning::
+
+    The ``range`` filter is deprecated as of Twig 3.15. Use the ``..`` operator
```
xabbuh:
```suggestion
    The ``range`` function is deprecated as of Twig 3.15. Use the ``..`` operator
```

:)

fabpot: 😂

xabbuh (approved):

fabpot:
Reverted, I didn't realize that `range` has one additional feature that `..` does not have: step.

PR #4410, opened 2024-10-24
Add return type to compiled macro

ruudk:
This makes it easier for TwigStan to analyze the return type.

### DIFF
--- a/src/Node/MacroNode.php
+++ b/src/Node/MacroNode.php
@@ -14,6 +14,7 @@ namespace Twig\Node;
 use Twig\Attribute\YieldReady;
 use Twig\Compiler;
 use Twig\Error\SyntaxError;
+use Twig\Markup;
 use Twig\Node\Expression\ArrayExpression;
 use Twig\Node\Expression\Variable\LocalVariable;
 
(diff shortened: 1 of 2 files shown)


ruudk:
/cc @fabpot @stof 

This would be good to support macros in TwigStan:
- https://github.com/twigstan/twigstan/issues/4

src/Node/MacroNode.php
```diff
@@ -76,7 +77,7 @@ public function compile(Compiler $compiler): void
 
         $compiler
             ->raw('...$varargs')
-            ->raw(")\n")
+            ->raw("): string | Markup\n")
```
fabpot:
```suggestion
            ->raw("): string|Markup\n")
```

fabpot:
I don't think we should merge it yet. I'd like for macros to yield instead.

ruudk:
Since it's compiled code, we change it to become iterable later?

src/Node/MacroNode.php
ruudk: Fixed.

fabpot:
Thank you @ruudk.

PR #4411, opened 2024-10-24
Add return type to getDebugInfo

ruudk:


fabpot:
Thank you @ruudk.

stof:
this should not be necessary though, as the compiled template overrides an existing abstract method of the Template, where we already define the phpdoc type.

ruudk:
You are right. Sorry about this. We can revert this.

fabpot:
Reverted

PR #4413, opened 2024-10-24
do not drop none digit characters from generated variable names

xabbuh:
I was wondering if it was really intended to cast the var name generated by the compiler to an int as done in #4398. Doesn't this create the risk of variable names being reused at some point?

### DIFF
--- a/src/Node/Expression/TempNameExpression.php
+++ b/src/Node/Expression/TempNameExpression.php
@@ -41,7 +41,7 @@ class TempNameExpression extends AbstractExpression
     public function compile(Compiler $compiler): void
     {
         if (null === $this->getAttribute('name')) {
-            $this->setAttribute('name', \sprintf('_l%d', $compiler->getVarName()));
+            $this->setAttribute('name', $compiler->getVarName());
         }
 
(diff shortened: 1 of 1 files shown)


src/Node/Expression/TempNameExpression.php
```diff
@@ -41,7 +41,7 @@ public function __construct(string|int|null $name, int $lineno)
     public function compile(Compiler $compiler): void
     {
         if (null === $this->getAttribute('name')) {
-            $this->setAttribute('name', \sprintf('_l%d', $compiler->getVarName()));
+            $this->setAttribute('name', \sprintf('_l%s', $compiler->getVarName()));
```
fabpot:
```suggestion
            $this->setAttribute('name', $compiler->getVarName());
```

Should be enough, right?

xabbuh: Probably, I wasn't sure if it was intended to introduce some kind of "namespace" here.

xabbuh: code updated as suggested, let's see if the tests agree

fabpot:
Thank you @xabbuh.

PR #4415, opened 2024-10-24
Prevent calling yield() on false

ruudk:
When Template::getParent returns false, calling yield() will fail.

> Cannot call method yield() on Twig\\Template|Twig\\TemplateWrapper|false.

Let's change the return type of `getTemplate` from false to null.

This allows us to use the null safe operator introduced in PHP 8.

### DIFF
--- a/src/Template.php
+++ b/src/Template.php
@@ -32,7 +32,7 @@ abstract class Template
     public const ARRAY_CALL = 'array';
     public const METHOD_CALL = 'method';
 
-    protected $parent;
+    protected ?self $parent = null;
     protected $parents = [];
     protected $blocks = [];
     protected $traits = [];
@@ -74,14 +74,14 @@ abstract class Template
      *
      * @return self|TemplateWrapper|false The parent template or false if there is no parent
      */
-    public function getParent(array $context): self|TemplateWrapper|false
+    public function getParent(array $context): self|TemplateWrapper|null
     {
         if (null !== $this->parent) {
             return $this->parent;
         }
 
         if (!$parent = $this->doGetParent($context)) {
-            return false;
+            return null;
         }
 
         if ($parent instanceof self || $parent instanceof TemplateWrapper) {
--- a/tests/Node/ModuleTest.php
+++ b/tests/Node/ModuleTest.php
@@ -92,8 +92,6 @@ class __TwigTemplate_%x extends Template
 
         \$this->source = \$this->getSourceContext();
 
-        \$this->parent = false;
-
         \$this->blocks = [
         ];
     }
@@ -120,7 +118,7 @@ class __TwigTemplate_%x extends Template
      */
     public function getDebugInfo(): array
     {
-        return array (  42 => 1,);
+        return array (  40 => 1,);
     }
 
     public function getSourceContext(): Source
@@ -281,7 +279,7 @@ class __TwigTemplate_%x extends Template
         // line 4
         \$context["foo"] = "foo";
         // line 2
-        yield from \$this->getParent(\$context)->unwrap()->yield(\$context, array_merge(\$this->blocks, \$blocks));
+        yield from \$this->getParent(\$context)?->unwrap()->yield(\$context, array_merge(\$this->blocks, \$blocks));
     }
 
     /**
--- a/src/Node/ModuleNode.php
+++ b/src/Node/ModuleNode.php
@@ -339,7 +339,6 @@ final class ModuleNode extends Node
                     ->raw(");\n")
                 ;
             }
-
             $compiler->write('yield from ');
 
             if ($parent instanceof ConstantExpression) {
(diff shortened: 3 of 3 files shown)


ruudk:
Another solution could be, to change the `false` return type to `null`, so that we can use [null safe operator](https://stitcher.io/blog/php-8-nullsafe-operator) like this:
```php
yield from $this->getParent($context)?->unwrap()->yield(\$context, array_merge($this->blocks, $blocks));
```
This would be my preference.

src/Node/EmbedNode.php
```diff
@@ -49,7 +49,11 @@ protected function addGetTemplate(Compiler $compiler, string $template = ''): vo
         if ($this->getAttribute('ignore_missing')) {
             $compiler
                 ->raw(";\n")
+                ->write(sprintf("if (\$%s) {\n", $template))
```
fabpot: This is always `true` as `$template` is never empty when `ignore_missing` is set.

ruudk: Ok, reverted.

fabpot:
Is it possible to add a test that would break without the patch?

tests/Node/ModuleTest.php
```diff
@@ -186,7 +186,9 @@ protected function doDisplay(array \$context, array \$blocks = []): iterable
         \$macros["macro"] = \$this->macros["macro"] = \$this->loadTemplate("foo.twig", "foo.twig", 2)->unwrap();
         // line 1
         \$this->parent = \$this->loadTemplate("layout.twig", "foo.twig", 1);
-        yield from \$this->parent->unwrap()->yield(\$context, array_merge(\$this->blocks, \$blocks));
+        if (\$this->parent) {
```
fabpot: `$this->parent` always exists as `loadTemplate` always returns an object or throws an exception.

ruudk: You're right. I just saw that too.

ruudk:
I pushed a new approach. 

Instead of introducing the variables, we change the return type from false to null and use the null safe operator. 

Please have a look.

I'll see if I can create a failing test case... might be hard.

ruudk:
It's not possible to create a failing test for this, because Twig itself does guard against it.

It's just that it can't be statically analyzed. 

Tried this test:
```
--TEST--
"extends" tag
--TEMPLATE--
{% extends foo %}

{% block content %}
FOO
{% endblock %}
--DATA--
return ['foo' => 'not-found.twig']
--EXPECT--
FOO
```

Gets rendered like:
```php
    protected function doGetParent(array $context): bool|string|Template|TemplateWrapper
    {
        // line 2
        return $this->loadTemplate((isset($context["foo"]) || array_key_exists("foo", $context) ? $context["foo"] : (function () { throw new RuntimeError('Variable "foo" does not exist.', 2, $this->source); })()), "index.twig", 2);
    }

    protected function doDisplay(array $context, array $blocks = []): iterable
    {
        $macros = $this->macros;
        yield from $this->getParent($context)?->unwrap()->yield($context, array_merge($this->blocks, $blocks));
    }
```

So getParent calls doGetParent that throws when it does not exist. 

```
1) Twig\Tests\IntegrationTest::testIntegration with data set "tags/inheritance/dynamic_invalid.test" ('tags/inheritance/dynamic_invalid.test', '"extends" tag', '', array('\n{% extends foo %}\n\n{% blo...ock %}'), false, array(array('--DATA--\nreturn ['foo' => 'n...-\nFOO', '\nreturn ['foo' => 'not-found.twig']\n', '', '\nFOO')), '')
"extends" tag (in tags/inheritance/dynamic_invalid.test)
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'FOO'
+'Twig\Error\Error: Twig\Error\LoaderError: Template "not-found.twig" is not defined in "index.twig" at line 2.'
```

src/Template.php
```diff
@@ -74,14 +74,14 @@ abstract public function getSourceContext(): Source;
      *
      * @return self|TemplateWrapper|false The parent template or false if there is no parent
```
stof, line 75: this needs to be updated as well

stof (commented):
`\Symfony\Bridge\Twig\Form\TwigRendererEngine` uses the internal Template object of Twig (which we should avoid if we can, as it is not covered by BC) and will be impacted by this PR.

ruudk:
> `\Symfony\Bridge\Twig\Form\TwigRendererEngine` uses the internal Template object of Twig (which we should avoid if we can, as it is not covered by BC) and will be impacted by this PR.

Ouch. 

https://github.com/symfony/symfony/blob/5feb7493c372384d78b4931bb76f73c83f49f50e/src/Symfony/Bridge/Twig/Form/TwigRendererEngine.php#L165

What do you suggest? Not changing the return type, but store the call that is problematic in a variable, and only call it when non false?

ruudk:
If this is not something to pursue, I can also try to create an extension to remove the false type from this `getParent`. I guess it will only be a static analysis issue.

stof:
@ruudk if you can do this with a phpstan extension, it might be easier (the `doDisplay` method will not contain a call to `getParent` for templates that don't have a parent anyway)

fabpot:
Let's close as I'm always very reluctant to change code "just" because of static analysis.

PR #4416, opened 2024-10-24
Define missing macros variable in constructor

ruudk:
When defining a macro in a template, the constructor looks like this:

```php
    public function __construct(\Twig\Environment $env)
    {
        parent::__construct($env);
        $this->source = $this->getSourceContext();
        $this->parent = \false;
        $this->blocks = [];
        $macros["_self"] = $this->macros["_self"] = $this;
    }
```

The `$macros` variable does not exist resulting in:

> Implicit array creation is not allowed - variable $macros does not exist.

/cc @fabpot Is this the best way to do it? Or would it be better to make this conditional instead?
https://github.com/twigphp/Twig/blob/f3db1d1a0c5cc21fad255bd6690bbde16d378459/src/Node/Expression/Variable/AssignTemplateVariable.php#L31-L36

### DIFF
--- a/tests/Node/ModuleTest.php
+++ b/tests/Node/ModuleTest.php
@@ -90,6 +90,7 @@ class __TwigTemplate_%x extends Template
     {
         parent::__construct(\$env);
 
+        \$macros = \$this->macros;
         \$this->source = \$this->getSourceContext();
 
         \$this->parent = false;
@@ -120,7 +121,7 @@ class __TwigTemplate_%x extends Template
      */
     public function getDebugInfo(): array
     {
-        return array (  42 => 1,);
+        return array (  43 => 1,);
     }
 
     public function getSourceContext(): Source
@@ -167,6 +168,7 @@ class __TwigTemplate_%x extends Template
     {
         parent::__construct(\$env);
 
+        \$macros = \$this->macros;
         \$this->source = \$this->getSourceContext();
 
         \$this->blocks = [
@@ -211,7 +213,7 @@ class __TwigTemplate_%x extends Template
      */
     public function getDebugInfo(): array
     {
-        return array (  48 => 1,  46 => 2,  39 => 1,);
+        return array (  49 => 1,  47 => 2,  40 => 1,);
     }
 
     public function getSourceContext(): Source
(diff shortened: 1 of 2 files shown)


tests/Node/ModuleTest.php
```diff
@@ -90,6 +90,7 @@ public function __construct(Environment \$env)
     {
         parent::__construct(\$env);
 
+        \$macros = \$this->macros;
```
fabpot, line 93: As `$macros` is not used in the constructor, I'm not sure I understand what it does/fixes.

ruudk: Before this PR, the compiled constructor looked like this:

```php
    public function __construct(\Twig\Environment $env)
    {
        parent::__construct($env);
        $this->source = $this->getSourceContext();
        $this->parent = \false;
        $this->blocks = [];
        $macros["_self"] = $this->macros["_self"] = $this;
    }
```

Look at the last line of the method. PHPStan reports this error:

> Implicit array creation is not allowed - variable $macros does not exist.

It's because by default, `$macros[$name] =` is prepended in 
https://github.com/twigphp/Twig/blob/f3db1d1a0c5cc21fad255bd6690bbde16d378459/src/Node/Expression/Variable/AssignTemplateVariable.php#L31-L36

So we have 2 options:

1) add `$macros = $this->macros;` as the beginning of the constructor, just like doDisplay and block_ methods. Now PHPStan sees that `$macros` exists, and that adding something to it is fine.
2) remove the `$macros[$name] = ` prefix when inside the constructor

ruudk: I feel option 2 makes more sense, shall I try that?

fabpot: This behavior comes from `MacroAutoImportNodeVisitor`, which has been removed in 3.15, so I don't think there is anything to fix.

ruudk: Ouch... then I'll wait for 3.15 to be released. Hopefully this can be done with https://github.com/twigphp/Twig/pull/4415 included 🙏

ruudk:
Closing as this will be solved with the release of 3.15 that removes `MacroAutoImportNodeVisitor`

PR #4419, opened 2024-10-25
Use the Spread operator everywhere

fabpot:


stof (approved):

stof:
This would require dropping support for PHP 8.0 (while the commit named "Drop support for 8.0" merged 2 months ago only dropped support for PHP 7.x)

PR #4420, opened 2024-10-25
Add `$this` return type to Template::unwrap

ruudk:
See https://phpstan.org/writing-php-code/phpdoc-types#static-and-%24this

This way, PHPStan understands that the returned template is the exact same instance, and not just a Template.

### DIFF
--- a/src/Template.php
+++ b/src/Template.php
@@ -318,6 +318,7 @@ abstract class Template
 
     /**
      * @internal
+     * @return $this
      */
     public function unwrap(): self
     {


stof:
As this method is meant to make this similar to `TemplateWrapper` to support easily cases dealing with a union `Template|TemplateWrapper`, which benefit does this provide ?

ruudk:
Because of this:
```php
$macros['greeter'] = $this->macros['greeter'] = $this->loadTemplate(
        'macros.twig', 
        '@__main__/EndToEnd/Macros/layout.twig',
        1
    )->unwrap();
```

I can instruct PHPStan using a `loadTemplate` extension, that this call returns the `__TwigTemplate_8a8feb39bb5837ab39c1764f9795e711` class.

But since the code calls `unwrap` on it, it becomes `Template` again.

That means that we loose the specific type immediately.

By adding `@return $this` on the `unwrap` method _only_ on the Template, we tell PHPStan: this returns the same instance. 

The TemplateWrapper is not changed. The `unwrap` method there, will still return Template.

ruudk:
Exactly for this scenario, the `@return $this` is invented/supported by PHPStan.

stof (approved):

ruudk:
@fabpot This one is ready for review.

fabpot:
Thank you @ruudk.

PR #4421, opened 2024-10-28
Add type for join method

ruudk:
Looking at the code, this method is very forgiving in what it accepts.

This helps PHPStan to not complain when something other than an array is passed.

Did I miss other types?

/cc @stof

### DIFF
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -766,9 +766,9 @@ final class CoreExtension extends AbstractExtension
      *  {{ [1, 2, 3]|join }}
      *  {# returns 123 #}
      *
-     * @param array       $value An array
-     * @param string      $glue  The separator
-     * @param string|null $and   The separator for the last pair
+     * @param iterable|array|string|float|int|bool|null  $value An array
+     * @param string                                     $glue  The separator
+     * @param string|null                                $and   The separator for the last pair
      *
      * @internal
      */


src/Extension/CoreExtension.php
```diff
@@ -766,9 +766,9 @@ public static function last(string $charset, $item)
      *  {{ [1, 2, 3]|join }}
      *  {# returns 123 #}
      *
-     * @param array       $value An array
-     * @param string      $glue  The separator
-     * @param string|null $and   The separator for the last pair
+     * @param iterable|array|string|float|int|bool|object|null  $value An array
```
stof: `object` should probably be `Stringable` instead, as non-stringable objects won't work in `implode`

ruudk: It is casted to array, https://3v4l.org/FqgLS

stof: oh. This is probably unexpected behavior.

ruudk: So what to do? 

/cc @fabpot

stof: I would suggest not including `object` in the type here (without changing the implementation for now). This would make TwigStan report cases triggering this weird behavior as mistakes.
I'm not sure about the scalar types, for which the casting is more usable.

ruudk: Good idea, done!

ruudk:
@fabpot this one is ready

ruudk:
You can see here how it works: 
https://github.com/twigstan/twigstan/pull/90

I had to ignore the error that is fixed by this PR:
https://github.com/twigstan/twigstan/blob/d7a85dab53b9d9146169185e9d0225d0b6c52c24/tests/EndToEnd/Filters/errors.json

ruudk:
@fabpot What's needed to get this PR merged? Do I need to do changes?

stof (approved):

fabpot:
Thank you @ruudk.

PR #3963, opened 2024-01-13
Add new "find" filter

smnandre:
This filter returns the first _value_ matching the given arrow function or null

Attempt to solve #3962 

(open to any feedback)

PR #4422, opened 2024-10-28
Add new “any” test

jdreesen:
PHP 8.4 will add the new [`array_find`, `array_find_key`, `array_any`, and `array_all`](https://php.watch/versions/8.4/array_find-array_find_key-array_any-array_all) functions.

A `find` filter was already added in #3963.

This PR adds the `any` test, which returns `true` if _any_ of the elements in the array return `true` when passed to the arrow function.


jdreesen:
The fabbot compliant seems to be unrelated.

fabpot:
Isn't it the same as the current `has some` test?

jdreesen:
Oh wow, I completely missed that because I didn't look at the “Operators” section in the docs :facepalm: 

Awesome! I'll close here.

PR #4424, opened 2024-10-28
Add new “all” test

jdreesen:
PHP 8.4 will add the new [`array_find`, `array_find_key`, `array_any`, and `array_all`](https://php.watch/versions/8.4/array_find-array_find_key-array_any-array_all) functions.

A `find` filter was already added in #3963.
An `any` test will be added with #4422.

This PR adds the `all` test, which returns `true` if _all_ the elements in the array return `true` when passed to the arrow function.


jdreesen:
The fabbot compliant seems to be unrelated.

stof:
I think those 2 (this one and `any` from #4422) might fit better as tests (maybe with a slightly different name) as they are about providing a boolean.

jdreesen:
That sounds reasonable. I have changed it. 
Unfortunately, I can't think of a better name right now and `is all` sounds a bit strange :thinking:

Also, `needs_environment` doesn't seem to work for tests. I get the following error: 

> 1) Twig\Tests\IntegrationTest::testIntegration with data set "tests/all.test" ('tests/all.test', '"all" filter', '', array('\n{{ [1, 2] is all((v) => v >...se' }}'), false, array(array('--DATA--\nreturn [\n    'it' ...\ntrue', '\nreturn [\n    'it' => new \...,\n]\n', '', '\nfalse\ntrue\ntrue\nfalse\nf...\ntrue')), '')
Twig\Error\Error: Twig\Error\SyntaxError: Value for argument "arrow" is required for test "all" in "index.twig" at line 2.

Do we need to check the arrow here?

fabpot:
Isn't it the same as the current `has every` test?

jdreesen:
Oh wow, I completely missed that because I didn't look at the “Operators” section in the docs :facepalm: 

Awesome! I'll close here.

fabpot:
One question though: What feels best: `has every` or `has all`? Same for `has some` vs `has any`.

stof:
The naming `every` and `some` is taken from Javascript, making it familiar to frontend devs.

jdreesen:
As a backend developer I'm more used to `any` and `all`, that's probably part of the reason I overlook the existing tests.

But as @stof already said, frontend devs are probably more familiar with `every` and `some`.

fabpot:
I'm going to say something I don't like, but what about supporting both?

fabpot:
A bit like `===` vs `same as`.

jdreesen:
Sure, why not. Maybe we should have a Twig-CS-Fixer rule then, to enforce a consistent usage within a project?

/cc @VincentLanglet

PR #4425, opened 2024-10-29
Rely on reflection to access null properties

nicolas-grekas:
Benchmark like https://gist.github.com/arnaud-lb/c4235164fb41e11672aa670da8fa4fc7 (which were done for another use case but is still relevant here) prove that using reflection instead of array-casts is significantly faster for accessing properties.

Checking for dynamic properties is done with property_exists.

### DIFF
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -1642,7 +1642,7 @@ final class CoreExtension extends AbstractExtension
 
         // array
         if (Template::METHOD_CALL !== $type) {
-            $arrayItem = \is_bool($item) || \is_float($item) ? (int) $item : $item;
+            $arrayItem = \is_bool($item) || \is_float($item) ? (int) $item : $item = (string) $item;
 
             if ($sandboxed && $object instanceof \ArrayAccess && !\in_array($object::class, self::ARRAY_LIKE_CLASSES, true)) {
                 try {
@@ -1652,9 +1652,11 @@ final class CoreExtension extends AbstractExtension
                 }
             }
 
-            if (((\is_array($object) || $object instanceof \ArrayObject) && (isset($object[$arrayItem]) || \array_key_exists($arrayItem, (array) $object)))
-                || ($object instanceof \ArrayAccess && isset($object[$arrayItem]))
-            ) {
+            if (match (true) {
+                \is_array($object) => \array_key_exists($arrayItem, $object),
+                $object instanceof \ArrayAccess => $object->offsetExists($arrayItem),
+                default => false,
+            }) {
                 if ($isDefinedTest) {
                     return true;
                 }
@@ -1697,6 +1699,8 @@ final class CoreExtension extends AbstractExtension
             }
         }
 
+        $item = (string) $item;
+
         if (!\is_object($object)) {
             if ($isDefinedTest) {
                 return false;
@@ -1731,12 +1735,24 @@ final class CoreExtension extends AbstractExtension
                 }
             }
 
-            if (isset($object->$item) || \array_key_exists((string) $item, (array) $object)) {
+            static $propertyCheckers = [];
+
+            if (isset($object->$item)
+                || ($propertyCheckers[$object::class][$item] ??= self::getPropertyChecker($object::class, $item))($object, $item)
+            ) {
                 if ($isDefinedTest) {
                     return true;
                 }
 
-                return isset($object->$item) ? $object->$item : ((array) $object)[(string) $item];
+                return $object->$item;
+            }
+
+            if ($object instanceof \DateTimeInterface && \in_array($item, ['date', 'timezone', 'timezone_type'], true)) {
+                if ($isDefinedTest) {
+                    return true;
+                }
+
+                return ((array) $object)[$item];
             }
 
             if (\defined($object::class.'::'.$item)) {
@@ -2098,5 +2114,28 @@ final class CoreExtension extends AbstractExtension
         */
 
         return new GetAttrExpression($args[0], $args[1], $args[2] ?? null, Template::ANY_CALL, $line);
+    }
+
+    private static function getPropertyChecker(string $class, string $property): \Closure
+    {
+        static $classReflectors = [];
+
+        $class = $classReflectors[$class] ??= new \ReflectionClass($class);
+
+        if (!$class->hasProperty($property)) {
+            static $propertyExists;
+
+            return $propertyExists ??= \Closure::fromCallable('property_exists');
+        }
+
+        $property = $class->getProperty($property);
+
+        if (!$property->isPublic()) {
+            static $false;
+
+            return $false ??= static fn () => false;
+        }
+
+        return static fn ($object) => $property->isInitialized($object);
     }
 }


src/Extension/CoreExtension.php
```diff
@@ -1706,7 +1717,7 @@ public static function getAttribute(Environment $env, Source $source, $object, $
                     $env->getExtension(SandboxExtension::class)->checkPropertyAllowed($object, $item, $lineno, $source);
                 }
 
-                return isset($object->$item) ? $object->$item : ((array) $object)[(string) $item];
+                return $propertyReflector ? $propertyReflector->getValue($object) : $object->$item;
```
stof: how will this behave for hooks ?

src/Extension/CoreExtension.php
```diff
@@ -1697,7 +1703,12 @@ public static function getAttribute(Environment $env, Source $source, $object, $
 
         // object property
         if (Template::METHOD_CALL !== $type) {
-            if (isset($object->$item) || \array_key_exists((string) $item, (array) $object)) {
+            $propertyReflector = false;
+            if (match (true) {
+                isset($object->$item) => true,
+                ($propertyReflector = self::$propertyReflectors[$object::class][$item] ??= self::getPropertyReflector($object, $item))->isInitialized($object) => true,
```
stof: this looks wrong to me. The custom reflector used as fallback using arrayCast assumes that `getValue()` is called on the same `$object` due to the internal memoization

src/Extension/CoreExtension.php
```diff
@@ -2055,4 +2066,39 @@ public static function parseAttributeFunction(Parser $parser, Node $fakeNode, $a
...
+
+    private static function getPropertyReflector(object $object, string $property): \ReflectionProperty
+    {
+        $class = self::$classReflectors[$object::class] ??= new \ReflectionClass($object::class);
+
+        if ($class->hasProperty($property) && ($reflector = $class->getProperty($property))->isPublic()) {
+            return $reflector;
+        }
+
+        return new class ($property) extends \ReflectionProperty {
```
stof: if the array cast is only meant for dynamic properties, should we do a fast path for declared non-public properties ? Those names cannot be reused by a dynamic property.

src/Extension/CoreExtension.php
nicolas-grekas: this will call them, as it should for twig IMHO

src/Extension/CoreExtension.php
nicolas-grekas: getValue() should always return `null` because of the isset check just before
the only exception I know about is DateTime/DateTimeImmutable
the memoization is only for them
which means: what's wrong?

src/Extension/CoreExtension.php
nicolas-grekas: good idea, updated

src/Extension/CoreExtension.php
stof: is there a guarantee that the Reflector will always be calling `getValue()` on the object passed to `isInitialized` just before to have the right memoized value ? The way the code is written currently avoids doing a second array cast in the special reflector, but it makes it easier to introduce bugs by breaking this implicit invariant assumed by the implementation.

nicolas-grekas: We don't really care actually, because the memoized value is going to be always just `null`, except for DateTime(Immutable), and then this is just strings, so no memleak. This saves an array cast, which is quite cosly operation for getting a `null` value most of the time... All this is local so I see no maintenance issue either.

stof: it is not about memory leaks. It is about using a value from an array cast of a **different** object, in case we break the implicit invariant of always calling `isInitialized` just before `getValue`

src/Extension/CoreExtension.php
```diff
@@ -1626,9 +1629,12 @@ public static function getAttribute(Environment $env, Source $source, $object, $
...
             $arrayItem = \is_bool($item) || \is_float($item) ? (int) $item : $item;
 
-            if (((\is_array($object) || $object instanceof \ArrayObject) && (isset($object[$arrayItem]) || \array_key_exists($arrayItem, (array) $object)))
-                || ($object instanceof \ArrayAccess && isset($object[$arrayItem]))
-            ) {
+            if (match (true) {
+                \is_array($object) => \array_key_exists($arrayItem, $object),
+                $object instanceof \ArrayObject => \array_key_exists($arrayItem, $object->getArrayCopy()),
+                $object instanceof \ArrayAccess => isset($object[$arrayItem]),
+                default => false
```
```suggestion
                default => false,
```

nicolas-grekas: updated

nicolas-grekas:
PR updated with a simpler and more performant approach, that makes DateTimeInterface a special case (because it is).

src/Extension/CoreExtension.php
```diff
@@ -2055,4 +2073,27 @@ public static function parseAttributeFunction(Parser $parser, Node $fakeNode, $a
...
+    private static function getPropertyChecker(string $class, string $property): \Closure
+    {
+        static $classReflectors = [];
+
+        $class = $classReflectors[$class] ??= new \ReflectionClass($class);
+
+        if (!$class->hasProperty($property)) {
+            static $propertyExists;
+
+            return $propertyExists ??= property_exists(...);
```
GromNaN: Twig 3 supports PHP 8.0, which doesn't support first-class callable syntax.
```suggestion
            return $propertyExists ??= static fn ($object, $property) => property_exists($object, $property);
```

nicolas-grekas: oops
updated using Closure::fromCallable

src/Extension/CoreExtension.php
```diff
@@ -2055,4 +2073,27 @@ public static function parseAttributeFunction(Parser $parser, Node $fakeNode, $a
...
+    private static function getPropertyChecker(string $class, string $property): \Closure
+    {
+        static $classReflectors = [];
+
+        $class = $classReflectors[$class] ??= new \ReflectionClass($class);
+
+        if (!$class->hasProperty($property)) {
+            static $propertyExists;
+
+            return $propertyExists ??= \Closure::fromCallable('property_exists');
```
stof, line 2128: Do we need a local cache of the closure ? isn't PHP optimizing this internally already ?

src/Extension/CoreExtension.php
```diff
@@ -2055,4 +2073,27 @@ public static function parseAttributeFunction(Parser $parser, Node $fakeNode, $a
...
+        }
+
+        $property = $class->getProperty($property);
+
+        if (!$property->isPublic()) {
+            static $false;
+
+            return $false ??= static fn () => false;
```
stof, line 2136: do we actually need to cache the closure ?

src/Extension/CoreExtension.php
nicolas-grekas: nope it's not optimizing: php is creating a new object each time

src/Extension/CoreExtension.php
```diff
@@ -1720,6 +1738,8 @@ public static function getAttribute(Environment $env, Source $source, $object, $
 
                 return \constant($object::class.'::'.$item);
             }
+        } else {
+            $item = (string) $item;
```
fabpot: Can we cast before the `if`?

nicolas-grekas: updated

fabpot:
Thank you @nicolas-grekas.

src/Extension/CoreExtension.php
```diff
@@ -1642,7 +1642,7 @@ public static function getAttribute(Environment $env, Source $source, $object, $
 
         // array
         if (Template::METHOD_CALL !== $type) {
-            $arrayItem = \is_bool($item) || \is_float($item) ? (int) $item : $item;
+            $arrayItem = \is_bool($item) || \is_float($item) ? (int) $item : $item = (string) $item;
```
gharlan, line 1645: The string casting breaks my ArrayAccess object using objects as keys, when using `myCollection[object]` in twig.

nicolas-grekas: Arf. Can you please submit a fix as a PR together with a test case?

PR #4426, opened 2024-10-30
[String] Add SpanishInflector support for singular and plural

dennistobar:
This PR implement a new Inflector using Spanish (ISO Code = es) implemented in this PR symfony/symfony#58228.

* I sort the list of supported Inflectors to follow some rule (alphabet?)
* I just change invalid test from "it" to "qq" (qq doesn't exist as valid ISO code, but it is italian... don't be evil if anyone want to create the italian inflector 😅 )

### DIFF
--- a/extra/string-extra/StringExtension.php
+++ b/extra/string-extra/StringExtension.php
@@ -15,17 +15,20 @@ use Symfony\Component\String\AbstractUnicodeString;
 use Symfony\Component\String\Inflector\EnglishInflector;
 use Symfony\Component\String\Inflector\FrenchInflector;
 use Symfony\Component\String\Inflector\InflectorInterface;
+use Symfony\Component\String\Inflector\SpanishInflector;
 use Symfony\Component\String\Slugger\AsciiSlugger;
 use Symfony\Component\String\Slugger\SluggerInterface;
 use Symfony\Component\String\UnicodeString;
+use Twig\Error\RuntimeError;
 use Twig\Extension\AbstractExtension;
 use Twig\TwigFilter;
 
 final class StringExtension extends AbstractExtension
 {
     private $slugger;
-    private $frenchInflector;
     private $englishInflector;
+    private $spanishInflector;
+    private $frenchInflector;
 
     public function __construct(?SluggerInterface $slugger = null)
     {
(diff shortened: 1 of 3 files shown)


extra/string-extra/StringExtension.php
```diff
@@ -79,10 +81,12 @@ public function singular(string $value, string $locale = 'en', bool $all = false
     private function getInflector(string $locale): InflectorInterface
     {
         switch ($locale) {
-            case 'fr':
-                return $this->frenchInflector ?? $this->frenchInflector = new FrenchInflector();
             case 'en':
                 return $this->englishInflector ?? $this->englishInflector = new EnglishInflector();
+            case 'es':
+                return $this->spanishInflector ?? $this->spanishInflector = new SpanishInflector();
```
stof, line 91: this needs to handle the case where SpanishInflector does not exist in `symfony/string`, as it does not exist for all versions supported by the requirement.

stof (changes requested):

extra/string-extra/StringExtension.php
dennistobar: Hi, thanks for the comment, do you any example to make the check?, and how can I test it?.

stof:
```php
if (!class_exists(SpanishInflector::class)) {
     // throw exception with an helpful message
}
```

For testing, you would have to define 2 different tests for the 2 cases, with different `CONDITION` sections to skip the one that does not correspond to the availability of the class

extra/string-extra/Tests/Fixtures/singular-invalid-language.test
```diff
@@ -1,10 +1,10 @@
...
 "singular" filter
 --TEMPLATE--
-{{ 'partitions'|singular('it') }}
+{{ 'partitions'|singular('qq') }}
 
 --DATA--
 return []
 
 --EXCEPTION--
-Twig\Error\RuntimeError: An exception has been thrown during the rendering of a template ("Locale "it" is not supported.") in "index.twig" at line 2.
```
fabpot: Can you revert these changes were you've removed the end line?

dennistobar: done

fabpot:
Thank you @dennistobar.

PR #4427, opened 2024-10-30
Use conditional return type for `slice`

ruudk:
This helps PHPStan understand the return type when a string is passed.

### DIFF
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -686,12 +686,13 @@ final class CoreExtension extends AbstractExtension
     /**
      * Slices a variable.
      *
-     * @param mixed $item         A variable
-     * @param int   $start        Start of the slice
-     * @param int   $length       Size of the slice
-     * @param bool  $preserveKeys Whether to preserve key or not (when the input is an array)
+     * @template TValue
+     * @param \Traversable<TValue>|array<TValue>|string $item         A variable
+     * @param int                                       $start        Start of the slice
+     * @param int                                       $length       Size of the slice
+     * @param bool                                      $preserveKeys Whether to preserve key or not (when the input is an array)
      *
-     * @return mixed The sliced variable
+     * @return ($item is string ? string : array<TValue>)
      *
      * @internal
      */


ruudk:
/cc @stof @fabpot 
Ready for review

src/Extension/CoreExtension.php
```diff
@@ -686,12 +686,12 @@ public static function merge(...$arrays): array
...
-     * @param int   $start        Start of the slice
-     * @param int   $length       Size of the slice
-     * @param bool  $preserveKeys Whether to preserve key or not (when the input is an array)
+     * @param \Traversable|array|string $item         A variable
+     * @param int                       $start        Start of the slice
+     * @param int                       $length       Size of the slice
+     * @param bool                      $preserveKeys Whether to preserve key or not (when the input is an array)
      *
-     * @return mixed The sliced variable
+     * @return ($item is string ? string : array)
```
fabpot: Any other method that would benefit from this change?

ruudk: Probably, and when I find them, I will bring it up for PR. 

It's too much to do at once. So I'm running TwigStan on our project and currently fixing the things I encounter.

fabpot: > Probably, and when I find them, I will bring it up for PR.
> 
> It's too much to do at once. So I'm running TwigStan on our project and currently fixing the things I encounter.

Ok, I will wait then.

ruudk: Can we merge this PR for now? I'm working occasionally on TwigStan and will propose new PR's when I find new things.

stof: This could be more precise for the case of array as well, by making it generic:

```
/**
 * @template T
 * @param \Traversable<T>|T[]|string $item
 * @return ($item is string ? string : T[])
 */
```

This would allow TwigStan to preserve the type of item values.

ruudk: Good idea. I updated the PR so that it also handles TKey.

stof: The handling of TKey is wrong. It will produce bad results when phpstan infers the key type as a union of literals rather than as `int`. phpstan uses a custom type extension to implement the return type of `array_slice`, not phpdoc.

ruudk: Ok, I changed it back to value only. We can address this later with an extension in TwigStan to further refine it like PHPStan does.

ruudk:
@fabpot this one is ready as well

ruudk:
Because the argument is already mixed, I can add proper typing through a PHPStan extension.

See https://github.com/twigstan/twigstan/pull/90

Therefore, this can be closed.

PR #4429, opened 2024-10-31
Add types for referenced extensions inside Template

ruudk:
By introducing a `getExtension` method in `Template` we can help PHPStan with the proper return type.

/cc @stof

### DIFF
--- a/src/Node/Expression/CallExpression.php
+++ b/src/Node/Expression/CallExpression.php
@@ -52,7 +52,7 @@ abstract class CallExpression extends AbstractExpression
                     // Compile a non-optimized call to trigger a \Twig\Error\RuntimeError, which cannot be a compile-time error
                     $compiler->raw(\sprintf('$this->env->getExtension(\'%s\')', $class));
                 } else {
-                    $compiler->raw(\sprintf('$this->extensions[\'%s\']', ltrim($class, '\\')));
+                    $compiler->raw(\sprintf('$this->getExtension(\'%s\')', ltrim($class, '\\')));
                 }
 
                 $compiler->raw(\sprintf('->%s', $callable[1]));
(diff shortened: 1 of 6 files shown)


src/Node/Expression/CallExpression.php
```diff
@@ -52,7 +52,7 @@ protected function compileCallable(Compiler $compiler)
                     // Compile a non-optimized call to trigger a \Twig\Error\RuntimeError, which cannot be a compile-time error
                     $compiler->raw(\sprintf('$this->env->getExtension(\'%s\')', $class));
                 } else {
-                    $compiler->raw(\sprintf('$this->extensions[\'%s\']', ltrim($class, '\\')));
+                    $compiler->raw(\sprintf('$this->getExtension(\'%s\')', ltrim($class, '\\')));
```
ruudk, line 55: I wonder, why aren't we always using `$this->env->getExtension`? That one already has proper typing.

stof: maybe this was a performance optimization to avoid extra method calls (based on the comment a file lines above).

This could actually be supported for the existing code if https://github.com/phpstan/phpstan/issues/9521 was implemented in phpstan, as the type of `Template::$extensions` is `class-string-map<T of ExtensionInterface, T>`

ruudk: Yes, unfortunately we don't have it yet.

ruudk:
@fabpot ready for review

ruudk:
@stof @fabpot what to do with this?

fabpot:
I'm 👎 on this one as I don't want to change the code just to please a static analyzer.

ruudk:
Understood, thanks!

PR #4432, opened 2024-11-02
Fix mistake in docs for `keys` filter

ruudk:


xabbuh (approved):

doc/filters/keys.rst
```diff
@@ -9,7 +9,7 @@ when you want to iterate over the keys of a sequence or a mapping:
     {% for key in [1, 2, 3, 4]|keys %}
         {{ key }}
     {% endfor %}
-    {# outputs: 1 2 3 4 #}
+    {# outputs: 0 1 2 3 #}
```
fabpot, line 12: I'm wondering if we should also change the example as it's quite confusing. Maybe we can use `['a', 'b', 'c', 'd']` instead?

ruudk: Good idea, done!

fabpot:
Thank you @ruudk.

PR #4434, opened 2024-11-04
Add the include_only function

fabpot:


doc/deprecated.rst
```diff
@@ -28,6 +28,11 @@ Functions
 
   Note that it won't be removed in 4.0 to allow a smoother upgrade path.
 
+* The ``include`` function is deprecated as of Twig 3.15, use ``render``
+  instead. The ``render`` function does the same as ``include``, but the
+  rendered template does **not** have access to the current context (all
+  variables must be passed explicitely).
```
jdreesen:
```suggestion
  variables must be passed explicitly).
```

doc/functions/render.rst
```diff
@@ -0,0 +1,90 @@
...
+    {{ render(['page_detailed.html', 'page.html']) }}
+
+If ``ignore_missing`` is set, it will fall back to rendering nothing if none
+of the templates exist, otherwise it will throw an exception.
+
+Variables
+---------
+
+Rendered templates **do not** have access to the variables defined in the
+context, but you can pass variables explicitely:
```
jdreesen:
```suggestion
context, but you can pass variables explicitly:
```

fabpot:
Maybe it would be good to also deprecate the `include` tag in 3.15.
My only "concern" is about the `embed` tag that is kind of a glorified include node behind the scene.

alexander-schranz:
@fabpot wanted to give some feedback. I really like the new include decision without context by default. Even I'm not a fan of renaming it `render`, but I understand from BC reasons for better transparency. I'm working the last two years where we only use `with_context = false` to make things more transparent what variables are accessed by components. Only case which was confusing is the `embed` tag which we use for things like `overlays`, `fieldsets`, ...

Typically looks like this:

```twig
{# overlay.html.twig #}
<dialog>
   <h3>
       {% block title %}{% endblock %}
   </h3>
   
   <div>
       {% block content %}{% endblock %}
   </div>
</dialog>
```

```twig
{# using.html.twig #}
{% set title = 'Title' %}
{% set content = 'Content' %}

{% embed "overlay.html.twig" only %}
    {% block title %}
        {{ title }} {# not available #}
    {% endblock %}
    {% block content %}
        <div>
            {{ content }} {# not available #}
        </div>
    {% endblock %}
{% endembed %}
```

To get this work I need give the variables also to overlay.html.twig which may have similar variables defined.

```twig
{# using.html.twig #}
{% set title = 'Title' %}
{% set content = 'Content' %}

{% embed "overlay.html.twig" {title: title, content: content} only %}
    {% block title %}
        {{ title }}
    {% endblock %}
    {% block content %}
        <div>
            {{ content }}
        </div>
    {% endblock %}
{% endembed %}
```

So it is a little bit strange. That I need to give the variables to the `overlay.html.twig` to access them in my own twig file. Better would be only have to give variables in it which are really used by the overlay.html.twig and block of the embed still has access to the one by its own file:


```twig
{# using.html.twig #}
{% set title = 'Title' %}
{% set content = 'Content' %}

{% embed "overlay.html.twig" only %}
    {% block title %}
        {{ title }} {# still have access to this one #}
    {% endblock %}
    {% block content %}
        <div>
            {{ content }} {# still have access to this one #}
        </div>
    {% endblock %}
{% endembed %}
```

Another solution would be maybe make it transparency on the embed block e.g.:

```twig
{# using.html.twig #}
{% set title = 'Title' %}
{% set content = 'Content' %}

{% embed "overlay.html.twig" only %}
    {% embedblock title with { title: title} %}
        {{ title }}
    {% endembedblock %}

    {% embedblock content with { content: content} %}
        <div>
            {{ content }}
        </div>
    {% endembedblock %}
{% endembed %}
```

I hope I did make it understandable where the issues with the embed tag is.

stof:
`render` will conflict with the [Symfony function](https://symfony.com/doc/current/reference/twig_reference.html#render) rendering the response of a sub-request, which exists since 10 years in the Symfony ecosystem ? https://github.com/symfony/symfony/blob/v7.1.6/src/Symfony/Bridge/Twig/Extension/HttpKernelExtension.php#L28

stof:
@alexander-schranz the `{% embed %}` tag actually defines an anonymous template extending the provided template name, with the body of the tag being the body of that anonymous template. The blocks inside it are normal blocks, but they are blocks defined by the anonymous template.

Your proposed syntax will not work, because blocks are rendered with the context from the caller (i.e. the parent template). they have no way to access the context of the place including the template.

alexander-schranz:
@stof I understand that it is a technical challange to change that behaviour. Just wanted to give feedback that from end user perspective it just felled for me strange how it behaves currently.

willrowe:
@fabpot is the new terminology `variables` as opposed to `context` or does `context` have a different meaning?

derrabus:
> `render` will conflict with the [Symfony function](https://symfony.com/doc/current/reference/twig_reference.html#render)

Maybe we can keep the include function, but add a flag to opt into the new behavior and deprecate not doing so.

stof:
@derrabus technically, that flag already exists in `include`: `with_context = false`

derrabus:
> @derrabus technically, that flag already exists in `include`: `with_context = false`

Perfect. So let's deprecate not setting that flag to `false` explicitly?

willrowe:
I thought @derrabus meant a flag in the configuration, like how you can ignore missing variables.

stof:
@willrowe setting a flag in the Environment is bad, as it expects that the migration is done all at once, including all third-party templates. Switching the behavior of a function through a global state basically makes it unreliable for third-party code.

CHANGELOG
```diff
@@ -1,5 +1,7 @@
 # 3.15.0 (2024-XX-XX)
- 
+
+ * Deprecate the `include` function, use `render` instead
```
lyrixx: >Maybe it would be good to also deprecate the include tag in 3.15.

Yes, Let's deprecate the include tag too

lyrixx (commented):
IMHO, twig should ship something (rector plugin, raw binary script), to update ours templates

alexander-schranz:
@lyrixx think twig-cs-fixer from @VincentLanglet can do that work. Upgrading include without_context = false to the new render I think should be possible without lot of effort.

Problem is more about include where the context was not false which was default at current state. Sure quickfix could be give everywhere `_context` in it with a fixer rule but think that is something I would not suggest todo, but is sure the "easiest" upgrade way.

Example: https://github.com/VincentLanglet/Twig-CS-Fixer/pull/327

For such rule it is sure easier if `include` get renamed to `render`. As the upgrade is then easier. Else a "fixer" is not able to skip already converted includes only if we would make still possible to set `with_context = false` to mark already converted includes,  even true throws an error.

doc/functions/render.rst
```diff
@@ -0,0 +1,91 @@
...
+Templates are loaded via the current Twig loader.
+
+Templates can be a string or an expression evaluating to a string:
+
+.. code-block:: twig
+
+    {{ render('template.html') }}
+    {{ render(template_var) }}
+
+A template can also be an instance of ``\Twig\Template`` or a
```
smnandre: Passing a Template instead of a TemplateWrapper is deprecated, afaik

tests/Fixtures/functions/render/ignore_missing.test
```diff
@@ -0,0 +1,9 @@
+--TEST--
+"include" function
```
smnandre:
```suggestion
"render" function
```

romainmenke:
Please don't make this change.
Neither including or excluding context is purely good or bad.
Many teams use either approach successfully.
Many teams run into issues with either approach.

It is typical for such cases to only hear about the teams that run into issues with the current default while not hearing anything from the other side as they are happy with how things are.

Switching the default (or worse, deprecating the current default) will not actually do any good for the wider community.

Is there any dedicated issue where this was discussed in detail?

------

From a business perspective this change alone would be devastating for us.
We have thousands of twig templates we actively maintain.

We can't rewrite all of these so that each variable is passed explicitly.
There also isn't any tool that can automate this refactor.

Please do not just make such massively breaking changes.
Including other templates is the core of any templating system. Changing this aspect is unthinkable imho.

-------

I think a more prudent approach could be to add new functionality without deprecating the old behavior.

Let users pick the behavior that best suites their context/project.

------

Needing to add `_context` to each and every template include also seems horrible to me.
The entire feature of always having the context available in included templates is that you don't need to do that. :)

Why place extra burden on template authors?
Why make the easy cases harder?

CHANGELOG
romainmenke: Please don't deprecate the include tag.
Many still use it all the time.

fabpot:
Rebased, deprection is gone, function renamed to `include_only` (why? similar to `{% include "foo.twig" only %}`).

doc/functions/include_only.rst
```diff
@@ -0,0 +1,82 @@
...
+
+.. code-block:: twig
+
+    {{ include_only('template.html.twig', {name, email}) }}
+
+    {# is equivalent to #}
+
+    {{ include_only('template.html.twig', {name: name, email: email}) }}
+
+And if the expression evaluates to a ``\Twig\TemplateWrapper`` instance, Twig
```
stof: Those sections seem to be related to the template loading, and so are related to line 21-22. Having the documentation about the context (line 24-41) between them looks really weird.
I would probably even make them different sections with proper headers to make the structure of the page easier to grasp

doc/functions/include_only.rst
```diff
@@ -0,0 +1,82 @@
...
+
+    {{ include_only('page.html.twig', sandboxed: true) }}
+
+Arguments
+---------
+
+* ``template``:       The template to render
+* ``variables``:      The variables to pass to the template
+* ``ignore_missing``: Whether to ignore missing templates or not
+* ``sandboxed``:      Whether to sandbox the template or not
```
stof: I don't think this new function should have a `sandboxed` argument at all due to #4854

src/Extension/CoreExtension.php
```diff
@@ -1532,6 +1533,21 @@ public static function include(Environment $env, $context, $template, $variables
...
+     * Renders a template without giving it access to the current context.
+     *
+     * @param string|array|TemplateWrapper $template      The template to render or an array of templates to try consecutively
+     * @param array                        $variables     The variables to pass to the template
+     * @param bool                         $ignoreMissing Whether to ignore missing templates or not
+     * @param bool                         $sandboxed     Whether to sandbox the template or not
+     *
+     * @internal
+     */
+    public static function includeOnly(Environment $env, $template, $variables = [], $ignoreMissing = false, $sandboxed = false): string
```
stof: this should return a Markup object to match #4825

src/Extension/CoreExtension.php
```diff
@@ -1532,6 +1533,21 @@ public static function include(Environment $env, $context, $template, $variables
         }
     }
 
+    /**
+     * Renders a template without giving it access to the current context.
+     *
+     * @param string|array|TemplateWrapper $template      The template to render or an array of templates to try consecutively
+     * @param array                        $variables     The variables to pass to the template
```
```suggestion
     * @param array<string, mixed>         $variables     The variables to pass to the template
```

This will help when using twigstan by providing the precise types.

src/Extension/CoreExtension.php
```diff
@@ -1532,6 +1533,21 @@ public static function include(Environment $env, $context, $template, $variables
         }
     }
 
+    /**
+     * Renders a template without giving it access to the current context.
+     *
+     * @param string|array|TemplateWrapper $template      The template to render or an array of templates to try consecutively
```
```suggestion
     * @param string|array<string|TemplateWrapper>|TemplateWrapper $template      The template to render or an array of templates to try consecutively
```

stof:
The PR title should be updated to reflect the new state.

doc/functions/include_only.rst
fabpot: indeed

doc/functions/include_only.rst
fabpot: reworked

fabpot:
> The PR title should be updated to reflect the new state.

Updated the title and took your comments into account.

fabpot:
Added an additional commit that might be controversial where I make include_only the prefered function and include redirects to it for most of the docs.

src/Extension/CoreExtension.php
```diff
@@ -1532,6 +1533,22 @@ public static function include(Environment $env, $context, $template, $variables
...
+     *
+     * @param string|array<string|TemplateWrapper>|TemplateWrapper $template      The template to render or an array of templates to try consecutively
+     * @param array<string, mixed>                                 $variables     The variables to pass to the template
+     * @param bool                                                 $ignoreMissing Whether to ignore missing templates or not
+     *
+     * @return string|Markup
+     *
+     * @internal
+     */
+    public static function includeOnly(Environment $env, $template, $variables = [], $ignoreMissing = false)
```
stof, line 1547:
```suggestion
    public static function includeOnly(Environment $env, $template, array $variables = [], bool $ignoreMissing = false)
```

fabpot: Done

PR #4435, opened 2024-11-04
Fix wrong type for cycle position

ruudk:
It also accepts 0, which is not a positive int.


ruudk:
@fabpot please have a look

fabpot:
Thank you @ruudk.

PR #4440, opened 2024-11-06
Limits ensureToStringAllowed recursion to arrays and objects

### LINKED ISSUE #4439: Regression: GHSA-6377-hfv9-hqf6 fix can exceed recursion limits / OOM
Several of us who updated twig/twig:3.14.1 on our Drupal web sites have seen that this new recursive logic causes OOM or stack limits. This seems related to [https://github.com/twigphp/Twig/security/advisories/GHSA-6377-hfv9-hqf6](GHSA-6377-hfv9-hqf6).

With xdebug enable, this is displayed as

    Error: Xdebug has detected a possible infinite loop, and aborted your script with a stack depth of '1000' frames in Twig\Extension\SandboxExtension->isSandboxed() (line 56 of /var/www/html/vendor/twig/twig/src/Extension/SandboxExtension.php).
    
    Twig\Extension\SandboxExtension->ensureToStringAllowed() (Line: 124)
    Twig\Extension\SandboxExtension->ensureToStringAllowed() (Line: 93)

Drupal form render arrays are often very nested and very deep. I do not quite understand the twig sandbox code so I do not know if this is fixable downstream, but it seems like we would only need to recurse if the value is an object rather than on every array/list value.

Any help would be appreciated.

**Steps to reproduce**

1. Install Drupal 10.3 or >11.0.1 (helpful to use Standard profile) 
2. Update to twig/twig:3.14.1
3. Log in as an administrative user.
4. Attempt to create a node.

[drupal.org issue](https://www.drupal.org/project/drupal/issues/3485956)

mradcliffe:
Resolves #4439 

Maybe.

This would need security review.

### DIFF
--- a/src/Extension/SandboxExtension.php
+++ b/src/Extension/SandboxExtension.php
@@ -121,7 +121,9 @@ final class SandboxExtension extends AbstractExtension
     {
         if (\is_array($obj)) {
             foreach ($obj as $v) {
-                $this->ensureToStringAllowed($v, $lineno, $source);
+                if (\is_object($v) || \is_array($v)) {
+                    $this->ensureToStringAllowed($v, $lineno, $source);
+                }
             }
 
             return $obj;


src/Extension/SandboxExtension.php
```diff
@@ -121,7 +121,9 @@ public function ensureToStringAllowed($obj, int $lineno = -1, ?Source $source =
     {
         if (\is_array($obj)) {
             foreach ($obj as $v) {
-                $this->ensureToStringAllowed($v, $lineno, $source);
+                if (\is_object($v) || \is_array($v)) {
```
mradcliffe, line 124: is_array still causes the issues in Drupal, and I don't think it would be acceptable _not_ to check that because we need to continue checking arrays.

heddn:
This didn't resolve the issue for me.

mradcliffe:
Thanks, @heddn 👋🏼 . I was too eager and also added in the is_array check, which I think is necessary. When I was playing around with it, only checking `is_object($v)` worked.

src/Extension/SandboxExtension.php
heddn: is_object works.

mradcliffe:
With Drupal, we explicitly allow __toString in \Drupal\Core\Template\TwigSandboxPolicy so do we even need to use this method at all? If we don't need the check here, would the solution be to maintain a separate DrupalSandboxExtension for the 3.x branch? :(

Perhaps in 4.x, SecurityPolicyInterface would have a new method that would help performance here since Drupal's policy allows __toString (among other methods). `SecurityPolicyInterface::isToStringAllowed()`?

larowlan:
See also https://github.com/twigphp/Twig/pull/4441

mradcliffe:
That seems a lot better. I'm closing this.

fabpot:
I've just released new versions with a fix.

PR #4441, opened 2024-11-07
Add recursion prevention

### LINKED ISSUE #4439: Regression: GHSA-6377-hfv9-hqf6 fix can exceed recursion limits / OOM

larowlan:
Resolves #4439

Drupal allows printing arrays direct in templates, as we have a node visitor/render function that can 'render' our 'render arrays'.

This adds some recursion prevention.

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -488,7 +488,7 @@ won't be allowed and will generate a ``\Twig\Sandbox\SecurityError`` exception.
 
 .. note::
 
-    As of Twig 1.14.1 (and on Twig 3.11.2), if the ``Article`` class implements
+    As of Twig 3.14.1 (and on Twig 3.11.2), if the ``Article`` class implements
     the ``ArrayAccess`` interface, the templates will only be able to access
     the ``title`` and ``body`` attributes.
 
--- a/src/Extension/SandboxExtension.php
+++ b/src/Extension/SandboxExtension.php
@@ -26,11 +26,14 @@ final class SandboxExtension extends AbstractExtension
     private $policy;
     private $sourcePolicy;
 
+    static array $recursionProjection = [];
+
     public function __construct(SecurityPolicyInterface $policy, $sandboxed = false, ?SourcePolicyInterface $sourcePolicy = null)
     {
         $this->policy = $policy;
         $this->sandboxedGlobally = $sandboxed;
         $this->sourcePolicy = $sourcePolicy;
+        static::$recursionProjection = [];
     }
 
     public function getTokenParsers(): array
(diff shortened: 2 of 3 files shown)


mstrelan:
This didn't work for me. Many of our tests started failing with 3.14.1. Picking one of those tests, with 3.14.0 it passes in 19 seconds. With 3.14.1 it produces a 500 error in around 4 seconds. With this PR I cancelled the test after 3 minutes or so. I guess instead of reaching a recursion limit we now just have an infinite loop. I have no looked in to the specifics or traced this with a debugger.

werk21:
We hotfixed this with this: https://www.drupal.org/files/issues/2024-11-07/3485956-twig.diff
Caution, only a hot fix, not a real solution. This will probably reduce the effectiveness of the original security patch.

nicolas-grekas (approved):
Thanks for the PR.
I improved the patch in the second commit.
:rocket:

stefan-korn:
why is this patch going against https://github.com/twigphp/Twig/tree/3.11.x, shouldn't this merge with https://github.com/twigphp/Twig/tree/3.x ?

nicolas-grekas:
We merge in the lowest maintained branch, and then we merge that branch up.

fabpot:
I've just released new versions.

larowlan:
Thank you, that was fast, all happened whilst I slept ❤️

PR #4447, opened 2024-11-11
Clarify documentation for escape filter

dmjohnsson23:
It was somewhat unclear from the documentation what the intended purpose of the 'js' escape strategy is. I wasn't certain *where* exactly in JavaScript such strings were intended to be output. Someone (by that I mean me--though maybe I'm just an idiot...) might inadvertently think the filter was meant to be used somewhere in actual JavaScript code, rather than simply in strings.


stof (approved):

fabpot:
Thank you @dmjohnsson23.

Issue #4453, opened 2024-11-19
[3.15] Public constants get precedence over `is*` method calls [BC Break]


ruudk:
We have the following class that is passed as context to the template:
```php
<?php
final class SomeClass {
    public const string available = 'available';

    public function __construct(private string $value) {}

    public function isAvailable() : bool {
        return $this->value === self::available;
    }
}
```

We use it like this:
```twig
{% if someObject.available %}
  Available
{% endif %}
```

On 3.14 this would go to `isAvailable()` and only be true when the value was `available`.

But on 3.15 this suddenly changed. 

We suspect that it now goes to the public `available` constant and therefore it's now always true, even when the value is something else.

On one hand, this feels logical. But it was unexpected, causing issues in production. We think this is a backwards compatibility break.

/cc @stof @fabpot

jdreesen:
See: https://github.com/twigphp/Twig/pull/4353#issuecomment-2377923327

/cc @nicolas-grekas

fabpot:
Using the following would fix the issue:

```php
public const string AVAILABLE = 'available';
```

Note the upper case `AVAILABLE` instead of `available`.

The (wrong) assumption referenced above was that constants are "always" uppercased.

ruudk:
Thanks. The class in question was automatically generated and thus not following the widely used upper snake case for constants.

So we can fix this on our side.

But I do think this was a BC break that can have bigger consequences when people are not aware.

nicolas-grekas:
Let's close for now. We can always reconsider but if this is easy enough to change on user's side, the BC break is worth it since it enables seamless usage of enums.

ericmorand:
> The (wrong) assumption referenced above was that constants are "always" uppercased.

God.

> the BC break is worth it

Oh dear god.

PR #4454, opened 2024-11-19
Add BC break note in CHANGELOG for 3.15

nicolas-grekas:
Let's add this hint for people affected by the behavior change, as in https://github.com/twigphp/Twig/issues/4453


fabpot (approved):

fabpot:
Thank you @nicolas-grekas.

PR #4455, opened 2024-11-20
Simplify enum usage in docs

ruudk:
There is no need to use `()`.

### DIFF
--- a/doc/functions/enum.rst
+++ b/doc/functions/enum.rst
@@ -13,7 +13,7 @@
     {{ enum('App\\MyEnum').SomeCase.value }}
 
     {# get all cases of an enum #}
-    {% enum('App\\MyEnum').cases() %}
+    {% enum('App\\MyEnum').cases %}
 
     {# call any methods of the enum class #}
     {% enum('App\\MyEnum').someMethod() %}


doc/functions/enum.rst
```diff
@@ -13,10 +13,10 @@
     {{ enum('App\\MyEnum').SomeCase.value }}
 
     {# get all cases of an enum #}
-    {% enum('App\\MyEnum').cases() %}
+    {% enum('App\\MyEnum').cases %}
 
     {# call any methods of the enum class #}
-    {% enum('App\\MyEnum').someMethod() %}
+    {% enum('App\\MyEnum').someMethod %}
```
fabpot: I would keep this one as a method might have some arguments (maybe we should add some here).

ruudk: What about

```suggestion
    {% enum('App\\MyEnum').someMethod %}
    {% enum('App\\MyEnum').someMethod(someArg) %}
```

fabpot: That would work for me even if this is really just how method calls work by default in Twig, so nothing specific here.

ruudk:
@fabpot I undid the change to the method call. Let's keep it like this.

fabpot:
Thank you @ruudk.

PR #4456, opened 2024-11-20
Add missing import

VincentLanglet:
Hi @fabpot, the phpdoc was updated in https://github.com/twigphp/Twig/pull/4367 but the import was forgotten.
This leads to an error with all static analysis tools when implementing ExtensionInterface.

If possible a patch version would be helpful.

(Tests are already failing on 3.x)


fabpot:
Thank you @VincentLanglet.

ruudk:
@fabpot Would it be an idea to set-up PHPStan for this project? This could have been detected earlier.

nvdbeek:
@fabpot when is this fix coming in a tag?

PR #4458, opened 2024-11-20
Add phpstan analysis

VincentLanglet:
As suggested by @ruudk here: https://github.com/twigphp/Twig/pull/4456#issuecomment-2488087625
I think it could be useful to add phpstan analysis on twig.

This is an example of all the things reported when checking level 3 PHPStan (more will get more errors to fix).


src/ExpressionParser.php
```diff
@@ -53,9 +53,9 @@ class ExpressionParser
     public const OPERATOR_LEFT = 1;
     public const OPERATOR_RIGHT = 2;
 
-    /** @var array<string, array{precedence: int, class: class-string<AbstractUnary>}> */
+    /** @var array<string, array{precedence: int, precedence_change?: OperatorPrecedenceChange, class: class-string<AbstractUnary>}> */
```
VincentLanglet, line 56: Phpdoc was forgotten to be updated here when introducing OperatorPrecedenceChange

src/ExpressionParser.php
```diff
@@ -125,7 +125,7 @@ public function parseExpression($precedence = 0)
 
             $expr->setAttribute('operator', 'binary_'.$token->getValue());
 
-            $this->triggerPrecedenceDeprecations($expr, $token);
+            $this->triggerPrecedenceDeprecations($expr);
```
VincentLanglet, line 128: The method only accepts one argument

src/ExpressionParser.php
```diff
@@ -246,7 +246,7 @@ private function getPrimary(): AbstractExpression
             $expr->setAttribute('operator', 'unary_'.$token->getValue());
 
             if ($this->deprecationCheck) {
-                $this->triggerPrecedenceDeprecations($expr, $token);
+                $this->triggerPrecedenceDeprecations($expr);
```
VincentLanglet, line 249: The method only accepts one argument

src/ExtensionSet.php
```diff
@@ -48,7 +48,7 @@ final class ExtensionSet
     private $unaryOperators;
     /** @var array<string, array{precedence: int, class?: class-string<AbstractExpression>, associativity: ExpressionParser::OPERATOR_*}> */
     private $binaryOperators;
-    /** @var array<string, mixed> */
+    /** @var array<string, mixed>|null */
```
VincentLanglet, line 51: We set null on this value sometimes.

src/Node/Expression/BlockReferenceExpression.php
```diff
@@ -28,7 +28,7 @@ class BlockReferenceExpression extends AbstractExpression
     public function __construct(Node $name, ?Node $template, int $lineno)
     {
         if (!$name instanceof AbstractExpression) {
-            trigger_deprecation('twig/twig', '3.15', 'Not passing a "%s" instance to the "node" argument of "%s" is deprecated ("%s" given).', AbstractExpression::class, static::class, get_class($node));
+            trigger_deprecation('twig/twig', '3.15', 'Not passing a "%s" instance to the "node" argument of "%s" is deprecated ("%s" given).', AbstractExpression::class, static::class, get_class($name));
```
VincentLanglet, line 31: $node was not defined.

src/Node/Node.php
```diff
@@ -186,17 +186,11 @@ public function removeAttribute(string $name): void
         unset($this->attributes[$name]);
     }
 
-    /**
-     * @param string|int $name
```
VincentLanglet: Int is not accepted

src/OperatorPrecedenceChange.php
```diff
@@ -35,7 +35,7 @@ public function getVersion(): string
         return $this->version;
     }
 
-    public function getNewPrecedence(): string
+    public function getNewPrecedence(): int
```
VincentLanglet, line 38: The constructor accepts int, so the return type should be int to no ?
cc @fabpot since you're the one who made this class.

src/Node/ImportNode.php
```diff
@@ -25,9 +25,6 @@
 #[YieldReady]
 class ImportNode extends Node
 {
-    /**
-     * @param bool $global
-     */
     public function __construct(AbstractExpression $expr, AbstractExpression|AssignTemplateVariable $var, int $lineno)
     {
         if (!\is_bool(\func_num_args() > 3)) {
```
stof: this check is broken. `\func_num_args() > 3` is always a boolean

VincentLanglet:
Tests were already failing and fabbot error is unrelated
https://fabbot.io/report/twigphp/Twig/4458/93ca91b81d9be93d0a21441c03ef24ed7c274ab0

src/Node/MacroNode.php
```diff
@@ -85,7 +87,9 @@ public function compile(Compiler $compiler): void
             ->indent()
         ;
 
-        foreach ($this->getNode('arguments')->getKeyValuePairs() as $pair) {
+        /** @var ArrayExpression $arguments */
+        $arguments = $this->getNode('arguments');
```
stof: no need to assign the variable a second time. You can reuse it from line 68

src/OperatorPrecedenceChange.php
stof: indeed

src/Node/ImportNode.php
VincentLanglet: Indeed. I changed to 
```
if (\func_num_args() > 3) {
```

This would have been reported by phpstan on a higher level I think.

phpstan.neon.dist
```diff
@@ -0,0 +1,9 @@
+includes:
+    - phpstan-baseline.neon
+
+parameters:
+    level: 3
```
stof, line 5: level 3 is quite weak. Can't we go at a higher level (possibly using `inferPrivatePropertyTypeFromConstructor` for the 3.x branch where property are not using native types yet) ?

VincentLanglet: The level 4 get `[ERROR] Found 75 errors`.

I plan to bump PHPStan level, but I think it's better
- to have a small pr introducing pr
- to have a pr for level 4
- to have another pr for level 5
- ...

This will give an easier review (and avoid myself to fix everything and getting the pr stale/closed)

src/Node/Node.php
fabpot: It is accepted explicitly in Twig 4.0 (same for all similar methods).

fabpot:
@VincentLanglet Can you also rebase to see (as stof fixed a bunch of test issues)

VincentLanglet:
> @VincentLanglet Can you also rebase to see (as stof fixed a bunch of test issues)

I rebased and added back the `int|string` phpdoc @fabpot 

In the next PR I'll try to bump level 4

fabpot:
Thank you @VincentLanglet.

VincentLanglet:
There is some fixes in this PR, would it be enough for a 3.15.1 version ?

PR #4459, opened 2024-11-20
Fix the exception message to match the expected one for not ready nodes

stof:
This fixes the tests that got broken in https://github.com/twigphp/Twig/pull/4446 because that PR updated the expected message to include the FQCN of the attribute for both the deprecation and the exception, but forgot to use the FQCN in the actual exception.


nicolas-grekas (approved):

xabbuh (approved):

fabpot:
Thank you @stof.

PR #4460, opened 2024-11-20
Fix the string-extra tests when running with older symfony/string

stof:
My review comment in https://github.com/twigphp/Twig/pull/4426#discussion_r1822452803 was not taken into account to handle the case of using a version of `symfony/string` without the SpanishInflector, and the PR was merged with broken tests.


nicolas-grekas (approved):

xabbuh (approved):

fabpot:
Thank you @stof.

PR #4461, opened 2024-11-20
Fix the intl-extra tests

stof:
symfony/intl has updated its data from ICU 75.1 to ICU 76.1, which includes new currency and scripts.


nicolas-grekas (approved):

xabbuh (approved):

fabpot:
Thank you @stof.

PR #4464, opened 2024-11-20
Use phpstan level 4

VincentLanglet:


src/Environment.php
```diff
@@ -139,10 +155,7 @@ public function __construct(LoaderInterface $loader, array $options = [])
         $escaperExt = new EscaperExtension($options['autoescape']);
         $escaperExt->setEnvironment($this, false);
         $this->addExtension($escaperExt);
-        if (\PHP_VERSION_ID >= 80000) {
```
VincentLanglet, line 142: Unused code since "php": ">=8.0.2", in composer.json

src/Environment.php
```diff
@@ -873,7 +888,7 @@ public function getUnaryOperators(): array
     /**
      * @internal
      *
-     * @return array<string, array{precedence: int, class: class-string<AbstractBinary>, associativity: ExpressionParser::OPERATOR_*}>
+     * @return array<string, array{precedence: int, class: class-string<AbstractExpression>, associativity: ExpressionParser::OPERATOR_*}>
```
VincentLanglet, line 891: Not always an AbstractBinary since CoreExtension::getOperators returns `NullCoalesceExpression` which is not extending AbstractBinary.

I did the same for Unary just in case.

src/ExpressionParser.php
```diff
@@ -753,7 +751,6 @@ private function parseTestExpression(Node $node): TestExpression
             $node = new MacroReferenceExpression($alias['node']->getNode('var'), $alias['name'], new ArrayExpression([], $node->getTemplateLine()), $node->getTemplateLine());
         }
 
-        $ready = $test instanceof TwigTest;
```
VincentLanglet, line 756: Unused code

src/Extension/CoreExtension.php
```diff
@@ -1223,7 +1223,7 @@ public static function length(string $charset, $thing): int
             return mb_strlen($thing, $charset);
         }
 
-        if ($thing instanceof \Countable || \is_array($thing) || $thing instanceof \SimpleXMLElement) {
```
VincentLanglet, line 1226: SimpleXMLElement is countable

src/Node/Expression/Binary/EqualBinary.php
```diff
@@ -15,23 +15,6 @@
 
 class EqualBinary extends AbstractBinary
 {
-    public function compile(Compiler $compiler): void
-    {
-        if (\PHP_VERSION_ID >= 80000) {
```
VincentLanglet, line 20: Always true since "php": ">=8.0.2", in composer.json ; same for others

src/Node/Expression/CallExpression.php
```diff
@@ -327,7 +327,10 @@ private function getTwigCallable(): TwigCallableInterface
     {
         $current = $this->getAttribute('twig_callable');
 
-        $this->setAttribute('twig_callable', match ($this->getAttribute('type')) {
+        /** @var 'test'|'function'|'filter' $type */
```
VincentLanglet, line 330: To avoid phpstan reporting that we don't support all match cases.

src/Node/Expression/NameExpression.php
```diff
@@ -41,20 +41,12 @@ public function compile(Compiler $compiler): void
         if ($this->getAttribute('is_defined_test')) {
             if (isset($this->specialVars[$name])) {
                 $compiler->repr(true);
-            } elseif (\PHP_VERSION_ID >= 70400) {
```
VincentLanglet, line 44: Always true

src/NodeVisitor/EscaperNodeVisitor.php
```diff
@@ -67,7 +67,7 @@ public function enterNode(Node $node, Environment $env): Node
         return $node;
     }
 
-    public function leaveNode(Node $node, Environment $env): ?Node
```
VincentLanglet, line 70: PHPStan report the null type as useless.

Since the class is final we can remove it.

src/Parser.php
```diff
@@ -408,7 +408,7 @@ private function filterBodyNodes(Node $node, bool $nested = false): ?Node
         // Node::class !== \get_class($node) should be removed in Twig 4.0
         $nested = $nested || (Node::class !== \get_class($node) && !$node instanceof Nodes);
         foreach ($node as $k => $n) {
-            if (null !== $n && null === $this->filterBodyNodes($n, $nested)) {
```
VincentLanglet, line 411: $n is never null ; when doing foreach on node, it's `\IteratorAggregate<int|string, Node>`

src/Util/ReflectionCallable.php
```diff
@@ -63,7 +63,7 @@ public function __construct(
             $this->name = $class->name.'::'.$r->name;
         } elseif (\PHP_VERSION_ID < 80111 && $class = $r->getClosureScopeClass()) {
             $callable = [\is_array($callable) ? $callable[0] : $class->name, $r->name];
-            $this->name = (\is_array($callable) ? $callable[0] : $class->name).'::'.$r->name;
```
VincentLanglet, line 66: Always an array.

fabpot:
I think we should focus on Twig 4 instead for the higher level.

VincentLanglet:
> I think we should focus on Twig 4 instead for the higher level.

Ok, I'll do others PR on Twig 4 then.

I'll wait for the 3.x merge into 4.x first.

VincentLanglet:
> I think we should focus on Twig 4 instead for the higher level.

I reopened the PR on 4.x. https://github.com/twigphp/Twig/pull/4494

PR #4466, opened 2024-11-22
Fix PHPStan for Twig 4.x

VincentLanglet:
Hi @fabpot, to help introducing PHPStan on 4.x, I made the merge 3.x into 4.x
- Fixing the conflits
- Fixing PHPStan build

I have
- A commit with the merge https://github.com/twigphp/Twig/pull/4466/commits/6a82e16617183c793bad4cfa672cee3c3923b714

- A commit with the update needed to get PHPStan build green in 4.x https://github.com/twigphp/Twig/pull/4466/commits/cb89b7528c997dd11a1a28192eba11a9fdbe5f5f


src/Environment.php
```diff
@@ -54,7 +55,7 @@ class Environment
     private LoaderInterface $loader;
     private bool $debug;
     private bool $autoReload;
-    private CacheInterface|string|false $cache;
```
VincentLanglet, line 57: Type was wrong, it's always CacheInterface unlike the originalCache.

src/Node/EmptyNode.php
```diff
@@ -26,7 +26,7 @@ public function __construct(int $lineno = 0)
         parent::__construct([], [], $lineno);
     }
 
-    public function setNode(string $name, Node $node): void
+    public function setNode(string|int $name, Node $node): void
```
VincentLanglet: Parent accepts string|int now.

src/Runtime/LoopIterator.php
```diff
@@ -45,11 +45,14 @@ public function __construct($seq)
             $this->seq = new \ArrayIterator($seq);
         } elseif ($seq instanceof \IteratorAggregate) {
             do {
+                /** @var \Iterator<TKey, TValue> $seq */
                 $seq = $seq->getIterator();
             } while ($seq instanceof \IteratorAggregate);
             $this->seq = $seq;
-        } elseif (is_iterable($seq)) {
```
VincentLanglet, line 51: Is iterable is not always an Iterator but a Traversable.

So we technically have to use IteratorIterator if it's not an Iterator.

fabpot:
@VincentLanglet I've just merged 3.x into 4.x to keep this PR only about phpstan.

src/ExpressionParser.php
```diff
@@ -51,9 +49,9 @@ class ExpressionParser
     public const OPERATOR_LEFT = 1;
     public const OPERATOR_RIGHT = 2;
 
-    /** @var array<string, array{precedence: int, precedence_change?: OperatorPrecedenceChange, class: class-string<AbstractUnary>}> */
+    /** @var array<string, array{precedence: int, precedence_change?: OperatorPrecedenceChange, class: class-string<AbstractExpression>}> */
     private array $unaryOperators;
-    /** @var array<string, array{precedence: int, precedence_change?: OperatorPrecedenceChange, class: class-string<AbstractBinary>, associativity: self::OPERATOR_*}> */
+    /** @var array<string, array{precedence: int, precedence_change?: OperatorPrecedenceChange, class: class-string<AbstractExpression>, associativity: self::OPERATOR_*}> */
```
fabpot: These 2 changes look wrong to me.

src/Runtime/LoopIterator.php
fabpot: Can you add a test that would break without this change?

src/ExpressionParser.php
VincentLanglet: Those values are set in construct
```
 $this->unaryOperators = $env->getUnaryOperators();
$this->binaryOperators = $env->getBinaryOperators();
```

They call the method on `$this->extensionSet`.
The ExtensionSet gets them from multiple ExtensionInterface, 
and getOperators is
```
/**
     * Returns a list of operators to add to the existing list.
     *
     *
     * @psalm-return array{
     *     array<string, array{precedence: int, precedence_change?: OperatorPrecedenceChange, class: class-string<AbstractExpression>}>,
     *     array<string, array{precedence: int, precedence_change?: OperatorPrecedenceChange, class?: class-string<AbstractExpression>, associativity: ExpressionParser::OPERATOR_*}>
     * }
     */
    public function getOperators(): array;
```

This is because 
```
'??' => ['precedence' => 5, 'class' => NullCoalesceExpression::class, 'associativity' => ExpressionParser::OPERATOR_RIGHT],
```
is not an AbstractBinary but just an AbtractExpression.

VincentLanglet:
> Can you add a test that would break without this change?

For the iterable/Iterator issue, I think it's an issue with PHPStan instead.
So I created an issue on https://github.com/phpstan/phpstan/issues/12102 (and a fix on phpstan side)

fabpot:
> > Can you add a test that would break without this change?
> 
> For the iterable/Iterator issue, I think it's an issue with PHPStan instead. So I created an issue on [phpstan/phpstan#12102](https://github.com/phpstan/phpstan/issues/12102) (and a fix on phpstan side)

Let's revert the change then.

VincentLanglet:
> > > Can you add a test that would break without this change?
> > 
> > 
> > For the iterable/Iterator issue, I think it's an issue with PHPStan instead. So I created an issue on [phpstan/phpstan#12102](https://github.com/phpstan/phpstan/issues/12102) (and a fix on phpstan side)
> 
> Let's revert the change then.

I reverted the extra if I added.

We still need `/** @var \Iterator<TKey, TValue> $seq */` to not lose the `template type`.

src/ExpressionParser.php
fabpot: Indeed. I suppose all binary operators should extend AbstractBinary. I'm going to have a look at it.

VincentLanglet: > Indeed. I suppose all binary operators should extend AbstractBinary. I'm going to have a look at it.

The issue already exists on 3.x if you want to take a look and fix it on the lower branch.
You can ping me after the fix, I'll update the PHPDoc if needed :)

fabpot: See #4482

fabpot: This PR can now be rebased.

VincentLanglet:
I rebased @fabpot

fabpot:
Thank you @VincentLanglet.

PR #4467, opened 2024-11-22
Add support for `Twig\Markup`

nlemoine:


ruudk:
I think this should be a generic solution, not just for this specific extension.

It also feels similar to this problem I tried to solve: https://github.com/twigphp/Twig/pull/4124

nlemoine:
In the meantime, that seems legit to me. @fabpot?

faizanakram99:
> I think this should be a generic solution, not just for this specific extension.
> 
> It also feels similar to this problem I tried to solve: #4124

It already works for core extension, {{ obj }} is casted into string

nlemoine:
For reminder, this PR is about `html_classes` only (part of `html-extra`), not Twig `{{ stringableObject }}` 🙂 

~It handles cases likes:~
(edit)
See: https://github.com/twigphp/Twig/pull/4467#issuecomment-2550687890

```twig
{{ html_classes(class_a, {
    (stringableObjectThatMightBeConvertedToAClass): true
}) }}
```
Which currently throws exception.

Can you please take a closer look at it? 🙏

fabpot:
I must be missing something here, but even without your patch, the new test passes. See #4506 where I've added your test without changing the code.

nlemoine:
Sorry, the test case was the wrong one. You should [get an error](https://php-play.dev/?c=DwfgDgFmBQ0GYFcB2BjALgSwPZIAQTQFsAbAYWIEMBnKgUyoAoA6FgEgoCcBzKgSgC5cVNBwxIu0AN7Rcs3KxSUa9XAF5cAbQC6Abhly4WDrQooIuBu25Vc1eRjUA%2BeZy69c0uV9wY4FgDoYVAD6wqLilq687p7ecQpKdFTaahZhYm4u3HpxsgC%2BuLTEdL4BQcGcHBQAnpHc0R76uQZGJmYWVly2NgnUNqrOCjgAJhiYODFNzd6lDACEgSHpEb00DbHT02gQHFgA7rhItAcAogAeKLRg40gM-lRg4WhwDADkACoQtLgARAQkwUUfXoP1wiFQN1s3AQhFoSDQuAApMMLABrWjVJHDdxUCBYBDEFEAI2%2BFCEIgyuCMtjwYmEFFQ3ywfgAyhTxBQicRaAAaXBcLAIn6Iqg-JivPmsDCSoE0PlcWhoYLDWhEhBcYJoapgWiWWV8aI5TbePJTY0%2BPzzIZIUY3Sbm5ooHCYJAIWhGh24U2e2SrJIpdQMZaZP0e6be3IFIp0Rrm7a7A5HU4XK43O4PJ4vD5fX7-YiAxIgsHIdDYPCuGFwhHIoR4gnE760MZfDi2cnhLh8hk%2BJD0xlU1nsric7lU1vdyo1eWC34isUS%2BzyxXK1XqzXa3WdaK8MNeCP5WBeYxoBAcWmEMDELAqt64BeT6rBZAYACOboYD%2BCcAwxDQtA4eqFlQfL0pgKDFhCZYdAAbjEuDHqeeCvK8uBzKo6isNBOhetueimtAfqkKkSa4PqPgXtysLwjYbIdiO3wbLgYAIFyGDgeCpY4LgwSalgtEZAwAjthkUyMUeiqIXeKCvGGpp5Ho0C0GYWD4EQZBAYBfSkDuQA&v=8.1&f=html) in the current code with:

```twig
{{ html_classes(stringableObjectThatMightBeConvertedToAClass) }}
```

The test fixtures have been updated accordingly.

extra/html-extra/HtmlExtension.php
```diff
@@ -92,20 +92,20 @@ public static function htmlClasses(...$args): string
     {
         $classes = [];
         foreach ($args as $i => $arg) {
-            if (\is_string($arg)) {
-                $classes[] = $arg;
+            if (\is_string($arg) || $arg instanceof \Stringable) {
```
fabpot: I don't think doing it for all `\Stringable` objects is what we want here.

But having a specific case for `Markup` instanceof might be enough:

```suggestion
            if (\is_string($arg) || $arg instanceof Markup) {
```

fabpot (changes requested):

extra/html-extra/HtmlExtension.php
```diff
@@ -102,10 +103,10 @@ public static function htmlClasses(...$args): string
                     if (!$condition) {
                         continue;
                     }
-                    $classes[] = $class;
+                    $classes[] = (string) $class;
                 }
             } else {
-                throw new RuntimeError(\sprintf('The "html_classes" function argument %d should be either a string or an array, got "%s".', $i, get_debug_type($arg)));
+                throw new RuntimeError(\sprintf('The "html_classes" function argument %d should be either a string, an instance of %s or an array, got "%s".', $i, Markup::class, get_debug_type($arg)));
```
fabpot: I would not change the exception message as the Markup class is more an implementation detail than something people needs to be aware.

extra/html-extra/HtmlExtension.php
```diff
@@ -102,10 +103,10 @@ public static function htmlClasses(...$args): string
                     if (!$condition) {
                         continue;
                     }
-                    $classes[] = $class;
+                    $classes[] = (string) $class;
```
fabpot: Only for Makup instances.

fabpot (changes requested):

extra/html-extra/HtmlExtension.php
nlemoine: Done!

extra/html-extra/HtmlExtension.php
nlemoine: Done!

fabpot:
Thank you @nlemoine.

PR #4469, opened 2024-11-23
Fix SafeAnalysisNodeVisitor::getSafe() return value

fabpot:
`SafeAnalysisNodeVisitor::getSafe()` can return `[]` or `null`. Both mean the same thing, but `null` is partially supported as `EscaperNodeVisitor` does not support `null` everywhere (as the node visitor never set `safe` to `null`). But a third party might return `null`. This PR makes the code more robust by deprecating using `null`.

### DIFF
--- a/src/NodeVisitor/EscaperNodeVisitor.php
+++ b/src/NodeVisitor/EscaperNodeVisitor.php
@@ -172,7 +172,7 @@ final class EscaperNodeVisitor implements NodeVisitorInterface
     {
         $safe = $this->safeAnalysis->getSafe($expression);
 
-        if (null === $safe) {
+        if (!$safe) {
             if (null === $this->traverser) {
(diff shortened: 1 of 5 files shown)


src/NodeVisitor/EscaperNodeVisitor.php
```diff
@@ -196,7 +196,7 @@ private function isSafeFor(string $type, AbstractExpression $expression, Environ
             $safe = $this->safeAnalysis->getSafe($expression);
         }
 
-        return \in_array($type, $safe) || \in_array('all', $safe);
+        return null !== $safe && (\in_array($type, $safe) || \in_array('all', $safe));
```
fabpot, line 199: Even if it's never `null` with core extensions, it's a possibility that we need to take into account.

PR #4471, opened 2024-11-23
Deprecate not passing a Source to TokenStream

### LINKED ISSUE #4457: Remove internal annotation for TokenStream::getSourceContext
In a Twig extension of mine I have code similar to this, where I call `TokenStream::getSourceContext` to initialize a SyntaxError:

```php
use Twig\Error\SyntaxError;
use Twig\Node\Node;
use Twig\TokenParser\AbstractTokenParser;

class SomeTokenParser extends AbstractTokenParser
{
    public function parse(Token $token): Node
    {

        // some code where $value is extracted from token stream
        
        if ($value > 5) {
            throw new SyntaxError(
                'some tag is being used with an unsupported value above 5.',
                $stream->getCurrent()->getLine(),
                $stream->getSourceContext(),
            );
        }
    }
}
```

Everything in that example is part of the Twig API, except for the call to `$stream->getSourceContext()`. One could remove that line, but would lose the context of the error (the uncompiled code and the source path) which can be quite helpful for somebody who encounters a syntax error.

`getSourceContext` is the only method in `TokenStream` that is marked as internal - the returned `Source` object is not internal. I am therefore wondering if it would not make sense to remove the internal flag for `getSourceContext` as an easy way to officially provide the source context when processing the token stream.

fabpot:
Closes #4457

### DIFF
--- a/doc/deprecated.rst
+++ b/doc/deprecated.rst
@@ -216,6 +216,12 @@ Parser
 * Passing ``null`` to ``Twig\Parser::setParent()`` is deprecated as of Twig
   3.12.
 
+Lexer
+-----
+
+* Not passing a ``Source`` instance to ``Twig\TokenStream`` constructor is
+  deprecated as of Twig 3.16.
+
 Templates
 ---------
 
--- a/src/TokenStream.php
+++ b/src/TokenStream.php
@@ -27,7 +27,11 @@ final class TokenStream
         private array $tokens,
         private ?Source $source = null,
     ) {
-        $this->source = $source ?: new Source('', '');
+        if (null === $this->source) {
+            trigger_deprecation('twig/twig', '3.16', \sprintf('Not passing a "%s" object to "%s" constructor is deprecated.', Source::class, __CLASS__));
+
+            $this->source = new Source('', '');
+        }
     }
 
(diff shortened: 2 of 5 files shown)


doc/deprecated.rst
```diff
@@ -216,6 +216,12 @@ Parser
 * Passing ``null`` to ``Twig\Parser::setParent()`` is deprecated as of Twig
   3.12.
 
+Lexer
+-----
+
+* Not passing a ``Source`` instance to ``Twig\Lexer`` constructor is deprecated
```
xabbuh: `TokenStream` instead of `Lexer`?

src/TokenStream.php
```diff
@@ -27,7 +27,11 @@ public function __construct(
         private array $tokens,
         private ?Source $source = null,
     ) {
-        $this->source = $source ?: new Source('', '');
+        if (null === $this->source) {
```
stof, line 30: This should change the property to a non-promoted one, so that the property type can be not-nullable

fabpot: I will do this in 4.0. As it's private, that doesn't change anything anyway, right (or maybe for static analyzers)?

PR #4472, opened 2024-11-24
[Twig 4] Add PHPUnit as dev dependency

ruudk:
This makes it easier to run the tests locally.

I understand we didn't do this previously (Twig 3) because of `symfony/simple-phpunit`. But that is now removed in Twig 4, so maybe we can do this as well?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -5,7 +5,6 @@
     "keywords": ["templating"],
     "homepage": "https://twig.symfony.com",
     "license": "BSD-3-Clause",
-    "minimum-stability": "dev",
     "authors": [
         {
             "name": "Fabien Potencier",
@@ -32,7 +31,8 @@
     },
     "require-dev": {
         "psr/container": "^1.0|^2.0",
-        "phpstan/phpstan": "^2.0"
+        "phpstan/phpstan": "^2.0",
+        "phpunit/phpunit": "^11.4"
     },
     "autoload": {
         "psr-4" : {
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,7 +33,6 @@ jobs:
                   coverage: "none"
                   php-version: ${{ matrix.php-version }}
                   ini-values: memory_limit=-1
-                  tools: phpunit:11.3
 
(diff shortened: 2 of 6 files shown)


fabpot:
@ruudk That works for me. We used to have PHPunit 11.3. Here, you're suggesting to bump to 11.5, but tests don't pass, can you have a look?

ruudk:
Reported problem upstream: 
- https://github.com/sebastianbergmann/phpunit/issues/6043

I'll report back here when it's resolved.

sebastianbergmann:
> you're suggesting to bump to 11.5

I am also interested in why you propose to use a development version of PHPUnit.

ruudk:
@sebastianbergmann I did not do this on purpose. I ran `composer require phpunit/phpunit --dev` and it picked ^11.5. Even when I manually choose ^11.4 is still installs the dev version.

I think we need to change this:
https://github.com/twigphp/Twig/blob/c0689e3e686f5c8b5e9b9bdb10a74ea6ccfea350/composer.json#L8

I was unaware of this setting.

ruudk:
PR is ready for review now.

composer.json
```diff
@@ -5,7 +5,6 @@
     "keywords": ["templating"],
     "homepage": "https://twig.symfony.com",
     "license": "BSD-3-Clause",
-    "minimum-stability": "dev",
```
stof, line 8: using dev stability for our dependencies was done on purpose. You should use `phpunit/phpunit": "^11.4@stable"` to override it only for phpunit instead.

ruudk: Is it really needed? Looking at the dependencies:
https://github.com/twigphp/Twig/blob/c192cbff74c3f46c6dbb3c7b70cf72a4c475d571/composer.json#L26-L35

ruudk: Btw, the same thing happened when @VincentLanglet added PHPStan. It was installing the dev version without anybody noticing it.

VincentLanglet: "phpstan/phpstan": "^2.0@stable" 

should be preferred too then

fabpot:
Thank you @ruudk.

PR #4473, opened 2024-11-24
[Twig 4] Introduce ForElseNode

ruudk:
Currently there is no line number information for `{% else %}` inside a `for` loop.

Normally, this wouldn't be such a big deal.

But TwigStan uses this line information to map errors in PHP back to the original Twig source.

Let's say you have the following template:

```twig
{% for product in products %}
    <h2>FOR</h2>
{% else %}
    <p>No products found</p>
{% endfor %}
```

And `products` is of type `non-empty-array<Product>`, it means that the else will never happen.

This is currently reported as:
```
Negated boolean expression is always false.
🔖 booleanNot.alwaysFalse
🐘 compiled_index.php:83
🌱 templates/product/index.html.twig:2
🎯 src/Controller/ProductController.php:15
```

But as you can see, it points to line number 2, instead of line number 3.

In the compiled code, it looks like this:
```php
// line 1
foreach ($context['_seq'] as $context["_key"] => $context["product"]) {
    // line 2
    yield "        <h2>FOR</h2>\n    ";
    $context['_iterated'] = \true;
}
if (!$context['_iterated']) {
    // line 4
    yield "        <p>No products found</p>\n    ";
}
```

With this change, we will change the compiled code to become:
```php
// line 1
foreach ($context['_seq'] as $context["_key"] => $context["product"]) {
    // line 2
    yield "        <h2>FOR</h2>\n    ";
    $context['_iterated'] = \true;
}
// line 3
if (!$context['_iterated']) {
    // line 4
    yield "        <p>No products found</p>\n    ";
}
```

### DIFF
--- a/src/Node/ForNode.php
+++ b/src/Node/ForNode.php
@@ -25,7 +25,7 @@ use Twig\Node\Expression\Variable\AssignContextVariable;
 #[YieldReady]
 class ForNode extends Node
 {
-    public function __construct(AssignContextVariable $keyTarget, AssignContextVariable $valueTarget, AbstractExpression $seq, ?AbstractExpression $ifexpr, Node $body, ?Node $else, int $lineno)
+    public function __construct(AssignContextVariable $keyTarget, AssignContextVariable $valueTarget, AbstractExpression $seq, ?AbstractExpression $ifexpr, Node $body, ?ForElseNode $else, int $lineno)
     {
         if ($ifexpr) {
             $body = new IfNode(new Nodes([$ifexpr, $body]), null, $lineno);
@@ -72,13 +72,7 @@ class ForNode extends Node
         ;
 
         if ($this->hasNode('else')) {
-            $compiler
-                ->write("if (0 === \$iterator->getIndex0()) {\n")
-                ->indent()
-                ->subcompile($this->getNode('else'))
-                ->outdent()
-                ->write("}\n")
-            ;
+            $compiler->subcompile($this->getNode('else'));
         }
 
         // remove some "private" loop variables (needed for nested loops)
--- a/src/Node/ForElseNode.php
+++ b/src/Node/ForElseNode.php
@@ -1,0 +1,44 @@
+<?php
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) Fabien Potencier
+ * (c) Armin Ronacher
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Twig\Node;
+
+use Twig\Attribute\YieldReady;
+use Twig\Compiler;
+use Twig\Node\Expression\AbstractExpression;
+use Twig\Node\Expression\Variable\AssignContextVariable;
+
+/**
+ * Represents an else node in a for loop.
+ *
+ * @author Fabien Potencier <fabien@symfony.com>
+ */
+#[YieldReady]
+class ForElseNode extends Node
+{
+    public function __construct(Node $body, int $lineno)
+    {
+        parent::__construct(['body' => $body], [], $lineno);
+    }
+
+    public function compile(Compiler $compiler): void
+    {
+        $compiler
+            ->addDebugInfo($this)
+            ->write("if (0 === \$iterator->getIndex0()) {\n")
+            ->indent()
+            ->subcompile($this->getNode('body'))
+            ->outdent()
+            ->write("}\n")
+        ;
+    }
+}
--- a/src/TokenParser/ForTokenParser.php
+++ b/src/TokenParser/ForTokenParser.php
@@ -13,6 +13,7 @@
 namespace Twig\TokenParser;
 
 use Twig\Node\Expression\Variable\AssignContextVariable;
+use Twig\Node\ForElseNode;
 use Twig\Node\ForNode;
 use Twig\Node\Node;
 use Twig\Token;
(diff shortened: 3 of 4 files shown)


src/Node/ForNode.php
```diff
@@ -25,7 +25,7 @@
 #[YieldReady]
 class ForNode extends Node
 {
-    public function __construct(AssignContextVariable $keyTarget, AssignContextVariable $valueTarget, AbstractExpression $seq, ?AbstractExpression $ifexpr, Node $body, ?Node $else, int $lineno)
+    public function __construct(AssignContextVariable $keyTarget, AssignContextVariable $valueTarget, AbstractExpression $seq, ?AbstractExpression $ifexpr, Node $body, ?ForElseNode $else, int $lineno)
```
stof, line 28: Restricting the type is a BC break. You should introduce the new class in 3.x with a deprecation instead (where passing something else than a `ForElseNode` or `null` would trigger a deprecation and wrap the node in a ForElseNode, using the line of that node as fallback line number)

ruudk: Sounds like a good idea. 

Do you think the direction of this PR is OK / makes sense? I can propose the V3 change once that is clear.

src/Node/ForNode.php
```diff
@@ -72,13 +72,7 @@ public function compile(Compiler $compiler): void
...
 
         if ($this->hasNode('else')) {
-            $compiler
-                ->write("if (0 === \$iterator->getIndex0()) {\n")
-                ->indent()
-                ->subcompile($this->getNode('else'))
-                ->outdent()
-                ->write("}\n")
-            ;
+            $compiler->subcompile($this->getNode('else'));
```
fabpot, line 75: IIUC, this is to get the `// line xxx` comment above the `if` instead of inside it, right?

What about adding `->addDebugInfo($this->getNode('else'))` after current line 75 instead?

ruudk: The goal of this PR is to get the line number of `{% else %}` above the PHP `if` condition.

Your suggestion would print the line number of the line after the `{% else %}`.

ruudk: You can see it in the example of the PR description.

fabpot: I might be missing something. Isn't it what you want?

```php
$_v0 = new \Twig\Runtime\LoopIterator([1, 2, 3, 4]);
yield from ($_v1 = function ($iterator, &$context, $blocks, $recurseFunc, $depth) {
    $macros = $this->macros;
    $parent = $context;
    foreach ($iterator as $context["_key"] => $context["key"]) {
        // line 3
        yield "    IF";
    }
    // line 5
    if (0 === $iterator->getIndex0()) {
        yield "    ELSE";
    }
    unset($context['_key'], $context['key']);
    $context = array_intersect_key($context, $parent) + $parent;
    yield from [];
})($_v0, $context, $blocks, $_v1, 0);
```

ruudk: I tried doing `->addDebugInfo($this->getNode('else'))` but that will print the wrong line (4 instead of 3). 

```twig
1 {% for product in products %}
2    <h2>FOR</h2>
3 {% else %}
4    <p>No products found</p>
5 {% endfor %}
```

Before this PR it compiles to:
```php
// line 1
foreach ($context['_seq'] as $context["_key"] => $context["product"]) {
    // line 2
    yield "        <h2>FOR</h2>\n    ";
    $context['_iterated'] = \true;
}
if (!$context['_iterated']) {
    // line 4
    yield "        <p>No products found</p>\n    ";
}
```

With this PR it compiles to:
```php
// line 1
foreach ($context['_seq'] as $context["_key"] => $context["product"]) {
    // line 2
    yield "        <h2>FOR</h2>\n    ";
    $context['_iterated'] = \true;
}
// line 3
if (!$context['_iterated']) {
    // line 4
    yield "        <p>No products found</p>\n    ";
}
```

Because `if (!$context['_iterated']) {` points to `{% else %}` on line 3.

fabpot: Got it now. Looks like your PR makes sense.

src/Node/ForNode.php
fabpot: Let's start with the v3 PR.

src/Node/ForNode.php
ruudk: Oke great, I just rebased the PR and it seems to be green and ready for review now.

src/Node/ForNode.php
fabpot: I've just submitted #4552 to deprecate passing something that is not-null for `$ifexpr`. I don't think having a `ForElseNode` in 3.x makes sense.

ruudk: So this comment can be resolved ?

fabpot: Oops, I did a PR for `$ifexpr`, but here, we're talking about `$else`. 🙇

fabpot:
@ruudk I've done the work here for 3.x: #4553

ruudk:
Thank you! I will rebase as soon as V4 is synced.

ruudk:
I guess this was closed accidentally?

ruudk:
Ah, it's already merged.

fabpot:
> Ah, it's already merged.

Yes, I made the needed changes when merging 3.x to 4.x (basically resolving the conflicts).

ruudk:
Got it, thank you for getting this merged 💙

PR #4474, opened 2024-11-25
Add more coding standard rules

fabpot:


stof:
IIRC, there was a discussion on the Symfony repo reporting that Twig coding standards don't define the spacing rule around the `=` operator in macro argument declarations and around the `=` or `:` separators for named arguments.
Maybe those could be defined as well.

VincentLanglet:
> IIRC, there was a discussion on the Symfony repo reporting that Twig coding standards don't define the spacing rule around the `=` operator in macro argument declarations and around the `=` or `:` separators for named arguments. Maybe those could be defined as well.

For the record, currently Twig-CS-Fixer consider it should be 
```
foo(name_arg=42) // no space at all
```
and
```
foo(name_args: 42) // one space after
```
cf
https://github.com/VincentLanglet/Twig-CS-Fixer/blob/main/src/Rules/Function/NamedArgumentSpacingRule.php

fabpot:
Updated

VincentLanglet (approved):
I'll work on implementing the missing rule in Twig-CS-Fixer :)

doc/coding_standards.rst
```diff
@@ -88,21 +88,34 @@ standards:
...
+* Do not put any spaces before and after ``=`` in macro argument declarations:
 
   .. code-block:: twig
 
-    {% set name = 'Fabien' %}
-    {% set first_name = 'Fabien' %}
+    {% macro html_input(class="input") %}
 
-* Use snake case for all function/filter/test names:
+* Put exactly one space after the ``:`` sign in macro argument declarations:
```
VincentLanglet, line 97: @fabpot It works too for named arguments
```
{{ html_input(class_name: 'pwd') }}
```

doc/coding_standards.rst
```diff
@@ -88,21 +88,34 @@ standards:
     [1, 2, 3]
     {'name': 'Fabien'}
 
-* Use snake case for all variable names (provided by the application and
-  created in templates):
+* Do not put any spaces before and after ``=`` in macro argument declarations:
```
VincentLanglet, line 91: It works too for named arguments
```
{{ html_input(class_name='pwd') }}
```

doc/coding_standards.rst
VincentLanglet: And I just noticed, but I don't think this macro definition is a valid one.
```
{% macro html_input(class: "input") %}
```

doc/coding_standards.rst
VincentLanglet: I was wrong for macro, seems like it should be ` = ` cf https://twig.symfony.com/doc/3.x/tags/macro.html

VincentLanglet:
I opened https://github.com/twigphp/Twig/pull/4488

PR #4475, opened 2024-11-25
Fix having macro variables starting with an underscore

### LINKED ISSUE #4452: Macro arguments using leading/tailing underscore (BC introduced in v3.15)
Hi. I discovered a side effect after upgrading to the latest version, related to [this line in PR 4391](https://github.com/twigphp/Twig/pull/4391/files#diff-60c82d266554727246bf978c6075764c0538da108886e554bb8b10fa3b21c751R88).
For me, it's a BC, which may affect others as well.
The usage of macro arguments having leading/tailing underscores worked well until 3.15.

Example:
```twig
{% macro testing(_message) %}
The message: '{{ _message }}'.
{% endmacro %}
{{ _self.testing('Hello World!') }}
```

Output:
3.14: `The message: 'Hello World!'.`
3.15: `The message: ''.`

Compiled template:
In the referred line of the PR, underscores are trimmed off from arguments as they are assigned onto `$context`. `_message` becomes `message`, but the original `_message` is requested in the `escape`-method later. That's an error.
```php
[...]
class __TwigTemplate_[...] extends Template
[...]
    // line 1
    public function macro_testing($_message = null, ...$varargs): string|Markup
    {
        $macros = $this->macros;
        $context = [
            "message" => $_message,
            "varargs" => $varargs,
        ] + $this->env->getGlobals();

        $blocks = [];

        return ('' === $tmp = \Twig\Extension\CoreExtension::captureOutput((function () use (&$context, $macros, $blocks) {
            // line 2
            yield "The message: '";
            yield $this->env->getRuntime('Twig\Runtime\EscaperRuntime')->escape(($context["_message"] ?? null), "html", null, true);
            yield "'.
";
            yield from [];
        })())) ? '' : new Markup($tmp, $this->env->getCharset());
    }
```

I'm curious how I should see the code change. If this is the intention, maybe a `trigger_deprecation` should be called while evaluating the macro arguments.
I just wanted to let you know - the affected templates will be fixed on my side.

fabpot:
Closes #4452

### DIFF
--- a/src/Node/MacroNode.php
+++ b/src/Node/MacroNode.php
@@ -49,7 +49,7 @@ class MacroNode extends Node
         }
 
         foreach ($arguments->getKeyValuePairs() as $pair) {
-            if ('_'.self::VARARGS_NAME.'_' === $pair['key']->getAttribute('name')) {
+            if ("\u{035C}".self::VARARGS_NAME === $pair['key']->getAttribute('name')) {
                 throw new SyntaxError(\sprintf('The argument "%s" in macro "%s" cannot be defined because the variable "%s" is reserved for arbitrary arguments.', self::VARARGS_NAME, $name, self::VARARGS_NAME), $pair['value']->getTemplateLine(), $pair['value']->getSourceContext());
             }
         }
@@ -89,9 +89,13 @@ class MacroNode extends Node
 
         foreach ($arguments->getKeyValuePairs() as $pair) {
             $name = $pair['key'];
+            $var = $name->getAttribute('name');
+            if (str_starts_with($var, "\u{035C}")) {
+                $var = substr($var, \strlen("\u{035C}"));
+            }
             $compiler
                 ->write('')
-                ->string(trim($name->getAttribute('name'), '_'))
+                ->string($var)
                 ->raw(' => ')
                 ->subcompile($name)
                 ->raw(",\n")
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
 # 3.16.0 (2024-XX-XX)
 
+ * Fix having macro variables starting with an underscore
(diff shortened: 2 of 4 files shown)


stof:
What if the original name starts and ends with underscores ?

stof:
What was the reason for the trimming anyway ?

fabpot:
> What was the reason for the trimming anyway ?

We allow people to use variables like `this`, `context`, ... So, we need to add a prefix on the PHP side, and trim it when using it in the Twig code.

fabpot:
> What if the original name starts and ends with underscores ?

Fixed

src/Node/MacroNode.php
```diff
@@ -89,9 +89,13 @@ public function compile(Compiler $compiler): void
 
         foreach ($arguments->getKeyValuePairs() as $pair) {
             $name = $pair['key'];
+            $var = $name->getAttribute('name');
+            if ("\u{035C}" === mb_substr($var, 0, 1)) {
+                $var = mb_substr($var, 1);
```
nicolas-grekas:
```suggestion
            if (str_starts_with($var, "\u{035C}")) {
                $var = substr($var, \strlen("\u{035C}"));
```

alexislefebvre:
`"\u{035C}"` is used in several places, which could lead to unexpected behaviours if one instance is updated but not the others. What do you think about using a constant to define and reuse it? I could open a PR if you agree to this proposal.

fabpot:
> `"\u{035C}"` is used in several places, which could lead to unexpected behaviours if one instance is updated but not the others. What do you think about using a constant to define and reuse it? I could open a PR if you agree to this proposal.

That won't happen and in any case, we have tests, so I don't think we need a constant here.

PR #4477, opened 2024-11-25
Add support for PHPUnit 11

MauricioFauth:
This is my last attempt to add support for PHPUnit 11 for Twig 3.

This pull request adds new test methods that only runs in PHPUnit >= 11 and makes the current test methods to only run with PHPUnit < 11.

I think this addresses the @stof concerns about BC break, as the there's no change for PHPUnit < 11 besides the skipped tests that requires PHPUnit >= 11, but no test is actually skipped as they all run with the test methods that requires PHPUnit < 11.

Running on PHPUnit 11 requires `Twig\Test\NodeTestCase::provideTests()` and `Twig\Test\IntegrationTestCase::getFixturesDirectory()` to be implemented.
So it's recommended to first fix the deprecations with PHPUnit 10 before running the tests with PHPUnit 11, or the tests will fail.

- https://github.com/twigphp/Twig/pull/4317
- https://github.com/twigphp/Twig/pull/4313

### DIFF
--- a/src/Test/IntegrationTestCase.php
+++ b/src/Test/IntegrationTestCase.php
@@ -11,6 +11,9 @@
 
 namespace Twig\Test;
 
+use PHPUnit\Framework\Attributes\DataProvider;
+use PHPUnit\Framework\Attributes\Group;
+use PHPUnit\Framework\Attributes\RequiresPhpunit;
 use PHPUnit\Framework\TestCase;
 use Twig\Environment;
 use Twig\Error\Error;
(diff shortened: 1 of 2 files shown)


MauricioFauth:
A note about this change that I forgot to mention, it will only work for PHPUnit >= 11.4.0 as I had to fix it in PHPUnit first to make this PR work. The patch was included in PHPUnit 10.5.36 as well, however this PR does not affect users with PHPUnit < 10.5.36 and only those who try to run the tests with PHPUnit 11.

- See https://github.com/sebastianbergmann/phpunit/pull/5957.

derrabus:
Note that the 4.x branch should already be compatible with PHPUnit 11. Not sure how long it'll take to push Twig 4 over the finish line, but can't we live with PHPUnit 10 until then?

MauricioFauth:
> Note that the 4.x branch should already be compatible with PHPUnit 11. Not sure how long it'll take to push Twig 4 over the finish line, but can't we live with PHPUnit 10 until then?

As this patch has no impact on current users (no BC break), it will make easier to users to upgrade to Twig 4. And Twig will be compatible with PHPUnit 11 right now instead of waiting for Twig 4.
And PHPUnit 11 is almost one year old already and [PHPUnit 12 is due by February 07, 2025](https://github.com/sebastianbergmann/phpunit/milestone/52).

PR #4482, opened 2024-11-29
Enforce AbstractBinary for all binary operators

fabpot:


src/ExpressionParser.php
```diff
@@ -278,13 +278,9 @@ private function parseConditionalExpression($expr): AbstractExpression
                     // Ternary without else (expr ? expr2)
                     $expr3 = new ConstantExpression('', $this->parser->getCurrentToken()->getLine());
                 }
-            } else {
-                // Ternary without then (expr ?: expr3)
```
fabpot, line 282: This is now a proper operator in CoreExtension.

src/Node/Expression/OperatorEscapeInterface.php
```diff
@@ -0,0 +1,22 @@
...
+ * (c) Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+ namespace Twig\Node\Expression;
+
+/**
+ * Interface implemented by n-ary operators for n > 1.
```
fabpot, line 15: This interface can be defined on 1-ary operators, but that would not make sense.

src/TokenParser/TypesTokenParser.php
```diff
@@ -63,9 +63,13 @@ private function parseSimpleMappingExpression(TokenStream $stream): array
             $first = false;
 
             $nameToken = $stream->expect(Token::NAME_TYPE);
-            $isOptional = null !== $stream->nextIf(Token::PUNCTUATION_TYPE, '?');
 
-            $stream->expect(Token::PUNCTUATION_TYPE, ':', 'A type name must be followed by a colon (:)');
+            if ($stream->nextIf(Token::OPERATOR_TYPE, '?:')) {
```
fabpot, line 67: There is an ambiguity when using `?:` in `foo? : boolean`.

VincentLanglet:
I'll update all the PHPdoc in 3.x after this PR (and then do the same in 4.x)

src/ExpressionParser.php
stof: You still need this `else` clause to define the variables (see the phpstan error) or to throw an exception about invalid syntax.

src/Node/Expression/Binary/ElvisBinary.php
```diff
@@ -0,0 +1,28 @@
...
+final class ElvisBinary extends AbstractBinary implements OperatorEscapeInterface
+{
+    public function operator(Compiler $compiler): Compiler
+    {
+        return $compiler->raw('?:');
+    }
+
+    public function getOperandNamesToEscape(): array
+    {
+        return ['left', 'right'];
```
stof, line 48: This implementation looks wrong to me. When applying auto-escaping, we need to escape the `then` part without escaping the condition itself (to avoid changing the condition), which means turning `a ?: b` into `a ? a|e : b|e`

src/ExpressionParser.php
fabpot: I've removed the `if` as it's not possible anymore to have a `?` operator followed by the : `operator`.

src/Node/Expression/Binary/ElvisBinary.php
fabpot: Right, fixed.

src/Node/Expression/OperatorEscapeInterface.php
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * @author Fabien Potencier <fabien@symfony.com>
+ */
+interface OperatorEscapeInterface
+{
+    public function getOperandNamesToEscape(): array;
```
stof, line 24: I would add `@return string[]` to have a more precise return type.

fabpot: Done

VincentLanglet:
I made the PR to update all the related phpdoc @fabpot https://github.com/twigphp/Twig/pull/4485

PR #4487, opened 2024-11-30
Support underscores in number literals

smnandre:
```twig
{{ 1000 == 1_000 ? 'yes' : 'no' }}
# now: syntax error
# this PR: "yes"
```

> As of PHP 7.4.0, integer literals may contain underscores (_) between digits, for better readability of literals. These underscores are removed by PHP's scanner.

https://www.php.net/manual/en/language.types.integer.php

This PR replicates that behaviour, using the regexp to match the literals and then remove the "_".

I'm targeting **Twig4** but maybe 3.x would be ok?

I cannot think of a real case that
- does not trigger an error currently
- would work differently after this PR

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,6 @@
 # 3.17.0 (2024-XX-XX)
 
- * n/a
+ * Support underscores in number literals
 
 # 3.16.0 (2024-11-29)
 
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -23,3 +23,9 @@ parameters:
 			identifier: parameter.phpDocType
 			count: 5
 			path: src/Node/Node.php
+
+		- # Adding 0 to the string representation of a number is valid and what we want here
+			message: '#^Binary operation "\+" between 0 and string results in an error\.$#'
+			identifier: binaryOp.invalid
+			count: 1
+			path: src/Lexer.php
--- a/src/Lexer.php
+++ b/src/Lexer.php
@@ -44,8 +44,16 @@ class Lexer
     public const STATE_INTERPOLATION = 4;
 
     public const REGEX_NAME = '/[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*/A';
-    public const REGEX_NUMBER = '/[0-9]+(?:\.[0-9]+)?([Ee][\+\-][0-9]+)?/A';
     public const REGEX_STRING = '/"([^#"\\\\]*(?:\\\\.[^#"\\\\]*)*)"|\'([^\'\\\\]*(?:\\\\.[^\'\\\\]*)*)\'/As';
+
+    public const REGEX_NUMBER = '/(?(DEFINE)
+        (?<LNUM>[0-9]+(_[0-9]+)*)               # Integers (with underscores)   123_456
+        (?<FRAC>\.(?&LNUM))                     # Fractional part               .456
+        (?<EXPONENT>[eE][+-]?(?&LNUM))          # Exponent part                 E+10
+        (?<DNUM>(?&LNUM)(?:(?&FRAC))?)          # Decimal number                123_456.456
+    )(?:(?&DNUM)(?:(?&EXPONENT))?)              #                               123_456.456E+10
+    /Ax';
+    
     public const REGEX_DQ_STRING_DELIM = '/"/A';
     public const REGEX_DQ_STRING_PART = '/[^#"\\\\]*(?:(?:\\\\.|#(?!\{))[^#"\\\\]*)*/As';
     public const REGEX_INLINE_COMMENT = '/#[^\n]*/A';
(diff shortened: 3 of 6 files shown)


fabpot:
LGTM on 3.x
Can you also add an entry in the changelog and add a note in the docs ?

CHANGELOG
```diff
@@ -1,6 +1,6 @@
 # 3.17.0 (2024-XX-XX)
 
- * n/a
+ * Support underscores in numbers literals
```
```suggestion
 * Support underscores in number literals
```

smnandre:
@fabpot updated to target 3.x, mention syntax in documentation & add a line in changelog

src/Lexer.php
```diff
@@ -44,7 +44,7 @@ class Lexer
     public const STATE_INTERPOLATION = 4;
 
     public const REGEX_NAME = '/[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*/A';
-    public const REGEX_NUMBER = '/[0-9]+(?:\.[0-9]+)?([Ee][\+\-][0-9]+)?/A';
+    public const REGEX_NUMBER = '/[0-9]+(?:_[0-9]+)*(?:\.[0-9]+(?:_[0-9]+)*)?([Ee][\+\-][0-9]+)?/A';
```
fabpot: Maybe we can do the same as in https://github.com/symfony/symfony/blob/7.3/src/Symfony/Component/ExpressionLanguage/Lexer.php#L42-L43 where the exponent can also accept `_`.

fabpot: https://github.com/symfony/symfony/pull/44073

fabpot:
What about doing the same changes as we’ve done on Symfony? Both the regex and the logic.

smnandre:
Yep i have been interrupted earlier.. will finish later today!

smnandre:
Update: i missread things.. i'm sorry.

smnandre:
Update: i missread the scanner code.. 🤕

smnandre:
@fabpot  it took time to hit me ... it should be as you expected now.

Only thing missing is the "float without integer" syntax (ie `.99`).  

What do you think we handle this in following PR ? 
I'm afraid it would require other (bigger?) changes.

-

So, i had to update a bit the symfony regep, to disallow the "decimal part without integer part".


```diff
    public const REGEX_NUMBER = '/
                (?(DEFINE)(?P<LNUM>[0-9]+(_[0-9]+)*))
-                (?:\.(?&LNUM)|(?&LNUM)(?:\.(?!\.)(?&LNUM)?)?)(?:[eE][+-]?(?&LNUM))?/Ax';
+                (?:(?&LNUM)(?:\.(?!\.)(?&LNUM)?)?)(?:[eE][+-]?(?&LNUM))?/Ax';
```                


The following tests _pass_ with the current state of this PR, but would  fail if  we kept the original Symfony regexp.

-

❌ nested.definedArray.0

> Twig\Error\Error: Twig\Error\SyntaxError: Unexpected token "number" ("end of print statement" expected) in "index.twig" at line 16.

Pretty much the same thing/cause

```twig
{{ nested.definedArray.0          is     defined ? 'ok' : 'ko' }}
```

https://github.com/twigphp/Twig/blob/6823b64cad3ce8f9300ff69cfa7d23e20d61f10c/tests/Fixtures/tests/defined.test


❌  ``if items.3 is defined``

> Twig\Error\Error: Twig\Error\SyntaxError: Unexpected token "number" of value "0.3" ("end of statement block" expected) in "index.twig" at line 7

https://github.com/twigphp/Twig/blob/6823b64cad3ce8f9300ff69cfa7d23e20d61f10c/tests/Fixtures/tags/apply/with_if_tag.test#L9

I guess we are in the same situation you explained me quite recently (with a block name "cb-1") 

Not sure how would fix this though 🤷‍♂️ 

--

src/Lexer.php
```diff
@@ -346,11 +354,7 @@ private function lexExpression(): void
         }
         // numbers
         elseif (preg_match(self::REGEX_NUMBER, $this->code, $match, 0, $this->cursor)) {
-            $number = (float) $match[0];  // floats
-            if (ctype_digit($match[0]) && $number <= \PHP_INT_MAX) {
-                $number = (int) $match[0]; // integers lower than the maximum
-            }
-            $this->pushToken(Token::NUMBER_TYPE, $number);
+            $this->pushToken(Token::NUMBER_TYPE, 0 + str_replace('_', '', $match[0]));
```
fabpot, line 357: @smnandre Let's ignore the phpstan error (`phpstan-baseline.neon`).

smnandre: Done (you may want to check the comment)

phpstan-baseline.neon
```diff
@@ -23,3 +23,9 @@ parameters:
 			identifier: parameter.phpDocType
 			count: 5
 			path: src/Node/Node.php
+
+		- # String is matched by the REGEX_NUMBER
```
fabpot:
```suggestion
		- # Adding 0 to the string representation of a number is valid and what we want here
```

fabpot (changes requested):

fabpot:
Thank you @smnandre.

PR #4488, opened 2024-11-30
Update coding standards

VincentLanglet:


doc/coding_standards.rst
```diff
@@ -88,17 +88,23 @@ standards:
     [1, 2, 3]
     {'name': 'Fabien'}
 
-* Do not put any spaces before and after ``=`` in macro argument declarations:
+* Put exactly one space before and after ``=`` in macro argument declarations:
```
VincentLanglet, line 91: Cf example in https://twig.symfony.com/doc/3.x/tags/macro.html
```
{% macro input(name, value, type = "text", size = 20) %}
```

This seems to be the current standard ; but if you prefer no space, I can update Twig-CS-Fixer.

doc/coding_standards.rst
```diff
@@ -88,17 +88,17 @@ standards:
...
 
   .. code-block:: twig
 
-    {% macro html_input(class="input") %}
+    {% macro html_input(class = "input") %}
 
-* Put exactly one space after the ``:`` sign in macro argument declarations:
+* Put exactly one space after the ``:`` sign when using named arguments:
```
VincentLanglet, line 97: I'm not sure it's worth mentioning a standard for named argument with `=` since 
```
Use : instead of = to separate argument names and values:

{{ data|convert_encoding(from: 'iso-2022-jp', to: 'UTF-8') }}
```

doc/coding_standards.rst
fabpot: Right, let's keep spaces around `=`.

doc/coding_standards.rst
```diff
@@ -88,17 +88,17 @@ standards:
...
 
   .. code-block:: twig
 
-    {% macro html_input(class: "input") %}
+    {% html_input(class: "input") %}
```
fabpot:
```suggestion
    {{ html_input(class: "input") }}
```
?

VincentLanglet: Indeed, updated

fabpot:
Thank you @VincentLanglet.

PR #4490, opened 2024-12-01
add back minimum stability

xabbuh:
enforce stable releases of dev tools instead


xabbuh:
see https://github.com/twigphp/Twig/pull/4472#discussion_r1856393363

nicolas-grekas (approved):
Definitely :+1:

fabpot:
Thank you @xabbuh.

PR #4493, opened 2024-12-02
Require "erusev/parsedown": "dev-master as 1.x-dev"

nicolas-grekas:
Should fix the CI until https://github.com/erusev/parsedown/pull/891 is merged.


fabpot:
Thank you @nicolas-grekas.

PR #4494, opened 2024-12-02
Bump Phpstan to level 4

VincentLanglet:


src/Error/Error.php
```diff
@@ -142,11 +142,7 @@ private function updateRepr(): void
         }
 
         if ($this->name) {
-            if (\is_string($this->name) || $this->name instanceof \Stringable) {
-                $name = \sprintf('"%s"', $this->name);
-            } else {
-                $name = json_encode($this->name);
-            }
```
VincentLanglet, line 149: This is weird.

The property is typed as `private ?string $name;` so it always a string in this `if`.

src/Extension/CoreExtension.php
```diff
@@ -1199,7 +1199,7 @@ public static function length(string $charset, $thing): int
             return mb_strlen($thing, $charset);
         }
 
-        if (is_countable($thing) || $thing instanceof \SimpleXMLElement) {
```
VincentLanglet, line 1202: SimpleXMLElement is countable.

src/Parser.php
```diff
@@ -362,7 +365,7 @@ private function filterBodyNodes(Node $node, bool $nested = false): ?Node
         // we need to discard the wrapping "Node" for the "body" node
         $nested = $nested || !$node instanceof Nodes;
         foreach ($node as $k => $n) {
-            if (null !== $n && null === $this->filterBodyNodes($n, $nested)) {
```
VincentLanglet, line 365: node is `@implements \IteratorAggregate<int|string, Node>` so `$n` is never null.

src/Extension/CoreExtension.php
fabpot: Since 8.0. As we bumped the min version, this change is valid.

src/Error/Error.php
```diff
@@ -142,11 +142,7 @@ private function updateRepr(): void
...
+            $name = \sprintf('"%s"', $this->name);
```
fabpot: As $name is not used anywhere else, maybe we can merge this line with the next?

src/Error/Error.php
fabpot: I've just had a quick look to see why this is done this way and this is indeed not valid anymore.

src/Error/Error.php
VincentLanglet: Yes, Done in https://github.com/twigphp/Twig/pull/4494/commits/8a2f2c0c1128831dd50a12b4d926f1f34ee8e693

fabpot:
Thank you @VincentLanglet.

PR #4495, opened 2024-12-02
Set return type of Node::count()

billynoah:
There seems to be no reason to use the #[\ReturnTypeWillChange] attribute here as the method should always return an int.

### DIFF
--- a/src/Node/Node.php
+++ b/src/Node/Node.php
@@ -256,8 +256,7 @@ class Node implements \Countable, \IteratorAggregate
     /**
      * @return int
      */
-    #[\ReturnTypeWillChange]
-    public function count()
+    public function count(): int
     {
         return \count($this->nodes);
     }


xabbuh:
Doing so would break child classes that don’t have the return type declared.

xabbuh:
see https://3v4l.org/r8Sa2

billynoah:
I see. In that case, I will update child classes when I have some time.  Thanks.

xabbuh:
That's not an option as this change would also affect child classes that are not provided by Twig but come from third-party libraries or are defined in applications themselves. This change can simply not happen in Twig 3.

billynoah:
Ok. I think at some point any third-party libraries or user level child classes should be required to conform to the new spec, but I can understand it may be premature since these are relatively new changes to php. I guess we can just trash this pull request.  thanks

fabpot:
The change has already been done in 4.x.
https://github.com/twigphp/Twig/blob/4.x/src/Node/Node.php#L226

PR #4496, opened 2024-12-05
Fix ArrayAccess with objects as keys

### LINKED ISSUE #4476: In version 3.15.0, when accessing SplObjectStorage by index (object), an error is generated
```
{% for q in sections[s] %}
```
where section is SplObjectStorage 
and s is a object
Raise error  An exception has been thrown during the rendering of a template ("Object of class App\Attempt\Section could not be converted to string").

gharlan:
fixes #4476
The bug was introduced in 3.15 via https://github.com/twigphp/Twig/pull/4425#discussion_r1869120001

### DIFF
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -1644,7 +1644,7 @@ final class CoreExtension extends AbstractExtension
 
         // array
         if (Template::METHOD_CALL !== $type) {
-            $arrayItem = \is_bool($item) || \is_float($item) ? (int) $item : $item = (string) $item;
+            $arrayItem = \is_bool($item) || \is_float($item) ? (int) $item : $item;
 
             if ($sandboxed && $object instanceof \ArrayAccess && !\in_array($object::class, self::ARRAY_LIKE_CLASSES, true)) {
                 try {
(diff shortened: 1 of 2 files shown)


src/Extension/CoreExtension.php
```diff
@@ -1644,7 +1644,7 @@ public static function getAttribute(Environment $env, Source $source, $object, $
 
         // array
         if (Template::METHOD_CALL !== $type) {
-            $arrayItem = \is_bool($item) || \is_float($item) ? (int) $item : $item = (string) $item;
+            $arrayItem = \is_bool($item) || \is_float($item) ? (int) $item : $item;
```
gharlan, line 1647: @nicolas-grekas I'm not sure if there was a good reason for the string casting. So maybe it should be done somewhere else? Or is it ok to just remove the casting?

nicolas-grekas (approved):
Thanks for the regression test, that's what was missing.

fabpot:
Thank you @gharlan.

brandonkelly:
This change has introduced a separate breaking change: previously if you did `obj[key]` and `key` was an object that implemented `__toString()`, it would work as expected. But now you get an error, `Cannot access offset of type <TypeName> on array`.

I realize that `ArrayAccess::offsetGet()` accepts `mixed $offset`, but this feels like bad behavior to me?

Arrays in PHP can only be indexed by integer or string keys, and the intent of `ArrayAccess` is to make objects act like arrays. So it seem reasonable to assume that the “key” passed to `offsetGet()` is either a string on an int.

**EDIT:** created an issue for this: #4701

PR #4500, opened 2024-12-12
Fix the null coalescing operator when the test returns null

### LINKED ISSUE #4499: Null-coalescing does not work when method returns null
```twig
{{ something.nice(category) ?? 'nope' }}
```
should return `nope` but the result is empty

tried to add a test to `tests/IntegrationTest.php` but not sure how to inject an object for `tests/Fixtures/tests/null_coalesce.test`

this broke my code after latest release: i had a function where it accepts only string type but then the null coalesce stopped working and null was passed instead of the fallback string

fabpot:
Closes #4499

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
 # 3.17.1 (2024-XX-XX)
 
+ * Fix the null coalescing operator when the test returns null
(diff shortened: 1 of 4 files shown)


norkunas:
Thank you for fixing so fast :)

xabbuh (approved):

PR #4501, opened 2024-12-12
adds support for invoking closures

### LINKED ISSUE #3848: Support for closures or at least `Closure::__invoke`
Calling a closure is not supported in twig right now, any variable passed as `\Closure` cannot be called directly as Twig looks up the function in global functions (and extensions), the alternative is to call it via `__invoke` method on closure.

https://3v4l.org/EBR2Y (`__invoke` method does exist on Closures and work the same way as `$closure()` works).

Unfortunately this doesn't work in twig as `twig_get_attribute` uses `get_class_methods` to determine available methods on an object and fallbacks to `__call` if it exists and php  being php `get_class_methods($closure)` doesn't return `__invoke` https://3v4l.org/VKjAF :man_shrugging:  even though it exists and [is documented](https://www.php.net/manual/en/class.closure.php).

I think this can be fixed by checking if `$object` is instance of \Closure and then allow calling `__invoke` or is there a better way ?

I am happy to contribute a PR if it makes sense.

faizanakram99:
- fixes #3848

### DIFF
--- a/tests/TemplateTest.php
+++ b/tests/TemplateTest.php
@@ -394,6 +394,10 @@ class TemplateTest extends TestCase
             [true, ['foo' => 'bar'], $arrayAccess, 'vars', [], $anyType],
         ]);
 
+        // test for Closure::__invoke()
+        $tests[] = [true, 'closure called', fn (): string => 'closure called', '__invoke', [], $anyType];
+        $tests[] = [true, 'closure called', fn (): string => 'closure called', '__invoke', [], $methodType];
+
         // tests when input is not an array or object
         $tests = array_merge($tests, [
             [false, null, 42, 'a', [], $anyType, 'Impossible to access an attribute ("a") on a int variable ("42") in "index.twig".'],
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,6 +2,7 @@
 
  * Fix the null coalescing operator when the test returns null
  * Fix the Elvis operator when used as '? :' instead of '?:'
+ * Support for invoking closures
 
 # 3.17.0 (2024-12-10)
 
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -1740,6 +1740,10 @@ final class CoreExtension extends AbstractExtension
 
             static $propertyCheckers = [];
 
+            if ($object instanceof \Closure && '__invoke' === $item) {
+                return $isDefinedTest ? true : $object();
+            }
+
             if (isset($object->$item)
                 || ($propertyCheckers[$object::class][$item] ??= self::getPropertyChecker($object::class, $item))($object, $item)
             ) {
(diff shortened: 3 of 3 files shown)


faizanakram99:
@fabpot errors are unrelated

tests/TemplateTest.php
```diff
@@ -394,6 +394,13 @@ public static function getGetAttributeTests()
             [true, ['foo' => 'bar'], $arrayAccess, 'vars', [], $anyType],
         ]);
 
+        // test for Closure::__invoke()
+        $closure = function (): string {
+            return 'closure called';
+        };
```
fabpot:
```suggestion
        // test for Closure::__invoke()
        $closure = fn (): string => 'closure called';
```

fabpot (changes requested):

tests/TemplateTest.php
```diff
@@ -394,6 +394,11 @@ public static function getGetAttributeTests()
             [true, ['foo' => 'bar'], $arrayAccess, 'vars', [], $anyType],
         ]);
 
+        // test for Closure::__invoke()
+        $closure = fn (): string => 'closure called';
+
+        $tests[] = [true, 'closure called', $closure, '__invoke', [], $anyType];
+
```
fabpot: Now that I see it, maybe we can even inline everything!

```suggestion
        $tests[] = [true, 'closure called', fn (): string => 'closure called', '__invoke', [], $anyType];
```

fabpot:
Last comment and I think we're good: can you add an item in the CHANGELOG file?

nicolas-grekas (approved):
LGTM
I'd suggest opening an issue on php-src also, get_class_methods() should return __invoke to me (and other non static methods BTW).

faizanakram99:
> Last comment and I think we're good: can you add an item in the CHANGELOG file?

Done

faizanakram99:
> LGTM I'd suggest opening an issue on php-src also, get_class_methods() should return __invoke to me (and other non static methods BTW).

Done https://github.com/php/php-src/issues/17126

lowwa132 (commented):
Thank you for the PR @faizanakram99

tests/TemplateTest.php
```diff
@@ -394,6 +394,9 @@ public static function getGetAttributeTests()
             [true, ['foo' => 'bar'], $arrayAccess, 'vars', [], $anyType],
         ]);
 
+        // test for Closure::__invoke()
+        $tests[] = [true, 'closure called', fn (): string => 'closure called', '__invoke', [], $anyType];
```
fabpot, line 398: Let's add another test:

```
$tests[] = [true, 'closure called', fn (): string => 'closure called', '__invoke', [], $methodType];
```

fabpot (changes requested):
For PHP 8.0, we need to check that line 1742, we not calling `isset($object->$item)` for Closure as it does not work.

fabpot:
Thank you @faizanakram99.

PR #4504, opened 2024-12-16
Add a way to stream template rendering

fabpot:
Even if we're using `yield` internally, there is no easy way to stream template rendering. This new method can be used like this to HTTP stream a template with Symfony:

```php
use Symfony\Component\HttpFoundation\StreamedResponse;
use Twig\Environment;

$twig = new Environment(/* ... */);

$response = new StreamedResponse($twig->load('index')->stream([]));
$response->send();
```

### DIFF
--- a/src/TemplateWrapper.php
+++ b/src/TemplateWrapper.php
@@ -30,6 +30,22 @@ final class TemplateWrapper
     ) {
     }
 
+    /**
+     * @return iterable<scalar|\Stringable|null>
+     */
+    public function stream(array $context = []): iterable
+    {
+        yield from $this->template->yield($context);
+    }
+
+    /**
+     * @return iterable<scalar|\Stringable|null>
+     */
+    public function streamBlock(string $name, array $context = []): iterable
+    {
+        yield from $this->template->yieldBlock($name, $context);
+    }
+
     public function render(array $context = []): string
     {
         return $this->template->render($context);
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,6 @@
 # 3.18.0 (2024-XX-XX)
 
- * n/a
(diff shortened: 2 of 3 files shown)


fabpot:
I thought about streaming blocks as well, but I'm not sure this is worth it.

yceruto:
This will allow us to improve this method https://github.com/symfony/symfony/blob/7.3/src/Symfony/Bundle/FrameworkBundle/Controller/AbstractController.php#L267 👍 

>I thought about streaming blocks as well, but I'm not sure this is worth it.

I think yes, since there is a `renderBlock()` method added recently https://github.com/symfony/symfony/blob/dd882dbbd4910dce577d5bd280920e259dd782a5/src/Symfony/Bundle/FrameworkBundle/Controller/AbstractController.php#L428

GromNaN (approved):
Will you add a helper `Twig\Environment::stream()` as well?

src/TemplateWrapper.php
```diff
@@ -42,6 +42,11 @@ public function display(array $context = [])
         $this->template->display($context, \func_get_args()[1] ?? []);
     }
 
+    public function stream(array $context = []): iterable
```
GromNaN: To declare a return type compatible with the type expected by `StreamedResponse`.
```suggestion
    /**
     * @return iterable<string>
     */
    public function stream(array $context = []): iterable
```

smnandre:
> I think yes, since there is a `renderBlock()` method added recentl

And it's used via the #[Template] attribute too.

smnandre:
And this will probably allow some decent optimizations in the TwigEngineRenderer / block systems :)

fabpot:
I've just added a `streamBlock()` method as well.

fabpot:
> Will you add a helper `Twig\Environment::stream()` as well?

No, I don't think this is needed. The Environment class has already too many methods. I would like to do the opposite actually and deprecate the `Environment::render()` method.

PR #4510, opened 2024-12-18
[Doc] Update the docs about attribute() function

javiereguiluz:
I Was fixing `filter` -> `function` error, but I realized that we could expand the docs of `attribute()` a bit, even if it's deprecated 🙏

### DIFF
--- a/doc/functions/attribute.rst
+++ b/doc/functions/attribute.rst
@@ -3,21 +3,30 @@
 
 .. warning::
 
-    The ``attribute`` filter is deprecated as of Twig 3.15. Use the
+    The ``attribute`` function is deprecated as of Twig 3.15. Use the
     :ref:`dot operator <dot_operator>` that now accepts any expression
     when wrapped with parenthesis.
 
-    Note that this filter will still be available in Twig 4.0 to allow a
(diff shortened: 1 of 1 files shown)


stof:
The note about the algorithm being the same than the dot operator should be updated, removing the `except` part, as the dot operator also supports arbitrary expression now (which is why the function is deprecated)

javiereguiluz:
You are right. I just removed the note. Thanks.

doc/functions/attribute.rst
```diff
@@ -26,11 +35,6 @@ attribute:
 
     {{ attribute(object, method) is defined ? 'Method exists' : 'Method does not exist' }}
 
-.. note::
-
-    The resolution algorithm is the same as the one used for the ``.``
-    operator, except that the item can be any valid expression.
```
stof, line 32: I would keep the note saying it is the same algorithm than the dot operator (as it avoids having to duplicate the doc to explain which algorithm is used by this function), removing only the end of the sentence.

fabpot: Done while merging.

fabpot:
Thank you @javiereguiluz.

PR #4511, opened 2024-12-19
Ignore exceptions from undefined handlers when using the guard tag

### LINKED ISSUE #4505: Issue with the `guard` tag
I updated a bundle to start using the new `guard` tag: https://github.com/EasyCorp/EasyAdminBundle/pull/6656

However, instead of the expected `guard` behavior, in tests you can now see a lot of messages like:

```
"Did you forget to run "composer require symfony/web-link"?
Unknown function "preload" in "@EasyAdmin/includes/_css_assets.html.twig"."
at UndefinedCallableHandler.php line 105
```

As @stof explained on Symfony Slack, this only happens when using Twig inside a Symfony application, where this class exists --> https://github.com/symfony/symfony/blob/7.3/src/Symfony/Bridge/Twig/UndefinedCallableHandler.php

@stof proposed to either change the implementation of `guard` in Twig to skip the undefined callback handler when checking for function existence ... or provide a way for such handlers to opt-in for such bypass.

Thanks!

fabpot:
Fixes #4505

### DIFF
--- a/src/TokenParser/GuardTokenParser.php
+++ b/src/TokenParser/GuardTokenParser.php
@@ -33,7 +33,11 @@ final class GuardTokenParser extends AbstractTokenParser
 
         $nameToken = $stream->expect(Token::NAME_TYPE);
 
-        $exists = null !== $this->parser->getEnvironment()->$method($nameToken->getValue());
+        try {
+            $exists = null !== $this->parser->getEnvironment()->$method($nameToken->getValue());
+        } catch (SyntaxError) {
+            $exists = false;
+        }
 
         $stream->expect(Token::BLOCK_END_TYPE);
         if ($exists) {
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
 # 3.18.0 (2024-XX-XX)
 
+ * Ignore `SyntaxError` exceptions from undefined handlers when using the `guard` tag
  * Add a way to stream template rendering (`TemplateWrapper::stream()` and `TemplateWrapper::streamBlock()` )
 
 # 3.17.1 (2024-12-12)
--- a/tests/TokenParser/GuardTokenParserTest.php
+++ b/tests/TokenParser/GuardTokenParserTest.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace Twig\Tests\TokenParser;
+
+use PHPUnit\Framework\TestCase;
+use Twig\Environment;
+use Twig\Error\SyntaxError;
+use Twig\Loader\ArrayLoader;
+use Twig\Parser;
+use Twig\Source;
+
+class GuardTokenParserTest extends TestCase
+{
+    public function testUndefinedHandlers()
+    {
+        $this->expectNotToPerformAssertions();
+
+        $env = new Environment(new ArrayLoader(), ['cache' => false, 'autoescape' => false]);
+        $env->registerUndefinedFunctionCallback(fn ($name) => throw new SyntaxError('boom.'));
+        (new Parser($env))->parse($env->tokenize(new Source('{% guard function boom %}{% endguard %}', '')));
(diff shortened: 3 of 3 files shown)


src/TokenParser/GuardTokenParser.php
```diff
@@ -33,7 +33,11 @@ public function parse(Token $token): Node
 
         $nameToken = $stream->expect(Token::NAME_TYPE);
 
-        $exists = null !== $this->parser->getEnvironment()->$method($nameToken->getValue());
+        try {
+            $exists = null !== $this->parser->getEnvironment()->$method($nameToken->getValue());
+        } catch (\Throwable) {
```
stof: should we catch all throwables or only `Twig\Error\Error` ?

fabpot: As there is no contract here (the undefined handlers are not supposed to throw exceptions), I think catching everything is fine, or maybe just `\Exception`?

stof: Well, the usage done in Symfony is to throw a SyntaxError, to provide a more precise message than the SyntaxError thrown in https://github.com/twigphp/Twig/blob/91e609663e208dec9b0a70b210d1716a89ca025f/src/ExpressionParser.php#L803 (telling people which extra package they need to install to have access to that function of the twig-bridge).
the `MissingExtensionSuggestor` of `twig/extra-bundle` does exactly the same.

so my suggestion would even be to catch only SyntaxError, to account only for this pattern.

fabpot: If we think Symfony is the only one to mis-behave, I'm fine with catching just Twig exceptions.

stof: Well, as I said, Symfony is not the only one. Twig itself does as well in its extra-bundle. However, they misbehave in a very specific way where they expect no handler to support those callables but want to provide a more precise error (and there is no API allowing a handler to only provide a suggestion for the error message when it knows where this function could come from, as `return false` does not distinguish the case of "I don't know anything about this function" vs "I don't have an implementation of this function but I know which package does so the error message can provide the info").

At least, those places throw a SyntaxError, which preserves the behavior of the ExpressionParser (throwing a SyntaxError on unknown functions). Any handler using a different exception type would leak this abuse to the behavior of the ExpressionParser. So I would catch only SyntaxError, not even all Twig exceptions.

fabpot: Done

stof (approved):

javiereguiluz:
Thanks for merging this and for releasing it as part of Twig 3.18.0.

Sadly, this change didn't fix the issue for me. See the failing tests:

https://github.com/EasyCorp/EasyAdminBundle/actions/runs/12550268707

xabbuh:
The reason for the failure Javier observes (and also the reason for https://github.com/twigphp/Twig/discussions/4528#discussioncomment-11724572) is that the following `subparseIgnoreUnknownTwigCallables()` in `GuardTokenParser` also throws a `SyntaxError`.

javiereguiluz:
Maybe we should reopen this issue to not forget about it. Thanks.

stof:
I found the root cause of the issue. https://github.com/twigphp/Twig/blob/6513bc48f04abb9b1ac712a6dcd5d9f7466c58c4/src/ExpressionParser.php#L799 relies on getting a falsy value for non-existent functions before deciding to ignore them or to fail. To fully support the case of undefined handlers throwing a SyntaxError to provide a custom message, we also need to catch exceptions in that case (this was not caught by tests because this PR did not implement a functional test covering that case but only a unit test, which does not catch that it handles only half of the places that need to be handled).
I will try to do a PR with the missing part.

@javiereguiluz we cannot reopen because this is not an issue. It is the merged PR implementing the change.

PR #4513, opened 2024-12-23
Add `getLastModified()` to extensions

GromNaN:
Give to extensions the ability to set a last modification date for cache invalidation.

### Runtime

Currently, the cache is not invalidated when the signature of a runtime method is modified. This is an issue for templates that use named arguments, as argument names have an impact on the generated class.

With this change, extensions using runtime classes can compute a modification date by including the files on which they depend.

By default, the `AbstractExtension` checks if there is a file for the runtime class with the same name of the extension suffixed by Runtime instead of Extension.
This is the convention applied in [Symfony](https://github.com/symfony/symfony/tree/7.3/src/Symfony/Bridge/Twig/Extension) and Twig Extra: `MarkdownExtension` has `MarkdownRuntime`.

### Attribute 

Contributing to https://github.com/twigphp/Twig/pull/3916.

The extension class that will get the configuration from attributes will be able to track the classes having attributes to find the last modification date of all this classes.

### ~BC break~

~In Twig 4.0, the method `getLastModified` will be added to `ExtensionInterface`. It is extremely rare to implement this interface without extending `AbstractExtension`. So adding this method to the interface shouldn't be a problem as the base class has an implementation.~

### DIFF
--- a/doc/advanced.rst
+++ b/doc/advanced.rst
@@ -868,7 +868,7 @@ must be autoload-able)::
             // implement the logic to create an instance of $class
             // and inject its dependencies
             // most of the time, it means using your dependency injection container
-            if ('CustomRuntimeExtension' === $class) {
+            if ('CustomTwigRuntime' === $class) {
                 return new $class(new Rot13Provider());
             } else {
                 // ...
@@ -884,9 +884,9 @@ must be autoload-able)::
     (``\Twig\RuntimeLoader\ContainerRuntimeLoader``).
 
 It is now possible to move the runtime logic to a new
-``CustomRuntimeExtension`` class and use it directly in the extension::
+``CustomTwigRuntime`` class and use it directly in the extension::
 
-    class CustomRuntimeExtension
+    class CustomTwigRuntime
     {
         private $rot13Provider;
 
@@ -906,12 +906,20 @@ It is now possible to move the runtime logic to a new
         public function getFunctions()
         {
             return [
-                new \Twig\TwigFunction('rot13', ['CustomRuntimeExtension', 'rot13']),
+                new \Twig\TwigFunction('rot13', ['CustomTwigRuntime', 'rot13']),
                 // or
-                new \Twig\TwigFunction('rot13', 'CustomRuntimeExtension::rot13'),
+                new \Twig\TwigFunction('rot13', 'CustomTwigRuntime::rot13'),
             ];
         }
     }
+
+.. note::
+
+    The extension class should implement the ``Twig\Extension\LastModifiedExtensionInterface``
+    interface to invalidate the template cache when the runtime class is modified.
+    The ``AbstractExtension`` class implements this interface and tracks the
+    runtime class if its name is the same as the extension class but ends with
+    ``Runtime`` instead of ``Extension``.
 
 Testing an Extension
 --------------------
--- a/src/Extension/AbstractExtension.php
+++ b/src/Extension/AbstractExtension.php
@@ -11,7 +11,7 @@
 
 namespace Twig\Extension;
 
-abstract class AbstractExtension implements ExtensionInterface
+abstract class AbstractExtension implements LastModifiedExtensionInterface
 {
     public function getTokenParsers()
     {
@@ -41,5 +41,22 @@ abstract class AbstractExtension implements ExtensionInterface
     public function getOperators()
     {
         return [[], []];
+    }
+
+    public function getLastModified(): int
+    {
+        $filename = (new \ReflectionClass($this))->getFileName();
+        if (!is_file($filename)) {
+            return 0;
+        }
+
+        $lastModified = filemtime($filename);
+
+        // Track modifications of the runtime class if it exists and follows the naming convention
+        if (str_ends_with($filename, 'Extension.php') && is_file($filename = substr($filename, 0, -13) . 'Runtime.php')) {
+            $lastModified = max($lastModified, filemtime($filename));
+        }
+
+        return $lastModified;
     }
 }
--- a/src/Extension/EscaperExtension.php
+++ b/src/Extension/EscaperExtension.php
@@ -57,6 +57,14 @@ final class EscaperExtension extends AbstractExtension
         ];
     }
 
+    public function getLastModified(): int
+    {
+        return max(
+            parent::getLastModified(),
+            filemtime((new \ReflectionClass(EscaperRuntime::class))->getFileName()),
+        );
+    }
+
     /**
      * @deprecated since Twig 3.10
      */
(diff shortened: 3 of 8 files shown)


src/ExtensionSet.php
```diff
@@ -116,14 +116,17 @@ public function getLastModified(): int
...
-            }
+            $lastModified = max(match(true) {
+                $r->hasMethod('getLastModified') => $extension->getLastModified(),
+                is_file($r->getFileName()) => filemtime($r->getFileName()),
+                default => 0,
+            }, $lastModified);
         }
 
-        return $this->lastModified;
+        return $this->lastModified = $lastModified;
```
GromNaN, line 129: To be replaced in 4.0 by:

```php
return $this->lastModified = max(array_map(static fn (ExtensionInterface $extension) => $extension->getLastModified(), $this->extensions));
```

src/Extension/EscaperExtension.php
```diff
@@ -57,6 +57,14 @@ public function getFilters(): array
         ];
     }
 
+    public function getLastModified(): int
+    {
+        return max(
+            parent::getLastModified(),
+            filemtime((new \ReflectionClass(EscaperRuntime::class))->getFileName()),
+        );
+    }
```
GromNaN, line 66: Example of extension using a runtime class.

src/ExtensionSet.php
```diff
@@ -116,14 +116,17 @@ public function getLastModified(): int
...
 
+        $lastModified = 0;
         foreach ($this->extensions as $extension) {
             $r = new \ReflectionObject($extension);
-            if (is_file($r->getFileName()) && ($extensionTime = filemtime($r->getFileName())) > $this->lastModified) {
-                $this->lastModified = $extensionTime;
-            }
+            $lastModified = max(match(true) {
+                $r->hasMethod('getLastModified') => $extension->getLastModified(),
+                is_file($r->getFileName()) => filemtime($r->getFileName()),
```
GromNaN: I'll trigger a deprecation when an extension doesn't implement the new method.

src/ExtensionSet.php
```diff
@@ -116,14 +116,22 @@ public function getLastModified(): int
...
+                trigger_deprecation('twig/twig', '3.18', 'Not implementing the "getLastModified" method in "%s" is deprecated.', \get_class($extension));
+
+                if (is_file($r->getFileName())) {
+                    $lastModified = max(filemtime($r->getFileName()), $lastModified);
+                }
             }
         }
 
-        return $this->lastModified;
+        return $this->lastModified = $lastModified;
```
GromNaN, line 132: To be replaced in 4.0 by the max of all `getLastModified`.

```php
return $this->lastModified = max(array_map(static fn (ExtensionInterface $extension) => $extension->getLastModified(), $this->extensions));
```

src/Extension/AbstractExtension.php
```diff
@@ -42,4 +42,11 @@ public function getOperators()
     {
         return [[], []];
     }
+
+    public function getLastModified(): int
+    {
+        $filename = (new \ReflectionClass($this))->getFileName();
+
+        return is_file($filename) ? filemtime($filename) : 0;
+    }
```
smnandre: Could we maybe check there if a Runtime exists ? 

Something very naive like 

```
if file_exists(str_replace('Extension.php', 'Runtime.php', $filename)
```

wdyt ?

smnandre:
Do we need to add it to the interface ? Could we not simply ignore extensions not implementing this interface ?

src/Extension/AbstractExtension.php
GromNaN: This would be a hard coded convention that matches with the Symfony implementation. https://github.com/symfony/symfony/tree/7.3/src/Symfony/Bridge/Twig/Extension

I would agree. Maybe using the class name instead?

GromNaN:
Is it a thing to implement the interface without extending the abstract class? Adding this method should not be an issue for the large majority of projects, but it helps keeping the code simple.

smnandre:
Well this may break external extensions in prod ...  but you're right:
* there is some time before 4.0
* no one is forced to upgrade day one
* and .... yes in the vast majority i guess AbstractExtension is behind :)

src/Extension/AbstractExtension.php
smnandre: Yep!

A step further could be to "call" the extension methods (probaby just reflect them) to extract the various callable type info and see wich classes they are dependant from ?

fabpot:
> Is it a thing to implement the interface without extending the abstract class? Adding this method should not be an issue for the large majority of projects, but it helps keeping the code simple.

"large majority" is not enough when talking about BC breaks :)

Let's add a proper interface as this is a new capability (we've done it in the past with `ExistsLoaderInterface` for instance).

GromNaN:
> Let's add a proper interface as this is a new capability

Ok, interface added.

I'll also add automatic detection of `Runtime` classes to make it automatically work with Symfony TwigBridge extensions.

doc/advanced.rst
```diff
@@ -906,13 +906,19 @@ It is now possible to move the runtime logic to a new
...
+                new \Twig\TwigFunction('rot13', 'CustomTwigRuntime::rot13'),
             ];
         }
     }
 
+.. note::
+
+    By naming the runtime with the ``Runtime`` suffix instead of ``Extension``,
+    Twig automatically tracks updates to its source file for cache invalidation.
+    Otherwise, you can implement the method ``getLastModified(): int``.
```
fabpot:
```suggestion
    Otherwise, you can implement the ``getLastModified(): int`` method on the extension class.
```

Or maybe we should mention that the extension should implement `LastModifiedExtensionInterface`, which is implemented by `AbstractExtension`?

src/Extension/EscaperExtension.php
GromNaN: This is the only extension that has a runtime class that is not in the same namespace.

src/Extension/LastModifiedExtensionInterface.php
```diff
@@ -0,0 +1,23 @@
...
+
+namespace Twig\Extension;
+
+interface LastModifiedExtensionInterface extends ExtensionInterface
+{
+    /**
+     * Returns the last modification time of the extension for cache invalidation.
+     *
+     * This timestamp should be the last time the source code of the extension class
+     * and all its dependencies were modified.
```
fabpot: Would it makes sense to specifically mention the Runtime classes here?

```suggestion
     * and all its dependencies were modified (including the Runtime class).
```

fabpot (changes requested):

fabpot (approved):

fabpot:
@GromNaN LGMT, can you add a note in the CHANGELOG (you will need to rebase as I've just bump the version to 3.19.0)?

GromNaN:
Rebased, squashed, changelog updated.

fabpot:
Thank you @GromNaN.

PR #4514, opened 2024-12-26
Remove `preg_quote` delimiter argument (minor)

smnandre:
Following code source instructions

[src/Lexer.php#L91](https://github.com/smnandre/Twig/blob/c1f7277b7b6dc66859686347d359229e9edd2be9/src/Lexer.php#L91)

```
// when PHP 7.3 is the min version, we will be able to remove the '#' part in preg_quote as it's part of the default
```

Same thing in ExtensionSet.php

I think this has already been done in 4.0 so please close if you did not want to update 3.x code for now


fabpot:
Correct, I don't want to do these changes in 3.x.

PR #4516, opened 2024-12-26
Fix MacroNode $name argument can be overridden

### LINKED ISSUE #4515: MacroNode `$name` argument overriden in constructor loop ?
In [MacroNode](https://github.com/twigphp/Twig/blob/c1f7277b7b6dc66859686347d359229e9edd2be9/src/Node/MacroNode.php#L34-L57) constructor, the `$name` argument can be overridden by the last "key" of the `$argments` argument

```php
    //                                     🟢 
    public function __construct(string $name, Node $body, Node $arguments, int $lineno)
    {
         // ...

        if (!$arguments instanceof ArrayExpression) {
             // ...
            
            //                       🟢 
            foreach ($arguments as $name => $default) { 
                $args->addElement($default, new LocalVariable($name, $default->getTemplateLine()));
            }

            // ...
        }

        // ...
        //                                                                            🟢 
        parent::__construct(['body' => $body, 'arguments' => $arguments], ['name' => $name], $lineno);
    }
```

https://github.com/twigphp/Twig/blob/c1f7277b7b6dc66859686347d359229e9edd2be9/src/Node/MacroNode.php#L34-L57

Not sure if this has any real occurence / consequence... if no sorry for the waste of time 🤷

smnandre:
Fix #4515

### DIFF
--- a/src/Node/MacroNode.php
+++ b/src/Node/MacroNode.php
@@ -41,8 +41,8 @@ class MacroNode extends Node
             trigger_deprecation('twig/twig', '3.15', \sprintf('Not passing a "%s" instance as the "arguments" argument of the "%s" constructor is deprecated ("%s" given).', ArrayExpression::class, static::class, $arguments::class));
 
             $args = new ArrayExpression([], $arguments->getTemplateLine());
-            foreach ($arguments as $name => $default) {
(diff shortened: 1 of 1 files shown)


fabpot:
Thank you @smnandre.

PR #4517, opened 2024-12-26
Replace Twigfiddle by the new Twig playground

fabpot:
https://symfony.com/blog/introducing-the-new-twig-playground


smnandre:
oO  This is amazing.

PR #4518, opened 2024-12-26
Add invoke filter

Lorenzschaef:
Adds a new filter `invoke`, to invoke arrow functions and other php callables.

See the discussion here: https://github.com/twigphp/Twig/pull/4378#issuecomment-2557389092

### DIFF
--- a/doc/filters/invoke.rst
+++ b/doc/filters/invoke.rst
@@ -1,0 +1,19 @@
+``invoke``
+=======
+
+The ``invoke`` filter invokes an arrow function with the the given arguments:
+
+.. code-block:: twig
+
+    {% set person = {first: "Bob", last: "Smith"} %}
+    {% set func = p => "#{p.first} #{p.last}" %}
+
+    {{ func|invoke(person) }}
+    {# outputs Bob Smith #}
+
+Note that the arrow function has access to the current context.
+
+Arguments
+---------
+
+All given arguments are passed to the arrow function
--- a/doc/filters/invoke.rst
+++ b/doc/filters/invoke.rst
@@ -1,7 +1,7 @@
 ``invoke``
 =======
 
-The ``invoke`` filter invokes an arrow function with the the given arguments:
+The ``invoke`` filter invokes an arrow function with the given arguments:
 
 .. code-block:: twig
 
--- a/doc/filters/invoke.rst
+++ b/doc/filters/invoke.rst
@@ -1,6 +1,10 @@
 ``invoke``
 ==========
 
+.. versionadded:: 3.19
+
+    The ``invoke`` filter has been added in Twig 3.19.
+
 The ``invoke`` filter invokes an arrow function with the given arguments:
 
 .. code-block:: twig
@@ -10,10 +14,3 @@ The ``invoke`` filter invokes an arrow function with the given arguments:
 
     {{ func|invoke(person) }}
     {# outputs Bob Smith #}
-
-Note that the arrow function has access to the current context.
-
-Arguments
----------
-
-All given arguments are passed to the arrow function.
--- a/doc/filters/invoke.rst
+++ b/doc/filters/invoke.rst
@@ -1,11 +1,11 @@
 ``invoke``
-=======
+==========
 
 The ``invoke`` filter invokes an arrow function with the given arguments:
 
 .. code-block:: twig
 
-    {% set person = {first: "Bob", last: "Smith"} %}
+    {% set person = { first: "Bob", last: "Smith" } %}
     {% set func = p => "#{p.first} #{p.last}" %}
 
     {{ func|invoke(person) }}
@@ -16,4 +16,4 @@ Note that the arrow function has access to the current context.
 Arguments
 ---------
 
-All given arguments are passed to the arrow function
+All given arguments are passed to the arrow function.
--- a/doc/templates.rst
+++ b/doc/templates.rst
@@ -1018,7 +1018,8 @@ The following operators don't fit into any of the other categories:
     Arrow function support for functions, macros, and method calls was added in
     Twig 3.15 (filters and tests were already supported).
 
-  Arrow functions can be called using the ``invoke`` filter:
+  Arrow functions can be called using the :doc:`invoke </filters/invoke>`
+  filter.
 
   .. versionadded:: 3.19
 
--- a/doc/templates.rst
+++ b/doc/templates.rst
@@ -1018,9 +1018,9 @@ The following operators don't fit into any of the other categories:
     Arrow function support for functions, macros, and method calls was added in
     Twig 3.15 (filters and tests were already supported).
 
-  Arrow functions can be invoked using the ``invoke`` filter.
+Arrow functions can be called using the ``invoke`` filter:
 
-    .. versionadded:: 3.19
+.. versionadded:: 3.19
 
     The ``invoke`` filter has been added in Twig 3.19.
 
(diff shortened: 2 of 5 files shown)


src/Extension/CoreExtension.php
```diff
@@ -915,6 +916,20 @@ public static function keys($array): array
         return array_keys($array);
     }
 
+    /**
+     * Invokes a callable
+     *
+     * @param callable $callable
+     * @param ...$arguments
+     * @return mixed
```
GromNaN: phpdoc is not added when it does not provide information beyond the native types. But you can add usage example.

```suggestion
     *
     *  {% set func = x => 'Hello '~x %}
     *  {{ func|invoke('World') }}
```

src/Extension/CoreExtension.php
```diff
@@ -915,6 +916,20 @@ public static function keys($array): array
...
+     *
+     * @internal
+     */
+    public static function invoke(callable $callable, ...$arguments): mixed
+    {
+        return $callable(...$arguments);
```
smnandre: Don't we want to limit / check the `$callable` argument ? 

Via `checkArrow()` for instance?

Lorenzschaef: The code now looks more like the other functions/filters that accept arrow functions, including a call to `checkArrow()`

xabbuh:
This change also deserves an entry in the changelog for 3.19.0.

Lorenzschaef:
@xabbuh, I can add that.

What about the documentation? Should I also add a page to the filter reference in the docs? And mention it in the section about arrow functions? Or do you have something different in mind?

xabbuh:
Doing what you suggested looks good to me.

src/Extension/CoreExtension.php
```diff
@@ -915,6 +916,20 @@ public static function keys($array): array
...
     }
 
+    /**
+     * Invokes a callable
+     *
+     * @param \Closure $arrow
+     *
+     * @internal
+     */
+    public static function invoke(Environment $env, $arrow, ...$arguments): mixed
```
```suggestion
    public static function invoke(Environment $env, \Closure $arrow, ...$arguments): mixed
```

src/Extension/CoreExtension.php
```diff
@@ -915,6 +916,20 @@ public static function keys($array): array
         return array_keys($array);
     }
 
+    /**
+     * Invokes a callable
+     *
+     * @param \Closure $arrow
```
fabpot: Can be removed

src/Extension/CoreExtension.php
```diff
@@ -915,6 +916,20 @@ public static function keys($array): array
...
+     *
+     * @internal
+     */
+    public static function invoke(Environment $env, $arrow, ...$arguments): mixed
+    {
+        self::checkArrow($env, $arrow, 'invoke', 'filter');
```
fabpot: Can be removed. It's used for BC on 3.x, but as this is a new method, we can directly type hint the param with `\Closure` like done in 4.x for the other existing methods.

fabpot (changes requested):

src/Extension/CoreExtension.php
```diff
@@ -266,6 +266,7 @@ public function getFilters(): array
             // iteration and runtime
             new TwigFilter('default', [self::class, 'default'], ['node_class' => DefaultFilter::class]),
             new TwigFilter('keys', [self::class, 'keys']),
+            new TwigFilter('invoke', [self::class, 'invoke']),
```
GromNaN, line 269: The native function `call_user_func` would be identical.

```php
            new TwigFilter('invoke', 'call_user_func'),
```

GromNaN: In fact, no. It would allow calling any callable (including all PHP functions), not only Closure.

doc/filters/invoke.rst
```diff
@@ -0,0 +1,19 @@
+``invoke``
+=======
+
+The ``invoke`` filter invokes an arrow function with the the given arguments:
```
xabbuh:
```suggestion
The ``invoke`` filter invokes an arrow function with the given arguments:
```

doc/filters/invoke.rst
```diff
@@ -0,0 +1,19 @@
+``invoke``
+=======
```
fabpot:
```suggestion
==========
```

doc/filters/invoke.rst
```diff
@@ -0,0 +1,19 @@
...
+
+The ``invoke`` filter invokes an arrow function with the given arguments:
+
+.. code-block:: twig
+
+    {% set person = {first: "Bob", last: "Smith"} %}
```
fabpot:
```suggestion
    {% set person = { first: "Bob", last: "Smith" } %}
```

doc/templates.rst
```diff
@@ -1018,6 +1018,12 @@ The following operators don't fit into any of the other categories:
     Arrow function support for functions, macros, and method calls was added in
     Twig 3.15 (filters and tests were already supported).
 
+  Arrow functions can be invoked using the ``invoke`` filter.
+
+    .. versionadded:: 3.19
+
+    The ``invoke`` filter has been added in Twig 3.19.
```
fabpot:
```suggestion
Arrow functions can be called using the ``invoke`` filter:

.. versionadded:: 3.19

    The ``invoke`` filter has been added in Twig 3.19.
```

doc/filters/invoke.rst
```diff
@@ -0,0 +1,19 @@
...
+
+    {{ func|invoke(person) }}
+    {# outputs Bob Smith #}
+
+Note that the arrow function has access to the current context.
+
+Arguments
+---------
+
+All given arguments are passed to the arrow function
```
fabpot:
```suggestion
All given arguments are passed to the arrow function.
```

fabpot (changes requested):

doc/templates.rst
Lorenzschaef: I suppose you meant to leave indentation of two spaces, so I fixed that.

doc/templates.rst
```diff
@@ -1018,6 +1018,12 @@ The following operators don't fit into any of the other categories:
     Arrow function support for functions, macros, and method calls was added in
     Twig 3.15 (filters and tests were already supported).
 
+  Arrow functions can be called using the ``invoke`` filter:
```
xabbuh:
```suggestion
  Arrow functions can be called using the :doc:`invoke </filters/invoke>` filter.
```

xabbuh, line 1021:
```suggestion
  Arrow functions can be called using the :doc:`invoke </filters/invoke>` filter.
```

doc/filters/invoke.rst
```diff
@@ -0,0 +1,19 @@
...
+
+.. code-block:: twig
+
+    {% set person = { first: "Bob", last: "Smith" } %}
+    {% set func = p => "#{p.first} #{p.last}" %}
+
+    {{ func|invoke(person) }}
+    {# outputs Bob Smith #}
+
+Note that the arrow function has access to the current context.
```
fabpot: I don't understand this sentence. Do you mean that the function does **not** have access to the context?

Lorenzschaef: I took this sentence is from the documentation of `map`. Maybe in this case we would have to be more specific, that the arrow function has access to the context of where it is _defined_? Or would you prefer to remove it alltogether?

fabpot: It's true for `map` because we explicitely pass the context to the anonymous function that Twig creates. But here, this is not the case. This sentence should be removed.

fabpot:
Thank you @Lorenzschaef.

PR #4519, opened 2024-12-28
Fix unary operator precedence change

### LINKED ISSUE #4503: Twig 3.15 introduces potential bc break for extensions with operators
The issue is caused by this PR: https://github.com/twigphp/Twig/pull/4367

Unfortunately I'm currently on the road and don't have much time to debug deeper into the issue, but it essentially comes down to the fact that classes extending `\Twig\Extension\AbstractExtension` and also implementing operators via `getOperators()` will have to be updated to `precedence_change`.

We end up running into this issue with one of our extensions in phpBB after updating twig to the latest version:
```
Twig\Error\SyntaxError: An exception has been thrown during the compilation of a template ("Undefined array key "precedence_change"") in "define.html".
```
(see this build: https://github.com/phpbb/phpbb/actions/runs/12332949057/job/34421397543?pr=6761)

As of right now this ends up being a backwards incompatible change and not just a deprecation.

### Sample code
I've created some sample code for reproducing this issue (test project called `twig_test`).

Base file we'll access is `twig_test/test.php`:
```php
<?php

require_once __DIR__ . '/vendor/autoload.php';

use Twig\Environment;
use Twig\Loader\FilesystemLoader;
use MyVendor\CustomExtension;

// Set up Twig loader and environment
$loader = new FilesystemLoader(__DIR__ . '/templates');
$twig = new Environment($loader);
$twig->addExtension(new CustomExtension());

// Render a template with the extension
echo $twig->render('example.twig', [
	'foo'	=> true,
]);

```

`twig_test/composer.json`:
```json
{
  "require": {
    "twig/twig": "3.15.*"
  },
  "autoload": {
    "psr-4": {
      "MyVendor\\": "src/"
    }
  }
}

```

`twig_test/src/CustomExtension.php`:
```php
<?php

namespace MyVendor;

use Twig\Extension\AbstractExtension;

class CustomExtension extends AbstractExtension
{
	public function getOperators()
	{
		return [
			[
				'!' => ['precedence' => 50, 'class' => '\Twig\Node\Expression\Unary\NotUnary'],
			],
			[]
		];
	}
}

```

Template file `twig_test/templates/example.twig`:
```twig
{% if !foo %}
    not foo
{% else %}
    foo
{% endif %}

```

Expected output when accessing `test.php` (works until 3.14):
`foo`

Actual output with 3.15:
```
Warning: Undefined array key "precedence_change" in /var/www/html/twig_test/vendor/twig/twig/src/ExpressionParser.php on line 152
foo
```

fabpot:
Closes #4503

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,7 +1,8 @@
 # 3.18.0 (2024-XX-XX)
 
+ * Fix unary operator precedence change
  * Ignore `SyntaxError` exceptions from undefined handlers when using the `guard` tag
- * Add a way to stream template rendering (`TemplateWrapper::stream()` and `TemplateWrapper::streamBlock()` )
+ * Add a way to stream template rendering (`TemplateWrapper::stream()` and `TemplateWrapper::streamBlock()`)
 
 # 3.17.1 (2024-12-12)
 
--- a/src/ExpressionParser.php
+++ b/src/ExpressionParser.php
@@ -149,7 +149,6 @@ class ExpressionParser
                 return;
             }
             $target = explode('_', $unaryOp)[1];
-            $change = $this->unaryOperators[$target]['precedence_change'];
             /** @var AbstractExpression $node */
             $node = $expr->getNode('node');
             foreach ($this->precedenceChanges as $operatorName => $changes) {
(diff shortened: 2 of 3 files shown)


rxu:
There's a similar entry but for binary operators on line 169
https://github.com/twigphp/Twig/blob/2cf1939d0142ba65cdea288f84624f9177da3fb7/src/ExpressionParser.php#L169
Probably needs adjusting as well.

fabpot:
> There's a similar entry but for binary operators on line 169
> 
> https://github.com/twigphp/Twig/blob/2cf1939d0142ba65cdea288f84624f9177da3fb7/src/ExpressionParser.php#L169
> 
> Probably needs adjusting as well.

I don't think so.

rxu:
Tested locally and the issue seems to be fixed. Thanks.

PR #4521, opened 2024-12-30
Fix typo in enum twig doc functions

BaptisteChabrol:


doc/functions/enum.rst
```diff
@@ -13,10 +13,10 @@
     {{ enum('App\\MyEnum').SomeCase.value }}
 
     {# get all cases of an enum #}
-    {% enum('App\\MyEnum').cases %}
+    {{ enum('App\\MyEnum').cases }}
```
stof: This does not make sense either. You cannot echo an array.

BaptisteChabrol: yes I know but enum tag in {% %} doesn't work if is not in for loop it's confusing when you read the doc. 
Maybe replace by `{% for case in enum('App\\MyEnum').cases %} ` ?

Like on the enum_cases page :
```
{% for case in enum_cases('App\\MyEnum') %}
    {{ case.value }}
{% endfor %}
```

fabpot: Yes, having a better example like what you mention would make more sense.

fabpot:
Thank you @BaptisteChabrol.

PR #4522, opened 2025-01-02
Add @template on RuntimeLoaderinterface (Twig 4)

smnandre:
Do php-stan annotation have an impact on BC promise? 

The interface uses the term "class" already.. i'm guessing it's ok ?


fabpot:
> Do php-stan annotation have an impact on BC promise?

Like phpdocs, no.

fabpot:
Thank you @smnandre.

xabbuh (approved):

PR #4523, opened 2025-01-02
Add last missing type hints on Twig4

smnandre:
This should be the last batch

### DIFF
--- a/src/Node/Expression/CallExpression.php
+++ b/src/Node/Expression/CallExpression.php
@@ -23,6 +23,9 @@ abstract class CallExpression extends AbstractExpression
         $this->compileCallable($compiler);
     }
 
+    /**
+     * @return void
+     */
     protected function compileCallable(Compiler $compiler)
     {
         $twigCallable = $this->getAttribute('twig_callable');
--- a/src/Markup.php
+++ b/src/Markup.php
@@ -27,7 +27,7 @@ class Markup implements \Countable, \JsonSerializable, \Stringable
         $this->charset = $charset;
     }
 
-    public function __toString()
+    public function __toString(): string
     {
         return $this->content;
     }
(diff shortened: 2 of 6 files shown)


smnandre:
(fabbot unrelated)

fabpot:
Thank you @smnandre.

src/Node/Expression/CallExpression.php
```diff
@@ -23,6 +23,9 @@ public function compile(Compiler $compiler): void
         $this->compileCallable($compiler);
     }
 
+    /**
+     * @return void
```
stof, line 27: This `@return` should be added in 3.x, to be able to add an actual return type in 4.x before the 4.0 release.

alexander-schranz:
@smnandre Not sure what the target is of the types to catch all missing, but if you want to find more all missing types I suggest have a look at: https://github.com/TomasVotruba/type-coverage

```diff
diff --git a/composer.json b/composer.json
index f6890151..e27523bc 100644
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,8 @@
     "require-dev": {
         "psr/container": "^1.0|^2.0",
         "phpstan/phpstan": "^2.0@stable",
-        "phpunit/phpunit": "^11.4@stable"
+        "phpunit/phpunit": "^11.4@stable",
+        "tomasvotruba/type-coverage": "dev-main"
     },
     "autoload": {
         "psr-4" : {
diff --git a/phpstan.neon.dist b/phpstan.neon.dist
index d3077dc4..e61f0150 100644
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -1,7 +1,15 @@
 includes:
     - phpstan-baseline.neon
+    - vendor/tomasvotruba/type-coverage/config/extension.neon

 parameters:
+    type_coverage:
+        return: 100
+        param: 100
+        property: 100
+        constant: 100
     level: 4
     paths:
         - src
```

smnandre:
Hi @alexander-schranz, thank for the tool, it will be a very nice addition for my projects at work!

Here it is not that helpful (i may have missed a configuration or something) as the docblocks are not used/looked up..

But it allowed me to spot some misses i'll fix soon!

PR #4525, opened 2025-01-02
Optimize NameExpression compilation

fabpot:


src/Node/Expression/NameExpression.php
```diff
@@ -39,7 +39,7 @@ public function compile(Compiler $compiler): void
         $compiler->addDebugInfo($this);
 
         if ($this->getAttribute('is_defined_test')) {
-            if (isset($this->specialVars[$name])) {
+            if (isset($this->specialVars[$name]) || $this->getAttribute('always_defined')) {
                 $compiler->repr(true);
             } elseif (\PHP_VERSION_ID >= 70400) {
```
GromNaN, line 44: This is always true since PHP 8.0 is required.

fabpot: I know, but I don't accept cleanup in 3.x anymore, it's already part of the 4.x branch.

GromNaN: Your project, your rules :wink:

PR #4527, opened 2025-01-02
Add VariableOptimizerNodeVisitor

fabpot:
This is the simplest useful node visitor that use the new types information to optimize the compiled code.

### DIFF
--- a/src/NodeVisitor/VariableOptimizerNodeVisitor.php
+++ b/src/NodeVisitor/VariableOptimizerNodeVisitor.php
@@ -1,0 +1,56 @@
+<?php
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Twig\NodeVisitor;
+
+use Twig\Environment;
+use Twig\Node\Expression\NameExpression;
+use Twig\Node\ModuleNode;
+use Twig\Node\Node;
+use Twig\Node\TypesNode;
+
+/**
+ * @author Fabien Potencier <fabien@symfony.com>
+ *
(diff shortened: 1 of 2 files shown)


fabpot:
I think that this node visitor must also enforce the availability of the variable at runtime to make this works correctly.
Like converting the `types` tag to checks that would throw exceptions based on the existence of the variables and maybe checking the types as well. The more I think about it, the more I think that this would be like a "strict" mode that would enable these performance optimizations.

WDYT /cc @jdreesen @ruudk?

jdreesen:
Did you mean @drjayvee instead?

fabpot:
> Did you mean @drjayvee instead?

Oops, sorry for the wrong ping.

drjayvee:
> This is the simplest useful node visitor that use the new types information to optimize the compiled code.

Sure, but it does add an exception for `strict_variables`, namely that typed variables never default to `null` even if `strict_variables` is disabled.

> I think that this node visitor must also enforce the availability of the variable at runtime to make this works correctly
 Like converting the types tag to checks that would throw exceptions based on the existence of the variables and maybe checking the types as well. The more I think about it, the more I think that this would be like a "strict" mode that would enable these performance optimizations.

~~When you say "checking the types", I'm assuming you mean a run-time check? The other meaning is compile-time validation of the type to see whether it is valid, and whether a class (or interface, trait) exists.~~

Sorry, I didn't read your comment properly.

Aside from the run-time checks, there's also the (possible) compile-time validation of the type to check whether it is valid, and whether a given class (or interface, trait) exists.

TwigQI does all of those already. I wouldn't mind bringing some of its functionality into Twig core.

*However*, all of these require that we first standardize the typing system and syntax. That discussion in #4532 proved to be controversial. I wouldn't mind having another go at it, however.

PR #4534, opened 2025-01-05
Add return type annotations on Twig 3 to prepare Twig 4

smnandre:
Started to add missing type annotation following @stof comment [here](https://github.com/twigphp/Twig/pull/4523#discussion_r1901915880).

Update:  I checked here every class (but one) in the `src` directory.

But you should ~~probably~~ **certainely** have another look.

Remains:
- src/ExpressionParser.php, where i'm not confident enough about the best return types (AbstractExpression|null works almost everytime, but lack some precision)
- extra directory
- tests directory

I'm happy to let them to someone else.... this was less fun than I expected 😅

### DIFF
--- a/src/AbstractTwigCallable.php
+++ b/src/AbstractTwigCallable.php
@@ -79,6 +79,9 @@ abstract class AbstractTwigCallable implements TwigCallableInterface
         return $this->dynamicName;
     }
 
+    /**
+     * @return callable|array{class-string, string}|null
+     */
     public function getCallable()
     {
         return $this->callable;
--- a/src/Extension/SandboxExtension.php
+++ b/src/Extension/SandboxExtension.php
@@ -72,7 +72,7 @@ final class SandboxExtension extends AbstractExtension
         return $this->sourcePolicy->enableSandbox($source);
     }
 
-    public function setSecurityPolicy(SecurityPolicyInterface $policy)
+    public function setSecurityPolicy(SecurityPolicyInterface $policy): void
     {
         $this->policy = $policy;
     }
@@ -117,6 +117,13 @@ final class SandboxExtension extends AbstractExtension
         }
     }
 
+    /**
+     * @param mixed $obj
+     *
+     * @return mixed
+     *
+     * @throws SecurityNotAllowedMethodError
+     */
     public function ensureToStringAllowed($obj, int $lineno = -1, ?Source $source = null)
     {
         if (\is_array($obj)) {
--- a/src/Markup.php
+++ b/src/Markup.php
@@ -27,7 +27,7 @@ class Markup implements \Countable, \JsonSerializable, \Stringable
         $this->charset = $charset;
     }
 
-    public function __toString()
+    public function __toString(): string
     {
         return $this->content;
     }
--- a/src/NodeVisitor/EscaperNodeVisitor.php
+++ b/src/NodeVisitor/EscaperNodeVisitor.php
@@ -157,7 +157,10 @@ final class EscaperNodeVisitor implements NodeVisitorInterface
         return \in_array($type, $safe) || \in_array('all', $safe);
     }
 
-    private function needEscaping()
+    /**
+     * @return string|false
+     */
+    private function needEscaping(): string|bool
     {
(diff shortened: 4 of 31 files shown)


smnandre:
(fabbot unrelated)

smnandre:
(fixed my mistakes, all seems ok now)

src/AbstractTwigCallable.php
```diff
@@ -79,6 +79,9 @@ public function getDynamicName(): string
         return $this->dynamicName;
     }
 
+    /**
+     * @return callable
```
stof: This is wrong. The callable of a AbstractTwigCallable might not be a PHP callable, due to the case of Twig runtimes.

See the constructor parameter type (on child classes, as it is missing on the constructor of that class currently)

src/Extension/SandboxExtension.php
```diff
@@ -117,6 +120,13 @@ public function checkPropertyAllowed($obj, $property, int $lineno = -1, ?Source
         }
     }
 
+    /**
+     * @param array|object $obj
```
stof: This function actually supports `mixed` as input (and as output)

src/Extension/SandboxExtension.php
```diff
@@ -72,6 +72,9 @@ private function isSourceSandboxed(?Source $source): bool
         return $this->sourcePolicy->enableSandbox($source);
     }
 
+    /**
+     * @return void
```
stof: As this class is final, the return type can be added as a native one directly.

src/Markup.php
```diff
@@ -27,6 +27,9 @@ public function __construct($content, $charset)
         $this->charset = $charset;
     }
 
+    /**
+     * @return string
```
stof: this type can be added directly (`__toString` is special in PHP and has an implicit `string` return type)

src/NodeVisitor/EscaperNodeVisitor.php
```diff
@@ -157,7 +157,7 @@ private function isSafeFor(string $type, AbstractExpression $expression, Environ
         return \in_array($type, $safe) || \in_array('all', $safe);
     }
 
-    private function needEscaping()
+    private function needEscaping(): string|bool
```
stof, line 163: `string|false` actually. `true` is not a valid strategy.

src/Sandbox/SecurityNotAllowedMethodError.php
```diff
@@ -33,6 +33,9 @@ public function getClassName(): string
         return $this->className;
     }
 
+    /**
+     * @return string
```
stof: Can be a native one as the class is final

src/Token.php
```diff
@@ -80,6 +80,9 @@ public function getType(): int
         return $this->type;
     }
 
+    /**
+     * @return string|int|float|bool|null
```
stof: This is wrong if the constructor does not have a parameter type restricting the type of the value.

src/Parser.php
```diff
@@ -307,6 +313,9 @@ public function addImportedSymbol(string $type, string $alias, ?string $name = n
         $this->importedSymbols[0][$type][$alias] = ['name' => $name, 'node' => $internalRef];
     }
 
+    /**
+     * @return string|null
```
stof: This is wrong. It returns `array{name: string, node: AssignTemplateVariable|null}`

src/NodeVisitor/EscaperNodeVisitor.php
smnandre: "true" will wait Twig4 / PHP 8.2 no ?

smnandre:
Thank you very much @stof! (and sorry i make that many mistakes 😮‍💨 )

I've fixed all your comments

fabpot:
Thank you @smnandre.

src/NodeVisitor/EscaperNodeVisitor.php
stof: I don't understand what you mean here. `true` would indeed require PHP 8.2 to be used. But the one we need here is `false`, not `true`

PR #4543, opened 2025-01-19
Move operator definitions to objects

fabpot:


doc/deprecated.rst
```diff
@@ -411,3 +411,26 @@ Operators
...
+          ];
+      }
+
+    After:
+
+      public function getOperators(): Operators {
+          return new Operators([
+              new NotUnaryOperator(),
+          ]);
+      }
```
smnandre:
```suggestion
      public function getOperators(): array {
          return [
              new NotUnaryOperator(),
          ];
      }
```

Not sure if voluntary or not..  if so the paragraph just before should be updated

src/ExtensionSet.php
```diff
@@ -482,12 +467,110 @@ private function initExtension(ExtensionInterface $extension): void
...
+            }
+
+            public function getPrecedenceChange(): ?OperatorPrecedenceChange
+            {
+                return $this->op['precedence_change'] ?? null;
+            }
+
+            public function getNodeClass(): string
+            {
+                return $this->op['class'] ?? '';
```
stof: The empty string is not a valid node class. If the node class is optional for an operator, I'd rather use `null` than the empty string.

src/Operator/OperatorInterface.php
```diff
@@ -0,0 +1,34 @@
...
+
+    public function getNodeClass(): string;
+
+    public function getArity(): OperatorArity;
+
+    public function getPrecedence(): int;
+
+    public function getPrecedenceChange(): ?OperatorPrecedenceChange;
+
+    public function getAliases(): array;
```
stof: this should have phpdoc describing the precise type of the array.

src/Operator/OperatorInterface.php
```diff
@@ -0,0 +1,34 @@
...
+
+namespace Twig\Operator;
+
+use Twig\OperatorPrecedenceChange;
+
+interface OperatorInterface
+{
+    public function getOperator(): string;
+
+    public function getNodeClass(): string;
```
stof: we should add `@return class-string<AbstractExpression>` here (or maybe another base class if we are stricter)

src/Operator/OperatorInterface.php
```diff
@@ -0,0 +1,41 @@
...
+
+use Twig\Node\Expression\AbstractExpression;
+use Twig\OperatorPrecedenceChange;
+
+interface OperatorInterface
+{
+    public function getOperator(): string;
+
+    /**
+     * @return class-string<AbstractExpression>
```
stof:
```suggestion
     * @return class-string<AbstractExpression>|null
```

Alternatively, as I see that you throw a LogicException anyway when the operator does not have a node class, maybe the return type could stay non-nullable, by moving the exception in the BC layer instead.

src/Extension/ExtensionInterface.php
```diff
@@ -66,12 +63,7 @@ public function getFunctions();
     /**
      * Returns a list of operators to add to the existing list.
      *
-     * @return array<array> First array of unary operators, second array of binary operators
-     *
-     * @psalm-return array{
-     *     array<string, array{precedence: int, precedence_change?: OperatorPrecedenceChange, class: class-string<AbstractUnary>}>,
-     *     array<string, array{precedence: int, precedence_change?: OperatorPrecedenceChange, class?: class-string<AbstractBinary>, associativity: ExpressionParser::OPERATOR_*}>
-     * }
+     * @return OperatorInterface[]
```
stof: in 3.x, the return type should probably be a union of the old and new type, so that static analyzers don't report old code as being totally broken (when using native return types, the type of the BC layer needs to be taken into account)

src/ExpressionParser.php
```diff
@@ -102,28 +78,30 @@ public function parseExpression($precedence = 0)
...
+        while ($token->test(Token::OPERATOR_TYPE) && ($op = $this->operators->getBinary($token->getValue())) && $op->getPrecedence() >= $precedence) {
             $this->parser->getStream()->next();
 
             if ('is not' === $token->getValue()) {
                 $expr = $this->parseNotTestExpression($expr);
             } elseif ('is' === $token->getValue()) {
                 $expr = $this->parseTestExpression($expr);
-            } elseif (isset($op['callable'])) {
-                $expr = $op['callable']($this->parser, $expr);
+            } elseif (null !== $op->getCallable()) {
```
stof: There is nothing enforcing that `getCallable` exists here. This method exists in `AbstractBinaryOperator` but not in `OperatorInterface`, and nothing enforces that operators using binary arity extend `AbstractBinaryOperator`

src/Operator/Binary/AbstractBinaryOperator.php
```diff
@@ -0,0 +1,34 @@
...
+    {
+        return OperatorArity::Binary;
+    }
+
+    public function getAssociativity(): OperatorAssociativity
+    {
+        return OperatorAssociativity::Left;
+    }
+
+    public function getCallable(): ?callable
```
stof: this should have a phpdoc `@return (callable(ExpressionParser, AbstractExpression): AbstractExpression)|null` to describe the expected callable signature (putting this phpdoc on the highest level defining the method, which is currently `AbstractBinaryOperator` but might be the interface if you add it in an interface)

src/ExpressionParser.php
```diff
@@ -102,28 +78,30 @@ public function parseExpression($precedence = 0)
...
+                $expr = $op->getCallable()($this->parser, $expr);
             } else {
                 $previous = $this->setDeprecationCheck(true);
                 try {
-                    $expr1 = $this->parseExpression(self::OPERATOR_LEFT === $op['associativity'] ? $op['precedence'] + 1 : $op['precedence']);
+                    $expr1 = $this->parseExpression(OperatorAssociativity::Left === $op->getAssociativity() ? $op->getPrecedence() + 1 : $op->getPrecedence());
```
stof: `getAssociativity` is also not enforced in `OperatorInterface`. I have the feeling that we might need a `BinaryOperatorInterface` extending `OperatorInterface` to add more methods

src/Operator/Ternary/AbstractTernaryOperator.php
```diff
@@ -0,0 +1,23 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Twig\Operator\Ternary;
+
+use Twig\Operator\AbstractOperator;
+use Twig\Operator\OperatorArity;
+
+abstract class AbstractTernaryOperator extends AbstractOperator
```
stof: what is the reason to add this abstract class if we never use it ?

src/Operator/Operators.php
```diff
@@ -0,0 +1,93 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Twig\Operator;
+
+use Twig\Operator\Binary\AbstractBinaryOperator;
+use Twig\Operator\Unary\AbstractUnaryOperator;
+
+class Operators implements \IteratorAggregate
```
stof: should this class be `@internal` ?

src/Operator/Operators.php
```diff
@@ -0,0 +1,93 @@
...
+    private ?\WeakMap $precedenceChanges = null;
+
+    public function __construct(
+        array $operators = [],
+    ) {
+        $this->add($operators);
+    }
+
+    /**
+     * @param array<AbstractOperator> $operators
```
stof: shouldn't this use `OperatorInterface` rather than `AbstractOperator` ?

src/Operator/Operators.php
```diff
@@ -0,0 +1,93 @@
...
+use Twig\Operator\Unary\AbstractUnaryOperator;
+
+class Operators implements \IteratorAggregate
+{
+    private array $operators = [];
+    private array $aliases = [];
+    private ?\WeakMap $precedenceChanges = null;
+
+    public function __construct(
+        array $operators = [],
```
stof: the type of this argument should be defined.

src/Operator/Operators.php
```diff
@@ -0,0 +1,93 @@
...
+ */
+
+namespace Twig\Operator;
+
+use Twig\Operator\Binary\AbstractBinaryOperator;
+use Twig\Operator\Unary\AbstractUnaryOperator;
+
+class Operators implements \IteratorAggregate
+{
+    private array $operators = [];
```
stof: I suggest adding types for those properties (so that the static analysis job running in our CI can analyze the code of this class more precisely, bring more value): `@var array<value-of<OperatorArity>, array<string, OperatorInterface>>`

src/Operator/Operators.php
```diff
@@ -0,0 +1,93 @@
...
+
+        return $this;
+    }
+
+    public function getUnary(string $name): ?AbstractUnaryOperator
+    {
+        return $this->operators[OperatorArity::Unary->value][$name] ?? ($this->aliases[OperatorArity::Unary->value][$name] ?? null);
+    }
+
+    public function getBinary(string $name): ?AbstractBinaryOperator
```
stof: this getter expects that a binary operator extends AbstractBinaryOperator, but there is no such validation when adding them. Neither the ExtensionInterface nor the OperatorInterface explain that requirement.

src/Operator/Operators.php
stof: I suggest adding `@template-implements \IteratorAggregate<string, OperatorInterface>` to define the type of the iterator.

fabpot:
New iteration where I have moved most pseudo-operators (logic hardcoded in `ExpressionParser`) to real operators like any other ones (think `|`, `.`,  `[`, `(`, `??`, ...).

doc/operators_precedence.rst
```diff
@@ -0,0 +1,56 @@
+Unary operators precedence:
```
fabpot: This doc is the most important part of this PR.
We need to review the precedence order carefully.

src/ExpressionParser.php
```diff
@@ -95,202 +89,91 @@ public function parseExpression($precedence = 0)
...
-            }
-        }
-        $stream->expect(Token::PUNCTUATION_TYPE, ')');
-        $stream->expect(Token::ARROW_TYPE);
-
-        return new ArrowFunctionExpression($this->parseExpression(), new Nodes($names), $line);
-    }
-
-    private function getPrimary(): AbstractExpression
+    public function getPrimary(): AbstractExpression
```
stof: should the public method be named `parse*` for consistency with other methods ?

src/Operator/Ternary/AbstractTernaryOperator.php
```diff
@@ -0,0 +1,62 @@
...
+        if (!$class = $this->getNodeClass()) {
+            throw new \LogicException(\sprintf('The "%s" %s operator must have a Node class.', $this->getOperator(), $this->getArity()->value));
+        }
+
+        $then = $parser->parseExpression($this->getPrecedence());
+        if ($parser->getStream()->nextIf(Token::PUNCTUATION_TYPE, $this->getElseOperator())) {
+            // Ternary operator (expr ? expr2 : expr3)
+            $else = $parser->parseExpression($this->getPrecedence());
+        } else {
+            // Ternary without else (expr ? expr2)
```
stof: Why would the abstract class enforce that all ternary operator have an optional `else` part that produces an empty string ? This seems to hardcode part of the behavior of the `? ... :` operator in the abstract class, making that abstract class not reusable (and so useless to have an abstract class)

src/Operator/Ternary/AbstractTernaryOperator.php
```diff
@@ -0,0 +1,62 @@
...
+            $else = new ConstantExpression('', $line);
+        }
+
+        return new $class($left, $then, $else, $line);
+    }
+
+    abstract public function getElseOperator(): string;
```
stof: why public ?

src/Operator/Unary/ParenthesisUnaryOperator.php
```diff
@@ -0,0 +1,67 @@
...
+        if (!$stream->test(Token::OPERATOR_TYPE, '=>')) {
+            throw new SyntaxError('A list of variables must be followed by an arrow.', $stream->getCurrent()->getLine(), $stream->getSourceContext());
+        }
+
+        return new ListExpression($names, $line);
+    }
+
+    public function getOperator(): string
+    {
+        return '(';
```
stof: I find it really weird to treat parenthesis as a unary operator (its parsing implementation does not look like the parsing of an expression either btw)

doc/operators_precedence.rst
```diff
@@ -0,0 +1,56 @@
...
+Binary and Ternary operators precedence:
+
+========== ========
+Precedence Operator
+========== ========
+
+300         |
+            .
+            [
+            (
```
stof: I think this needs more explanation to indicate that this parenthesis is representing the function call. I know no language that describe function calls as being a binary operator

src/Operator/Binary/FunctionBinaryOperator.php
```diff
@@ -0,0 +1,69 @@
...
+use Twig\Node\Expression\NameExpression;
+
+class FunctionBinaryOperator extends AbstractBinaryOperator
+{
+    private $readyNodes = [];
+
+    public function parse(ExpressionParser $parser, AbstractExpression $expr, int $line): AbstractExpression
+    {
+        if (!$expr instanceof NameExpression) {
+            throw new SyntaxError('Function name must be a string.', $line, $parser->getStream()->getSourceContext());
```
stof:
```suggestion
            throw new SyntaxError('Function name must be an identifier.', $line, $parser->getStream()->getSourceContext());
```

In terms of syntax, a string would have quotes around it.

doc/operators_precedence.rst
stof: this precedence looks weird to me compared to unary operators. How is `- myfunc()` parsed, before and after this change ? Same for `- var[0]`

src/ExpressionParser.php
```diff
@@ -864,10 +629,13 @@ private function setDeprecationCheck(bool $deprecationCheck): bool
         return $current;
     }
 
-    private function createArguments(int $line): ArrayExpression
+    /**
+     * @internal
+     */
+    public function createCallableArguments(int $line, bool $parseOpenParenthesis = true): ArrayExpression
```
stof: The name `create*` is weird to me. This method is parsing callable arguments, so I would name it `parse*` for consistency with other methods

src/Node/Expression/ListExpression.php
```diff
@@ -0,0 +1,48 @@
...
+
+class ListExpression extends AbstractExpression
+{
+    /**
+     * @param array<ContextVariable> $items
+     */
+    public function __construct(array $items, int $lineno)
+    {
+        foreach ($items as $item) {
+            if (!$item instanceof ContextVariable) {
```
stof: Do we need to validate this here ? Isn't this also ensured by the parser when using parsing ?
For custom constructed nodes, we already have the parameter type defining the contract and we generally don't add runtime checks for array values when we have such contract.

src/Lexer.php
```diff
@@ -360,17 +361,17 @@ private function lexExpression(): void
         // punctuation
         elseif (str_contains(self::PUNCTUATION, $this->code[$this->cursor])) {
             // opening bracket
-            if (str_contains('([{', $this->code[$this->cursor])) {
+            if (str_contains('{', $this->code[$this->cursor])) {
                 $this->brackets[] = [$this->code[$this->cursor], $this->lineno];
             }
             // closing bracket
-            elseif (str_contains(')]}', $this->code[$this->cursor])) {
+            elseif (str_contains('}', $this->code[$this->cursor])) {
```
stof: Do we still ensure we have proper balanced brackets for each type of brackets in the new logic ?

src/Lexer.php
```diff
@@ -337,14 +337,15 @@ private function lexExpression(): void
...
-        elseif ('=' === $this->code[$this->cursor] && ($this->cursor + 1 < $this->end) && '>' === $this->code[$this->cursor + 1]) {
-            $this->pushToken(Token::ARROW_TYPE, '=>');
-            $this->moveCursor('=>');
-        }
         // operators
         elseif (preg_match($this->regexes['operator'], $this->code, $match, 0, $this->cursor)) {
-            $this->pushToken(Token::OPERATOR_TYPE, preg_replace('/\s+/', ' ', $match[0]));
+            $operator = preg_replace('/\s+/', ' ', $match[0]);
+            $type = Token::OPERATOR_TYPE;
+            // to be removed in 4.0
```
stof: This removal will likely make things weird for the token stream (which will impact custom token parsers). After this removal, `(` would generator an `OPERATOR_TYPE` token, while `)` will not (will it stay a `PUNCTUATION_TYPE` ? it depends how the new `PUNCTUATION` constant will look like)

src/TokenParser/AbstractTokenParser.php
```diff
@@ -29,4 +33,26 @@ public function setParser(Parser $parser): void
     {
         $this->parser = $parser;
     }
+
+    protected function parseAssignmentExpression()
```
stof: this should have a return type (and probably some phpdoc explaining what it parsed in more details)

doc/operators_precedence.rst
stof: Btw, what happens if an extension defines an operator with higher precedence that this "function call" operator ? Using a function call near this operator would suddenly not parse the function call in the expected way.

src/ExpressionParser.php

src/Operator/Binary/FunctionBinaryOperator.php
fabpot: Correct

src/Operator/Ternary/AbstractTernaryOperator.php
fabpot: protected now

src/TokenParser/AbstractTokenParser.php
fabpot: Done

src/ExpressionParser.php
fabpot: Updated

src/Operator/Ternary/AbstractTernaryOperator.php
fabpot: Indeed, I've moved the parse() method to the implementation.

src/Lexer.php
fabpot: Fixed, with some new tests to cover the code.

CHANGELOG
```diff
@@ -1,6 +1,9 @@
 # 3.20.0 (2025-XX-XX)
 
- * n/a
+ * Bump minimum PHP version to 8.1
```
stof: You forgot to update the PHP requirement in composer.json to make Composer aware of this bump.

src/Extension/ExtensionInterface.php
```diff
@@ -25,6 +23,8 @@
  * Interface implemented by extension classes.
  *
  * @author Fabien Potencier <fabien@symfony.com>
+ *
+ * @method array<ExpressionParserInterface> getExpressionParsers()
```
stof, line 27: should the abstract extension define this method, to avoid reporting deprecations for all extensions that are using other extension points and don't need any migration work ? It does not make sense to force them to implement a method returning an empty array.

src/Extension/ExtensionInterface.php
```diff
@@ -66,11 +66,11 @@ public function getFunctions();
     /**
      * Returns a list of operators to add to the existing list.
      *
-     * @return array<array> First array of unary operators, second array of binary operators
+     * @return ExpressionParserInterface[]|array<array>
      *
-     * @psalm-return array{
-     *     array<string, array{precedence: int, precedence_change?: OperatorPrecedenceChange, class: class-string<AbstractUnary>}>,
-     *     array<string, array{precedence: int, precedence_change?: OperatorPrecedenceChange, class?: class-string<AbstractBinary>, associativity: ExpressionParser::OPERATOR_*}>
+     * @psalm-return ExpressionParserInterface[]|array{
```
stof: Should we change the return type of `getOperators` or should expression parsers always be returned by `getExpressionParsers` ? The ExtensionSet seems to consider that `getOperators` always return the old version.

src/ExtensionSet.php
```diff
@@ -479,8 +467,11 @@ private function initExtension(ExtensionInterface $extension): void
             $this->visitors[] = $visitor;
         }
 
-        // operators
-        if ($operators = $extension->getOperators()) {
+        // expression parsers
+        if (method_exists($extension, 'getExpressionParsers')) {
+            $this->expressionParsers->add($extension->getExpressionParsers());
+        } else {
```
stof: I would suggest to handle `getOperators` even when `getExpressionParsers` exists, to make it possible for the abstract extension to provide a default implementation of `getExpressionParsers` returning an empty array.

src/ExtensionSet.php
```diff
@@ -489,8 +480,53 @@ private function initExtension(ExtensionInterface $extension): void
...
+
+                if ($op['callable']) {
+                    $expressionParsers[] = $this->convertInfixExpressionParser($op['class'], $operator, $op['precedence'], $op['associativity'], $op['precedence_change'] ?? null, $op['aliases'] ?? [], $op['callable']);
+                } else {
+                    $expressionParsers[] = new BinaryOperatorExpressionParser($op['class'], $operator, $op['precedence'], $op['associativity'], $op['precedence_change'] ?? null, $op['aliases'] ?? []);
+                }
+            }
+
+            if (count($expressionParsers)) {
+                trigger_deprecation('twig/twig', '3.20', \sprintf('Extension "%s" uses the old signature for "getOperators()", please update it to return an array of "ExpressionParserInterface" objects.', \get_class($extension)));
```
stof: the deprecation message needs to be updated, asking to migrate to `getExpressionParsers` instead.

src/OperatorPrecedenceChange.php
```diff
@@ -11,32 +11,24 @@
...
-    public function getPackage(): string
-    {
-        return $this->package;
-    }
-
-    public function getVersion(): string
-    {
-        return $this->version;
-    }
+        trigger_deprecation('twig/twig', '1.20', 'The "%s" class is deprecated since Twig 1.20. Use "%s" instead.', self::class, PrecedenceChange::class);
```
stof: the versiion should be 3.20, not 1.20

src/Parser.php
```diff
@@ -337,11 +343,47 @@ public function popLocalScope(): void
...
+        trigger_deprecation('twig/twig', '3.20', 'Method "%s()" is deprecated, use "parseExpression()" instead.', __METHOD__);
+
+        if (null === $this->expressionParser) {
+            $this->expressionParser = new ExpressionParser($this, $this->env);
+        }
+
         return $this->expressionParser;
     }
 
+    public function parseExpression($precedence = 0)
```
stof:
```suggestion
    public function parseExpression(int $precedence = 0): AbstractExpression
```

src/Parser.php
```diff
@@ -380,6 +422,79 @@ public function getCurrentToken(): Token
         return $this->stream->getCurrent();
     }
 
+    public function getFunction(string $name, int $line): TwigFunction
+    {
+        if (!$function = $this->env->getFunction($name)) {
```
stof: to fix the issue discussed in https://github.com/twigphp/Twig/pull/4511#issuecomment-2598774060, those methods should be handling SyntaxError happening in those calls, to either rethrow them or ignore the unknown callable.
I was planning to send a PR to fix this in ExpressionParser.

src/TokenParser/AbstractTokenParser.php
```diff
@@ -29,4 +33,31 @@ public function setParser(Parser $parser): void
     {
         $this->parser = $parser;
     }
+
+    /**
+     * Parses an assignment expression like "a, b".
+     *
+     * @return Nodes<AssignContextVariable>
```
stof: `Nodes` is not defined as a generic class, so this type does not make sense.

src/TokenParser/SetTokenParser.php
```diff
@@ -70,4 +71,17 @@ public function getTag(): string
     {
         return 'set';
     }
+
+    private function parseMultitargetExpression()
```
stof:
```suggestion
    private function parseMultitargetExpression(): Nodes
```

tests/ExpressionParserTest.php
```diff
@@ -570,7 +571,7 @@ public function testUnaryPrecedenceChange()
     {
         $env = new Environment(new ArrayLoader(), ['cache' => false, 'autoescape' => false]);
         $env->addExtension(new class extends AbstractExtension {
-            public function getOperators()
+            public function getExpressionParsers()
```
stof:
```suggestion
            public function getExpressionParsers(): array
```

src/ExtensionSet.php
```diff
@@ -46,10 +52,8 @@ final class ExtensionSet
     private $functions;
     /** @var array<string, TwigFunction> */
     private $dynamicFunctions;
-    /** @var array<string, array{precedence: int, precedence_change?: OperatorPrecedenceChange, class: class-string<AbstractUnary>}> */
-    private $unaryOperators;
-    /** @var array<string, array{precedence: int, precedence_change?: OperatorPrecedenceChange, class?: class-string<AbstractBinary>, associativity: ExpressionParser::OPERATOR_*}> */
-    private $binaryOperators;
+    /** @var ExpressionParsers */
+    private $expressionParsers;
```
stof: the new property could be a typed property directly

src/ExpressionParser/Prefix/OperandExpressionParser.php
```diff
@@ -0,0 +1,256 @@
...
+use Twig\Node\Expression\ArrayExpression;
+use Twig\Node\Expression\Binary\ConcatBinary;
+use Twig\Node\Expression\ConstantExpression;
+use Twig\Node\Expression\Variable\ContextVariable;
+use Twig\Parser;
+use Twig\Token;
+
+class OperandExpressionParser implements PrefixExpressionParserInterface
+{
+    private string $type = 'OPERAND';
```
stof: why making this type use capital letters when all others are lowercase ?

src/ExpressionParser/Prefix/OperandExpressionParser.php
```diff
@@ -0,0 +1,256 @@
...
+
+    public function parse(Parser $parser, Token $token): AbstractExpression
+    {
+        $stream = $parser->getStream();
+        switch (true) {
+            case $token->test(Token::NAME_TYPE):
+                $stream->next();
+                switch ($token->getValue()) {
+                    case 'true':
+                    case 'TRUE':
```
stof: what about other cases ? If the boolean literals are case insensitive, shouldn't `True` or `trUE` also be accepted ?

src/ExpressionParser/Infix/ArgumentsTrait.php
```diff
@@ -0,0 +1,81 @@
...
+use Twig\Node\Expression\Unary\SpreadUnary;
+use Twig\Node\Expression\Variable\ContextVariable;
+use Twig\Node\Expression\Variable\LocalVariable;
+use Twig\Node\Nodes;
+use Twig\Parser;
+use Twig\Token;
+
+trait ArgumentsTrait
+{
+    protected function parseCallableArguments(Parser $parser, int $line, bool $parseOpenParenthesis = true): ArrayExpression
```
stof: should those be protected or private ? I find it weird that using the trait automatically forces the class to have some protected API (a class *wanting* to have it protected or public can still override the visibility when using the trait)

fabpot:
Operators are just a subset of "expression parsers". I've reworked this PR to introduce expression parsers instead of operators. It's more modular and more flexible. That allowed me to extract all logic from ExpressionParser to expression parser classes.

src/Operator/Unary/ParenthesisUnaryOperator.php
fabpot: Fixed by the new implementation.

src/ExpressionParser/Prefix/OperandExpressionParser.php
fabpot: It has never been supported.

fabpot:
@stof Most comments adressed.

src/Parser.php
fabpot: @stof Can you please submit a PR on this one? I don't plan to merge this PR anytime soon, so that would be great to be able to fix this issue outside of this PR. Thanks.

stof: Sure. I will (and I'll let you move the code here when rebasing your branch)

src/ExpressionParser/ExpressionParserType.php
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * (c) Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Twig\ExpressionParser;
+
+enum ExpressionParserType: string
```
stof, line 17: should this be tagged as `@internal` ?

src/ExpressionParser/ExpressionParsers.php
```diff
@@ -0,0 +1,138 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Twig\ExpressionParser;
+
+/**
+ * @template-implements \IteratorAggregate<ExpressionParserInterface>
+ */
+final class ExpressionParsers implements \IteratorAggregate
```
stof, line 19: should this be `@internal` ?

src/ExpressionParser/ExpressionParsers.php
```diff
@@ -0,0 +1,138 @@
...
+            foreach ($operator->getAliases() as $alias) {
+                $this->aliases[$type->value][$alias] = $operator;
+            }
+        }
+
+        return $this;
+    }
+
+    /**
+     * @param class-string<ExpressionParserInterface> $name
```
stof: should this be `class-string<PrefixExpressionParserInterface>` instead ?

src/ExpressionParser/ExpressionParsers.php
```diff
@@ -0,0 +1,138 @@
...
+    {
+        return
+            $this->parsers[ExpressionParserType::Prefix->value][$name]
+            ?? $this->aliases[ExpressionParserType::Prefix->value][$name]
+            ?? null
+        ;
+    }
+
+    /**
+     * @param class-string<ExpressionParserInterface> $name
```
stof: should also enforce getting a `InfixExpressionParserInterface` name.

And maybe those `*byClass` method should actually be generic method, taking a `class-string<T>` and returning `T|null`

src/ExpressionParser/Infix/ArrowExpressionParser.php
```diff
@@ -0,0 +1,44 @@
...
+
+use Twig\ExpressionParser\AbstractExpressionParser;
+use Twig\ExpressionParser\InfixAssociativity;
+use Twig\ExpressionParser\InfixExpressionParserInterface;
+use Twig\Node\Expression\AbstractExpression;
+use Twig\Node\Expression\ArrowFunctionExpression;
+use Twig\Parser;
+use Twig\Token;
+
+class ArrowExpressionParser extends AbstractExpressionParser implements InfixExpressionParserInterface
```
stof: I suggest making all those classes final. And I'm wondering whether they should be `@internal` as well.

src/ExpressionParser/Infix/BinaryOperatorExpressionParser.php
```diff
@@ -0,0 +1,66 @@
...
+use Twig\ExpressionParser\InfixExpressionParserInterface;
+use Twig\ExpressionParser\PrecedenceChange;
+use Twig\Node\Expression\AbstractExpression;
+use Twig\Parser;
+use Twig\Token;
+
+class BinaryOperatorExpressionParser extends AbstractExpressionParser implements InfixExpressionParserInterface
+{
+    public function __construct(
+        /** @var class-string<AbstractExpression> */
```
stof: should it support any AbstractExpression or should it enforce `AbstractBinary` as it expects the constructor signature of this class ? And should we make `AbstractBinary`  implement an (internal) interface enforcing the constructor signature ?

stof (commented):
My questions asked previously on the operators_precedence documentation (so called the most important part of the PR) have not been answered yet. It would be great to get some feedback on them.

src/ExpressionParser/Infix/BinaryOperatorExpressionParser.php
stof: Note that if we don't add an internal interface enforcing the constructor, we should probably tag AbstractBinary with `@phpstan-consistent-constructor`, so that at least projects using static analysis get such validation for their custom node classes

fabpot: Done

src/Node/Expression/ListExpression.php
fabpot: Removed

doc/operators_precedence.rst
fabpot: The last commit starts the work on this one.
I've merged unary and binary operator precedence tables.
I've added a way to add a description for each operator.
I've tweaked the precedence of () and [] as you mentioned.
I've deprecated precedence outside of -512/512 (arbitrary) to be able to "own" the lowest/highest precedence.

fabpot: Next step is to add a specific test suite for operators.

doc/operators_precedence.rst
```diff
@@ -0,0 +1,49 @@
+
+=========== ============== ======= ============= ===========
+Precedence  Operator       Type    Associativity Description
+=========== ============== ======= ============= ===========
+512         `.`            infix   Left          Get an attribute on a variable
+            `[`                                  Array access
+            `(`                                  Function call
+500         `-`            prefix  n/a
+            `+`
+300         `|`            infix   Left          Filter call
```
stof: is this describing the current precedence (in the 3.x version) or the future precedence defined by a PrecedenceChange ?

bin/generate_operators_precedence.php
```diff
@@ -0,0 +1,50 @@
+<?php
```
stof, line 1: The `bin` folder should be added in the list of folders ignored from exports in `.gitattributes` to avoid shipping this internal script in dist archives downloaded by composer.

doc/operators_precedence.rst
fabpot: I've just pushed more tests and changed the precedence of the filter operator to make an expression like ``{{ -1|abs }}`` works as expected.

doc/operators_precedence.rst
fabpot: The future precedence... which is what you "get" when you use parentheses (which is required to avoid deprecation notices). Maybe we should have both the current and the new precedence here? I didn't do it to avoid having too much information.

bin/generate_operators_precedence.php
fabpot: Good idea

src/ExpressionParser/ExpressionParsers.php
fabpot: I've modified these methods to generic ones.

doc/operators_precedence.rst
stof: when you use parentheses, the precedence of the operator does not matter anymore, as the parenthesis win anyway.

I think that the documentation in the 3.x branch needs to document the existing precedence as this is how parsing happens (we might add an info about the planned precedence change though)

fabpot: Updated

src/Parser.php
```diff
@@ -380,6 +415,79 @@ public function getCurrentToken(): Token
...
```
stof: Those methods need to be updated to preserve the fix done in https://github.com/twigphp/Twig/pull/4575

fabpot: Rebased

src/Lexer.php
fabpot: Code removed as not needed anymore.

fabpot:
I think I've taken all comments into account now.
Ready for maybe a lst round of reviews?

src/ExpressionParser/ExpressionParserInterface.php
```diff
@@ -0,0 +1,28 @@
...
+    public function __toString(): string;
+
+    public function getName(): string;
+
+    public function getPrecedence(): int;
+
+    public function getPrecedenceChange(): ?PrecedenceChange;
+
+    /**
+     * @return array<string>
```
nicolas-grekas, line 25: ?
```suggestion
     * @return list<string>
```

src/ExpressionParser/ExpressionParsers.php
```diff
@@ -0,0 +1,128 @@
...
+     * @var \WeakMap<ExpressionParserInterface, array<ExpressionParserInterface>>|null
+     */
+    private ?\WeakMap $precedenceChanges = null;
+
+    /**
+     * @param array<ExpressionParserInterface> $parsers
+     */
+    public function __construct(
+        array $parsers = [],
+    ) {
```
nicolas-grekas:
```suggestion
    public function __construct(array $parsers = [])
    {
```

src/ExpressionParser/ExpressionParsers.php
```diff
@@ -0,0 +1,128 @@
...
+     */
+    private array $parsersByClass = [];
+
+    /**
+     * @var \WeakMap<ExpressionParserInterface, array<ExpressionParserInterface>>|null
+     */
+    private ?\WeakMap $precedenceChanges = null;
+
+    /**
+     * @param array<ExpressionParserInterface> $parsers
```
nicolas-grekas, line 37: list instead of array everywhere?

src/ExpressionParser/ExpressionParsers.php
```diff
@@ -0,0 +1,128 @@
...
+    ) {
+        $this->add($parsers);
+    }
+
+    /**
+     * @param array<ExpressionParserInterface> $parsers
+     *
+     * @return $this
+     */
+    public function add(array $parsers): self
```
nicolas-grekas: for clarity:
```suggestion
    public function add(array $parsers): static
```

src/ExpressionParser/ExpressionParsers.php
```diff
@@ -0,0 +1,128 @@
...
+    {
+        foreach ($parsers as $parser) {
+            if ($parser->getPrecedence() > 512 || $parser->getPrecedence() < 0) {
+                trigger_deprecation('twig/twig', '3.20', 'Precedence for "%s" must be between 0 and 512, got %d.', $parser->getName(), $parser->getPrecedence());
+                // throw new \InvalidArgumentException(\sprintf('Precedence for "%s" must be between 0 and 512, got %d.', $parser->getName(), $parser->getPrecedence()));
+            }
+            $interface = $parser instanceof PrefixExpressionParserInterface ? PrefixExpressionParserInterface::class : InfixExpressionParserInterface::class;
+            $this->parsersByName[$interface][$parser->getName()] = $parser;
+            $this->parsersByClass[get_class($parser)] = $parser;
```
nicolas-grekas:
```suggestion
            $this->parsersByClass[\get_class($parser)] = $parser;
```

nicolas-grekas (commented):
Very very high overview :)

src/ExpressionParser/ExpressionParsers.php
stof: As your own code does not care about the keys, I don't see the reason to force the caller to provide a list.

src/ExpressionParser/ExpressionParsers.php
stof: This does not change anything as it is a final class anyway

Issue #2819, opened 2019-01-22
twig_date_converter timezone with time


anuri73:
Hello, I want to understand the logic of code in file twig/twig/lib/Twig/Extension/Core.php, line 433:

> $asString = (string) $date;
    if (ctype_digit($asString) || (!empty($asString) && '-' === $asString[0] && ctype_digit(substr($asString, 1)))) {
        $date = new DateTime('@'.$date);
    } else {
        $date = new DateTime($date, $env->getExtension('Twig_Extension_Core')->getTimezone());
    }

Why we initialize Datetime on second condition with default timezone if we got it in **$timezone** variable earlier? 

When I used this function with only times ("H:i") this function returns value in another timezone(UTC) and not what I expected(Ex: Asia/Almaty).

Steps to reproduce:
{{ train.Departure.Time | date('H:i', 'Asia/Almaty') }} return time in UTC timezone. But as you can see, I want to get time in Asia/Almaty timezone
where train.Departure.Time is some DateTime value

xabbuh:
Are you sure that `train.Departure.Time` is a `DateTime` object? Because in this case the function should exit earlier and never reach that point.

anuri73:
I think, You mean:
> if ($date instanceof DateTimeInterface) {

I'll check it again, but I'm pretty sure that $date was instance of DateTime

fabpot:
@anuri73 I'm closing as you weren't sure. Feel free to reopen if needed.

PR #4546, opened 2025-01-20
Fix timezone conversion on strings

### LINKED ISSUE #4545: Automatic timezone conversion from date_default_timezone to CoreExtension timezone broken for strings
When using a default time zone:
```php
date_default_timezone_set('UTC');
```

And specifying a different timezone for twig to display:
```php
$twig->getExtension(\Twig\Extension\CoreExtension::class)->setTimezone('Europe/Amsterdam');
```

Objects get converted from the default_timezone to the displayTimezone set with `setTimezone`:
```twig
[
    'dateObject' => new \DateTime('2000-01-01 12:00'),
]

{{ dateObject|date }} // January 1, 2000 13:00
```

This is currently not the case for strings however:
```twig
[
    'dateString' => '2000-01-01 12:00',
]

{{ dateString|date }} // January 1, 2000 12:00 => this should be 13:00
```

This is because for all DateTime(Immutable) objects, the [setTimeZone method gets called when the timezone is not false](https://github.com/twigphp/Twig/blob/4f3770ea781f4bf8055240a0e951a3fa2205c734/src/Extension/CoreExtension.php#L583), but [for strings the timezone is also passed when initializing a dateTime object](https://github.com/twigphp/Twig/blob/4f3770ea781f4bf8055240a0e951a3fa2205c734/src/Extension/CoreExtension.php#L579).

PrinsFrank:
When setting a default application timezone, and a display timezone:
```php
date_default_timezone_set('UTC');
$twig->getExtension(\Twig\Extension\CoreExtension::class)->setTimezone('Europe/Paris');
```

Date objects get converted, but strings don't. See #4545. 

For most other paths in `convertDate`, the dateTime object is initialized, and afterwards the timezone gets set. For strings, the timezone is passed as a constructor argument before having the timezone set again, losing its converting behaviour. This makes the convertDate method work identical for DateTime objects as strings. This different behaviour was already spotted in #3568.

Fixes #4545 #3568 #2819

### DIFF
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -576,7 +576,7 @@ final class CoreExtension extends AbstractExtension
         if (ctype_digit($asString) || ('' !== $asString && '-' === $asString[0] && ctype_digit(substr($asString, 1)))) {
             $date = new \DateTime('@'.$date);
         } else {
-            $date = new \DateTime($date, $this->getTimezone());
+            $date = new \DateTime($date);
         }
 
         if (false !== $timezone) {
--- a/tests/Fixtures/filters/date_time_zone_conversion.test
+++ b/tests/Fixtures/filters/date_time_zone_conversion.test
@@ -1,0 +1,91 @@
+--TEST--
+"date" filter with time zone conversion
+--TEMPLATE--
+{{ date1|date }}
+{{ date1|date('d/m/Y') }}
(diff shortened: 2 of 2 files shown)


PrinsFrank:
@fabpot Can I clarify anything or help with anything on this PR?

fabpot:
Thank you @PrinsFrank.

PR #4548, opened 2025-01-22
Ignore static properties when using the dot operator

### LINKED ISSUE #4547: Accessing Undefined Variable with Getter and Static Property with same name
In Twig, when using a class that has a getter and a public static property with the same name, Twig tries to access the static property instead of using the getter. This behavior has changed between versions 3.14 and 3.15. In version 3.14, the getter is used, but in version 3.15, the same code leads to an "undefined property" warning.

Minimal reproduction example:
```php
<?php

require __DIR__ . '/../vendor/autoload.php';

class Demo {
    public static $testName = 'Test Name';

    public function getTestName(): string {
        return self::$testName;
    }
}

$loader = new \Twig\Loader\ArrayLoader([
    'index' => 'Hello {{ demo.testName }}!',
]);
$twig = new \Twig\Environment($loader);

echo $twig->render('index', ['demo' => new Demo()]);
```

### Output twig version 3.14.0
```
Hello Test Name!
```

### Output twig version 3.15.0
```
PHP Notice:  Accessing static property Demo::$testName as non static in /tmp/twig/vendor/twig/twig/src/Extension/CoreExtension.php on line 1747
PHP Warning:  Undefined property: Demo::$testName in /tmp/twig/vendor/twig/twig/src/Extension/CoreExtension.php on line 1747
Hello !
```

### Question
Is this change an intentional adjustment related to the introduction of accessing class constants with the dot operator, or is it a regression/bug?

--------------------------

Thanks for your work!

nicolas-grekas:
Fix #4547

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,6 +4,7 @@
  * Add the `invoke` filter
  * Make `{}` optional for the `types` tag
  * Add `LastModifiedExtensionInterface` and implementation in `AbstractExtension` to track modification of runtime classes
+ * Ignore static properties when using the dot operator
 
 # 3.18.0 (2024-12-29)
 
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -2151,7 +2151,7 @@ final class CoreExtension extends AbstractExtension
 
         $property = $class->getProperty($property);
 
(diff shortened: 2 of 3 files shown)


fabpot:
Thank you @nicolas-grekas.

PR #4551, opened 2025-01-24
Fix inheritance logic to be able to use instanceof ContextVariable instead of NameExpression

fabpot:


src/Node/Expression/AssignNameExpression.php
```diff
@@ -15,8 +15,9 @@
 use Twig\Compiler;
 use Twig\Error\SyntaxError;
 use Twig\Node\Expression\Variable\AssignContextVariable;
+use Twig\Node\Expression\Variable\ContextVariable;
 
-class AssignNameExpression extends NameExpression
+class AssignNameExpression extends ContextVariable
```
fabpot, line 20: This is the main changes that allows us to remove any `instanceof NameExpression` in the code (as `NameExpression` is deprecated).

PR #4555, opened 2025-01-24
add missing parentheses to deprecation message

xabbuh:


fabpot:
Good catch, thanks @xabbuh.

PR #4560, opened 2025-01-28
Fix typo for html_cva code example

matthieupetel:
Fix a typo where 'red' should be 'blue' according to the code above.


stof (approved):

xabbuh (approved):

fabpot:
Good catch, thanks @matthieupetel.

PR #4562, opened 2025-01-30
Add some tests on balanced brackets

fabpot:


stof:
should we also have tests for other balancing errors ? For instance a closing bracket that has no corresponding opening bracket ?

PR #4568, opened 2025-02-03
Reimplement the cache tag in an adhoc way

fabpot:


extra/cache-extra/TokenParser/CacheTokenParser.php
```diff
@@ -31,24 +31,22 @@ public function parse(Token $token): Node
         $tags = null;
         while ($stream->test(Token::NAME_TYPE)) {
             $k = $stream->getCurrent()->getValue();
+            if (!in_array($k, ['ttl', 'tags'])) {
+                throw new SyntaxError(\sprintf('Unknown "%s" configuration.', $k), $stream->getCurrent()->getLine(), $stream->getSourceContext());
+            }
+
             $stream->next();
-            $args = $expressionParser->parseNamedArguments();
```
fabpot, line 35: This is really a hack that is not needed.

stof:
The PR title (and the commit message) should be updated to mention *which* tag it talks about.

fabpot:
> The PR title (and the commit message) should be updated to mention _which_ tag it talks about.

Indeed

extra/cache-extra/TokenParser/CacheTokenParser.php
```diff
@@ -31,24 +31,22 @@ public function parse(Token $token): Node
...
+            $stream->expect(Token::PUNCTUATION_TYPE, '(');
+            if ($stream->test(Token::PUNCTUATION_TYPE, ')')) {
+                throw new SyntaxError(\sprintf('The "%s" modifier takes exactly one argument (0 given).', $k), $stream->getCurrent()->getLine(), $stream->getSourceContext());
```
stof: shouldn't we keep this a similar better error message after the arg, when passing multiple args, instead of saying `Unexpected token "punctuation" of value "," ("punctuation" expected with value ")"` ?

fabpot: Sure, done

PR #4569, opened 2025-02-03
Add BinaryInterface

fabpot:
@stof That's what you had in mind for https://github.com/twigphp/Twig/pull/4543#discussion_r1939732064?

### DIFF
--- a/src/Node/Expression/Binary/BinaryInterface.php
+++ b/src/Node/Expression/Binary/BinaryInterface.php
@@ -1,0 +1,22 @@
+<?php
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) Fabien Potencier
+ *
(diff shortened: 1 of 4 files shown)


src/Node/Expression/Binary/BinaryInterface.php
```diff
@@ -0,0 +1,27 @@
...
+use Twig\Node\Node;
+
+/**
+ * @internal
+ */
+interface BinaryInterface
+{
+    /**
+     * @param AbstractExpression $left
+     * @param AbstractExpression $right
```
stof: I think the interface should define `AbstractExpression` instead. Variance rules will allow some classes to use `Node` for BC reasons (as it is a wider type)

fabpot: Updated

PR #4571, opened 2025-02-04
Add support for registered undefined element callbacks in tests

stof:
This also defines the proper type for the callable arguments in those methods, allowing to know the expected signature (and allowing static analysis tools to check it) instead of having to inspect the implementation to see how those callables are used.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,6 +2,7 @@
 
  * Fix wrong array index
  * Bump minimum PHP version to 8.1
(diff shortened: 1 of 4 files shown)


stof:
This will be useful to write integration tests for the guard tag ensuring that the case of handlers throwing a SyntaxError (which was supposed to be supported in 3.18.0) actually works (hint: it does not work in 3.18.0)

fabpot (approved):

fabpot:
Thank you @stof.

PR #4575, opened 2025-02-09
Fix support for ignoring syntax erros in an undefined handler in guard

stof:
This fixes the issue discussed in https://github.com/twigphp/Twig/pull/4511#issuecomment-2598774060 so that ignoring SyntaxError from undefined handlers wanting to provide a custom message actually works in the `{% guard %}` tag.

This adds an integration test covering that behavior to ensure it works.

### DIFF
--- a/src/ExpressionParser.php
+++ b/src/ExpressionParser.php
@@ -799,7 +799,17 @@ class ExpressionParser
 
     private function getFunction(string $name, int $line): TwigFunction
     {
-        if (!$function = $this->env->getFunction($name)) {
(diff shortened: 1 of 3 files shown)


fabpot (commented):
Shouldn't we do the same for tests as well?

stof:
There is no undefined handler for tests (and adding such an handler would be hard due to the fact that tests can be named either with 1 or 2 identifiers)

fabpot:
Thank you @stof.

PR #4577, opened 2025-02-13
Make the raw filter more "sticky"

fabpot:
When using `{{ foo|raw }}`, the expression won't be escaped as expected.

But if you store the result and reuse it afterwards, the value is not considered "safe" anymore:

`{% set bar = foo|raw %}{{ bar }}`

Here `bar` will be escaped.

For Twig 4.x, I propose to make the `raw` filter "sticky" by wrapping the value with `Markup`. That way, the second example won't be escaped as expected.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
 # 4.0.0 (2024-XX-XX)
 
+ * Make the `raw` filter sticky
  * Add back the `if` condition on `for` loops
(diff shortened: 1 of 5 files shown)


stof:
Is the `raw` filter properly optimized away when auto-escaping gets applied on a location, to avoid wrapping the value in a Markup object to have it cast to string right after (both as final filter before display and as filter in a location expecting a pre-escaped value ?)

PR #4581, opened 2025-02-14
Fix CS

fabpot:


src/Markup.php
```diff
@@ -46,9 +46,6 @@ public function count()
         return mb_strlen($this->content, $this->charset);
     }
 
-    /**
```
xabbuh, line 49: @fabpot Looks like Symfony's `DebugClassLoader` is not happy with this change: https://github.com/symfony/symfony/actions/runs/13329105184/job/37228898654?pr=59776#step:8:2832

xabbuh: see #4582

PR #4583, opened 2025-02-14
Remove template source from Template::loadTemplate()

### LINKED ISSUE #3266: Errors thrown for an invalid parent don't report their file and line as the Twig template
See the fiddle at https://twigfiddle.com/hmgy4z

The template name and the line are reported as part of the message, not as part of the exception location, as done for most other Twig errors (and so the Symfony error page would not extract a code snippet from the Twig file).

The cause is that the location is set by using `\Twig\Template::loadTemplate` in the compile code ([see the compilation logic](https://github.com/twigphp/Twig/blob/19f98c7298359f62a65ccf670b123e8836582394/src/Node/ModuleNode.php#L317-L325)) but that method takes only a template name, not a source context. And when it builds a source context from it, it cannot give it a source path, even if the template had the info.

A suggestion to fix that would be to change `\Twig\Template::loadTemplate` to accept a SourceContext rather than a template name (we might support both if needed) for cases which have it (most places now, as the compiled code always extract the name from the source context anyway).

fabpot:
Closes #3266

The first commit supports string and Source. But I think we don't even need to pass anything (that's the other commits).

### DIFF
--- a/src/Template.php
+++ b/src/Template.php
@@ -89,7 +89,7 @@ abstract class Template
         }
 
         if (!isset($this->parents[$parent])) {
-            $this->parents[$parent] = $this->loadTemplate($parent);
+            $this->parents[$parent] = $this->load($parent, -1);
         }
 
         return $this->parents[$parent];
@@ -270,21 +270,15 @@ abstract class Template
     /**
      * @param string|TemplateWrapper|array<string|TemplateWrapper> $template
      */
-    protected function loadTemplate($template, $templateName = null, $line = null, $index = null): self|TemplateWrapper
+    protected function load(string|TemplateWrapper|array $template, int $line, int|null $index = null): self
     {
         try {
             if (\is_array($template)) {
-                return $this->env->resolveTemplate($template);
+                return $this->env->resolveTemplate($template)->unwrap();
             }
 
             if ($template instanceof TemplateWrapper) {
-                return $template;
-            }
-
-            if ($template instanceof self) {
-                trigger_deprecation('twig/twig', '3.9', 'Passing a "%s" instance to "%s" is deprecated.', self::class, __METHOD__);
-
-                return $template;
+                return $template->unwrap();
             }
 
             if ($template === $this->getTemplateName()) {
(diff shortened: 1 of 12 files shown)


stof:
The test does not ensure the issue is solved as it asserts only the exception message, not the exception location.

fabpot:
> The test does not ensure the issue is solved as it asserts only the exception message, not the exception location.

Yes, I know, I will update it.

src/Template.php
```diff
@@ -270,7 +270,7 @@ public function getBlockNames(array $context, array $blocks = []): array
     /**
      * @param string|TemplateWrapper|array<string|TemplateWrapper> $template
      */
-    protected function loadTemplate($template, $templateName = null, $line = null, $index = null): self|TemplateWrapper
+    protected function loadTemplate($template, int|null $line = null, int|null $index = null): self|TemplateWrapper
```
smnandre: Ouch... this will break current implementation of Twig Components (yes, i know....) 

Do you have any vision on when this could be released ?

stof: Is Twig Component relying on extending an internal class ? And is it using a caret constraint for its Twig constraint ? If answering yes to both questions, there is a big issue in the Twig Components package as internal APIs are not covered by BC (and so relying on the caret constraint that relies on semver guarantees is wrong).

fabpot: We can maybe support both signatures until 4.0.

fabpot: @smnandre Should be good now.

fabpot:
@stof Should be good now.

src/Template.php
Kocal: Thanks @fabpot ❤️

smnandre: Thank you very much @fabpot! 

@stof this is what my "(yes, i know....)" implied. 

But my priorities here are and will always be
1- maintain the existing codebase / adapt how I can and when I can 
2- then if I have time and still a bit of energy after work... try to "fix" or improve things for the future

As you also know I have one other pressing issue to deal with... and I try to do so with the same order of priorities,  time, energy and help.

PR #4589, opened 2025-02-19
Make `in_array()` calls strict

alexandre-daubois:
I propose to add `strict: true` to all calls to `in_array()`, making the calls faster.


fabpot:
Thank you @alexandre-daubois.

Issue #3651, opened 2022-02-20
Trim filter marks string as unsafe; it is not HTML-aware


tremby:
I was expecting to be able to pass a safe string of HTML through `trim` and have it still marked as safe. However, as far as I can tell, the `trim` filter is returning a plain string, and Twig no longer treats it as safe HTML.

This means I cannot use the same code/macro to trim an input string which could be either safe `\Twig_Markup` or an unsafe string, and have a predictable output.

Consider this. It's at https://twigfiddle.com/hpnf8v

```twig
{% set myhtml %}
  Here is<br>my HTML  
{% endset %}

{% set myunsafestring = " i <3 u " %}

Plain output (fine, but I want it trimmed):
{{ myhtml }}
{{ myunsafestring }}

Raw (bad):
{{ myhtml | raw }}
{{ myunsafestring | raw }}

Trimmed (bad):
{{ myhtml | trim }}
{{ myunsafestring | trim }}

Trimmed and raw (bad):
{{ myhtml | trim | raw }}
{{ myunsafestring | trim | raw }}
```

- When passed through `raw`, the HTML is fine but not trimmed and the unsafe string breaks my markup.
- When passed through `trim`, the HTML is now trimmed but escaped (it should not be escaped), and the unsafe string is fine (escaped and trimmed)
- When passed through `trim` then `raw`, the HTML is trimmed and fine but the unsafe string breaks my markup.

I see this as a bug, if `\Twig_Markup` is meant to be a first-class data type. The built-in filters should be aware of it.

tremby:
Thank you!

lal65:
Just commenting here that this is kind of a backwards-incompatible change.

Consider this:
```
{% set x %}
  {% block someblock %}
  {% endblock %}
{% endset %}

{{ x|trim ? 'Show sensitive data' : 'Hide sensitive data' %}
```

previous to this change, you'd see this output:

```
Hide sensitive data
```

but afterwards...

```
Show sensitive data
```

since truthiness checks on \Twig\Markup objects will always be true.

fabpot:
> {% set x %}
>   {% block someblock %}
>   {% endblock %}
> {% endset %}

See #4591

PR #4591, opened 2025-02-21
Fix testing and expression when it evaluates to an instance of Markup

fabpot:
Refs #3651


fabpot:
Just in case it might help you @ruudk @drjayvee with some static analysis.

PR #4592, opened 2025-02-21
Simplify Error implementation

fabpot:


src/Error/Error.php
```diff
@@ -165,34 +148,25 @@ private function updateRepr(): void
...
-        if (null !== $template && null === $this->sourcePath) {
-            $src = $template->getSourceContext();
-            $this->sourceCode = $src->getCode();
-            $this->sourcePath = $src->getPath();
-        }
-
-        if (null === $template || $this->lineno > -1) {
+        if ($template) {
+            $this->source = $template->getSourceContext();
+        } elseif ($this->lineno > -1) {
```
smnandre, line 169: @fabpot  FYI 

I have _one_ failure due to this,  in a very specific case (where we did not set the line of file)

In this situation, line 150, `$template` is null 

I'll fix on our side, and then check if I can reproduce in real-life situation.

But there _may_ be a blind spot here that could concern other people out there.

Image

smnandre: There is a problem indeed. Ii have two potential fixes to suggest

## 1 - revert removal of early exit  in `Error.php` // works 100%


```diff
+        // Can happen if an exception is thrown during the rendering of a template
+        if (null === $template) {
+           return;
+       }

        $r = new \ReflectionObject($template);
        $file = $r->getFileName();
```

https://github.com/twigphp/Twig/blob/7cde13f6c98ea43a85b09c4bb3e0faed6921ab68/src/Error/Error.php


## 2 - wrap the exception in `Template.php` // still not sure

https://github.com/twigphp/Twig/blob/7cde13f6c98ea43a85b09c4bb3e0faed6921ab68/src/Template.php#L200-L206

Something like this fixes ~the~ my problem apparently

```php
            try {
                $this->display($context);
            } catch (\Throwable $e) {
                while (ob_get_level() > $level) {
                    ob_end_clean();
                }

+                throw new RuntimeError('Foo bar', previous: $e);
            }
```

smnandre: Example here: https://github.com/symfony/ux/actions/runs/13509786686/job/37747519361?pr=2593

fabpot: @smnandre Can you create a small reproducer?

fabpot: I've a reproducer now.

smnandre: I'm sorry did not see your comments. Will test your PR in my original scenario

PR #4594, opened 2025-02-22
Simplify Error

smnandre:
Follows #4593

### DIFF
--- a/src/Error/Error.php
+++ b/src/Error/Error.php
@@ -124,16 +124,10 @@ class Error extends \Exception
         }
 
         if ($this->source && $this->source->getName()) {
-            $name = $this->source->getName();
-            if (\is_string($name) || $name instanceof \Stringable) {
-                $name = \sprintf('"%s"', $name);
-            } else {
-                $name = json_encode($name);
-            }
-            $this->message .= \sprintf(' in %s', $name);
+            $this->message .= \sprintf(' in "%s"', $this->source->getName());
         }
 
-        if ($this->lineno && $this->lineno >= 0) {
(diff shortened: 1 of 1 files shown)


src/Error/Error.php
```diff
@@ -124,16 +124,10 @@ private function updateRepr(): void
         }
 
         if ($this->source && $this->source->getName()) {
-            $name = $this->source->getName();
-            if (\is_string($name) || $name instanceof \Stringable) {
```
smnandre, line 128: Can be simplified as` is_string($name)` is always true

src/Error/Error.php
```diff
@@ -124,16 +124,10 @@ private function updateRepr(): void
...
-                $name = \sprintf('"%s"', $name);
-            } else {
-                $name = json_encode($name);
-            }
-            $this->message .= \sprintf(' in %s', $name);
+            $this->message .= \sprintf(' in "%s"', $this->source->getName());
         }
 
-        if ($this->lineno && $this->lineno >= 0) {
```
smnandre, line 136: As we cannot have `!$this->lineno` and `$this->lineno = 0` at the same time

src/Error/Error.php
fabpot: Correct, this is something we've done already in 4.x.

src/Error/Error.php
fabpot: Indeed, it's all because we've added type hints which makes such code irrelevant.

fabpot:
Thank you @smnandre.

PR #4595, opened 2025-02-22
Improve error reporting

fabpot:


src/Error/Error.php
```diff
@@ -106,41 +106,26 @@ public function appendMessage($rawMessage): void
...
     private function updateRepr(): void
     {
-        $this->message = $this->rawMessage;
-
-        if ($this->source && $this->source->getPath() && $this->lineno > 0) {
-            $this->file = $this->source->getPath();
+        if ($this->lineno > 0) {
             $this->line = $this->lineno;
-
-            return;
```
fabpot, line 115: Not sure why we returned here, which means no file and line in the error message.

PR #4596, opened 2025-02-22
Remove uneeded property on Error

fabpot:


stof:
IIRC, this was done this way to override the native `line` property only when we can also override the `file`, to avoid having a line number not corresponding to the file being mentioned in the stack trace.

fabpot:
Right, see #4598

PR #4599, opened 2025-02-26
Fix Error when the trace has Twig file/line information instead of the original PHP info

fabpot:
See https://github.com/twigphp/Twig/pull/4592/files#r1966616393
/cc @smnandre


src/Error/Error.php
```diff
@@ -111,6 +115,8 @@ private function updateRepr(): void
             $this->file = $this->source->getPath();
             if ($this->lineno > 0) {
                 $this->line = $this->lineno;
+            } else {
+                $this->line = -1;
```
fabpot, line 119: Not related, but let's not keep the PHP line if we override the file with a Twig one.

stof:
is it possible to write a test avoiding regression on that case, as it seems not to be an obvious case given it has been missed previously ?

smnandre:
This does fix my scenario 👍

fabpot:
> is it possible to write a test avoiding regression on that case, as it seems not to be an obvious case given it has been missed previously ?

I've added many tests to ensure we're covering a lot of different scenarios.

src/Error/Error.php
```diff
@@ -144,6 +151,8 @@ private function guessTemplateInfo(): void
                 if ($this->source->getName() === $trace['object']->getTemplateName() && !$isEmbedContainer) {
                     $template = $trace['object'];
                     $templateClass = \get_class($trace['object']);
+
+                    break;
```
fabpot, line 155: Not related, but a small perf improvement

tests/ErrorTest.php
```diff
@@ -253,6 +259,112 @@ public function testTwigExceptionUpdateFileAndLineTogether()
...
+            {
+                return 'foo';
+            }
+        });
+
+        try {
+            $twig->render('no_line_and_context_exception.twig', ['line' => $errorLine]);
+            $this->fail();
+        } catch (RuntimeError $e) {
+            if (1 === $errorLine && !$addDebugInfo && !$exceptionWithLineAndContext) {
```
stof, line 317: Why are we reporting the exception without any line number when we don't have debug info or context and the exception is on first line of the template while we report it saying `at line 1` when the error happens on line 5 without the contextual info ? Where does this difference come from ? Is it because of the text node coming before that has some debug info on line 1 ?

fabpot: Correct, when the code is not on line 1, there is a TextNode that calls `addDebugInfo()`.

stof: I suggest adding a comment in the test explaining that, to avoid having future contributors wondering what could cause this (I spent quite some times reading this test before figuring out that the text node could be responsible for that)

fabpot: Done

PR #4603, opened 2025-02-27
remove not needed code

xabbuh:


xabbuh:
I looked at the errors reported by PHPStan. But I am not sure if the code really should be simplified like this or if adding the `break` in #4599 wasn't a mistake instead.

fabpot:
I've just added some more tests to this PR to ensure error reporting works well for `embed`.

fabpot:
Thank you @xabbuh.

PR #4605, opened 2025-02-28
reduce the number of deprecations being triggered

xabbuh:
Triggering deprecations for specific arguments is not needed as the method itself is deprecated as well.


stof (approved):

fabpot:
Thank you @xabbuh.

PR #4607, opened 2025-02-28
Fix `ModuleNode` instanciation when `$embeddedTemplates` is null

alexandre-daubois:
Related to https://github.com/symfony/symfony/pull/59886, in case `$embeddedTemplates` is null

### DIFF
--- a/src/Node/ModuleNode.php
+++ b/src/Node/ModuleNode.php
@@ -41,7 +41,7 @@ final class ModuleNode extends Node
         if (!$embeddedTemplates instanceof Node) {
             trigger_deprecation('twig/twig', '3.21', \sprintf('Not passing a "%s" instance as the "embedded_templates" argument of the "%s" constructor is deprecated.', Node::class, static::class));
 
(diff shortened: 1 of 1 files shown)


fabpot:
Thank you @alexandre-daubois.

src/Node/ModuleNode.php
```diff
@@ -41,7 +41,7 @@ public function __construct(Node $body, ?AbstractExpression $parent, Node $block
         if (!$embeddedTemplates instanceof Node) {
             trigger_deprecation('twig/twig', '3.21', \sprintf('Not passing a "%s" instance as the "embedded_templates" argument of the "%s" constructor is deprecated.', Node::class, static::class));
 
-            $embeddedTemplates = new Nodes($embeddedTemplates);
+            $embeddedTemplates = new Nodes($embeddedTemplates ?? []);
```
xabbuh, line 44: What about leveraging `EmptyNode` if `$embeddedTemplates` is `null` (see #4608)?

alexandre-daubois: Thank you for the PR!

PR #4610, opened 2025-03-06
Use `:` instead of `=` for named argument in the docs

jdreesen:


fabpot:
Good catch, thanks @jdreesen.

PR #4611, opened 2025-03-06
Update phpstan-baseline.neon

stof:
Ignore errors about Node protected properties that could be overridden in child classes with properties using hooks that would not allow unsetting array items. For now, we ignore this potential issue.

The solution to fix this would be to define the properties as private in 4.0, forcing child classes to use the public methods to access them. Another solution could be to mark the property as final, but this requires PHP 8.4+

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -23,3 +23,8 @@ parameters:
 			identifier: binaryOp.invalid
 			count: 1
 			path: src/Lexer.php
+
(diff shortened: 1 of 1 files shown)


stof:
Closing this as the failure seen in https://github.com/twigphp/Twig/actions/runs/13700986354/job/38314455486 was actually an issue in phpstan itself (unset array indexes is not an issue here) and it has already been fixed.

PR #4613, opened 2025-03-12
Update html_cva.rst

seb-jean:


VincentLanglet:
defaultVariant should also be updated to `default_variant`

and some context might help to review this PR:
Twig coding standard ask for snake case named parameters but the doc was using pascal case for html_cva.

fabpot:
Thank you @seb-jean.

PR #4615, opened 2025-03-14
fix: deprecated documentation

Garfield-fr:


fabpot:
Thank you @Garfield-fr.

PR #4617, opened 2025-03-20
[Doc] Fix `code-block` in html_cva

smnandre:
There is no `terminal` code-block in the current Twig documentation.

Using `bash` here instead is more in line with the other pages.

(...and will allow to simplify CSS needed for code blocks)


xabbuh (approved):

stof (approved):

fabpot:
Thank you @smnandre.

PR #4620, opened 2025-03-24
Update include.rst

checkk22:
twig with_context works with =


stof:
-1 for that. The recommended syntax for named arguments is now `:` rather than `=` (which is why the documentation has been updated to use it).

PR #4622, opened 2025-03-28
This PR adds support for the optional chaining operator (?.) to Twig.

shokhaa:
This PR adds support for the optional chaining operator (?.) to Twig.

The optional chaining operator (?.) enables developers to read the value of a property
located deep within a chain of connected objects without having to check that each
reference in the chain is valid.

Example usage:
```twig
{{ foo?.bar?.baz }}
```
This is equivalent to:
```twig
{{ foo.bar.baz is defined ? foo.bar.baz }}`

### DIFF
--- a/src/ExpressionParser/Infix/OptionalChainExpressionParser.php
+++ b/src/ExpressionParser/Infix/OptionalChainExpressionParser.php
@@ -1,0 +1,107 @@
+<?php
+
+namespace Twig\ExpressionParser\Infix;
+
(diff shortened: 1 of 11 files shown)


shokhaa:
./vendor/bin/simple-phpunit tests/Extension/OptionalChainingTest.php 


Testing Twig\Tests\Extension\OptionalChainingTest
......S                                                             7 / 7 (100%)

Time: 00:00.036, Memory: 10.00 MB

PR #4623, opened 2025-03-28
This PR adds support for the optional chaining operator (?.) to Twig.

shokhaa:
This PR adds support for the optional chaining operator (?.) to Twig.

The optional chaining operator (?.) enables developers to read the value of a property
located deep within a chain of connected objects without having to check that each
reference in the chain is valid.

Example usage:
```twig
{{ foo?.bar?.baz }}
```
This is equivalent to:
```twig
{{ foo.bar.baz is defined ? foo.bar.baz }}`

### DIFF
--- a/src/ExpressionParser/Infix/OptionalChainExpressionParser.php
+++ b/src/ExpressionParser/Infix/OptionalChainExpressionParser.php
@@ -35,11 +35,7 @@ final class OptionalChainExpressionParser extends AbstractExpressionParser imple
         $arguments = new ArrayExpression([], $lineno);
         $type = Template::ANY_CALL;
         $isOptionalChain = true;
-
-        // Проверка, является ли левое выражение переменной
         $isVariable = $expr instanceof NameExpression;
-
-        // Для обработки квадратных скобок
         if ($stream->test(Token::OPERATOR_TYPE, '[')) {
             $token = $stream->next();
             $attribute = $parser->parseExpression();
@@ -73,11 +69,11 @@ final class OptionalChainExpressionParser extends AbstractExpressionParser imple
                 || '_self' === $expr->getAttribute('name') && $attribute instanceof ConstantExpression
             )
         ) {
-            // Для макросов не используем optional chaining
+   
             return new MacroReferenceExpression(new TemplateVariable($expr->getAttribute('name'), $expr->getTemplateLine()), 'macro_'.$attribute->getAttribute('value'), $arguments, $expr->getTemplateLine());
         }
 
-        // Создаем специальный флаг для проверки существования переменной
+  
         if ($isVariable && $expr instanceof NameExpression) {
             $expr->setAttribute('optional_chain', true);
         }
--- a/src/ExpressionParser/Infix/OptionalChainExpressionParser.php
+++ b/src/ExpressionParser/Infix/OptionalChainExpressionParser.php
@@ -97,7 +97,7 @@ final class OptionalChainExpressionParser extends AbstractExpressionParser imple
 
     public function getPrecedence(): int
     {
-        return 512; // Тот же приоритет, что и у оператора "."
+        return 512;
     }
 
     public function getAssociativity(): InfixAssociativity
--- a/src/ExpressionParser/Infix/OptionalChainExpressionParser.php
+++ b/src/ExpressionParser/Infix/OptionalChainExpressionParser.php
@@ -1,0 +1,107 @@
+<?php
+
+namespace Twig\ExpressionParser\Infix;
+
+use Twig\Error\SyntaxError;
+use Twig\ExpressionParser\AbstractExpressionParser;
+use Twig\ExpressionParser\ExpressionParserDescriptionInterface;
+use Twig\ExpressionParser\InfixAssociativity;
+use Twig\ExpressionParser\InfixExpressionParserInterface;
+use Twig\Lexer;
+use Twig\Node\Expression\AbstractExpression;
+use Twig\Node\Expression\ArrayExpression;
+use Twig\Node\Expression\ConstantExpression;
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

**Comment by shokhaa** [comment]
./vendor/bin/simple-phpunit tests/Extension/OptionalChainingTest.php

Testing Twig\Tests\Extension\OptionalChainingTest
......S 7 / 7 (100%)

Time: 00:00.036, Memory: 10.00 MB

File: src/ExpressionParser/Infix/OptionalChainExpressionParser.php [hunk 1]
```diff
@@ -0,0 +1,107 @@
...
+    public function parse(Parser $parser, AbstractExpression $expr, Token $token): AbstractExpression
+    {
+        $stream = $parser->getStream();
+        $token = $stream->getCurrent();
+        $lineno = $token->getLine();
+        $arguments = new ArrayExpression([], $lineno);
+        $type = Template::ANY_CALL;
+        $isOptionalChain = true;
+
+        // Проверка, является ли левое выражение переменной
```
alexisLefebvre: What does that mean? It probably should be in English.

shokhaa: Yes, I think I forgot to make a translation

src/ExpressionParser/Infix/OptionalChainExpressionParser.php
```diff
@@ -0,0 +1,103 @@
...
+
+        if (
+            $expr instanceof NameExpression
+            && (
+                null !== $parser->getImportedSymbol('template', $expr->getAttribute('name'))
+                || '_self' === $expr->getAttribute('name') && $attribute instanceof ConstantExpression
+            )
+        ) {
+   
+            return new MacroReferenceExpression(new TemplateVariable($expr->getAttribute('name'), $expr->getTemplateLine()), 'macro_'.$attribute->getAttribute('value'), $arguments, $expr->getTemplateLine());
```
stof, line 73: this seems to ignore the fact that this is an optional chain. Is it expected ?

src/Extension/CoreExtension.php
```diff
@@ -391,6 +392,7 @@ public function getExpressionParsers(): array
 
             // all literals
             new LiteralExpressionParser(),
+            new OptionalChainExpressionParser(),
```
stof, line 395: this should be registered near other `get attribute operators`, not in the section labelled as `all literals`

src/Lexer.php
```diff
@@ -61,6 +61,7 @@ class Lexer
     public const REGEX_INLINE_COMMENT = '/#[^\n]*/A';
     public const PUNCTUATION = '()[]{}?:.,|';
 
+
```
stof, line 64: should be reverted

src/Node/Expression/GetAttrExpression.php
```diff
@@ -36,7 +36,7 @@ public function __construct(AbstractExpression $node, AbstractExpression $attrib
             trigger_deprecation('twig/twig', '3.15', \sprintf('Not passing a "%s" instance as the "arguments" argument of the "%s" constructor is deprecated ("%s" given).', ArrayExpression::class, static::class, $arguments::class));
         }
 
-        parent::__construct($nodes, ['type' => $type, 'ignore_strict_check' => false, 'optimizable' => true], $lineno);
+        parent::__construct($nodes, ['type' => $type, 'ignore_strict_check' => false, 'optimizable' => true, 'is_optional_chain' => $isOptionalChain], $lineno);
```
stof, line 39: Do we need another attribute for the optional chain ? Couldn't it be a matter of the `ignore_strict_check` attribute ?

src/Node/Expression/GetAttrExpression.php
```diff
@@ -113,7 +202,7 @@ public function compile(Compiler $compiler): void
 
         $compiler->raw(', ')
             ->repr($this->getAttribute('type'))
-            ->raw(', ')->repr($this->definedTest)
+            ->raw(', ')->repr($this->definedTest ?? false)
```
stof, line 205: why this change ?

src/Node/Expression/GetAttrExpression.php
```diff
@@ -124,7 +213,6 @@ public function compile(Compiler $compiler): void
             $compiler->raw(')');
         }
     }
-
```
stof, line 127: this empty line between methods should not be removed

tests/Extension/OptionalChainingTest.php
```diff
@@ -0,0 +1,14 @@
+<?php
+// tests/Extension/OptionalChainingTest.php
+
+namespace Twig\Tests\Extension;
+
+use Twig\Test\IntegrationTestCase;
+
+class OptionalChainingTest extends IntegrationTestCase
```
stof, line 8: There is no `OptionalChaining` extension. The operator is registered by CoreExtension. This should be tested by the existing integration tests by putting the fixtures in the existing folder.

stof (commented):
You need to run the `bin/generate_operators_precedence.php` script to regenerate the list of operators in the documentation to include your new operator.

Bilge:
So we doing this or nah?

ruudk:
@fabpot What do you think of this idea? Today I had again a situation where this would have helped tremendously by doing something like `{% if app.user.account.id == some.other.id %}` and `account` being null.

ryanleichty:
I'm dying for optional chaining in twig 🥵

felds:
@shokhaa Since this is stuck for a long time, do you thing we can package it so it can be installed as an extension? I've made an implementation that extends the `.` operator, but I think yours is easier to package.

fabpot:
Closing in favor of #4717. Thank you for starting this.

PR #4625, opened 2025-03-29
[Doc] Fix intro for operator precedence table ?

hellomedia:
The table introductory paragraph says operator with lowest precedence are listed first ( I assume it means "listed first in the operator table" ).

However, the table starts with a precedence of 500, which seems to be the highest.

Am I completely misreading this ?


fabpot (approved):

fabpot (approved):

fabpot:
Good catch, thanks @hellomedia.

PR #4629, opened 2025-04-24
Add base64_encode twig filter

alexander-schranz:
I'm using base64_encode a few times to bring data from backend to my JS code where it only is encoded on click, also binary strings of svg can be base64 encode for data string in img.

### DIFF
--- a/tests/Fixtures/filters/base64_encode.test
+++ b/tests/Fixtures/filters/base64_encode.test
@@ -1,0 +1,10 @@
+--TEST--
+"base64_encode" filter
+--TEMPLATE--
+{{ "foo"|base64_encode|raw }}
+{{ foo|base64_encode|raw }}
+--DATA--
+return ['foo' => new \Twig\Markup('foo', 'UTF-8')]
+--EXPECT--
+Zm9v
+Zm9v
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -246,6 +246,7 @@ final class CoreExtension extends AbstractExtension
             new TwigFilter('url_encode', [self::class, 'urlencode']),
             new TwigFilter('json_encode', 'json_encode'),
             new TwigFilter('convert_encoding', [self::class, 'convertEncoding']),
+            new TwigFilter('base64_encode', 'base64_encode'),
 
             // string filters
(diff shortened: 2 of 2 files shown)


stof:
For data URIs, we already have an existing `data_uri` filter in the `twig/html-extra` package (which takes care of everything needed, not just performing base64 encoding.

tests/Fixtures/filters/base64_encode.test
```diff
@@ -0,0 +1,10 @@
+--TEST--
+"base64_encode" filter
+--TEMPLATE--
+{{ "foo"|base64_encode|raw }}
+{{ foo|base64_encode|raw }}
+--DATA--
+return ['foo' => new \Twig\Markup('foo', 'UTF-8')]
```
stof, line 7: why using a Markup object here ?

tests/Fixtures/filters/base64_encode.test
```diff
@@ -0,0 +1,10 @@
+--TEST--
+"base64_encode" filter
+--TEMPLATE--
+{{ "foo"|base64_encode|raw }}
```
stof, line 4: why using the `raw` filter here ?

stof (commented):
This PR misses the documentation for the filter.

tests/Fixtures/filters/base64_encode.test
alexander-schranz: @stof copied test from json_encode can remove it.

tests/Fixtures/filters/base64_encode.test
alexander-schranz: used the test from json_encode as reference

alexander-schranz:
> For data URIs, we already have an existing data_uri filter in the twig/html-extra package (which takes care of everything needed, not just performing base64 encoding.

I'm using it outside of data_uri usage. Really just `data-my="{{ privateText|base64_encode }}"`. But nice to see there is data_uri which we can then reference in the docs also.

fabpot:
I'm not sure the use cases are common enough to justify having it in core.

alexander-schranz:
@fabpot but would it hurt :)?

fabpot:
We're trying to limit the number of core features to the ones that are useful for "many" apps.
This is always subjective of course, but as Twig is mature now, I'm more strict about what we had.

Issue #3802, opened 2023-01-13
reduce filter doesn't pass the key to the arrow function


ju1ius:
Hi,

The `filter` and `map` filters both pass the value and the key of the iterable to their callback.
However the `reduce` filter does not.
And _should_, at least WRT api consistency.

An example where it's useful:
```twig
{% set status_classes = {
  'success': 200,
  'warning': 400,
  'error': 500,
} %}

<span class="{{ status_classes|reduce((acc, v, k) => response.status_code >= v ? k : acc, '') }}">
  {{ response.status_code }}
</span>
```

stof:
PHP's `array_reduce` does not support receiving the key in the callback (unlike `array_map` and `array_filter`): https://www.php.net/array_reduce

ju1ius:
> PHP's array_reduce does not support receiving the key in the callback (unlike array_map and array_filter): https://www.php.net/array_reduce

Indeed, but then `twig_array_reduce` shouldn't be implemented using `array_reduce`.
Just like `twig_array_map` doesn't use `array_map`.

PR #4638, opened 2025-05-12
Update reduce.rst

AndCycle:
remove note, better argument for example
reduce doesn't use `array_reduce` internally

https://github.com/twigphp/Twig/issues/3802
https://github.com/twigphp/Twig/commit/609794c37668989b670d79e97baf8d5ce05e0a97


fabpot:
Good catch, thanks @AndCycle.

PR #4639, opened 2025-05-15
[Docs] Add TwigCsFixer in tools list

smnandre:
Twig CS Fixer is already present in the code style / conventions page, but it feels to me very logical we see it on the "integration / tools" page.


fabpot:
Good catch, thanks @smnandre.

PR #4640, opened 2025-05-17
replace typographic quote with straigt single quote

xabbuh:
make DOCtor-RST 1.68 happy and account for OskarStark/doctor-rst#2011


fabpot:
Thank you @xabbuh.

PR #3737, opened 2022-08-23
Added configuration for commonmark use in twig-extra-bundle.

doekenorg:
Following this tweet: https://twitter.com/ramsey/status/1561894460560138241, this PR adds the configuration options for `commonmark` to the `twig-extra-bundle`.

The configuration is added to the `LeagueCommonMarkConverterFactory` and in turn injected in the `CommonMarkConverter`. This way a user can provide the configuration for usage inside twig.

I'm not sure if adding the configuration inside the extension is the way to go.

Hope this helps out at least @ramsey :-)

Added `ignoreExtraKeys` to be somewhat future proof in case extra configuration is added later on, and also to be able to provide de configuration for extensions.

PR #4642, opened 2025-05-23
[CommonMark] Update configuration keys + allow extra keys for extensions

doekenorg:
This PR adds missing keys from the current version of CommonMark (2.7).

It also fixes a bug where extra keys are removed. The original intent was to keep the extra keys, so extensions could be configured as well. Currently, these keys are removed, meaning extensions can not be configured.

Original PR #3737.


fabpot:
Thank you @doekenorg.

PR #4643, opened 2025-05-28
Fix path when possible when an exception occurs

lyrixx:
I found an error in the integration between Twig and Symfony
If a template throw an error, Symfony will enhance the message.
If a path is found in the message, it'll turn it to a link

image

https://github.com/symfony/symfony/blob/7.4/src/Symfony/Component/ErrorHandler/ErrorRenderer/HtmlErrorRenderer.php#L304-L307
https://github.com/symfony/symfony/blob/7.4/src/Symfony/Component/ErrorHandler/Resources/views/exception.html.php#L18

However, in twig, we don't put an absolute path, but a relative one.
So it does not work. This PR puts an absolute link if possible

>[!CAUTION]
> I did not updated the test yet, I'm waiting for feedback

### DIFF
--- a/src/Error/Error.php
+++ b/src/Error/Error.php
@@ -122,8 +122,12 @@ class Error extends \Exception
         if ($punctuation = '.' === $last || '?' === $last ? $last : '') {
             $this->message = substr($this->message, 0, -1);
         }
-        if ($this->source && $this->source->getName()) {
-            $this->message .= \sprintf(' in "%s"', $this->source->getName());
+        if ($this->source) {
+            if ($this->source->getPath()) {
+                $this->message .= \sprintf(' in "%s"', $this->source->getPath());
+            } else if ($this->source->getName()) {
+                $this->message .= \sprintf(' in "%s"', $this->source->getName());
+            }
         }
         if ($this->lineno > 0) {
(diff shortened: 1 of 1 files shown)


src/Error/Error.php
```diff
@@ -122,8 +122,12 @@ private function updateRepr(): void
         if ($punctuation = '.' === $last || '?' === $last ? $last : '') {
             $this->message = substr($this->message, 0, -1);
         }
-        if ($this->source && $this->source->getName()) {
-            $this->message .= \sprintf(' in "%s"', $this->source->getName());
+        if ($this->source) {
```
fabpot, line 125: This change looks wrong to me. Here, we want to use the name of the template, not the path. The path/line information is part of the regular Exception properties.

lyrixx: But we need to do that to make Symfony works. I don't know how we can do better.

stof:
As discussed in the Symfony Slack channel, the HtmlErrorRenderer of Symfony must be fixed to stop assuming that anything looking like `in "..."` in an exception message corresponds to `...` being a valid absolute path.

> However, in twig, we don't put an absolute path, but a relative one.

We actually don't put a path at all (even relative). We put a Twig template name.

PR #4645, opened 2025-06-05
Update templates.rst: Removing duplication

ThomasLandauer:
Page: https://twig.symfony.com/doc/3.x/templates.html#dot_operator

Reason: This is repeated a few lines further down.


stof (approved):

fabpot:

PR #4651, opened 2025-06-17
Update date.rst: Fixing `null` return value

ThomasLandauer:
Page: https://twig.symfony.com/doc/3.x/functions/date.html

I'm getting `null` - isn't this the default behavior?

If you merge this, maybe the example should be changed too, since `somedate < null` isn't obviously true.
Plus (more important): The recommended way to actually get the current date should be shown ;-)

### DIFF
--- a/doc/functions/date.rst
+++ b/doc/functions/date.rst
@@ -19,7 +19,7 @@ You can pass a timezone as the second argument:
         {# do something #}
     {% endif %}
 
-If no argument is passed, the function returns the current date:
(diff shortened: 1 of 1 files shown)


stof:
The implementation returns the current date, not `null`: https://github.com/twigphp/Twig/blob/403bd9d73c2a010e5b26689f2f2eb9d7ddf391af/src/Extension/CoreExtension.php#L600-L606

ThomasLandauer:
Indeed, I overrode this in a custom function.
Sorry! So I'm closing here.

PR #4653, opened 2025-06-17
Allow usage of other Markdown converters than CommonMark in LeagueMarkdown

### LINKED ISSUE #3581: LeagueMarkdown converter is explicitly requiring CommonMarkConverter
League CommonMark: v1.6.6
Twig Markdown-Extra: v3.3.3
Twig: v3.3.3

The `LeagueMarkdown`  constructor is requiring the `$converter` parameter to explicitly be an instance of `CommonMarkConverter`, however with League CommonMark v1.6, the converter looks to have been split into two different instance types:
1. `MarkdownConverter`
2. `CommonMarkConverter` which extends `MarkdownConverter`

For our usage, we're needing to pass-in the `$environment` to the converter (GitHub flavor with some other slight modifications) So, with v1.6, we can't use `CommonMarkConverter` since it triggers a deprecation error:
> Passing an $environment into the "League\CommonMark\CommonMarkConverter" constructor is deprecated in 1.6 and will not be supported in 2.0; use MarkdownConverter instead. See https://commonmark.thephpleague.com/2.0/upgrading/consumers/#commonmarkconverter-and-githubflavoredmarkdownconverter-constructors for more details.

Would it be possible to change the constructor for `LeagueMarkdown` to only require `MarkdownConverter` or maybe even just the `ConverterInterface`?

xJuvi:
Since the `CommonMarkConverter` from `League\CommonMark` doesn't allow customizing the `Environment` class at construct-time, this simple change allows providing an instance the parent class, so we can both inject a config to the converter AND a custom list of extensions (like `CommonMarkCoreExtension`) via the constructor.

This PR fixes #3581

### DIFF
--- a/extra/markdown-extra/LeagueMarkdown.php
+++ b/extra/markdown-extra/LeagueMarkdown.php
@@ -12,13 +12,14 @@
 namespace Twig\Extra\Markdown;
 
 use League\CommonMark\CommonMarkConverter;
+use League\CommonMark\MarkdownConverter;
 
 class LeagueMarkdown implements MarkdownInterface
 {
     private $converter;
     private $legacySupport;
 
-    public function __construct(?CommonMarkConverter $converter = null)
+    public function __construct(?MarkdownConverter $converter = null)
     {
         $this->converter = $converter ?: new CommonMarkConverter();
         $this->legacySupport = !method_exists($this->converter, 'convert');


xJuvi:
Hey @stof and @fabpot,
the failing check isn't caused by my change - it happens before and is another bug. All other checks were successful.

The failig checks will be fixed in #4661 

Can you approve this PR?

xJuvi:
Hey @fabpot 
do you need any additional changes or is this ok and can be done by approval?

Kind regards
Hannes

fabpot:
Thank you @xJuvi.

PR #4659, opened 2025-07-01
Add integer cast to core extension

jasperP98:
In twig, there is no clean way to cast a value to an integer. For example, to cast the string '123' to an integer, you are obligated to do it via a hacky way like doing '123' + 0. This way twig converts it to an integer but this does not seem like the proper way.

### DIFF
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -241,6 +241,7 @@ final class CoreExtension extends AbstractExtension
             new TwigFilter('number_format', [$this, 'formatNumber']),
             new TwigFilter('abs', 'abs'),
             new TwigFilter('round', [self::class, 'round']),
+            new TwigFilter('int_cast', [$this, 'castInt']),
 
             // encoding
             new TwigFilter('url_encode', [self::class, 'urlencode']),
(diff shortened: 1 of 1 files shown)


src/Extension/CoreExtension.php
```diff
@@ -692,6 +693,22 @@ public function formatNumber($number, $decimal = null, $decimalPoint = null, $th
...
     }
 
+    /**
+     * Cast a value to an integer.
+     *
+     * @param mixed $value
+     * @return int
+     * @throws RuntimeError
+     */
+    public function castInt($value): intcastI
```
stof: this is invalid code

src/Extension/CoreExtension.php
```diff
@@ -241,6 +241,7 @@ public function getFilters(): array
             new TwigFilter('number_format', [$this, 'formatNumber']),
             new TwigFilter('abs', 'abs'),
             new TwigFilter('round', [self::class, 'round']),
+            new TwigFilter('int_cast', [$this, 'nt']),
```
stof: `nt` does not exist

stof (commented):
this is untested.

stof:
What is the use case for which you need an int cast though ?

jasperP98:
The use case is a string casting to an integer. In our specific use case, we want to push a numeric string to the datalayer. The datalayer expects an integer but there is no clean way to cast this in twig

src/Extension/CoreExtension.php
jasperP98: My IDE did a wrong copy paste when clicking through the function

src/Extension/CoreExtension.php
jasperP98: My IDE did a wrong copy paste when clicking through the function

fabpot:
I'm 👎 for adding in core as the transformation should probably not be done in the templates themselves.
You can still create your own custom extension.

PR #4661, opened 2025-07-03
add twig extra bundle missign return types

xJuvi:
Add missing return type on the twig extra bundle for symfony integration. At the moment the unit tests for this bundle are failig on php 8.4.


xJuvi:
Fixed as result of #4665

PR #4663, opened 2025-07-05
Escaper performance: avoid static variables

gharlan:
Replace two static variables with `match` expressions. 
And bypass a third static variable for the common charset `UTF-8`.

### DIFF
--- a/src/Runtime/EscaperRuntime.php
+++ b/src/Runtime/EscaperRuntime.php
@@ -140,6 +140,10 @@ final class EscaperRuntime implements RuntimeExtensionInterface
             case 'html':
                 // see https://www.php.net/htmlspecialchars
 
+                if ('UTF-8' === $charset) {
+                    return htmlspecialchars($string, \ENT_QUOTES | \ENT_SUBSTITUTE, 'UTF-8');
+                }
+
                 // Using a static variable to avoid initializing the array
                 // each time the function is called. Moving the declaration on the
                 // top of the function slow downs other escaping strategies.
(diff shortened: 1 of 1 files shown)


src/Runtime/EscaperRuntime.php
```diff
@@ -288,18 +293,13 @@ public function escape($string, string $strategy = 'html', ?string $charset = nu
                         * entities that XML supports. Using HTML entities would result in this error:
                         *     XML Parsing Error: undefined entity
                         */
-                        static $entityMap = [
+                        return match ($ord) {
```
gharlan, line 296: The `match` expression is faster: https://3v4l.org/5HYQE

src/Runtime/EscaperRuntime.php
```diff
@@ -140,6 +140,10 @@ public function escape($string, string $strategy = 'html', ?string $charset = nu
             case 'html':
                 // see https://www.php.net/htmlspecialchars
 
+                if ('UTF-8' === $charset) {
+                    return htmlspecialchars($string, \ENT_QUOTES | \ENT_SUBSTITUTE, $charset);
+                }
```
gharlan, line 145: Comparison (when using only the common charset `UTF-8`): https://3v4l.org/hTE8X

src/Runtime/EscaperRuntime.php
```diff
@@ -140,6 +140,10 @@ public function escape($string, string $strategy = 'html', ?string $charset = nu
             case 'html':
                 // see https://www.php.net/htmlspecialchars
 
+                if ('UTF-8' === $charset) {
+                    return htmlspecialchars($string, \ENT_QUOTES | \ENT_SUBSTITUTE, $charset);
```
fabpot:
```suggestion
                    return htmlspecialchars($string, \ENT_QUOTES | \ENT_SUBSTITUTE, 'UTF-8');
```

Probably not faster, but maybe easier to read.

fabpot (approved):

fabpot:
Thank you @gharlan.

PR #4665, opened 2025-07-10
Fix compatibility with Symfony 8

nicolas-grekas:


extra/twig-extra-bundle/DependencyInjection/TwigExtraExtension.php
```diff
@@ -18,13 +18,36 @@
 use Symfony\Component\DependencyInjection\Loader\PhpFileLoader;
 use Twig\Extra\TwigExtraBundle\Extensions;
 
+if (!method_exists(ContainerBuilder::class, 'getAutoconfiguredAttributes')) {
+    /** @internal */
+    trait TwigExtraExtensionTrait
+    {
+        public function load(array $configs, ContainerBuilder $container): void
```
gharlan, line 25: why not just adding the return type (-> #4661)?
Especially since the trait is internal.

jdreesen: Because the `TwigExtraExtension` in which it is used is not `@internal` and thus adding the return type would be a BC break.

nicolas-grekas: Because that'd be a BC break if anyone extends the class and uses Symfony 7

fabpot:
Thank you @nicolas-grekas.

PR #4670, opened 2025-07-28
Document that the render function can throw \ErrorException

gnat42:
The twig `render` function can throw an exception of type `\ErrorException` in the case of errors such as an array to string conversion. This documents that its a valid exception that can be thrown. Otherwise, tools like phpstan complain that you are catching an exception that can never be thrown

### DIFF
--- a/src/Environment.php
+++ b/src/Environment.php
@@ -327,6 +327,7 @@ class Environment
      * @throws LoaderError  When the template cannot be found
      * @throws SyntaxError  When an error occurred during compilation
      * @throws RuntimeError When an error occurred during rendering
+     * @throws \ErrorException When there syntax/php error, i.e. array to string
      */
     public function render($name, array $context = []): string
     {


fabpot:
Can you share a small template that would throw an `\ErrorException`?

gnat42:
Hello Fabien,

   Making changes to one of the unit tests shows the behaviour I'm seeing.  The test itself has a number of variables passed in through the context, as follows:
   
```
        $context = [
            'string' => 'foo',
            'null' => null,
            'empty_array' => [],
            'array' => ['foo' => 'foo'],
            'array_access' => new TemplateArrayAccessObject(),
            'magic_exception' => new TemplateMagicPropertyObjectWithException(),
            'object' => new \stdClass(),
        ]; 
``` 

So if in the twig template you had `{{ array }}` you get a "Array to string conversion" error. In our project it came as an \ErrorException. In the test modification below I had to catch \Exception.   I'm not 100% sure why catching \ErrorException in the unit test doesn't work. When looking at the exception type, I was getting a class of type `PHPUnit\Framework\Error\Warning` so I'm not sure if its a PHPUnit thing or a php configuration thing. I suppose the other way to "fix" this is to catch exceptions like these "Array to string conversion" and re-throw a Twig exception instead.

```
diff --git a/tests/TemplateTest.php b/tests/TemplateTest.php
index eb0e2db7..2378e008 100644
--- a/tests/TemplateTest.php
+++ b/tests/TemplateTest.php
@@ -58,6 +58,8 @@ class TemplateTest extends TestCase
             $this->fail('Accessing an invalid attribute should throw an exception.');
         } catch (RuntimeError $e) {
             $this->assertSame(\sprintf($message, 'index'), $e->getMessage());
+        } catch (\Exception $e) {
+            $this->assertSame(\sprintf($message, 'index'), $e->getMessage());
         }
     }
 
@@ -81,6 +83,7 @@ class TemplateTest extends TestCase
             ['{% from _self import foo %}{% macro foo(obj) %}{{ obj.missing_method() }}{% endmacro %}{{ foo(array_access) }}', 'Neither the property "missing_method" nor one of the methods "missing_method()", "getmissing_method()"/"ismissing_method()"/"hasmissing_method()" or "__call()" exist and have public access in class "Twig\Tests\TemplateArrayAccessObject" in "%s" at line 1.'],
             ['{{ magic_exception.test }}', 'An exception has been thrown during the rendering of a template ("Hey! Don\'t try to isset me!") in "%s" at line 1.'],
             ['{{ object["a"] }}', 'Impossible to access a key "a" on an object of class "stdClass" that does not implement ArrayAccess interface in "%s" at line 1.'],
+            ['{{ array }}', 'Array to string conversion'],
         ];
     }
```

stof:
ErrorException is not something thrown by Twig. It is something thrown by the ErrorHandler in Symfony.

That's why you get a different exception in tests: PHPUnit uses its own error handler, not the Symfony error handler.

And this means that this PR is wrong. Twig does not provide any guarantee about how PHP warnings and errors are handled (not even that they are turned into exceptions at all).

PR #4674, opened 2025-08-02
Move some @throws phpdocs where they belong

fabpot:


src/Environment.php
```diff
@@ -323,10 +323,6 @@ public function getTemplateClass(string $name, ?int $index = null): string
      * Renders a template.
      *
      * @param string|TemplateWrapper $name The template name
-     *
-     * @throws LoaderError  When the template cannot be found
-     * @throws SyntaxError  When an error occurred during compilation
-     * @throws RuntimeError When an error occurred during rendering
```
stof, line 329: This should stay there on the public API IMO, as this documents the public behavior of Twig.

PR #4678, opened 2025-08-04
Add missing use statements in ExtensionInterface

### LINKED ISSUE #4677: Forgotten use statements in ExtensionInterface
The `src/Extension/ExtensionInterface::getOperators()` references the AbstractUnary, AbstractBinary and ExpressionParser classes in the psalm return.

The interface does not contain use statements for these classes though. In my case i ran into this when creating a unit test specifically for twig extensions, where we test if our twig extensions override methods (including the getOperators), to ensure we do not forget to add tests for them.

While performing our regular Psalm checks the following error popped up:
```
ERROR: UndefinedDocblockClass - src/AbstractTests/AbstractTwigExtensionTestCase Docblock-defined class, interface or enum named Twig\Extension\ExpressionParser does not exist (see https://psalm.dev/200)
            $this->getExtension()->getOperators(),
```

Mock example of our test class:
```php
<?php

abstract class AbstractTwigExtensionTestCase extends TestCase
{
    final public function testGetOperators(): void
    {
        self::assertEquals(
            $this->getExpectedOperators(),
            $this->getExtension()->getOperators(),
        );
    }

    protected function getExpectedOperators(): array
    {
        return [[], []];
    }

    abstract protected function getExtension(): AbstractExtension;
}

KevinVanSonsbeek:
Fixes #4677 

Note: The classes were used specifically in a `@psalm-return` doc statement. But as the repository does not have any psalm config, and does not check it in the CI, this seems to have gone unnoticed for a few months.

### DIFF
--- a/src/Extension/ExtensionInterface.php
+++ b/src/Extension/ExtensionInterface.php
@@ -11,8 +11,11 @@
 
 namespace Twig\Extension;
 
+use Twig\ExpressionParser;
 use Twig\ExpressionParser\ExpressionParserInterface;
 use Twig\ExpressionParser\PrecedenceChange;
+use Twig\Node\Expression\Binary\AbstractBinary;
+use Twig\Node\Expression\Unary\AbstractUnary;
 use Twig\NodeVisitor\NodeVisitorInterface;
 use Twig\TokenParser\TokenParserInterface;
 use Twig\TwigFilter;


fabpot:
Good catch, thanks @KevinVanSonsbeek.

PR #4687, opened 2025-08-29
Introduce registerUndefinedTestCallback

VincentLanglet:
This is similar to `registerUndefinedFilterCallback` and `registerUndefinedFunctionCallback`

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,7 @@
 # 3.22.0 (2025-XX-XX)
 
  * Add support for two words test in guard tag
+ * Add `Environment::registerUndefinedTestCallback()`
 
 # 3.21.1 (2025-05-03)
 
--- a/src/Environment.php
+++ b/src/Environment.php
@@ -828,6 +828,14 @@ class Environment
     }
 
     /**
+     * @param callable(string): (TwigTest|false) $callable
+     */
+    public function registerUndefinedTestCallback(callable $callable): void
+    {
+        $this->extensionSet->registerUndefinedTestCallback($callable);
+    }
+
+    /**
      * @return void
      */
     public function addFunction(TwigFunction $function)
--- a/src/Test/IntegrationTestCase.php
+++ b/src/Test/IntegrationTestCase.php
@@ -102,6 +102,14 @@ abstract class IntegrationTestCase extends TestCase
     }
 
     /**
+     * @return array<callable(string): (TwigTest|false)>
+     */
+    protected function getUndefinedTestCallbacks(): array
+    {
+        return [];
+    }
+
+    /**
      * @return array<callable(string): (TokenParserInterface|false)>
      */
     protected function getUndefinedTokenParserCallbacks(): array
@@ -253,6 +261,10 @@ abstract class IntegrationTestCase extends TestCase
 
             foreach ($this->getUndefinedFunctionCallbacks() as $callback) {
                 $twig->registerUndefinedFunctionCallback($callback);
+            }
+
+            foreach ($this->getUndefinedTestCallbacks() as $callback) {
+                $twig->registerUndefinedTestCallback($callback);
             }
 
             foreach ($this->getUndefinedTokenParserCallbacks() as $callback) {
--- a/doc/recipes.rst
+++ b/doc/recipes.rst
@@ -271,13 +271,19 @@ Defining undefined Functions, Filters, and Tags on the Fly
     The ``registerUndefinedTokenParserCallback()`` method was added in Twig
     3.2.
 
-When a function/filter/tag is not defined, Twig defaults to throw a
+.. versionadded:: 3.22
+
+    The ``registerUndefinedTestCallback()`` method was added in Twig
+    3.22.
+
+When a function/filter/test/tag is not defined, Twig defaults to throw a
 ``\Twig\Error\SyntaxError`` exception. However, it can also call a `callback`_
-(any valid PHP callable) which should return a function/filter/tag.
+(any valid PHP callable) which should return a function/filter/test/tag.
 
 For tags, register callbacks with ``registerUndefinedTokenParserCallback()``.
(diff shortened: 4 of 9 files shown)


fabpot (changes requested):
Can you add a note in CHANGELOG?

stof:
How does this behave regarding the fact that test names can be either 1 identifier or 2 identifiers ? Will this call the undefined test callback for the first identifier _before_ trying to find a known test named in 2 words ?

And doesn't this require a special handling for guard nodes for the case where the callback is throwing its own SyntaxError to provide a custom error message ?

VincentLanglet:
> How does this behave regarding the fact that test names can be either 1 identifier or 2 identifiers ? Will this call the undefined test callback for the first identifier _before_ trying to find a known test named in 2 words ?

Yes, the same way `getTest` already works.

Currently, if I define a `divisible` test, the `divisible by` test won't work anymore.
I think it could be better to only use single-word tests or to reserve some words, but that another debate.

Here, when implementing registerUndefinedTestCallback, I'll have to return false for `divisible` and `same` ; and a mock for others.

> And doesn't this require a special handling for guard nodes for the case where the callback is throwing its own SyntaxError to provide a custom error message ?

I copied the implementation of registerUndefinedTokenParserCallback, registerUndefinedFunctionCallback and registerUndefinedFilterCallback ; why would I need a special handling here ?

stof:
> Yes, the same way `getTest` already works.

The difference is that currently, there is nothing special being run when the first identifier does not correspond to an undefined test. The fact that the parser checks 2 different test names is not an observable behavior currently.
Calling the undefined handler callback for `starts` when parsing `{% if foo starts with('bar') %}` looks wrong to me as `starts with` is a defined test.

> I copied the implementation of registerUndefinedTokenParserCallback, registerUndefinedFunctionCallback and registerUndefinedFilterCallback ; why would I need a special handling here ?

Because those undefined callbacks *also* required special handling to account for the case where the callback is throwing. See https://github.com/twigphp/Twig/pull/4575 where this was implemented to fix https://github.com/twigphp/Twig/issues/4505

stof:
> Here, when implementing registerUndefinedTestCallback, I'll have to return false for `divisible` and `same` ; and a mock for others.

This makes it possible for an undefined test callback to break core tests. This looks very fragile to me, and is not something I would like to see being shipped in Twig (it will be a maintenance nightmare, especially to preserve BC on it).

VincentLanglet:
> The difference is that currently, there is nothing special being run when the first identifier does not correspond to an undefined test. The fact that the parser checks 2 different test names is not an observable behavior currently. Calling the undefined handler callback for `starts` when parsing `{% if foo starts with('bar') %}` looks wrong to me as `starts with` is a defined test.

After a new look @stof, this issue doesn't exist since the Twig Parser is doing the right check (2 word or 1 word) based on the number of name token.
https://github.com/twigphp/Twig/blob/98b664a0ca0a76149ac9306b23d5e827f2bcb761/src/Parser.php#L493-L502

So you can mock the tests without expecting issue with one/two words tests. I tried on my project.

BTW, `starts with` is an operaor.

I think tests are used in the format `{% if foo is myTest('bar') %}`

> > I copied the implementation of registerUndefinedTokenParserCallback, registerUndefinedFunctionCallback and registerUndefinedFilterCallback ; why would I need a special handling here ?
> 
> Because those undefined callbacks _also_ required special handling to account for the case where the callback is throwing. See #4575 where this was implemented to fix #4505

Should be updated

stof:
Indeed, I used the wrong example when using `starts with`. I should have used `same as` or `divisible by`

> So you can mock the tests without expecting issue with one/two words tests. I tried on my project.

would be great to add tests covering this to prevent regressions (add a undefined test handler doing something for `divisible` for instance)

stof:
And I suggest also adding a test guarding a 2-word test name to ensure it works.

VincentLanglet:
> And I suggest also adding a test guarding a 2-word test name to ensure it works.

GuardTokenParser didn't have support for 2-word test in the first place. I had to add it.

All tests are added.

src/Environment.php
```diff
@@ -827,6 +827,14 @@ public function getTest(string $name): ?TwigTest
         return $this->extensionSet->getTest($name);
     }
 
+    /**
+     * @param callable(string): (TokenParserInterface|false) $callable
```
stof: this type is wrong. It must be `TwigTest|false` as return type.

src/Test/IntegrationTestCase.php
```diff
@@ -101,6 +101,14 @@ protected function getUndefinedFunctionCallbacks(): array
         return [];
     }
 
+    /**
+     * @return array<callable(string): (TwigTest|false)>
+     */
+    protected function getUndefinedTestCallbacks(): array
```
stof, line 107: you need to actually register those callbacks in `doIntegrationTest`, so that overriding this method has an effect.

VincentLanglet:
If preferred @fabpot @stof I extracted the support for two words test in guard tag in a PR https://github.com/twigphp/Twig/pull/4689 which can be reviewed/merged first.

VincentLanglet:
Rebased @fabpot @stof 

Ci failure seems unrelated.

CHANGELOG
```diff
@@ -1,6 +1,7 @@
 # 3.22.0 (2025-XX-XX)
 
  * Add support for two words test in guard tag
+ * Add `Environment::registerUndefinedTestCallback()` method.
```
```suggestion
 * Add `Environment::registerUndefinedTestCallback()`
```

fabpot (changes requested):
I think you also need to update the docs for that new feature.

VincentLanglet:
> I think you also need to update the docs for that new feature.

Indeed, I updated the doc

fabpot (approved):

fabpot:
Thank you @VincentLanglet.

PR #4690, opened 2025-09-12
Improve documentation examples for `enum` and `enum_cases`

Kocal:
It may be obvious, but I first thought initializing an BackendEnum with a specific value could be possible with `enum(FQCN, value)`, but it does not, it always returns the first case  instance (if exists).

Instead, `enum(FQCN).from(value)` must be used, and I think it's worth to improve `enum()` examples.

### DIFF
--- a/doc/functions/enum.rst
+++ b/doc/functions/enum.rst
@@ -10,15 +10,21 @@
 .. code-block:: twig
 
     {# display one specific case of a backed enum #}
-    {{ enum('App\\MyEnum').SomeCase.value }}
+    {{ enum('App\\CardSuite').Clubs.value }} {# "clubs" #}
 
     {# get all cases of an enum #}
-    {% for case in enum('App\\MyEnum').cases %}
+    {% for case in enum('App\\CardSuite').cases %}
         {{ case.value }}
     {% endfor %}
+    {# "clubs", "spades", "hearts", "diamonds" #}
+
+    {# get a specific case of an enum by value #}
+    {% set card_suite = enum('App\\CardSuite').from('hearts') %}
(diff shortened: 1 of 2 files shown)


doc/functions/enum_cases.rst
```diff
@@ -9,9 +9,10 @@
 
 .. code-block:: twig
 
-    {% for case in enum_cases('App\\MyEnum') %}
+    {% for case in enum('App\\CardSuite').cases %}
```
stof: this is wrong. The documentation for the `enum_cases` function must use the `enum_cases` function.

stof (changes requested):

doc/functions/enum.rst
```diff
@@ -10,15 +10,21 @@
...
+    {{ enum('App\\CardSuite').Clubs.value }} {# "clubs" #}
 
     {# get all cases of an enum #}
-    {% for case in enum('App\\MyEnum').cases %}
+    {% for case in enum('App\\CardSuite').cases %}
         {{ case.value }}
     {% endfor %}
+    {# "clubs", "spades", "hearts", "diamonds" #}
+
+    {# get a specific case of an enum #}
```
stof:
```suggestion
    {# get a specific case of an enum by value #}
```

Otherwise, it is not clear what the difference is with the first example.

doc/functions/enum_cases.rst
Kocal: Oops I went too quick on this one, nice catch

fabpot (approved):

fabpot:
Thank you @Kocal.

PR #4693, opened 2025-09-25
Compile 'index' with repr (not string) in EmbedNode

smnandre:
Before this PR, the generated Template code had quotes around the index (integer) parameter value.


fabpot (approved):

fabpot:
Thank you @smnandre.

PR #4694, opened 2025-09-25
[Doc] Tweaks in the escaping article

javiereguiluz:
AS you can see in the redenred page (https://twig.symfony.com/doc/3.x/filters/escape.html) the tip about `html_attr` contains some uneeded line breaks.

This also tweaks another explanation to make it more clear.


fabpot (approved):

fabpot:
Good catch, thanks @javiereguiluz.

PR #4698, opened 2025-10-18
Add throw tag to parse methods

VincentLanglet:
Hi @fabpot, with PHPStan improving every day his exception checking, it's useful to have well documented method.

I didn't update all of them (there are too many), and focus on the one I have issue with:
- Almost all the InfixExpressionParserInterface/PrefixExpressionParserInterface implements throws a SyntaxError in parse method
- Parser::parse method is catching SyntaxError and rethrowing them so it should be added to the phpdoc
- Since Parser::parse catch Parser::subparse, it would be useful to add the phpdoc to subparse

### DIFF
--- a/src/ExpressionParser/InfixExpressionParserInterface.php
+++ b/src/ExpressionParser/InfixExpressionParserInterface.php
@@ -11,12 +11,16 @@
 
 namespace Twig\ExpressionParser;
 
+use Twig\Error\SyntaxError;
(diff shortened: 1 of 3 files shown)


VincentLanglet:
Cs failure seems unrelated ; should I fix it @fabpot ?

fabpot (approved):

fabpot:
Thank you @VincentLanglet.

PR #4700, opened 2025-10-19
Update format_datetime.rst: Adding link to `knplabs/knp-time-bundle`

ThomasLandauer:
Page: https://twig.symfony.com/doc/3.x/filters/format_datetime.html

If you agree, I'd add the same note to all date/time filters.


fabpot:
Instead, I would prefer to have a page/section somewhere that would list interesting extensions.

fabpot:
Closing for now.

PR #4703, opened 2025-10-24
Add ext-intl requirement to composer.json

### LINKED ISSUE #4702: Twig intl-extra depends on intl PHP extension, but does not declare such dependency
`twig/intl-extra` depends on intl PHP extension. However, it is possible to install even on systems without that extension (`composer install` succeeds). In that case, its usage would end up with a quite mysterious error:

```
PHP Fatal error:  Uncaught Error: Class "IntlDateFormatter" not found
```

As https://www.php.net/manual/en/class.intldateformatter.php says, the `IntlDateFormatter` class is defined by the extension. 

`composer.json` supports declaring dependencies on PHP extensions as well (by depending on `ext-intl`). I propose we add that dependency, which would make `composer.json` complain directly when installing. 

Maybe the goal is to provide a fallback implementation _somewhere_ (https://github.com/symfony/polyfill-intl-icu/tree/1.x?tab=readme-ov-file seems to be providing it, although that package is deprecated), but it doesn't seem intended to fail on usage. Whether the expected behaviour is "fail on install" or "use a fallback", I'm not sure.

urbanecm:
intl-extra makes use of the `IntlDateFormatter` class, which is only available in the `intl` extension of PHP. However, this dependency is not noted in `composer.json`, which means installation succeeds, but usage fails with a rather mysterious error:

```
PHP Fatal error:  Uncaught Error: Class "IntlDateFormatter" not found
```

To avoid this misunderstanding, add an explicit dependency on `intl` to composer.json, to make composer complain that the dependency is not met during installation.

Fixes #4702.

### DIFF
--- a/extra/intl-extra/composer.json
+++ b/extra/intl-extra/composer.json
@@ -17,7 +17,8 @@
     "require": {
         "php": ">=8.1.0",
         "twig/twig": "^3.13|^4.0",
-        "symfony/intl": "^5.4|^6.4|^7.0"
+        "symfony/intl": "^5.4|^6.4|^7.0",
+        "ext-intl": "*"
     },
     "require-dev": {
         "symfony/phpunit-bridge": "^6.4|^7.0"


stof:
Technically, it supports using `ext-intl` or `symfony/polyfill-intl-icu` (if you care only about English formatting rules in your projects but you want to use reusable templates that support localized formatting)

This PR makes that second use case impossible.

nicolas-grekas:
👎 as explained by @stof and in the linked issue. This is not ideal, but merging would make the situation worse, not better, on an important aspect of the package.

PR #4709, opened 2025-11-04
Add caution note for random function usage

xelan:
See https://www.php.net/manual/en/function.array-rand.php and https://www.php.net/manual/en/function.mt-rand.php. Users should know that the Twig function is a convenience feature, but unusable e.g. for serious gaming/gambling apps due to the potential predictability and limited value range.


stof (approved):

xabbuh (approved):

fabpot (approved):

fabpot:
Thank you @xelan.

PR #4713, opened 2025-11-23
Fix cycle() with non-countable ArrayAccess+Traversable objects

yoeunes:
When using `cycle()` with an object that implements `\ArrayAccess` and `\Traversable` but is not `\Countable`, the function currently returns the object instance immediately after triggering the deprecation notice.

This prevents the value from being converted to an array, causing the cycle logic to fail or return the object itself.

This PR removes the early return to ensure `self::toArray()` is called, allowing these objects to be cycled correctly as expected.

**How to test**

```php
$seq = new class implements \ArrayAccess, \IteratorAggregate {
    public function offsetExists($offset): bool { return true; }
    public function offsetGet($offset): mixed { return 'val'; }
    public function offsetSet($offset, $value): void {}
    public function offsetUnset($offset): void {}
    public function getIterator(): \Traversable { yield 'odd'; yield 'even'; }
};

// Should return 'odd', currently returns the $seq object
$result = CoreExtension::cycle($seq, 0);
```

Related to #4241

### DIFF
--- a/src/Extension/CoreExtension.php
+++ b/src/Extension/CoreExtension.php
@@ -417,10 +417,8 @@ final class CoreExtension extends AbstractExtension
 
                 trigger_deprecation('twig/twig', '3.12', 'Passing a non-countable sequence of values to "%s()" is deprecated.', __METHOD__);
 
-                return $values;
+                $values = self::toArray($values, false);
             }
-
-            $values = self::toArray($values, false);
         }
 
         if (!$count = \count($values)) {
--- a/tests/Extension/CoreTest.php
+++ b/tests/Extension/CoreTest.php
@@ -21,6 +21,7 @@ namespace Twig\Tests\Extension;
  */
 
 use PHPUnit\Framework\TestCase;
+use Symfony\Bridge\PhpUnit\ExpectDeprecationTrait;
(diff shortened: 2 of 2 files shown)


src/Extension/CoreExtension.php
```diff
@@ -416,8 +416,6 @@ public static function cycle($values, $position): mixed
                 // throw new RuntimeError('The "cycle" function expects a countable sequence as first argument.');
 
                 trigger_deprecation('twig/twig', '3.12', 'Passing a non-countable sequence of values to "%s()" is deprecated.', __METHOD__);
-
-                return $values;
             }
 
             $values = self::toArray($values, false);
```
stof: shouldn't this line be inside the `!is_countable` block (in the place that had the broken return). AFAICT, the intent of the is_countable check was to allow using the ArrayAccess&Countable object without converting it.

yoeunes: Hi @stof, thanks for the feedback! The conversion to array is now only performed inside the `!is_countable` block as suggested.

fabpot (approved):

fabpot:
Thank you @yoeunes.

PR #4714, opened 2025-11-26
Add 'invoke' to filters index

webda2l:
Missing from https://github.com/twigphp/Twig/pull/4518 I think.


fabpot (approved):

fabpot:
Good catch, thanks @webda2l.

Issue #3260, opened 2020-02-07
[Feature] Optional chaining operator


hason:
Add support for the optional chaining operator `?.` that is available in:

- JavaScript: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining
- TypeScritp 3.7: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining
- Latte 2.6: https://blog.nette.org/en/new-in-latte-2-6-optional-chaining-and-custom-functions#toc-optional-chaining
- PHP 8: https://www.php.net/releases/8.0/en.php#nullsafe-operator

```twig
{# Before #}
{{ foo.bar.baz is defined ? foo.bar.baz }}

{# After #}
{{ foo?.bar?.baz }}
```

VincentLanglet:
it should even be the default behaviour.

stof:
-1 for being the default behavior. That would make it a lot harder to identify mistakes (typo in a property name) when you don't expect it to be optional.

VincentLanglet:
> -1 for being the default behavior. That would make it a lot harder to identify mistakes (typo in a property name) when you don't expect it to be optional.

In php, there is tools like phpstan returning error for
```
$foo->getBar()->getBaz()
```
if getBar can be null. But there is nothing like this for twig. So mistake is easier to make.

I find sad the fact that
```
{% if foo.bar.baz %}
...
{% endif %}
```
throw an exception and display an error 500 in production when foo.bar is null.

This should be user-first and not dev-first.

What about a strict mode, enable in dev ?

stof:
Well, first, it does *not* throw an exception by default in production, as `strict_variables` is opt-in and will make `.` return `null` for non-existent attributes.

If what you want is to have everything optional in prod and strict in dev, Twig already has that feature since the 1.0 release (and probably even before in some 0.x releases).

To me, the benefit of a `?.` operator would be that it could be used even in strict mode

VincentLanglet:
> Well, first, it does _not_ throw an exception by default in production, as `strict_variables` is opt-in and will make `.` return `null` for non-existent attributes.
> 
> If what you want is to have everything optional in prod and strict in dev, Twig already has that feature since the 1.0 release (and probably even before in some 0.x releases).
> 
> To be, the benefit of a `?.` operator would be that it could be used even in strict mode

Oh sorry, didn't know that. That's great. I agree with you then.

[...]

PR #4717, opened 2025-12-01
Add null-safe operator

felds:
Adds native support for Twig’s null-safe `?.` operator so templates can traverse objects and arrays without defensive plumbing: `{{ user?.address?.city }}` now renders the value when present and stays silent when any hop is `null`.

`?.` reuses the existing `GetAttrExpression` operator, so it should have otherwise the same precedence and behavior as `.`.

Usage cases can be found in the tests:
https://github.com/twigphp/Twig/blob/7d98d45f58df040f41a81f7f8d4cce6c42be781e/tests/ExpressionParserTest.php#L304-L343

Tests cover both the rendered output and the generated PHP.

Let me know if any tests are missing or if the implementation should be done any other way.

### DIFF
--- a/extra/intl-extra/Tests/Fixtures/script_names.test
+++ b/extra/intl-extra/Tests/Fixtures/script_names.test
@@ -2,15 +2,15 @@
 "script_names" function
 --TEMPLATE--
 {{ script_names('UNKNOWN')|length }}
-{{ script_names() is iterable }}
-{{ script_names('fr') is iterable }}
+{{ script_names()|length > 200 ? 'more than 200' : 'less than 200' }}
+{{ script_names('fr')|length > 200 ? 'more than 200' : 'less than 200' }}
 {{ script_names()['Marc'] }}
 {{ script_names('fr')['Marc'] }}
 --DATA--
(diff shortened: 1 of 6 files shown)


felds:
This should fix issue #3260

felds:
The failing tests are already present in [twigphp:3.x](https://github.com/twigphp/Twig/tree/3.x). I have no idea what they refer to, and I don't think they are related to the changes in this PR.

felds:
I've found the bug.

It was not in this repo, but it appeared when `symfony/intl` had some data updated.

I think testing for exact results from dependencies is not a good idea, since it would require updating the tests in this lib every time the depended package was updated, so I changed the test to expect at least N results instead of exactly N results.

willrowe:
This is a duplicate of #4623

felds:
@willrowe thanks for pointing that out. I wonder why that PR is stuck. It looks pretty good too...

johndodev:
up

fabpot (approved):

fabpot:
Thank you @felds.

PR #4718, opened 2025-12-02
Add .gitignore & .gitattributes to all .gitattributes

jmsche:


extra/cache-extra/.gitattributes
```diff
@@ -1,2 +1,4 @@
 /Tests export-ignore
+/.gitattributes export-ignore
+/.gitignore export-ignore
```
GromNaN: On Symfony packages, we use a glob pattern:
```suggestion
/.git* export-ignore
```

jmsche: I followed the same format used in the base .gitattributes file: https://github.com/twigphp/Twig/blob/8074ddcab8e0a82a19213c8f4e546359dd10e6ed/.gitattributes#L7-L8

Should I also update the main one?

GromNaN: I think the main file should be updated too.

jmsche: Done :)

fabpot (approved):

fabpot:
Thank you @jmsche.

PR #4724, opened 2025-12-14
Avoid ord deprecation notice in PHP 8.5

iquito:
Change the one usage of `ord()` in Twig where a multibyte character can trigger the new deprecation in PHP 8.5, which is in the `html_attr` escape functionality in EscaperRuntime, currently leading to:

```
ord(): Providing a string that is not one byte long is deprecated. Use ord($str[0]) instead"
```

This small change fixes it. There are no other usages of `ord()` in Twig which lead to this deprecation notice.

### DIFF
--- a/src/Runtime/EscaperRuntime.php
+++ b/src/Runtime/EscaperRuntime.php
@@ -272,7 +272,7 @@ final class EscaperRuntime implements RuntimeExtensionInterface
                      * @license   https://framework.zend.com/license/new-bsd New BSD License
                      */
                     $chr = $matches[0];
-                    $ord = \ord($chr);
+                    $ord = \ord($chr[0]);
 
                     /*
                     * The following replaces characters undefined in HTML with the


fabpot (approved):

fabpot:
Thank you @iquito.

iquito:
@fabpot I saw you released a new version, but only `twig/extra-bundle` is available as a new version on packagist. Maybe because some of the Github workflows didn't succeed?

stof:
@iquito for patch releases, the splitter creates tags only for subtree repos that have changed since the previous release. That's expected.

xabbuh:
@stof The issue is that this PR touched a file from the `twig/twig` package, the tag was also created (see https://github.com/twigphp/Twig/releases/tag/v3.22.2) but it is not listed on Packagist: https://packagist.org/packages/twig/twig

iquito:
It might also be a Packagist issue, I am not sure, but I thought it might be worth it to find out why there is no new release.

stof:
@fabpot could you check the github webhook setup ?

packagist.org was showing the last update date of twig/twig on November 30th until I clicked the "Update now" button (which made the 3.22.2 tag appear)

fabpot:
Fixed now.

PR #4726, opened 2025-12-22
Enhance enum.rst with dynamic case example

Kocal:
Added example for dynamic enum case display.

### DIFF
--- a/doc/functions/enum.rst
+++ b/doc/functions/enum.rst
@@ -12,6 +12,10 @@
     {# display one specific case of a backed enum #}
     {{ enum('App\\CardSuite').Clubs.value }} {# "clubs" #}
 
+    {# display one specific case of a backed enum, with a dynamic name #}
+    {% set case_name = 'Spades' %}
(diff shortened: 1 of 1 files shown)


doc/functions/enum.rst
```diff
@@ -12,6 +12,10 @@
     {# display one specific case of a backed enum #}
     {{ enum('App\\CardSuite').Clubs.value }} {# "clubs" #}
 
+    {# display one specific case of a backed enum, with a dynamic name #}
+    {% set case_name = 'spades' %}
```
yceruto:
```suggestion
    {% set case_name = 'Spades' %}
```
It should match the case symbol, right?

yceruto (changes requested):

doc/functions/enum.rst
Kocal: Oops, yes you're right

stof (approved):

yceruto (approved):

fabpot (approved):

fabpot:
Thank you @Kocal.

PR #4730, opened 2026-01-08
Fix intl-extra tests

Jean-Beru:
Fix CI since `Scripts::getNames()` returns 4 new values. It just tests if the value is an array to limit maintenance on future updates.


fabpot (approved):

fabpot:
Thank you @Jean-Beru.

PR #4733, opened 2026-01-12
Add support for XDebug path mapping

### LINKED ISSUE #4697: Add support for XDebug native path mapping for compiled templates
Xdebug 3.5.0-alpha is shipping support for native path mapping.

Twig templates are a perfect use case for this feature, which would allow developers to define breakpoints on the Twig files and have XDebug map those to compiled templates.

See https://xdebug.org/announcements/2025-10-07 for the annoucement and https://xdebug.org/funding/001-native-path-mapping for the project description (useful until the documentation is updated to include that new feature).

We should explore how Twig could integrate support for this feature, generating Xdebug path mapping files when compiling templates.
Given that the mapping files cannot be in the Twig cache to be discovered directly, this will likely need some kind of opt-in. And you might not want to generate such files on production servers where XDebug won't be used (especially given that the Twig use case involves mapping line ranges, which can generate big files)

fabpot:
Closes #4697

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,6 @@
-# 3.22.2 (2026-XX-XX)
+# 3.23.0 (2026-XX-XX)
 
- * n/a
+ * Add support for Xdebug path mapping
 
 # 3.22.2 (2025-12-14)
 
--- a/doc/recipes.rst
+++ b/doc/recipes.rst
@@ -553,4 +553,61 @@ safe to avoid any escaping. You can do so by wrapping your expression with a
 
     $safeExpr = new RawFilter(new YourSafeNode());
 
+.. _debugging-templates:
+
+Debugging Twig Templates with Xdebug
+------------------------------------
+
+Xdebug 3.5+ supports native path mapping, which allows setting breakpoints
+directly in ``.twig`` files and having Xdebug map them to the correct lines
+in the compiled PHP files.
+
+When ``debug`` is enabled and a filesystem-based cache is configured, Twig
+automatically generates Xdebug source map files in the ``.xdebug`` subdirectory
+of the cache directory::
+
+    $twig = new \Twig\Environment($loader, [
+        'debug' => true,
+        'cache' => '/path/to/cache',
+    ]);
+
+To disable source map generation while keeping debug enabled, set
+``xdebug_source_map`` to ``false``.
+
+Configure Xdebug in ``php.ini``::
+
+    xdebug.mode=debug
+    xdebug.start_with_request=yes
+    xdebug.path_mapping=1
+
+You can then set breakpoints in ``.twig`` files. When debugging, template
+variables are available in the ``$context`` array (e.g., ``$context['name']``).
+
+Twig automatically registers source map files with Xdebug.
+
+VSCode-based IDE Setup
+~~~~~~~~~~~~~~~~~~~~~~
+
+1. Enable breakpoints in Twig files by adding to ``.vscode/settings.json``:
+
+.. code-block:: json
+
+       {
+           "debug.allowBreakpointsEverywhere": true
+       }
+
+2. Create a ``.vscode/launch.json`` with a basic configuration in your project:
+
+.. code-block:: json
+
+       {
+           "version": "0.2.0",
+           "configurations": [{
+               "name": "Listen for Xdebug",
+               "type": "php",
+               "request": "launch",
+               "port": 9003
+           }]
+       }
+
 .. _callback: https://www.php.net/manual/en/function.is-callable.php
--- a/src/Cache/ChainCache.php
+++ b/src/Cache/ChainCache.php
@@ -19,7 +19,7 @@ namespace Twig\Cache;
  *
  * @author Quentin Devos <quentin@devos.pm>
  */
-final class ChainCache implements CacheInterface, RemovableCacheInterface
+final class ChainCache implements CacheInterface, DirectoryCacheInterface, RemovableCacheInterface
 {
     /**
      * @param iterable<CacheInterface> $caches The ordered list of caches used to store and fetch cached items
@@ -76,6 +76,18 @@ final class ChainCache implements CacheInterface, RemovableCacheInterface
                 $cache->remove($name, $cls);
             }
         }
+    }
+
+    public function getDirectories(): array
+    {
+        $directories = [];
+        foreach ($this->caches as $cache) {
+            if ($cache instanceof DirectoryCacheInterface) {
+                $directories[] = $cache->getDirectories();
+            }
+        }
+
+        return array_merge(...$directories);
     }
 
     /**
--- a/src/Cache/XdebugSourceMapCache.php
+++ b/src/Cache/XdebugSourceMapCache.php
@@ -1,0 +1,161 @@
+<?php
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Twig\Cache;
+
+/**
+ * Cache decorator that writes Xdebug source map files alongside compiled templates.
+ *
+ * Xdebug 3.5+ supports native path mapping via .map files in .xdebug directories.
+ * This allows setting breakpoints in Twig templates and having Xdebug map them
+ * to the correct lines in the compiled PHP files.
+ *
+ * @author Fabien Potencier <fabien@symfony.com>
+ *
+ * @internal
+ */
+final class XdebugSourceMapCache implements CacheInterface, RemovableCacheInterface
+{
+    private ?string $pendingTemplatePath = null;
+    private ?array $pendingDebugInfo = null;
+
+    public function __construct(
+        private DirectoryCacheInterface&CacheInterface $cache,
+    ) {
+        $this->registerExistingSourceMaps();
+    }
+
+    /**
+     * Registers all existing source map files with Xdebug.
+     *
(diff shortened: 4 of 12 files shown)


CHANGELOG
```diff
@@ -1,6 +1,6 @@
 # 3.22.2 (2026-XX-XX)
```
jdreesen:
```suggestion
# 3.23.0 (2026-XX-XX)
```

doc/recipes.rst
```diff
@@ -553,4 +553,60 @@ safe to avoid any escaping. You can do so by wrapping your expression with a
...
+    xdebug.mode=debug
+    xdebug.start_with_request=yes
+    xdebug.path_mapping=1
+
+You can then set breakpoints in ``.twig`` files. When debugging, template
+variables are available in the ``$context`` array (e.g., ``$context['name']``).
+
+.. note::
+
+    XDebug looks for the ``.xdebug`` directory in the grandparent, parent, or
```
stof: The Twig cache will be none of those 3 locations in 99.9999% of projects, as you generally don't configure the Twig cache as `%kernel.project_dir%` (or the equivalent location in other frameworks than Symfony).

doc/recipes.rst
```diff
@@ -553,4 +553,60 @@ safe to avoid any escaping. You can do so by wrapping your expression with a
...
+    same directory as your entry PHP script. Ensure the cache directory is
+    located where XDebug can discover it.
+
+**VSCode-based IDE Setup**
```
stof: instead of a bold sentence, this should probably be a sub-heading instead.

doc/recipes.rst
```diff
@@ -553,4 +553,60 @@ safe to avoid any escaping. You can do so by wrapping your expression with a
...
+
+1. Enable breakpoints in Twig files by adding to ``.vscode/settings.json``::
```
stof: this should not use `::` but an explicit `..code-block: json` directive so that the syntax highlighting uses JSON.
The `::` shortcut is not about auto-detecting the language but about using the default language (which is likely to be `php` when building the Twig website)

doc/recipes.rst
```diff
@@ -553,4 +553,60 @@ safe to avoid any escaping. You can do so by wrapping your expression with a
...
+
+       {
+           "debug.allowBreakpointsEverywhere": true
+       }
+
+2. Create a ``.vscode/launch.json`` with a basic configuration in your project::
```
stof: same here

src/Cache/ChainCache.php
```diff
@@ -78,6 +78,18 @@ public function remove(string $name, string $cls): void
...
+    public function getDirectories(): array
+    {
+        $directories = [];
+        foreach ($this->caches as $cache) {
+            if ($cache instanceof DirectoryCacheInterface) {
+                $directories = array_merge($directories, $cache->getDirectories());
+            }
+        }
+
+        return $directories;
```
Kocal: The following code is more performant:
```suggestion
        foreach ($this->caches as $cache) {
            if ($cache instanceof DirectoryCacheInterface) {
                $directories []= $cache->getDirectories();
            }
        }

        return array_merge(...$directories);
```

doc/recipes.rst
fabpot: Fixed

src/Cache/XdebugSourceMapCache.php
```diff
@@ -0,0 +1,161 @@
...
+    private function registerExistingSourceMaps(): void
+    {
+        foreach ($this->cache->getDirectories() as $directory) {
+            $mapPath = $directory.'/.xdebug';
+            if (!is_dir($mapPath)) {
+                continue;
+            }
+
+            foreach (glob($mapPath.'/*.map') as $mapFile) {
+                xdebug_set_source_map($mapFile);
```
fabpot, line 51: This function is not available yet but this is the cleaneast way to do it IMHO.

fabpot:
@derickr I would love to get your feedback on how I've integrated the source map feature with Twig.

src/Cache/XdebugSourceMapCache.php
```diff
@@ -0,0 +1,161 @@
...
+            $mapFile = $directory.'/.xdebug/'.$filename;
+            if (is_file($mapFile)) {
+                @unlink($mapFile);
+            }
+        }
+    }
+
+    private function buildMapContent(string $cacheKey, string $templatePath, array $debugInfo): string
+    {
+        $lines = "# XDebug source map for Twig template\n";
```
jdreesen: Xdebug is spelled “Xdebug” with a lowercase “d”.

```suggestion
        $lines = "# Xdebug source map for Twig template\n";
```

src/Cache/ChainCache.php
```diff
@@ -78,6 +78,18 @@ public function remove(string $name, string $cls): void
         }
     }
 
+    public function getDirectories(): array
+    {
+        $directories = [];
+        foreach ($this->caches as $cache) {
+            if ($cache instanceof DirectoryCacheInterface) {
+                $directories []= $cache->getDirectories();
```
stof:
```suggestion
                $directories[] = $cache->getDirectories();
```

src/Cache/XdebugSourceMapCache.php
stof: is there any public discussion on the Xdebug side about this function ?

Btw, the verb `set` in the name is confusing if it registers a new source map without removing the previous one, as that's not the usual meaning of `set` in function names.

derickr: As this is a feature that I haven't added yet, I am more than happy to discuss how it should work. I agree that using `xdebug_add_source_map` is probably a better choice. Right now, the internals will reject a new source map if it touches the same files though, and that seems not so easy to change.

Alternatively, I could add a new function `xdebug_add_source_map_directory`, which you would only have to call once before the compiled template files are loaded into PHP's memory through include.

Then Xdebug can (additionally) scan this directory to pick up any source maps that exist.

src/Cache/XdebugSourceMapCache.php
```diff
@@ -0,0 +1,161 @@
...
+            // Use a reasonable max for last range (Xdebug can't handle PHP_INT_MAX)
+            $endLine = isset($compiledLines[$i + 1]) ? $compiledLines[$i + 1] - 1 : 999999;
+            $lines .= \sprintf(
+                "%s:%d-%d = %s:%d\n",
+                basename($cacheKey),
+                $startLine,
+                $endLine,
+                basename($templatePath),
+                $debugInfo[$startLine],
+            );
```
derickr, line 156: I imagine, this gets large, as there is a separate entry in the source file for each line. 

I am planning on optimising this, using how JavaScript source maps approach the line-to-line mappings, if possible. 

I would also recommend that if `$startLine` and `$endline` are the same, you should not include the `$endLine` in the mapping. It saves some processing on the Xdebug side (runs many times), where doing the test here needs to be done once per map-write.

src/Cache/XdebugSourceMapCache.php
```diff
@@ -0,0 +1,161 @@
...
+        $lines = "# Xdebug source map for Twig template\n";
+        $lines .= \sprintf("remote_prefix: %s/\n", \dirname($cacheKey));
+        $lines .= \sprintf("local_prefix: %s/\n\n", \dirname($templatePath));
+
+        $compiledLines = array_keys($debugInfo);
+        for ($i = 0, $count = \count($compiledLines); $i < $count; ++$i) {
+            $startLine = $compiledLines[$i];
+            // End line is exclusive, so use next start line minus 1
+            // Use a reasonable max for last range (Xdebug can't handle PHP_INT_MAX)
+            $endLine = isset($compiledLines[$i + 1]) ? $compiledLines[$i + 1] - 1 : 999999;
```
derickr, line 148: I don't like it that you need to use a random largish constant (`999999`) here. Perhaps it would make sense if I add an `EOF` marker to the parser, so that you can do:

```
compiled_file.php:5-EOF = orig_file.twig:4
```

Would that be helpful?

derickr (commented):
Thanks for looking into this!

I left some comments and questions to ask how things can be improved.

I'm also still waiting on https://youtrack.jetbrains.com/issue/WI-81344/Cant-Set-Breakpoints-in-Template-Files-Even-Though-I-Added-The-File-Name-Pattern-to-PHP to make it into a release of PhpStorm (it's already fine with VS Code's Plugin)

src/Cache/XdebugSourceMapCache.php
stof: it would definitely be helpful here.

src/Cache/XdebugSourceMapCache.php
stof: `xdebug_add_source_map_directory` might be better for the Twig use case indeed (and might maybe have less overhead for requests _not_ using breakpoints by skipping the filesystem scanning, depending on when xdebug actually traverses those directories)

derickr: It currently scans the files (if path mapping is enabled) at the start of every request. Although for now, it only works with the step debugger, in the future I want to extend the support to maps to other features too: https://xdebug.org/funding/001-native-path-mapping#future-scope

doc/recipes.rst
derickr: It should be spelled `Xdebug`, and not `XDebug`

derickr:
I've created the following tickets for myself:

- https://bugs.xdebug.org/view.php?id=2398
- https://bugs.xdebug.org/view.php?id=2399
- https://bugs.xdebug.org/view.php?id=2400

PR #4734, opened 2026-01-13
Update extends.rst

MAHADEV-J:
Semantically correct HTML elements instead of `<div>`itis


fabpot (approved):

fabpot:
Thank you @MAHADEV-J.

PR #4735, opened 2026-01-14
Add the = assignment operator

fabpot:


src/ExpressionParser/Infix/AssignmentExpressionParser.php
```diff
@@ -0,0 +1,55 @@
...
+    public function parse(Parser $parser, AbstractExpression $left, Token $token): AbstractExpression
+    {
+        if (!$left instanceof ContextVariable) {
+            throw new SyntaxError(\sprintf('Cannot assign to "%s", only variables can be assigned.', $left::class), $token->getLine(), $parser->getStream()->getSourceContext());
+        }
+        $right = $parser->parseExpression(InfixAssociativity::Left === $this->getAssociativity() ? $this->getPrecedence() + 1 : $this->getPrecedence());
+        $right = match ($this->getName()) {
+            '=' => $right,
+            default => throw new \LogicException(\sprintf('Unknown operator: %s.', $this->getName())),
+        };
```
fabpot, line 46: This could be simplified, but I'd maybe add more assignment operators in another PR.

src/Node/Expression/Binary/SetBinary.php
```diff
@@ -0,0 +1,44 @@
...
+     * @param ContextVariable    $left
+     * @param AbstractExpression $right
+     */
+    public function __construct(Node $left, Node $right, int $lineno)
+    {
+        $name = $left->getAttribute('name');
+        if (!\is_string($name)) {
+            throw new \LogicException('The "name" attribute must be a string.');
+        }
+        $left = new AssignContextVariable($name, $left->getTemplateLine());
```
smnandre, line 35: Are you not bypassing the [AssignNameExpression](https://github.com/fabpot/Twig/blob/c68c1ceffbe8775f03f881b186b899cb7fea87c8/src/Node/Expression/AssignNameExpression.php#L27-L31) checks (ensuring "name" is different of  `"false"` `"true" ` ... )   Not about their role right now :)

fabpot: I don't get your comment, can you clarify?

smnandre: To be honest, not sure I still can right. This now seems perfectly clear..  #insomnia

PR #4737, opened 2026-01-15
Update u.rst to clarify truncate method's third argument behavior

AlwaysVictorious:
Current wording "to preserve whole words" suggests setting the argument to ``true`` preserves words. However, the parameter is ``$cut`` (defaults to ``true`` for cutting at exact length), and ``false`` is required to preserve whole words.


fabpot (approved):

fabpot:
Thank you @Always-Victorious.

PR #4743, opened 2026-01-20
Add `html_attr_relaxed` escaping strategy

mpdude:
This adds `html_attr_relaxed`, a relaxed variant of the `html_attr` escaping strategy. The difference is that `html_attr_relaxed` does not escape the `:`, `@`, `[` and `]` characters. These are used by some front-end frameworks in attribute names to wire special handling/value binding. See https://v2.vuejs.org/v2/guide/syntax.html#v-bind-Shorthand for an example.

The HTML 5 spec does not exclude all those characters from attribute names ([html.spec.whatwg.org/multipage/syntax.html#attributes-2](https://html.spec.whatwg.org/multipage/syntax.html#attributes-2)).

However, at least XML processors will treat the colon as the XML namespace separator.

HTML 5 allows XML only on SVG and MathML elements, and only for pre-defined namespace-prefixes ([developer.mozilla.org/en-US/docs/Web/API/Attr/localName#:~:text=That means that the local,different from the qualified name](https://developer.mozilla.org/en-US/docs/Web/API/Attr/localName#:~:text=That%20means%20that%20the%20local,different%20from%20the%20qualified%20name)). For other something: prefixes, these will simply be passed on as part of the local attribute name.

According to [engine.sygnal.com/research/html5-attribute-names](https://engine.sygnal.com/research/html5-attribute-names), all current browser implementations handle at least the colon fine, and the aforementioned Vue.js documentation suggests that this is also the case for @.

Note also that Symfony UX only conditionally escapes attribute names, and it has `:` and `@` in its safe list:
https://github.com/symfony/ux/blob/c9a3e66b8ac53e870097e8a828913e57204398e7/src/TwigComponent/src/ComponentAttributes.php#L82

Closes #3614.

### DIFF
--- a/doc/filters/escape.rst
+++ b/doc/filters/escape.rst
@@ -57,6 +57,16 @@ documents:
   also when used as the value of an HTML attribute **without quotes**
   (e.g. ``data-attribute={{ some_value }}``).
 
+* ``html_attr_relaxed``: like ``html_attr``, but **does not** escape the ``@``, ``:``,
+  ``[`` and ``]`` characters. You may want to use this in combination with front-end
+  frameworks that use attribute names like ``v-bind:href`` or ``@click``. But, be
+  aware that in some processing contexts like XML, characters like the colon ``:``
+  may have meaning like for XML namespace separation.
+
+.. versionadded:: 3.24
+
+    The ``html_attr_relaxed`` strategy has been added in 3.23.
+
 Note that doing contextual escaping in HTML documents is hard and choosing the
 right escaping strategy depends on a lot of factors. Please, read related
 documentation like `the OWASP prevention cheat sheet
--- a/src/NodeVisitor/SafeAnalysisNodeVisitor.php
+++ b/src/NodeVisitor/SafeAnalysisNodeVisitor.php
@@ -54,6 +54,11 @@ final class SafeAnalysisNodeVisitor implements NodeVisitorInterface
 
             if (\in_array('html_attr', $bucket['value'], true)) {
                 $bucket['value'][] = 'html';
+                $bucket['value'][] = 'html_attr_relaxed';
+            }
+
+            if (\in_array('html_attr_relaxed', $bucket['value'], true)) {
+                $bucket['value'][] = 'html';
             }
 
             return $bucket['value'];
--- a/tests/Runtime/EscaperRuntimeTest.php
+++ b/tests/Runtime/EscaperRuntimeTest.php
@@ -179,6 +179,13 @@ class EscaperRuntimeTest extends TestCase
         }
     }
 
+    public function testHtmlAttributeRelaxedEscapingConvertsSpecialChars()
+    {
+        foreach ($this->htmlAttrSpecialChars as $key => $value) {
+            $this->assertEquals($value, (new EscaperRuntime())->escape($key, 'html_attr_relaxed'), 'Failed to escape: '.$key);
+        }
+    }
+
     public function testJavascriptEscapingConvertsSpecialChars()
     {
         foreach ($this->jsSpecialChars as $key => $value) {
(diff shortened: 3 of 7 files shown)


src/NodeVisitor/SafeAnalysisNodeVisitor.php
```diff
@@ -52,7 +52,7 @@ public function getSafe(Node $node)
                 continue;
             }
 
-            if (\in_array('html_attr', $bucket['value'], true)) {
+            if (\in_array('html_attr', $bucket['value'], true) || \in_array('html_attr_relaxed', $bucket['value'], true)) {
```
stof: `html_attr` should be considered safe for both `html` and `html_attr_relaxed` (while `html_attr_relaxed` should be safe for `html`)

stof: and this behavior needs to have regression tests covering the cases related to `html_attr_relaxed` to ensure future changes don't break it.

stof (changes requested):

src/NodeVisitor/SafeAnalysisNodeVisitor.php
mpdude: TBH, I did not understand what this code was doing in the first place, so I will need to :eyes:

stof: What this code was doing is saying that if the expression is safe for `html_attr` (determined by the `in_array` check), it should also be marked as safe for `html` (by adding `html` in the value). This is because the `html_attr` escaping is a superset of the `html` escaping.

mpdude:
@fabpot WDYT, would you endorse such an additional strategy?

src/NodeVisitor/SafeAnalysisNodeVisitor.php
mpdude: Like so c2529e01?

tests/Runtime/EscaperRuntimeTest.php
```diff
@@ -330,6 +337,29 @@ public function testHtmlAttributeEscapingEscapesOwaspRecommendedRanges()
...
+                $literal = $this->codepointToUtf8($chr);
+                $this->assertEquals($literal, (new EscaperRuntime())->escape($literal, 'html_attr_relaxed'));
+            } else {
+                $literal = $this->codepointToUtf8($chr);
+                if (\in_array($literal, $immune)) {
+                    $this->assertEquals($literal, (new EscaperRuntime())->escape($literal, 'html_attr_relaxed'));
+                } else {
+                    $this->assertNotEquals(
+                        $literal,
+                        (new EscaperRuntime())->escape($literal, 'html_attr'),
```
fabpot: Typo?

```suggestion
                        (new EscaperRuntime())->escape($literal, 'html_attr_relaxed'),
```

tests/Runtime/EscaperRuntimeTest.php
```diff
@@ -330,6 +337,29 @@ public function testHtmlAttributeEscapingEscapesOwaspRecommendedRanges()
...
+            || $chr >= 0x41 && $chr <= 0x5A
+            || $chr >= 0x61 && $chr <= 0x7A) {
+                $literal = $this->codepointToUtf8($chr);
+                $this->assertEquals($literal, (new EscaperRuntime())->escape($literal, 'html_attr_relaxed'));
+            } else {
+                $literal = $this->codepointToUtf8($chr);
+                if (\in_array($literal, $immune)) {
+                    $this->assertEquals($literal, (new EscaperRuntime())->escape($literal, 'html_attr_relaxed'));
+                } else {
+                    $this->assertNotEquals(
```
fabpot: Should be on one line.

tests/Runtime/EscaperRuntimeTest.php
```diff
@@ -179,6 +179,13 @@ public function testHtmlAttributeEscapingConvertsSpecialChars()
         }
     }
 
+    public function testHtmlAttributeRelaxedEscapingConvertsSpecialChars()
+    {
+        foreach ($this->htmlAttrSpecialChars as $key => $value) {
```
fabpot, line 184: Shouldn't we test :, @, [, or ] as well here (they are not part of `$this->htmlAttrSpecialChars`)?

doc/filters/escape.rst
```diff
@@ -57,6 +57,16 @@ documents:
   also when used as the value of an HTML attribute **without quotes**
   (e.g. ``data-attribute={{ some_value }}``).
 
+* ``html_attr_relaxed``: like ``html_attr``, but **does not** escape the ``@``, ``:``,
+  ``[`` and ``]`` characters. You may want to use this in combination with front-end
+  frameworks that use attribute names like ``v-bind:href`` or ``@click``. But, be
+  aware that in some processing contexts like XML, characters like the colon ``:``
+  may have meaning like for XML namespace separation.
+
+.. versionadded:: 3.23
```
fabpot:
```suggestion
.. versionadded:: 3.24
```

Can you also add an entry in CHANGELOG? Something like: Add html_attr_relaxed escaping strategy that preserves :, @, [, and ] for front-end framework attribute names

fabpot (changes requested):

tests/Runtime/EscaperRuntimeTest.php
mpdude: Short: I don't think we need to.

I don't fully understand why the tests are written and organized the way they are. It seems that has been inherited from old Zend Framework Escaper tests.

`testHtmlAttributeRelaxedEscapingConvertsSpecialChars` and `testHtmlAttributeEscapingConvertsSpecialChars` use `$htmlAttrSpecialChars` to test for a few samples that
* alnums are not escaped
* a few "immune" chars (common to both `html_attr` and `html_attr_relaxed`) are not escaped
* two examples beyond ASCII 0xFF are escaped
* other examples like `<>&"` are being escaped.

In addition to that, we have `testHtmlAttributeEscapingEscapesOwaspRecommendedRanges` and `testHtmlAttributeRelaxedEscapingEscapesOwaspRecommendedRanges`. Those tests will fully iterate the ASCII 0x01 to 0xFF range and test every single character from it. The test expects escaping to happen for all cases except the `0-9`, `a-z`, `A-Z` ranges and explicitly given whitelists of characters.

So, I think we're safe as-is.

mpdude:
Feedback addressed. Thank you!

When this gets merged, I'd like to use it as the escaping strategy for attribute names in #3930.

fabpot (approved):

fabpot:
Thank you @mpdude.

Issue #3399, opened 2020-09-23
Add syntax for destructuring assignment


hason:
Add destructuring assignment syntax that makes it possible to unpack values from arrays, or properties from objects, into distinct variables (as same is in Javascript):

A twig template
```twig
{% set [a, b] = [1, 2] %} 
{% for {id, firstName, ...rest} in data %}
{% endfor %}
```

should be converted to

```php
[$a, $b] = [1, 2];
foreach ($data as ['id' => $id, 'firstName' => $firstName]) {
    // TODO support for ...rest
}
```

Anubarak:
Is there any plan to include this is the near future?
I would really like to use it

fabpot:
This has been implemented.

mbenoukaiss:
I believe this issue was mistakenly closed with the merge of #4300 as the feature requested in this issue is different (destructuring assignment instead of unpacking)

RafaelKr:
@fabpot as @mbenoukaiss correctly pointed out the feature requested in this issue wasn't implemented with #4300 

This issue is about the concept of "Array Destructuring", see https://www.php.net/manual/en/language.types.array.php#language.types.array.syntax.destructuring

I've taken the examples from the PHP documentation and "translated" them to show how they could look in twig. You can always click on `Example X` in the heading to jump to the corresponding PHP example code.

### Indexed Arrays

#### [Example 11](https://www.php.net/manual/en/language.types.array.php#example-82): Array Destructuring
```twig
{% set source_array = ['foo', 'bar', 'baz'] %}

{% set [foo, bar, baz] = $source_array %}

{{ foo }}    {# prints "foo" #}
{{ bar }}    {# prints "bar" #}
{{ baz }}    {# prints "baz" #}
```

#### [Example 12](https://www.php.net/manual/en/language.types.array.php#example-83): Array Destructuring in Foreach
```twig
# Example #12
{% set source_array = [
    [1, 'John'],
    [2, 'Jane'],
] %}

{% for [id, name] in source_array %}
  {{ id }}: {{ name }}
{% endfor %}
```

#### [Example 13](https://www.php.net/manual/en/language.types.array.php#example-84): Ignoring Elements
```twig
{% set source_array = ['foo', 'bar', 'baz'] %}

{% set [, , baz] = $source_array %}

{{ baz }}    {# prints "baz" #}
```

#### [Example 15](https://www.php.net/manual/en/language.types.array.php#example-86): Swapping Two Variables
[...]

PR #4744, opened 2026-01-21
Add support for object and mapping destructuring

fabpot:
Refs https://github.com/twigphp/Twig/issues/3399

### DIFF
--- a/doc/templates.rst
+++ b/doc/templates.rst
@@ -1147,6 +1147,25 @@ You can skip values by leaving a slot empty:
     {# only assign the second value #}
     {% do [, last] = ['Fabien', 'Potencier'] %}
 
+Object Destructuring
+~~~~~~~~~~~~~~~~~~~~
+
+Use curly braces on the left side of an assignment to destructure an object
+or mapping by extracting values based on property/key names:
+
+.. code-block:: twig
+
+    {% do {name, email} = user %}
+
+    {{ name }}  {# user.name #}
+    {{ email }} {# user.email #}
+
+.. note::
+
+    Object destructuring uses the :ref:`dot operator <dot_operator>` to access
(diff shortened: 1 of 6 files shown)


doc/templates.rst
```diff
@@ -1147,6 +1147,25 @@ You can skip values by leaving a slot empty:
...
 
+Object Destructuring
+~~~~~~~~~~~~~~~~~~~~
+
+Use curly braces on the left side of an assignment to destructure an object
+or mapping by extracting values based on property/key names:
+
+.. code-block:: twig
+
+    {% do {name, email} = user %}
```
stof, line 1158: why doing that with `{% do %}` rather than with `{%set %}` to make it clear that we are doing assignments of variables ?

fabpot: My reasonning is that it's just an expression like any other one (it can be "embed" in a another expression).
I was thinking about maybe deprecating the `set` tag and add a `capture` tag for when you want to capture a chunk of a template.

nicolas-grekas: the cost of the deprecation might too high and thus not worth it ;)

stof: Oh, I missed the fact that assignment expressions are now a thing in the dev version of Twig.

I actually liked the fact that arbitrary expressions don't change the context (it also makes it simpler to think about scopes, especially in combination with expressions used in custom tags, as the affected scope there might depend on the internal implementation of the tag regarding its usage of the expression).
Having `{% set %}` to assign variables gives a much more explicit intent when reading the code than using `{% do %}` for everything (not counting the fact that it is too early to deprecate `{% set %}` given the huge impact on the ecosystem and the fact that assignment expressions are fairly new and so would require dropping support for many Twig versions in the ecosystem to perform the migration)

PR #4749, opened 2026-01-25
Fix null-safe operator test

HypeMC:
Noticed this while working on #4748.

Without `strict_variables`, the tests always pass, even when the null-safe operator is not used.


stof (approved):

fabpot (approved):

fabpot:
Thank you @HypeMC.

PR #4762, opened 2026-02-07
Fix CS

fabpot:


.php-cs-fixer.dist.php
```diff
@@ -11,19 +15,11 @@
         'no_unreachable_default_argument_value' => false,
         'braces' => ['allow_single_line_closure' => true],
         'heredoc_to_nowdoc' => false,
+        'single_line_throw' => false,
         'ordered_imports' => true,
         'phpdoc_types_order' => ['null_adjustment' => 'always_last', 'sort_algorithm' => 'none'],
-        'header_comment' => [
```
fabpot, line 16: Removed as we have 2 different headers depending on the files.

PR #4582, opened 2025-02-14
re-add mixed return type

xabbuh:
see https://github.com/twigphp/Twig/pull/4581#discussion_r1956080587 and https://github.com/symfony/symfony/actions/runs/13329105184/job/37228898654?pr=59776#step:8:2832

PR #4764, opened 2026-02-07
re-add mixed return type

xabbuh:
re-doing #4582, see https://github.com/symfony/symfony/actions/runs/21779345389/job/62840897500#step:8:3635

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -18,6 +18,7 @@ return (new Config())
         'single_line_throw' => false,
         'ordered_imports' => true,
         'phpdoc_types_order' => ['null_adjustment' => 'always_last', 'sort_algorithm' => 'none'],
+        'no_superfluous_phpdoc_tags' => ['allow_mixed' => true],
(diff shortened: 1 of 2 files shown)


fabpot:
Isn't it somehing that should be fixed in Symfony instead?

xabbuh:
I think the CS fixer is wrong here. Not having a docblock from what I understand is equivalent to having `void` as the return type not `mixed`.

fabpot:
> I think the CS fixer is wrong here. Not having a docblock from what I understand is equivalent to having `void` as the return type not `mixed`.

I would need to double-check, but I'm pretty sure that `mixed` is the default.

fabpot:
Apparently, we can do `'no_superfluous_phpdoc_tags' => ['allow_mixed' => true]`

xabbuh:
> Apparently, we can do `'no_superfluous_phpdoc_tags' => ['allow_mixed' => true]`

:+1: added

fabpot:
Thank you @xabbuh.

PR #4766, opened 2026-02-18
Add is valid json Twig test

qdequippe:
Add is valid json Twig test using [json_validate](https://www.php.net/manual/en/function.json-validate.php) PHP function


stof:
What is the use case for having this in Twig templates ?

And this looks like a good candidate for a Twig test instead, exposing an API like `value is valid json` instead of `json_validate(value)`

qdequippe:
> What is the use case for having this in Twig templates ?

My use case is when I want to display http response content, if content is a valid json a CSS class is applied
 
> And this looks like a good candidate for a Twig test instead, exposing an API like `value is valid json` instead of `json_validate(value)`

Good point Twig test is a better choice for this, I will update the PR

stof:
I'm not sure this use case is common enough to ship that in Twig itself. I will let @fabpot give his opinion on this too.

However, this does not prevent registering a Twig test in your own project needing this.

qdequippe:
> I'm not sure this use case is common enough to ship that in Twig itself. I will let @fabpot give his opinion on this too.

Of course it's just a proposal :)

src/Extension/CoreExtension.php
```diff
@@ -2175,6 +2175,27 @@ public static function parseAttributeFunction(Parser $parser, Node $fakeNode, $a
...
+     *        {# ... #}
+     *    {% endif %}
+     *
+     * @param mixed $value A variable
+     *
+     * @internal
+     */
+    public static function testValidJson($value): bool
+    {
+        if (!$value instanceof \Stringable) {
```
stof: this won't accept strings

stof (approved):

src/Extension/CoreExtension.php
```diff
@@ -2174,6 +2175,27 @@ public static function parseAttributeFunction(Parser $parser, Node $fakeNode, $a
...
+     *        {# ... #}
+     *    {% endif %}
+     *
+     * @param mixed $value A variable
+     *
+     * @internal
+     */
+    public static function testValidJson($value): bool
+    {
+        if (false === is_string($value) && !$value instanceof \Stringable) {
```
Kocal:
```suggestion
        if (!\is_string($value) && !$value instanceof \Stringable) {
```

fabpot:
I don't think it should be part of core.

PR #4770, opened 2026-02-23
Add support for `docs` property for `BlockNode` and `TypesNode`

### LINKED ISSUE #4768: Standardize Twig comment annotations for documenting template variables and blocks
```twig
```

## Why standardize?

```twig
{# _card.html.twig #}

```

WDYT? Thanks!

Kocal:
Close #4768

(Code generated almost entirely using Claude Opus 4.5.)

The following code is now valid (essential parts of our UX Toolkit Shadcn Button component):
```twig
{%- types {
    variant: "'default'|'secondary'|'destructive'|'outline'|'ghost'|'link'" docs="Visual style variant.",
    size: "'default'|'xs'|'sm'|'lg'|'icon'|'icon-xs'|'icon-sm'|'icon-lg'" docs="The button size.",
    button: '"button"' docs="The HTML tag to render.",
} -%}

{%- block content docs="The button label and/or icon" %}{% endblock -%}
```

### DIFF
--- a/src/Node/BlockNode.php
+++ b/src/Node/BlockNode.php
@@ -33,6 +33,7 @@ class BlockNode extends Node
         $compiler
             ->addDebugInfo($this)
             ->write("/**\n")
+            ->write($this->hasAttribute('docs') ? ' * '.str_replace("\n", "\n * ", $this->getAttribute('docs'))."\n" : '')
             ->write(" * @return iterable<null|scalar|\Stringable>\n")
             ->write(" */\n")
             ->write(\sprintf("public function block_%s(array \$context, array \$blocks = []): iterable\n", $this->getAttribute('name')), "{\n")
--- a/src/TokenParser/BlockTokenParser.php
+++ b/src/TokenParser/BlockTokenParser.php
@@ -42,6 +42,14 @@ final class BlockTokenParser extends AbstractTokenParser
         $this->parser->pushLocalScope();
         $this->parser->pushBlockStack($name);
 
+        // Check for optional docs="..." attribute
+        $docs = null;
+        if ($stream->test(Token::NAME_TYPE, 'docs')) {
+            $stream->next();
+            $stream->expect(Token::OPERATOR_TYPE, '=');
+            $docs = $stream->expect(Token::STRING_TYPE)->getValue();
+        }
+
         if ($stream->nextIf(Token::BLOCK_END_TYPE)) {
             $body = $this->parser->subparse([$this, 'decideBlockEnd'], true);
             if ($token = $stream->nextIf(Token::NAME_TYPE)) {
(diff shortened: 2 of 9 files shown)


src/Node/BlockNode.php
```diff
@@ -33,6 +33,7 @@ public function compile(Compiler $compiler): void
         $compiler
             ->addDebugInfo($this)
             ->write("/**\n")
+            ->write($this->hasAttribute('docs') ? ' * '.str_replace("\n", "\n * ", $this->getAttribute('docs'))."\n" : '')
```
Kocal, line 36: I think it could be a nice addition, it costs nothing:

Image

stof, line 36: This need to handle cases where the content of the attribute contains `*/`, to make sure it does not ends the PHP comment.

src/TokenParser/BlockTokenParser.php
```diff
@@ -42,6 +42,14 @@ public function parse(Token $token): Node
         $this->parser->pushLocalScope();
         $this->parser->pushBlockStack($name);
 
+        // Check for optional docs="..." attribute
+        $docs = null;
+        if ($stream->test(Token::NAME_TYPE, 'docs')) {
```
stof: this is breaking existing code if the block is using an expression starting with a `docs` variable or function.

I think it needs to be parsed *after* the expression, not before it, for the case using the expression-based content.

stof (commented):
With the introduction of the assignment operator, I think this proposal is ambiguous for blocks as `{% block content docs="Hello world!" %}` is a valid code already, using ` docs="Hello world!"` as the expression (even though assigning a `docs` variable in that scope is quite useless, as the block is the scope boundary)

willrowe:
I think `desc` or `description` would be more clear than `docs`.

Kocal:
>  I think this proposal is ambiguous for blocks as `{% block content docs="Hello world!" %}` is a valid code already

Are you sure? Because using `{%- block content docs="The button label and/or icon" %}{% endblock -%}` on 3.x gives me `Unknown "endblock" tag in components/Button.html.twig at line 35.`: 
image

Or did I miss something?

stof:
@Kocal the `{% endblock %}` tag is forbidden when using the shortcut in the start tag. But the switch between both variants is done purely based on the start tag (we need to know the variant *before* trying to parse the tag body as only one variant has it).
What is valid on 3.x is `{%- block content docs="The button label and/or icon" %}`

Kocal:
Ah I see, I didn't think about this case. Thank you!

I've pushed a new commit that look for `=`token , a string token, and then `endblock` token.

willrowe:
I want to reiterate that I do not think that `docs` is a good name for this.

fabpot:
Why not use the quivalent of inline comments (see #4349)?

Kocal:
This is what my first suggestion at https://github.com/twigphp/Twig/issues/4768#issue-3970582201 was about, should I go back to it?

Kocal:
I'm closing this PR, I believe things can stay in UX Toolkit with a better format for `@prop`/`@block` comments, and in https://github.com/Haehnchen/idea-php-symfony2-plugin/pull/2545 too.

**EDIT:** moved to https://github.com/symfony/ux/pull/3694

PR #4773, opened 2026-02-23
Move to PHPStan level 5

fabpot:
Not a big fan adding more phpdocs, but I don't have a better idea for now (except stop use `getNode()` and add proper getters).

### DIFF
--- a/src/Util/CallableArgumentsExtractor.php
+++ b/src/Util/CallableArgumentsExtractor.php
@@ -12,6 +12,7 @@
 namespace Twig\Util;
 
 use Twig\Error\SyntaxError;
+use Twig\Node\Expression\AbstractExpression;
 use Twig\Node\Expression\ArrayExpression;
 use Twig\Node\Expression\ConstantExpression;
 use Twig\Node\Expression\VariadicExpression;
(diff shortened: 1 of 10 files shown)


src/Util/CallableArgumentsExtractor.php
```diff
@@ -35,15 +36,16 @@ public function __construct(
...
      */
     public function extractArguments(Node $arguments): array
     {
-        /** @var array<int|string, Node> $extractedArguments */
+        /** @var array<int|string, AbstractExpression> $extractedArguments */
         $extractedArguments = [];
         $extractedArgumentNameMap = [];
         $named = false;
         foreach ($arguments as $name => $node) {
+            /** @var AbstractExpression $node */
```
nicolas-grekas, line 48: should be replaced by `@param` I suppose

nicolas-grekas: `@param Node<AbstractExpression>` + ensure the Node class is a template?

stof: making Node generic would be hard, because we mix 2 different kind of usages of the Node classes:

- nodes taking a list of child nodes (with a consistent type, even though that type might just be `Node` when it is not `AbstractExpression`)
- nodes storing named child nodes, where the node storage is used as an array shape.

PR #4774, opened 2026-02-23
Ensure filters/attributes aren't mistaken for operators

### LINKED ISSUE #4767: `literal` filter/operator conflict
Craft CMS defines a [`literal` filter](https://craftcms.com/docs/5.x/reference/twig/filters.html#literal), which conflicts with the new `literal` operator introduced in Twig 3.21 via #4543.

For example, the following template will trigger the error: `Unexpected token "operator" of value "literal" ("name" expected)`.

```twig
{{ 'foo' | literal }}
```

`Lexer::getOperatorRegex()` is defensive against filters being confused for operators here:

https://github.com/twigphp/Twig/blob/a64dc5d2cc7d6cafb9347f6cd802d0d06d0351c9/src/Lexer.php#L545-L548

However the lookbehind doesn’t account for filters that have spaces around them, which is pretty commonplace.

### Suggested fix

Changing that lookbehind code to the following fixes the problem:

```php
$r = '(?<![\.\|]\s|.[\.\|])'.$r;
```

(It’s not possible to have variable-length lookbehinds, so I believe this is the only way to do it, with that `.` and all, rather than just `(?<![\.\|]\s?)`.)

brandonkelly:
Updates the regex in `Lexer::getOperatorRegex()` to account for filters/attributes that have a space between their `|`/`.` operator and the filter/attribute name, to ensure they aren’t mistaken for operators.

A test is included that checks the following template.

```twig
{{ 'foo'|and }}
{{ 'bar' | and }}
{{ foo.and }}
{{ bar . and }}
{{ foo and bar }}
```

(Only the `and` in the last tag should be considered an operator.)

Fixes #4767

### DIFF
--- a/src/Lexer.php
+++ b/src/Lexer.php
@@ -544,7 +544,7 @@ class Lexer
 
             // an operator that begins with a character must not have a dot or pipe before
             if (ctype_alpha($expressionParser[0])) {
-                $r = '(?<![\.\|])'.$r;
+                $r = '(?<![\.\|]\s|.[\.\|])'.$r;
             }
 
             // an operator with a space can be any amount of whitespaces
--- a/tests/LexerTest.php
+++ b/tests/LexerTest.php
@@ -420,6 +420,43 @@ class LexerTest extends TestCase
         $this->addToAssertionCount(1);
     }
 
+    public function testFilterAndAttributeNamedAfterOperator()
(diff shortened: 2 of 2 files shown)


fabpot:
Thank you @brandonkelly.

PR #4775, opened 2026-02-23
Add getOperatorTokens() to ExpressionParserInterface to separate operator token registration from parser identity

### LINKED ISSUE #4767: `literal` filter/operator conflict

fabpot:
Closes #4767
Closes #4774


src/ExpressionParser/ExpressionParserInterface.php
```diff
@@ -11,6 +11,14 @@
 
 namespace Twig\ExpressionParser;
 
+/**
+ * @method array<string> getOperatorTokens() Returns the operator token strings that this expression parser handles.
```
stof: I would use `list<string>` to define a more precise type.

stof:
@fabpot I'm not sure this replaces #4774 (the test being added uses `and` as reproducer, which will still be an operator token)

src/ExpressionParser/ExpressionParserInterface.php
fabpot: Changed

stof (approved):

PR #4784, opened 2026-03-17
Add two tests for error conditions in #3930

mpdude:
Here are two test cases for potential errors in `html_attr_merge`, asked for by @fabpot in https://github.com/twigphp/Twig/pull/3930#pullrequestreview-3956645279.


fabpot:
Thank you @mpdude.

PR #4787, opened 2026-03-19
Reword parent-child analogy in `doc/tags/extends.rst`

llupa:


llupa:
This might be a language / cultural thing. But the analogy about filling holes, was ... weird?

doc/tags/extends.rst
```diff
@@ -71,10 +71,10 @@ value from the parent template is used instead.
...
 You can't define multiple ``block`` tags with the same name in the same
 template. This limitation exists because a block tag works in "both"
-directions. That is, a block tag doesn't just provide a hole to fill - it also
-defines the content that fills the hole in the *parent*. If there were two
-similarly-named ``block`` tags in a template, that template's parent wouldn't
-know which one of the blocks' content to use.
+directions. Think of each block as a named puzzle piece: it defines a slot
+in the *parent* layout **and** provides the content that snaps into that slot
+from the *child*. If there were two similarly-named ``block`` tags in a
+template, the *parent* wouldn't know which puzzle piece to use.
```
stof: I think this sentence would be more clear with the existing `which one of the blocks' content to use` than the puzzle piece analogy.

doc/tags/extends.rst
```diff
@@ -71,10 +71,10 @@ value from the parent template is used instead.
 
 You can't define multiple ``block`` tags with the same name in the same
 template. This limitation exists because a block tag works in "both"
-directions. That is, a block tag doesn't just provide a hole to fill - it also
-defines the content that fills the hole in the *parent*. If there were two
-similarly-named ``block`` tags in a template, that template's parent wouldn't
-know which one of the blocks' content to use.
+directions. Think of each block as a named puzzle piece: it defines a slot
```
stof: not sure we need this `puzzle piece` analogy.

Talking about `slot` instead of `hole` makes sense though. It will allow people to relate that to their knowledge of HTML shadow root which also uses that naming (it does not work exactly the same, but the idea of defining a place where content will be inserted is related)

llupa: I restored the sentence how it was and replaced `hole` with `slot`.

fabpot:
Thank you @llupa.

PR #4791, opened 2026-03-22
Allow `html_attr()` to return a spreadable object

### LINKED ISSUE #4790: Allow `html_attr()` to return a spreadable object for use with Twig Component attributes
Hi, I'm currently working on https://github.com/symfony/ux/issues/3386 and I'm facing an issue with the new `html_attr()` function (cc @mpdude)

## Context

When using `html_attr_merge() `with `html_attr_type('sst')` (from Twig 3.24), the merged result contains `SeparatedTokenList` objects as attribute values. This works perfectly when rendering onto native HTML elements:
```twig
{# Works fine — html_attr() stringifies everything #}
<button {{ html_attr(merged_attrs) }}>Click</button>
```

However, it is not possible to spread the merged attributes onto a Twig Component:
```twig
{# html_attr() returns a string: error #}
<twig:Button {{ html_attr(merged_attrs) }}>Click</twig:Button>

{# ...merged_attrs will try to use `Twig\Extra\Html\HtmlAttr\SeparatedTokenList` as a value: error #} 
<twig:Button {{ ...merged_attrs }}>Click</twig:Button>
```

> Symfony\UX\Toolkit\Tests\Functional\ComponentsRenderingTest::testComponentRendering with data set "Kit shadcn, component alert-dialog, code file Demo.html" ('shadcn', 'alert-dialog', '<twig:AlertDialog id="delete_...log>\n')
Twig\Error\RuntimeError: An exception has been thrown during the rendering of a template ("A "data-action" prop was passed when creating the component. No matching "data-action" property or mount() argument was found, so we attempted to use this as an HTML attribute. But, the value is not a scalar (it's a "Twig\Extra\Html\HtmlAttr\SeparatedTokenList"). Did you mean to pass this to your component or is there a typo on its name?") in "templates/components/Button.html.twig" at line 32

[ComponentAttributes](https://github.com/symfony/ux/blob/2.x/src/TwigComponent/src/ComponentAttributes.php) only accepts scalar values (strings, ints, bools, etc.). 

When `html_attr_merge()` produces `SeparatedTokenList` objects for attributes declared with `html_attr_type('sst')`, spreading them onto a component triggers the error:
```
(...) But, the value is not a scalar (it's a "Twig\Extra\Html\HtmlAttr\SeparatedTokenList")
```

## Real-world use case

In the Symfony UX Toolkit, components like `AlertDialog`, `Dialog`, and `Tooltip` expose trigger attributes (`data-action`, `data-*-target`, etc.) that need to be applied to another element (e.g.: a simple HTML element like `<button>`, or a Twig component like `<twig:Button>`). 

When composing multiple triggers on the same button (e.g., `AlertDialog` + `Tooltip`), users need to merge `data-action` values - which is exactly what `html_attr_merge` +` html_attr_type('sst') `was designed for, but the result can't be spread onto <twig:Button>.

See: https://github.com/symfony/ux/issues/3386

## Proposal

Modify `\Twig\Extra\Html\HtmlExtension::htmlAttr()` so that it returns an object that supports both string rendering and array-like access:
- `__toString()` — returns the HTML attribute string (current behavior: `class="foo bar" data-action="click->dialog#open")`
- `IteratorAggregate`: returns the attributes as a key-value array, with `SeparatedTokenList` values automatically cast to their string representation

This would allow:
```twig
{# On a native HTML element (already works today) #}
<button {{ html_attr(merged) }}>Click</button>

{# On a Twig Component (would work with this proposal) #}
<twig:Button {{ ...html_attr(merged) }}>Click</twig:Button>
```

> [!NOTE]
> This is exactly what we are doing with our [ComponentAttributes](https://github.com/symfony/ux/blob/2.x/src/TwigComponent/src/ComponentAttributes.php), we can use `<button{{ attributes }}>` and `<twig:Button {{ ...attributes }}>` and it works super fine.

Example
```twig
{% set alert_trigger = {
    'data-action': html_attr_type('sst', 'click->alert-dialog#open'),
    'data-alert-dialog-target': 'trigger',
} %}

{% set tooltip_trigger = {
    'data-action': html_attr_type('sst', 'mouseenter->tooltip#show mouseleave->tooltip#hide'),
    'data-tooltip-target': 'trigger',
} %}

{% set merged = html_attr_merge(alert_trigger, tooltip_trigger) %}

{# Today: only works on native elements #}
<button {{ html_attr(merged) }}>Click</button>

{# With this proposal: would also work on Twig Components #}
<twig:Button {{ ...html_attr(merged) }}>Click</twig:Button>

{# Spread would produce scalar values:
     data-action="click->alert-dialog#open mouseenter->tooltip#show mouseleave->tooltip#hide"
     data-alert-dialog-target="trigger"
     data-tooltip-target="trigger"
#}
```

## Current workaround

Without this feature, users must manually concatenate `data-action` strings using Twig's `|merge` filter:
```twig
<twig:Button {{ ...alert_trigger_attrs|merge(tooltip_trigger_attrs)|merge({
    'data-action': alert_trigger_attrs['data-action'] ~ ' ' ~ tooltip_trigger_attrs['data-action'],
}) }}>
    Click
</twig:Button>
```

Kocal:
Close https://github.com/twigphp/Twig/issues/4790

Generated by Claude Opus 4.6 and manully reviewed/tweaked.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,6 @@
-# 3.24.1 (2026-XX-XX)
+# 3.25.0 (2026-XX-XX)
 
- * n/a
+ * Make `html_attr()` return an `HtmlAttributes` object that is both `Stringable` and `IteratorAggregate`, allowing spreading onto Twig Components via `{{ ...html_attr(merged) }}`
 
 # 3.24.0 (2026-03-17)
 
--- a/extra/html-extra/HtmlAttributes.php
+++ b/extra/html-extra/HtmlAttributes.php
@@ -1,0 +1,136 @@
+<?php
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Twig\Extra\Html;
+
+use Twig\Error\RuntimeError;
+use Twig\Extra\Html\HtmlAttr\AttributeValueInterface;
+use Twig\Extra\Html\HtmlAttr\InlineStyle;
+use Twig\Extra\Html\HtmlAttr\SeparatedTokenList;
+use Twig\Runtime\EscaperRuntime;
+
+/**
+ * Represents a set of HTML attributes that can be rendered as a string
+ * or iterated over for spreading onto Twig Components.
+ *
+ * Usage:
+ *   {# On a native HTML element (renders as string) #}
+ *   <button {{ html_attr(merged) }}>Click</button>
+ *
+ *   {# On a Twig Component (spread as key-value pairs) #}
+ *   <twig:Button {{ ...html_attr(merged) }}>Click</twig:Button>
+ *
+ * @implements \IteratorAggregate<string, string>
+ */
+final class HtmlAttributes implements \Stringable, \IteratorAggregate, \Countable
+{
+    /**
+     * @param array<string, mixed> $attributes The raw merged attributes
+     */
+    public function __construct(
+        private readonly array $attributes,
+        private readonly EscaperRuntime $escaper,
+    ) {
+    }
+
+    public function __toString(): string
+    {
+        $result = '';
+
+        foreach ($this->resolveAttributes() as $name => $value) {
+            $result .= $this->escaper->escape($name, 'html_attr_relaxed').'="'.$this->escaper->escape($value).'" ';
+        }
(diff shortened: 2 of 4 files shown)


stof:
See my comments on the issue. I don't think think this is the right solution to the problem.

Kocal:
Replaced by https://github.com/symfony/ux/pull/3408

PR #4792, opened 2026-04-08
Fix date filter not passing timezone to DateTime constructor for string dates

### LINKED ISSUE #4746: The timezone argument is missing when instantiating from strings other than "now".
https://github.com/twigphp/Twig/blob/76c404ec67a6c8dcbb68e82b38a2dc6074c503de/src/Extension/CoreExtension.php#L618

Is there a reason not to pass the timezone as the second argument in that case (unlike previous cases) ? Timezone is set a few lines later but it's not wat we could expect 

Let see with an exemple : 

Right now it's 

- In Paris: 2026-01-22 it's 05:00 PM in Europe/Paris (+01:00)
- In London: 2026-01-22 04:00 PM UTC
- In Sydney: 2026-01-**23** 03:00 AM Australia/Sydney (+11:00) // **Notice it's day +1**

👉  php.ini timezone is set to UTC

Usage in Twig template 

```twig
{{ "today"|date('Y-m-d', 'Australia/Sydney') }}
```

Current version result :

```php
$today = new \DateTime("today"); // 2026-01-22 00:00:00.0 UTC (+00:00)
$today->setTimezone(new \DateTimeZone('Australia/Sydney')); // 2026-01-22 11:00:00.0 Australia/Sydney (+11:00)
echo $today->format('Y-m-d');
// output 2025-01-22 💥 
```

Expecting result

```php
$today = new \DateTime("today", new \DateTimeZone('Australia/Sydney')); // 2026-01-23 00:00:00.0 Australia/Sydney (+11:00)
echo $today->format('Y-m-d');
// output 2025-01-23 👍 
```

False positive case 

```twig
{{ "today"|date('Y-m-d', 'Europe/Paris') }}
```

```php
$today = new \DateTime("today"); // today set time to 00:00:00 (https://www.php.net/manual/en/datetime.formats.php)
$today->setTimezone(new \DateTimeZone('Europe/Paris'));
echo $today->format('Y-m-d');
// output 2025-01-22
```

Amoifr:
| Q             | A
| ------------- | ---
| Branch?       | 3.x
| Bug fix?      | yes
| New feature?  | no
| Deprecations? | no
| Issues        | Fix #4746
| License       | MIT

When using the `date` filter with a timezone and a relative date string like `"today"`, the timezone was not passed to the `DateTime` constructor. This caused `"today"|date('Y-m-d', 'Australia/Sydney')` to return the wrong date when the target timezone is ahead of the PHP default timezone.

The fix passes the timezone to the `DateTime` constructor for string dates, consistent with the existing behavior for `null`/`"now"` values.


stof:
This is changing the behavior for strings that are *not* relative times, as they are now parsed in the display timezone rather than in the PHP default timezone (see the change in existing tests that you had to do).

Amoifr:
Good point @stof! Yes, this changes behavior for absolute date strings too. I considered restricting the fix to relative dates only, but reliably detecting relative vs absolute strings in PHP is tricky.

That said, I think the new behavior is actually more consistent:

1. **The `null`/`"now"` path already passes timezone to the constructor** (line 611) without `setTimezone()` — so `null|date('H:i', 'Asia/Tokyo')` already interprets in the target timezone
2. **For `DateTime` objects with an explicit timezone**, `setTimezone()` still converts correctly since the embedded timezone takes precedence over the constructor argument
3. **For absolute strings without timezone info** (like `'2010-01-28 15:00:00'`), the string is inherently ambiguous — it could mean 15:00 in any timezone. Interpreting it in the target timezone feels more intuitive than interpreting it in whatever `date_default_timezone_set()` happens to be

The only case where `setTimezone()` conversion still matters is strings with embedded timezone like `'2010-01-28T15:00:00+04:00'`, and those work correctly because PHP ignores the constructor timezone when the string has its own.

But I understand this is a BC break for absolute dates. If you'd prefer to keep the current behavior for those, I can try to limit the scope — maybe by only passing the timezone to the constructor when it differs from `date_default_timezone_get()` and the string doesn't contain timezone indicators? Let me know your preference!

PR #4795, opened 2026-04-30
Lazy load `EscaperRuntime` in `EscaperExtension`

GromNaN:
This allows overriding `EscaperRuntime` via a custom runtime loader.

Previously, `EscaperExtension::setEnvironment()` was calling `$environment->getRuntime(EscaperRuntime::class)` eagerly. Since this method is called from `Environment::__construct()`, the runtime was resolved before any custom runtime loader could be injected, making it impossible to override `EscaperRuntime`.

- Required by https://github.com/symfony/symfony/pull/63929

### DIFF
--- a/src/Extension/EscaperExtension.php
+++ b/src/Extension/EscaperExtension.php
@@ -76,7 +76,7 @@ final class EscaperExtension extends AbstractExtension
         }
 
         $this->environment = $environment;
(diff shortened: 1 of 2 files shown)


fabpot (approved):

fabpot:
Thank you @GromNaN.

PR #4797, opened 2026-05-06
Make embeds deterministic

itsalmostchristmas:
Using `{% embed 'foo.twig' %}...{% endembed %}` makes compiled templates non-deterministic. This poses a problem for developers who wish to provide reproducible, pre-compiled builds of Twig templates.

The cause is `mt_rand()` used to generate template indices. The change I propose uses an incrementing counter instead. To keep tests passing, the number must be non-zero and unique per PHP process (reusing indices across `Twig\Parser` instances causes fatal errors).

A demonstration of the problem is available on the [Twig Playground](https://twig.symfony.com/play?data=eyJ0ZW1wbGF0ZXMiOltbImluZGV4LnR3aWciLCJ7JSBlbWJlZCAnZGlhbG9nLnR3aWcnICV9XG4gICAgeyUgYmxvY2sgbWVzc2FnZSAlfVxuICAgICAgICBUYXNrIGZhaWxlZCBzdWNjZXNzZnVsbHlcbiAgICB7JSBlbmRibG9jayAlfVxueyUgZW5kZW1iZWQgJX0iXSxbImRpYWxvZy50d2lnIiwiPGRpYWxvZz57JSBibG9jayBtZXNzYWdlICV9eyUgZW5kYmxvY2sgJX08L2RpYWxvZz4iXV0sImNvbnRleHQiOnt9LCJ2ZXJzaW9uIjoiMy4yMS4xIiwib3B0aW9ucyI6eyJzdHJpY3RfdmFyaWFibGVzIjp0cnVlLCJjaGFyc2V0IjoiVVRGLTgiLCJhdXRvZXNjYXBlIjoiIn19), where the compiled output of `index.twig` differs on every recompilation.

### DIFF
--- a/src/Parser.php
+++ b/src/Parser.php
@@ -54,6 +54,7 @@ class Parser
     private $importedSymbols;
     private $traits;
     private $embeddedTemplates = [];
+    private int $lastEmbedIndex = 0;
     private $varNameSalt = 0;
     private $ignoreUnknownTwigCallables = false;
     private ExpressionParsers $parsers;
@@ -81,8 +82,13 @@ class Parser
      */
     public function parse(TokenStream $stream, $test = null, bool $dropNeedle = false): ModuleNode
     {
+        // reset on root parse() calls only, so the counter spans nested/reentrant parses
+        if (!$this->stack) {
+            $this->lastEmbedIndex = 0;
+        }
+
         $vars = get_object_vars($this);
-        unset($vars['stack'], $vars['env'], $vars['handlers'], $vars['visitors'], $vars['expressionParser'], $vars['reservedMacroNames'], $vars['varNameSalt']);
+        unset($vars['stack'], $vars['env'], $vars['handlers'], $vars['visitors'], $vars['expressionParser'], $vars['reservedMacroNames'], $vars['lastEmbedIndex'], $vars['varNameSalt']);
         $this->stack[] = $vars;
 
         // node visitors
@@ -319,7 +325,7 @@ class Parser
      */
     public function embedTemplate(ModuleNode $template)
     {
-        $template->setIndex(mt_rand());
+        $template->setIndex(++$this->lastEmbedIndex);
 
         $this->embeddedTemplates[] = $template;
     }
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,6 @@
-# 3.24.1 (2026-XX-XX)
+# 3.25.0 (2026-XX-XX)
 
- * n/a
+ * Use deterministic suffixes for generated embed classes
 
 # 3.24.0 (2026-03-17)
 
--- a/doc/templates.rst
+++ b/doc/templates.rst
@@ -794,6 +794,21 @@ operand is contained in the right:
     You can use this operator to perform a containment test on strings,
     sequences, mappings, or objects implementing the ``Traversable`` interface.
 
+.. note::
+
+    For sequences, mappings, and ``Traversable`` objects, ``in`` uses a loose
+    comparison (similar to ``==``); use :doc:`same as <tests/sameas>` for a
+    strict comparison. Like PHP's ``in_array()``, this can yield unexpected
+    results when the left operand is a boolean:
+
+    .. code-block:: twig
+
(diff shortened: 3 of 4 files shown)


stof:
AFAICT, reusing the index should not be an issue as long as you ensure that it is not reused for embeds belonging to the same template (the final generated class name uses both the template name and the index to generate the class name). The hard thing is that this would have to account for *nested* `embed` nodes, while the parser state is reset when re-entering `parse()` to parse the inner `{% embed %}` (otherwise, we could just count `$this->embeddedTemplates` as a way to get an index).

And the index must indeed start at `1` for embedded templates due to the fact that falsy index values are treating the same than `null` in the compilation of `ModuleNode` currently: https://github.com/twigphp/Twig/blob/99d95dd6b8b0cc3c2397d15d45a4bb87eff89ddc/src/Node/ModuleNode.php#L98
This restriction could probably be lifted by using proper `null` checks instead.

The current solution with a static index is **not** reproducible either, as the compilation of a template depends on which templates have been parsed before it (or at least how many `{% embed %}` tag they contain)

itsalmostchristmas:
Thanks for your feedback.

> The current solution with a static index is **not** reproducible either, as the compilation of a template depends on which templates have been parsed before it (or at least how many `{% embed %}` tag they contain)

Indeed, I had only considered a full rebuild (e.g. `bin/console cache:clear` in Symfony), and not compiling individual templates. I've updated the PR to add a counter that's only reset upon leaving the root module. To my knowledge, this will result in deterministic output regardless of parsing order.

> This restriction could probably be lifted by using proper `null` checks instead.

I've decided not to change this, as in my mind it makes more sense for numbering of the embed classes to start from 1 anyway.

src/Parser.php
```diff
@@ -83,6 +84,12 @@ public function parse(TokenStream $stream, $test = null, bool $dropNeedle = fals
     {
         $vars = get_object_vars($this);
         unset($vars['stack'], $vars['env'], $vars['handlers'], $vars['visitors'], $vars['expressionParser'], $vars['reservedMacroNames'], $vars['varNameSalt']);
+
+        // restore on root modules only
+        if (0 < \count($this->stack)) {
+            unset($vars['nextEmbedIndex']);
```
stof: I would always unset it so that it is not stored in the stack, and then doing `if (!$this->stack) { $this->nextEmbedIndex = 1; }`. It will provide the same result by with clearer code IMO.

src/Parser.php
```diff
@@ -142,6 +143,11 @@ public function parse(TokenStream $stream, $test = null, bool $dropNeedle = fals
             $this->$key = $val;
         }
 
+        // restore on root modules only
```
stof: I would rather do that at the beginning of the method than at the end, to be safer in case some exception happens and the parser is reused later.

stof (approved):

src/Parser.php
```diff
@@ -81,8 +82,13 @@ public function getVarName(): string
      */
     public function parse(TokenStream $stream, $test = null, bool $dropNeedle = false): ModuleNode
     {
+        // restore on root modules only
```
fabpot:
```suggestion
        // reset on root parse() calls only, so the counter spans nested/reentrant parses
```

src/Parser.php
```diff
@@ -319,7 +325,7 @@ public function hasTraits(): bool
      */
     public function embedTemplate(ModuleNode $template)
     {
-        $template->setIndex(mt_rand());
+        $template->setIndex($this->nextEmbedIndex++);
```
fabpot:
```suggestion
        $template->setIndex(++$this->nextEmbedIndex);
```

fabpot (changes requested):
It looks good to me. Thank you.

Can you add a small regression test?
And a new entry in CHANGELOG?

src/Parser.php
itsalmostchristmas: Also changed the initial value to zero.

stof: the property name should be changed though, as it does not store the *next* index anymore, but the *last* index.

fabpot (approved):

fabpot:
Thank you @itsalmostchristmas.

PR #4798, opened 2026-05-07
[Doc] Reword whitespace control note about first-newline removal

### LINKED ISSUE #4720: The counter intuitive removal of first newline immediately after the regular comment closing #}
The doc/templates.rst states: "... default removal of the first newline inherited from PHP".

What are you talking about?  Where PHP removes the first newline?   I cannot google up anything like such a newline removal of PHP.

I think, this "default removal of the first newline" should be removed both from the Lexer class and from the doc.

Amoifr:
Fixes #4720.

The wording on line 1224 of `doc/templates.rst` ("removal of the first newline **inherited from PHP**") was misleading — Twig does not inherit anything from PHP, it deliberately implements the same behavior (just like Jinja does, which uses *"like in PHP"* in its own documentation).

The note already exists a few lines above (line 1210: *"The first newline after a template tag is removed automatically (like in PHP)."*), so the second mention can just point back to it instead of duplicating the (incorrect) attribution.

### DIFF
--- a/doc/templates.rst
+++ b/doc/templates.rst
@@ -1221,7 +1221,7 @@ Twig supports two modifiers:
 
 * *Line whitespace trimming* via the ``~`` modifier: Removes all whitespace
   (excluding newlines). Using this modifier on the right disables the default
-  removal of the first newline inherited from PHP.
+  removal of the first newline mentioned above.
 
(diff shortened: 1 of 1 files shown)


fabpot (approved):

fabpot:
Thank you @Amoifr.

Issue #4754, opened 2026-02-04
[RFC] Drop `|raw` in favor of a more secure approach


Toflar:
I would like to propose deprecating (and in future drop) the use of `|raw` which is quite a radical change of the current concept, I'm very much aware of that 😊 But bear with me for a second here - I named it `RFC` for a reason 😉 

Here's why I think the concept of `|raw` in templates is fundamentally flawed.
If you take a step back and think of what `{{ variable|raw }}` does, the first thought that comes to your mind because of how the filter is named is probably:

> Hello Twig, output `variable` exactly as it is, I know it is safe here.

And this thought is exactly what I would like to challenge. 
Because in projects where **you** write the controllers and **you** control the templates. **you** know exactly what input you have, where you pass the variables, in what context etc. - you are in **full control**.
Hence, this makes you think (!) that `{{ variable|raw }}` is safe and you can ouput it `|raw`.
But can you?

Here's the thing: I think, this assumption is wrong.
Because very often, the ones that write the templates and the ones that use them (aka write the controllers), are **not** the same people. 

Imagine you have template designer Alice that designs a `table.html.twig`:

```twig
<table>
    {% for row in rows %}
        <tr>
            {% for cell in row %}
                <td>{{ cell }}</td>
            {% endfor %}
        </tr>
    {% endfor %}
</table>
```

That's great, developer Bob can now use this template when writing his controller:

```php
$rows = [
    ['Cell 1', 'Cell 2', 'Cell 3'],
    ['Cell 4', 'Cell 5', 'Cell 6'],
];

echo $twig->render('table.html.twig', [
    'rows' => $rows,
]);
```

Nice! Now two weeks later, developer Eve also needs to output content formatted as a table, in a completely different context. She needs to output user input in one of the cells, though:

```php
$rows = [
    ['Cell 1', 'Cell 2', 'Cell 3'],
    ['Cell 4', 'Cell 5', $request->query->get('user-input')],
];

echo $twig->render('table.html.twig', [
    'rows' => $rows,
]);
```

Well, no problem. She checks the template, sees that it contains `<td>{{ cell }}</td>` and knows that Twig's output encoding is going to take care of that properly. All good!

Now another two weeks later, developer Bob comes back to his controller and his task is to wrap `Cell 1` into a `<strong>Cell 1</strong>`.
So he asks template designer Alice to adjust the template to allow for that.
Now what does template designer Alice do?
[...]

PR #4801, opened 2026-05-18
Allow Markup to be marked as safe for escaping

aschempp:
Draft implementation of https://github.com/twigphp/Twig/issues/4754

Let's keep the discussion on _why and how_ to the original issue, but feel free to discuss the suggested code changes here 😊 

/cc @Toflar

### DIFF
--- a/src/Markup.php
+++ b/src/Markup.php
@@ -20,11 +20,16 @@ class Markup implements \Countable, \JsonSerializable, \Stringable
 {
     private $content;
     private ?string $charset;
+    private array $options;
 
-    public function __construct($content, $charset)
+    public function __construct($content, $charset, array $options = [])
     {
         $this->content = (string) $content;
         $this->charset = $charset;
+
+        $this->options = array_merge([
+            'is_safe' => null,
+        ], $options);
     }
 
     public function __toString(): string
@@ -35,6 +40,11 @@ class Markup implements \Countable, \JsonSerializable, \Stringable
     public function getCharset(): string
     {
         return $this->charset;
+    }
+
+    public function getSafe(): array|null
+    {
+        return $this->options['is_safe'] ?? null;
     }
 
     /**
--- a/src/Runtime/EscaperRuntime.php
+++ b/src/Runtime/EscaperRuntime.php
@@ -99,7 +99,7 @@ final class EscaperRuntime implements RuntimeExtensionInterface
      */
     public function escape($string, string $strategy = 'html', ?string $charset = null, bool $autoescape = false)
     {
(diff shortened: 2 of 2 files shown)


src/Markup.php
```diff
@@ -20,11 +20,16 @@ class Markup implements \Countable, \JsonSerializable, \Stringable
 {
     private $content;
     private ?string $charset;
+    private array $options;
 
-    public function __construct($content, $charset)
+    public function __construct($content, $charset, array $options = [])
```
stof, line 25: I think we should pass an array of strategies there, not an array of options.

src/Runtime/EscaperRuntime.php
```diff
@@ -99,7 +99,7 @@ public function addSafeClass(string $class, array $strategies)
      */
     public function escape($string, string $strategy = 'html', ?string $charset = null, bool $autoescape = false)
     {
-        if ($autoescape && $string instanceof Markup) {
+        if ($autoescape && $string instanceof Markup && (null === $string->getSafe() || \in_array($strategy, $string->getSafe(), true))) {
```
stof, line 102: This should probably take into account the case of strategies that are subset of others (`html_attr` is safe for `html_attr_relaxed` and `html`, and `html_attr_relaxed`  is safe for `html`)

stof (commented):
The `{% set %}` tag in capture mode should probably be updated to mark its Markup output as safe for the current autoescaping strategy rather than all.
But I'm not sure whether we can do that in 3.x or should wait for 4.0.

src/Runtime/EscaperRuntime.php
aschempp: I didn't know about that! Does that also apply to functions and filters? If my function `[is_safe => 'html_attr']`, does a `foo.html.twig` skip autoescaping? Could you point to any code to look at?

src/Markup.php
aschempp: I thought about that too, but was unsure whether there can be more options. Also, this would make the syntax kinda identical with functions and filters. But I'm totally fine with either way.

src/Runtime/EscaperRuntime.php
stof: See https://github.com/twigphp/Twig/blob/3.x/src/NodeVisitor/SafeAnalysisNodeVisitor.php which does the analysis for auto-escaping.

src/Markup.php
stof: I don't think the Markup value object need to be consistent in syntax with functions and filters. They have totally different purposes.

PR #4818, opened 2026-05-25
Deprecate Twig\Sandbox\SecurityPolicy::setStrict()

fabpot:


stof:
In 4.0, I would only deprecate `->setStrict(false)`, keeping the method non-deprecated for now, to make it easier to support 3.x and 4.0 at the same time in a deprecation-free way.

PR #4822, opened 2026-05-29
Fix inconsistent array access with a Stringable key

### LINKED ISSUE #4804: \Stringable when strict_variables: true silently becomes string in array key. When strict_variables: false exception thrown
Try following template (test.twig):

`{{ array[object] }}`

For context:
```
$object = new class implements \Stringable {
    public function __toString()
    {
        return 'string';
    }
};
$this->render('test.twig', ['array' => ['string' => 'value'], 'object' => $object]);
```

Rendered template, `strict_variables: false`
```
(($_v0 = ($context["array"] ?? null)) && is_array($_v0) || $_v0 instanceof ArrayAccess ?
        ($_v0[($context["object"] ?? null)] ?? null) : null)
```
No casting key into string here. Exception: `Cannot access offset of type Stringable@anonymous on array`.

Rendered template, `strict_variables: true`
```
 CoreExtension::getAttribute(
        $this->env,
        $this->source,
        (isset($context["array"]) || array_key_exists("array", $context) ? $context["array"] : (function () {
            throw new RuntimeError('Variable "array" does not exist.', 1, $this->source);
        })()),
        (isset($context["object"]) || array_key_exists("object", $context) ? $context["object"] : (function () {
            throw new RuntimeError('Variable "object" does not exist.', 1, $this->source);
        })()),
        [],
        "array",
        false,
        false,
        false,
        1
    )
``` 

And inside `CoreExtension::getAttribute`:
```
if (match (true) {
    \is_array($object) => \array_key_exists($arrayItem = (string) $arrayItem, $object),
    $object instanceof \ArrayAccess => $object->offsetExists($arrayItem),
    default => false,
}) {

    ....

    return $object[$arrayItem];
}
```
Array key casts into string on https://github.com/twigphp/Twig/blob/91b3d5182d69dbedd090a6bcc2913ab1a7ee9d89/src/Extension/CoreExtension.php#L1706
no exception thrown.

Expected behaviour: No cast `$arrayItem` into string in `CoreExtension::getAttribute`.
If render throws exception in `strict_variables: false` mode,  it must be thrown in `strict_variables: true` too.

fabpot:
Closes #4804

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
 # 3.27.1 (2026-XX-XX)
 
+ * Fix array access with a `Stringable` key to coerce the key to string consistently instead of throwing in the optimized path
  * Fix sandbox replacing `IteratorAggregate` arguments (e.g. Symfony's `FormView`) by a plain array
 
 # 3.27.0 (2026-05-27)
--- a/src/Node/Expression/GetAttrExpression.php
+++ b/src/Node/Expression/GetAttrExpression.php
@@ -73,9 +73,9 @@ class GetAttrExpression extends AbstractExpression implements SupportDefinedTest
                     ->raw(' instanceof ArrayAccess ? (')
                     ->raw($var)
                     ->raw('[')
-                    ->subcompile($this->getNode('attribute'))
-                    ->raw('] ?? null) : null)')
                 ;
+                $this->compileArrayKey($compiler);
+                $compiler->raw('] ?? null) : null)');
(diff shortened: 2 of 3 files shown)


stof (approved):

stof:
@fabpot it would be great to fix the php-cs-fixer config to match the expectation of the codebase, or to fix existing violations if they are not intended, so that the fabbot CI job could be green again.

Issue #4820, opened 2026-05-28
Sandbox mode form_errors exception


pgoddijn123inkt:
Hi everybody,

With the release of twig 3.27.0 `form_errors()` is broken in the sandbox environment when you pass it a `Symfony\Component\Form\FormView`.

`An exception has been thrown during the rendering of a template ("Symfony\Component\Form\FormRenderer::searchAndRenderBlock(): Argument #1 ($view) must be of type Symfony\Component\Form\FormView, array given`

The cause seems to be changes to `ensureToStringAllowed()` / `doEnsureToStringAllowed()` in https://github.com/twigphp/Twig/commit/baebc46b67d641438d4bf276f6f7323cd90b71d6

Because `FormView` is Traversable, it gets cast to array, which causes the provided exception.

stof:
@fabpot I think we need to allow functions and filters to define explicitly whether they need to check toString on arguments (i.e. defining whether they will not do string casting).
Alternatively, we might want to infer this from the argument types of the PHP callable (not doing this array conversion of iterables if the function argument does not accept PHP arrays)

fabpot:
> [@fabpot](https://github.com/fabpot) I think we need to allow functions and filters to define explicitly whether they need to check toString on arguments (i.e. defining whether they will not do string casting). Alternatively, we might want to infer this from the argument types of the PHP callable (not doing this array conversion of iterables if the function argument does not accept PHP arrays)

I'd like to avoid the explicit flag as it it puts the burden on doing the right thing on all third-parties (private and OSS)
I like the type-inference suggestion more but it's a larger change (for well-typed callables, it would work great)

My PR is more like a quick fix one specific case we can fix easily that woudl fix this this issue.

neclimdul:
I've been struggling to piece together this as well. There's a complicated setup where you can recreate this in Drupal too and any Entity and a couple other objects will trigger this when being passed into a Function which, depending on the site, can cause a lot of issues.

The documented logic above https://github.com/twigphp/Twig/blob/3.x/src/Extension/SandboxExtension.php#L194-L199 makes sense in a rendering context, but not when passing to a function that might rely on methods of the object regardless of its interfaces its a bit more complicated. Modifying every function definition that uses this is pretty troublesome but I understand how this might be at the root of the sandbox bypasses and needed.

neclimdul:
[...]

PR #4823, opened 2026-06-01
Skip the sandbox `__toString` check on arguments whose PHP parameter type cannot implicitly coerce to string

fabpot:
The sandbox visitor currently wraps every argument of every Twig callable with `CheckToStringNode.

As an optimization, we are now only wrapping when needed (based on the callable type hints). This is a conservative approach (untyped, mixed, string, array, iterable, object, Stringable, Traversable, self/static/parent and unknown class names all keep wrapping).

Here is a concrete before/after for template `{{ demo(a, b) }}` under the sandbox, with the following signature on the PHP side `demo(int $a, string $b)`:

**Before**:

```php
    yield $this->sandbox->ensureToStringAllowed(
       $this->env->getFunction('demo')->getCallable()(
           $this->sandbox->ensureToStringAllowed(($context["a"] ?? null), 1, $this->source),
           $this->sandbox->ensureToStringAllowed(($context["b"] ?? null), 1, $this->source),
       ),
       1, $this->source,
   );
```

**After**

```php
   yield $this->sandbox->ensureToStringAllowed(
       $this->env->getFunction('demo')->getCallable()(
           ($context["a"] ?? null),                                                            // int: bare, skipped
           $this->sandbox->ensureToStringAllowed(($context["b"] ?? null), 1, $this->source),   // string: still wrapped
       ),
       1, $this->source,
   );
```

### DIFF
--- a/src/Util/CallableParameters.php
+++ b/src/Util/CallableParameters.php
@@ -1,0 +1,161 @@
+<?php
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Twig\Util;
+
+use Twig\Environment;
+use Twig\Node\Expression\FilterExpression;
+use Twig\Node\Expression\FunctionExpression;
+use Twig\Node\Expression\TestExpression;
+use Twig\Node\Node;
+use Twig\TwigCallableInterface;
+
+/**
+ * Reflects the PHP parameters backing a Twig callable expression.
+ *
+ * @author Fabien Potencier <fabien@symfony.com>
+ *
+ * @internal
+ */
+final class CallableParameters
+{
+    /**
+     * Returns the PHP parameters of a Filter/Function/Test call mapped to its
+     * template-level arguments.
+     *
+     * The parameters Twig injects automatically
+     * (`needs_charset/environment/context/is_sandboxed`) and the bound
+     * `arguments` are stripped. For a filter, the first returned parameter is
+     * the filter's input value. Returns null when reflection fails or the node
+     * is not a callable expression.
+     *
+     * @return list<\ReflectionParameter>|null
+     */
+    public static function fromNode(Node $node, Environment $env): ?array
+    {
+        if (!$node instanceof FilterExpression && !$node instanceof FunctionExpression && !$node instanceof TestExpression) {
+            return null;
+        }
+
+        $callable = self::resolveTwigCallable($node, $env);
+        if (null === $callable || null === $callable->getCallable()) {
+            return null;
+        }
+
+        try {
+            return (new ReflectionCallable($callable))->getTwigParameters();
+        } catch (\LogicException) {
+            return null;
+        }
+    }
+
+    /**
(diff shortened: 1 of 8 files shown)


src/Util/CallableParameters.php
```diff
@@ -0,0 +1,143 @@
...
+
+        if ($type->isBuiltin()) {
+            return match ($name) {
+                'int', 'float', 'bool', 'true', 'false', 'void', 'never' => true,
+                default => false, // string, array, iterable, object, mixed, callable
+            };
+        }
+
+        // Relative types can resolve to a subtype at runtime, so never safe.
+        if (\in_array($name, ['self', 'static', 'parent'], true)) {
```
stof: `self` and `parent` should be resolved to the actual class name, and be subject to the same logic than the class name. And `static` should probably be handled that way as well (probably treating it like `self` in this case).

src/Util/CallableParameters.php
```diff
@@ -0,0 +1,143 @@
...
+            return false;
+        }
+        // Interfaces and non-final classes are open: a subtype could add
+        // Stringable/Traversable, so only a final non-coercible class is safe.
+        if (!class_exists($name)) {
```
stof: Enums should be considered safe, as they cannot be stringable.

src/Util/CallableParameters.php
```diff
@@ -0,0 +1,153 @@
...
+     */
+    public static function isStringCoercionSafe(?\ReflectionType $type, ?\ReflectionClass $scope = null): bool
+    {
+        if (null === $type) {
+            return false;
+        }
+
+        // For a union the value is one of the members; for an intersection it
+        // satisfies all of them and thus has every member's capabilities. In
+        // both cases the value can coerce unless every member is itself safe.
```
stof: Due to the definition of "safe" for object types (a final class that is not stringable nor traversable), an intersection type is actually safe if it contains a safe type.

src/Util/CallableParameters.php
```diff
@@ -0,0 +1,153 @@
...
+            return true;
+        }
+
+        if (!$type instanceof \ReflectionNamedType) {
+            return false;
+        }
+
+        $name = $type->getName();
+        // `null` on a nullable type is always safe (no __toString).
+        if ('null' === $name) {
```
stof: Why not letting this be handled in the `$type->isBuiltin()` case below ? It is also a builtin type.

stof (commented):
For final classes implementing Stringable, would it make sense to try checking the policy at compile-time and mark it safe if we know that the policy allows casting them to string ?
This might help with the kind of logic currently done by Drupal to optimize the sandbox: https://github.com/twigphp/Twig/issues/4820#issuecomment-4566333505

fabpot:
> For final classes implementing Stringable, would it make sense to try checking the policy at compile-time and mark it safe if we know that the policy allows casting them to string ? This might help with the kind of logic currently done by Drupal to optimize the sandbox: [#4820 (comment)](https://github.com/twigphp/Twig/issues/4820#issuecomment-4566333505)

This is not possible as `SecurityPolicy` can be swapped, so we cannot make that decision at compile-time. The fact that security policies can be swapped is annoying for sure: https://github.com/twigphp/Twig/commit/99a10384ff0

stof (approved):

PR #4824, opened 2026-06-02
Cast printed expressions to string so values that cannot be converted to a string (arrays, non-`Stringable` objects, ...) report a usable stack trace at the print location

### LINKED ISSUE #4644: Improve error reporting in template render
We have a use case where we render templates that may contain data in array format and we use a filter to process that data when it is rendered.
If we forget the the filter, the template render throws "Array to string conversion" ErrorException in `Template.src#359`.
This ErrorException thrown flows up until our app catches it, but at that time there is no source context or line number information available.
It would be really helpful is twig could enhance this kind of errors with source context and line like it does for other template rendering errors.

### LINKED ISSUE #4765: Stack traces are unusable for array-to-string conversions when using the yield-based system
I got a case where a template was doing `{{ value }}` and received an empty array as `value` (due to a bug in my code). The stack trace for that was totally useless:

Image

In the old buffer-based architecture, converting the array to a string would happen in the compiled template at the location of this `{{ value }}`, allowing to debug this.
The new architecture documents that `Template::doDisplay` should return an `iterable<scalar|\Stringable|null>`, but the compiled code does not actually ensure this is respected. `PrintNode` does not ensure that the expression result satisfies that type before yielding it.

This has 2 impact:
- when using `TemplateWrapper::render` or `TemplateWrapper::display` (or the equivalent block-related methods), you will get warning with a useless stack trace (see my example)
- when using `TemplateWrapper::stream` (or `TemplateWrapper::streamBlock`), the returned iterable will not satisfy the contract (and whether you get an error or warning will depend on how the iterable is consumed, but it won't have a stack trace reaching to the root cause in the PrintNode).

fabpot:
Closes #4765
Closes #4780
Closes #4644

Is it what you had in mind @stof?

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,7 @@
 # 3.27.2 (2026-XX-XX)
 
  * Make the `IntegrationTestCase` and `NodeTestCase` test helpers compatible with PHPUnit 11
+ * Cast printed expressions to string so values that cannot be converted to a string (arrays, non-`Stringable` objects, ...) report a usable stack trace at the print location
  * Skip the sandbox `__toString` check on arguments whose PHP parameter type cannot implicitly coerce to string
 
 # 3.27.1 (2026-05-30)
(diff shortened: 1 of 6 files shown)


fabpot:
Thank you @stof.

stof:
@fabpot yes

PR #4825, opened 2026-06-02
Make the include() function return a Markup object

### LINKED ISSUE #4802: include() function must return a Markup object instead of a string
Hi everybody,

In a project, we have a node visitor altering the values of some template variables. We know if the value must be escaped or not thanks to the type: plain PHP string or `\Twig\Markup`

When setting a variable with a `set` tag "capturing" a chunk of text, the value of the variable is a `\Twig\Markup`.

```
{% set x %}{{ include('inner_template.twig') }}{% endset %}
{{ include('outer_template.twig', {slot: x}) }}
```

However, when setting a variable with an assignment, the value of the variable is a plain PHP string.

```
{% set x = include('inner_template.twig') %}
{{ include('outer_template.twig', {slot: x}) }}
````

It is not possible to just ask the template owner to use set in a way instead of the other, or to use `embed` tag with blocks. We have no control on the template authoring.

So, is it possible to make  `\Twig\Extension\CoreExtension::include()` returning a Markup instead of string?

```
- public static function include(Environment $env, $context, $template, $variables = [], $withContext = true, $ignoreMissing = false, $sandboxed = false): string
+ public static function include(Environment $env, $context, $template, $variables = [], $withContext = true, $ignoreMissing = false, $sandboxed = false): Markup
```

(thanks @vever001 for the suggestion and the examples)

fabpot:
Closes #4802

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -4,6 +4,7 @@
  * Stop reporting a skipped test in `IntegrationTestCase` when there is no legacy test to run
  * Make the `IntegrationTestCase` and `NodeTestCase` test helpers compatible with PHPUnit 11
  * Cast printed expressions to string so values that cannot be converted to a string (arrays, non-`Stringable` objects, ...) report a usable stack trace at the print location
+ * Make the `include()` function return a `Markup` object so an assigned result is not re-escaped when printed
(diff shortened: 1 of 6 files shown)


pdureau:
Merci Fabien 🤗

PR #4841, opened 2026-06-11
Fix Markup truthiness in boolean expressions

xtrime-ru:
## Summary

Fixes Markup truthiness handling in boolean expressions. 

`trim` can return a `Twig\Markup` instance for safe strings. Empty `Markup` objects must behave like empty strings in Twig truth tests, but PHP treats all objects as truthy. This caused expressions like this to incorrectly evaluate as true:
```twig
{% set x %}    {% endset %}
{% if x|trim and x|trim %}fail{% else %}ok{% endif %}
```

This case was working properly in https://github.com/twigphp/Twig/releases/tag/v3.14.2 and earlier.

## Related commits
- Bug was introduced in [v3.15.0](https://github.com/twigphp/Twig/releases/tag/v3.15.0) in this commit https://github.com/twigphp/Twig/commit/10c3142d3b036910f63080070c101bfff61e0743
- Partially fixed here:  https://github.com/twigphp/Twig/commit/10c3142d3b036910f63080070c101bfff61e0743

## Changes

- Added `TrueTest::wrap()` to centralize wrapping non-primitive expressions with Twig’s Markup-aware true test.
- Reused `TrueTest::wrap()` in:
  - `IfNode`
  - conditional ternary expressions
  - `and`, `or`, and `xor` binary expressions
  - Elvis expressions
  - unary `not`
- Added regression coverage for boolean operators (`and`, `or`, `xor`, `not`) and ternary/Elvis expressions whose operands evaluate to empty `Markup`.

## Tests
```bash
./vendor/bin/simple-phpunit tests/IntegrationTest.php --filter markup_test
```

### DIFF
--- a/src/Node/Expression/Binary/AndBinary.php
+++ b/src/Node/Expression/Binary/AndBinary.php
@@ -14,9 +14,16 @@ namespace Twig\Node\Expression\Binary;
 
 use Twig\Compiler;
 use Twig\Node\Expression\ReturnBoolInterface;
+use Twig\Node\Expression\Test\TrueTest;
+use Twig\Node\Node;
 
 class AndBinary extends AbstractBinary implements ReturnBoolInterface
 {
+    public function __construct(Node $left, Node $right, int $lineno)
+    {
+        parent::__construct(TrueTest::wrap($left), TrueTest::wrap($right), $lineno);
+    }
+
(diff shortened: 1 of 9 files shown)


fabpot:
Thank you @xtrime-ru.

PR #4845, opened 2026-06-17
Add documention note about variable scope of override blocks in {% embed ... only %}

### LINKED ISSUE #4844: Document (and possibly improve) variable scope of override blocks in {% embed ... only %}
## Summary

When overriding a block inside {% embed ... only %}, the block body executes in the embedded template's scope, not the calling template's. So variables available at the call site aren't available inside the block unless explicitly passed via `with`. This is surprising for component development, and as far as I can tell it isn't documented.

## Minimal reproduction

```twig
{# component.twig #}
{% block content %}{% endblock %}
```
```twig
{# page.twig #}
{% set msg = 'hello' %}
{% embed 'child.twig' with {} only %}
  {% block content %}{{ msg }}{% endblock %}  {# msg is undefined #}
{% endembed %}
```

Remove `only` and it renders `hello`, which confirms the behavior is specifically the context-trimming of `only` — the override block resolves `msg` against the embedded scope, which only has emptied.

## Why it's surprising

The block is written at the call site, but it runs in the embedded template's context. That write/run split is not terribly obvious, and it's the piece the [embed docs](https://twig.symfony.com/doc/3.x/tags/embed.html) don't currently explain.

Based on the related issues linked below, I understand this is intentional, but I'm flagging that it's undocumented and ergonomically rough for a growing use case.

## Motivation: Drupal's Single Directory Components

Drupal's Single Directory Components build on `embed`, giving components "props" (the `with` mapping) and "slots" (blocks) — a model very similar to React or web components. The [SDC docs explicitly recommend `only`](https://www.drupal.org/docs/develop/theming-drupal/using-single-directory-components/using-your-new-single-directory-component#s-use-embed-) "to avoid unexpected side-effects," and Drupal theme contexts are large, so `only` is the norm. The result is that slot content can't see call-site variables unless they're also threaded through props, which is unexpected coming from other component systems.

## Ask

1. Docs: Could the embed page state explicitly that override blocks execute in the embedded template's scope, and that `only` therefore excludes call-site variables from those blocks?
2. Design (open question): Is there any way to alleviate this perceived quirk?

## Related

- #2435 — explicit per-block variables (closest; closed as sugar). This issue adds the docs angle + the SDC/component motivation.
- #3665 — overriding embed blocks from a child template (different: outer override, not scope).
- #1360 — overriding blocks in included files (different tag).

andy-blum:
Closes #4844

### DIFF
--- a/doc/tags/embed.rst
+++ b/doc/tags/embed.rst
@@ -73,7 +73,7 @@ two boxes side by side:
     ┌─── page layout ─────────────────────┐
     │                                     │
     │           ┌── block "content" ──┐   │
-    │           │                     │   │    
+    │           │                     │   │
     │           │ ┌ block ┐ ┌ block ┐ │   │
     │           │ │"left" │ │"right"│ │   │
     │           │ │       │ │       │ │   │
@@ -163,6 +163,30 @@ The ``embed`` tag takes the exact same arguments as the ``include`` tag:
     {% embed "base" ignore missing %}
         ...
     {% endembed %}
+
+.. note::
+
+    Blocks you override inside an ``embed`` are evaluated in the embedded
+    template's context, not the context of the template containing the
+    ``embed`` tag. As with ``include``, embedded templates have access to the
+    variables of the active context. You can disable access to the context by
+    appending the ``only`` keyword. This prevents override blocks
+    from accessing the surrounding context, so any variable they need must also
+    be passed explicitly through ``with``:
+
+    .. code-block:: twig
+
+        {% set name = 'Fabien' %}
+
+        {# "name" is undefined inside the block #}
+        {% embed "base" only %}
+            {% block content %}{{ name }}{% endblock %}
+        {% endembed %}
+
+        {# "name" is passed explicitly and is available #}
+        {% embed "base" with {'name': name} only %}
+            {% block content %}{{ name }}{% endblock %}
+        {% endembed %}
 
 .. warning::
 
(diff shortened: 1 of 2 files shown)


doc/tags/embed.rst
```diff
@@ -164,6 +164,30 @@ The ``embed`` tag takes the exact same arguments as the ``include`` tag:
...
+    appending the ``only`` keyword. This prevents override blocks
+    from accessing the surrounding context, so any variable they need must also
+    be passed explicitly through ``with``:
+
+    .. code-block:: twig
+
+        {% set name = 'Fabien' %}
+
+        {# "name" is undefined inside the block #}
+        {% embed "base" only %}
```
e0ipso, line 182:
```suggestion
        {% embed "base" only %}
            {# "name" is undefined inside the block #}
```

doc/tags/embed.rst
```diff
@@ -164,6 +164,30 @@ The ``embed`` tag takes the exact same arguments as the ``include`` tag:
...
+            {% block content %}{{ name }}{% endblock %}
+        {% endembed %}
+
+        {# "name" is passed explicitly and is available #}
+        {% embed "base" with {'name': name} only %}
```
e0ipso, line 187:
```suggestion
        {% embed "base" with {'name': name} only %}
            {# "name" is passed explicitly and is available #}
```

e0ipso (commented):
Good clarifications. I suggested an über-nit pick.

fabpot:
Thank you @andy-blum.

PR #4851, opened 2026-06-27
Redesign macro calls and argument handling

fabpot:
This is my attempt to make macros "better". It uses modern PHP features that didn't exist when I designed macros a long time ago.

The first objective is to close the gap between their behavior and the behavior of Twig callables: functions, filters, and tests.

Here are some important changes:

* Calling a macro without passing a value for an argument that has no default value is deprecated; it is currently silently passed as `null`.
* Passing extra positional arguments or unknown named arguments to a macro without an explicit variadic argument is deprecated; these arguments are currently silently accepted through the implicit `varargs` variable.
* Explicit variadic macro arguments are now supported with `...name`.
* Macros are compiled as closures stored in the macro registry, instead of public generated `macro_*` methods.

The refactor introduces `TwigMacro` and `MacroArgument` to represent template-defined macros with an explicit signature, similar to the existing Twig callable model.

### DIFF
--- a/doc/tags/macro.rst
+++ b/doc/tags/macro.rst
@@ -11,26 +11,62 @@ via macros (called ``forms.twig``):
 
 .. code-block:: html+twig
 
-    {% macro input(name, value, type = "text", size = 20) %}
-        <input type="{{ type }}" name="{{ name }}" value="{{ value|e }}" size="{{ size }}"/>
+    {% macro input(name, value = "", type = "text", size = 20) %}
+        <input
+            type="{{ type }}"
+            name="{{ name }}"
+            value="{{ value|e }}"
+            size="{{ size }}"
+        />
     {% endmacro %}
 
-    {% macro textarea(name, value, rows = 10, cols = 40) %}
-        <textarea name="{{ name }}" rows="{{ rows }}" cols="{{ cols }}">{{ value|e }}</textarea>
+    {% macro textarea(name, value = "", rows = 10, cols = 40) %}
+        <textarea
+            name="{{ name }}"
+            rows="{{ rows }}"
+            cols="{{ cols }}"
+        >{{ value|e }}</textarea>
     {% endmacro %}
 
-Each macro argument can have a default value (here ``text`` is the default value
+A macro argument can have a default value (here ``text`` is the default value
 for ``type`` if not provided in the call).
 
-Macros differ from native PHP functions in a few ways:
+As with PHP function arguments, a macro argument is required unless it declares
+a default value. Here, ``name`` is required while ``value``, ``type``, and
+``size`` are optional.
 
-* Arguments of a macro are always optional.
+.. deprecated:: 3.29
 
-* If extra positional arguments are passed to a macro, they end up in the
-  special ``varargs`` variable as a list of values.
+    Calling a macro without a value for an argument that has no default value is
+    deprecated as of Twig 3.29; the argument will be required in Twig 4.0 (until
+    then, it defaults to ``null``). Give every optional argument an explicit
+    default value.
 
-But as with PHP functions, macros don't have access to the current template
-variables.
+To accept an arbitrary number of extra arguments, declare an explicit variadic
+argument as described below.
+
+Note that macros don't have access to the current template variables.
+
+A macro can declare an explicit variadic argument to collect any extra
+positional and named arguments into a named variable, using the same ``...``
+notation as PHP:
+
+.. code-block:: html+twig
+
+    {% macro tag(element, ...attributes) %}
+        <{{ element }}
+        {%- for key, value in attributes %} {{ key|e('html_attr') }}="{{ value }}"{% endfor -%}
+        >
+    {% endmacro %}
+
+    {{ _self.tag("input", type: "text", name: "username") }}
+
+The variadic argument must be the last one and cannot have a default value.
+
+.. versionadded:: 3.29
+
+    Support for declaring an explicit variadic macro argument was added in Twig
+    3.29.
 
 .. tip::
 
@@ -110,8 +146,13 @@ via the ``from`` tag:
 
         <p>{{ _self.input('password', '', 'password') }}</p>
 
-        {% macro input(name, value, type = "text", size = 20) %}
-            <input type="{{ type }}" name="{{ name }}" value="{{ value|e }}" size="{{ size }}"/>
+        {% macro input(name, value = "", type = "text", size = 20) %}
+            <input
+                type="{{ type }}"
+                name="{{ name }}"
+                value="{{ value|e }}"
+                size="{{ size }}"
+            />
         {% endmacro %}
 
 Macros Scoping
@@ -162,3 +203,16 @@ readability (the name after the ``endmacro`` word must match the macro name):
     {% macro input() %}
         ...
     {% endmacro input %}
+
+Deprecating a Macro
+-------------------
+
+Use the :doc:`deprecated <deprecated>` tag at the top of a macro to deprecate
+it; a deprecation notice is triggered whenever the macro is called:
+
+.. code-block:: html+twig
+
+    {% macro input(name, value = "") %}
+        {% deprecated 'The "input" macro is deprecated, use "field" instead.' %}
+        <input name="{{ name }}" value="{{ value|e }}"/>
+    {% endmacro %}
--- a/src/ExpressionParser/Infix/DotExpressionParser.php
+++ b/src/ExpressionParser/Infix/DotExpressionParser.php
@@ -60,34 +60,25 @@ final class DotExpressionParser extends AbstractExpressionParser implements Infi
             }
         }
 
-        if ($stream->test(Token::OPERATOR_TYPE, '(')) {
-            $type = Template::METHOD_CALL;
-            $arguments = $this->parseCallableArguments($parser, $token->getLine());
-        }
-
         $isMacroTarget = $expr instanceof NameExpression
             && (
                 null !== $parser->getImportedSymbol('template', $expr->getAttribute('name'))
                 || '_self' === $expr->getAttribute('name')
             );
 
-        if (
-            $isMacroTarget
-            && $attribute instanceof ConstantExpression
-            && \is_string($name = $attribute->getAttribute('value'))
-            && preg_match('#^[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*$#D', $name)
-        ) {
-            $node = new MacroReferenceExpression(new MacroVariable($expr->getAttribute('name'), $expr->getTemplateLine()), 'macro_'.$name, $arguments, $expr->getTemplateLine());
-            $node->setHasParentheses(Template::METHOD_CALL === $type);
-
-            return $node;
+        if ($stream->test(Token::OPERATOR_TYPE, '(')) {
+            $type = Template::METHOD_CALL;
+            $arguments = $this->parseCallableArguments($parser, $token->getLine(), preserveNames: $isMacroTarget);
         }
 
-        if ($isMacroTarget && !$attribute instanceof ConstantExpression) {
-            $node = new MacroReferenceExpression(new MacroVariable($expr->getAttribute('name'), $expr->getTemplateLine()), $attribute, $arguments, $expr->getTemplateLine());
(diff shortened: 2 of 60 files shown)


src/ExpressionParser/Infix/FunctionExpressionParser.php
```diff
@@ -42,8 +42,12 @@ public function parse(Parser $parser, AbstractExpression $expr, Token $token): A
 
         $name = $expr->getAttribute('name');
 
+        // To be removed in 4.0 ("from -> import" unification, see FromTokenParser):
+        // a bare call to a macro imported via "from" is resolved here through the
+        // dedicated "function" imported symbol. In 4.0 this branch goes away once
+        // "from" registers a regular "template" import symbol like "import" does.
```
stof: How would this work ? A macro called from an import is referenced as `m.my_macro()`, while a macro called from `from` looks like the function call `my_macro()`

src/ExpressionParser/Infix/DotExpressionParser.php
```diff
@@ -60,24 +60,24 @@ public function parse(Parser $parser, AbstractExpression $expr, Token $token): A
...
+        if ($stream->test(Token::OPERATOR_TYPE, '(')) {
+            $type = Template::METHOD_CALL;
+            $arguments = $this->parseCallableArguments($parser, $token->getLine(), preserveNames: $isMacroTarget);
+        }
+
         if (
             $isMacroTarget
             && $attribute instanceof ConstantExpression
             && \is_string($name = $attribute->getAttribute('value'))
             && preg_match('#^[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*$#D', $name)
```
stof: do we still need this name validation now that `MacroReferenceExpression` removed the validation ?

stof: btw, we might want to unify both branches creating MacroReferenceExpression now

src/TokenParser/MacroTokenParser.php
```diff
@@ -84,6 +88,22 @@ private function parseDefinition(): ArrayExpression
                 }
             }
 
+            if ($stream->nextIf(Token::OPERATOR_TYPE, '...')) {
+                $token = $stream->expect(Token::NAME_TYPE, null, 'A variadic argument must be a name');
+                $variadicName = (new LocalVariable($token->getValue(), $token->getLine()))->getAttribute('name');
+                if (str_starts_with($variadicName, "\u{035C}")) {
```
stof: why this special case ignoring this Unicode char ?

src/Template.php
```diff
@@ -506,37 +511,88 @@ public function yieldParentBlock($name, array $context, array $blocks = []): ite
...
-    protected function hasMacro(string $name, array $context): bool
+    /**
+     * Tells whether the macro of the given name is defined in this template or one of
+     * its parents. Used by the "is defined" test.
+     *
+     * Matching is case-insensitive on 3.x to mirror the historical behavior (macros used
+     * to compile to PHP methods, whose names fold ASCII case); it becomes case-sensitive
+     * in 4.0.
+     *
+     * @internal
```
stof, line 511: isn't the whole class internal already ?

tests/Fixtures/macros/varargs.test
```diff
@@ -3,11 +3,11 @@ macro with arbitrary arguments
 --TEMPLATE--
 {% from _self import test1, test2 %}
 
-{% macro test1(var) %}
+{% macro test1(var, ...varargs) %}
```
stof, line 6: Do we still have tests covering the deprecated BC layer of the implicit `varargs` ?

tests/Fixtures/macros/simple.test
```diff
@@ -4,7 +4,7 @@ macro
 {% import _self as test %}
 {% from _self import test %}
 
-{% macro test(a, b) -%}
+{% macro test(a = null, b = null) -%}
```
stof, line 7: Do we still have tests covering the deprecated BC layer of implicit optional arguments ?

src/TokenParser/FromTokenParser.php
```diff
@@ -55,8 +55,15 @@ public function parse(Token $token): Node
         $internalRef = new AssignTemplateVariable(new TemplateVariable(null, $token->getLine()), $this->parser->isMainScope());
         $node = new ImportNode($macro, $internalRef, $token->getLine());
 
+        // 4.0 "from -> import" unification: a "from x import a" alias is registered as a
+        // "function" imported symbol carrying the bare macro name, and resolved by
+        // FunctionExpressionParser into a MacroReferenceExpression. Since macro calls from
+        // both "import" and "from" now route through Template::callMacro(), the two paths
+        // only differ in this parse-time symbol plumbing. In 4.0, unify "from" onto the
+        // "template" import symbol and remove the dedicated "function"-symbol branch in
+        // FunctionExpressionParser.
```
stof: How can this actually happen ? The source being parsed looks like a function expression in the case of a usage based on `from`.

src/TokenParser/MacroTokenParser.php
stof: OK, I see that this char is used as a prefix for names corresponding to reserved names. I would suggest introducing a constant holding it to make the intent of the code clearer instead of having `"\u{035C}"` in many places in the code.

stof (changes requested):

fabpot:
@stof I think I've taken all your comments into account.

src/ExpressionParser/Infix/FunctionExpressionParser.php
fabpot: You're right, the bare call always parses as a function call, so this symbol mapping stays; I've fixed the comments accordingly.

src/ExpressionParser/Infix/DotExpressionParser.php
fabpot: Fixed both.

Bonus: `t.('foo-bar')()` now fails with a clear `Macro "foo-bar" is not defined` error instead of a template            attribute-access error.

src/TokenParser/MacroTokenParser.php
fabpot: Done

src/Template.php
fabpot: Removed

tests/Fixtures/macros/varargs.test
fabpot: Coverage existed in TwigMacroTest/CallMacroTest; I've also restored the old fixtures as legacy integration tests.

tests/Fixtures/macros/simple.test
fabpot: Coverage existed in TwigMacroTest/CallMacroTest; I've also restored the old fixtures as legacy integration tests.

doc/tags/macro.rst
```diff
@@ -11,26 +11,62 @@ via macros (called ``forms.twig``):
...
+
+A macro can declare an explicit variadic argument to collect any extra
+positional and named arguments into a named variable, using the same ``...``
+notation as PHP:
+
+.. code-block:: html+twig
+
+    {% macro tag(element, ...attributes) %}
+        <{{ element }}
+        {%- for key, value in attributes %} {{ key }}="{{ value }}"{% endfor -%}
```
stof:
```suggestion
        {%- for key, value in attributes %} {{ key|e('html_attr') }}="{{ value }}"{% endfor -%}
```

Even though this is not strictly necessary for this example, it avoids showing bad examples of how to render dynamic attributes (which can be part of the LLM training data out of context).

src/ExpressionParser/Infix/DotExpressionParser.php
```diff
@@ -60,28 +60,21 @@ public function parse(Parser $parser, AbstractExpression $expr, Token $token): A
...
-        ) {
-            return new MacroReferenceExpression(new TemplateVariable($expr->getAttribute('name'), $expr->getTemplateLine()), 'macro_'.$name, $arguments, $expr->getTemplateLine());
+        if ($stream->test(Token::OPERATOR_TYPE, '(')) {
+            $type = Template::METHOD_CALL;
+            $arguments = $this->parseCallableArguments($parser, $token->getLine(), preserveNames: $isMacroTarget);
         }
 
-        if ($isMacroTarget && !$attribute instanceof ConstantExpression) {
-            return new MacroReferenceExpression(new TemplateVariable($expr->getAttribute('name'), $expr->getTemplateLine()), $attribute, $arguments, $expr->getTemplateLine());
+        if ($isMacroTarget) {
```
stof, line 74: what happens if we detect a macro target but we don't have the braces for a call (i.e. `$type !== Template::METHOD_CALL)` ? Should we create a MacroReferenceExpression, throw a SyntaxError or create a GetAttrExpression ?

I would say creating a GetAttrExpression is probably not a good idea, but creating a MacroReferenceExpression calling the macro without braces might be unexpected.

src/Node/MacroNode.php
```diff
@@ -54,53 +64,80 @@ public function __construct(string $name, Node $body, Node $arguments, int $line
...
 
     public function compile(Compiler $compiler): void
     {
-        $compiler
-            ->addDebugInfo($this)
-            ->write(\sprintf('public function macro_%s(', $this->getAttribute('name')))
-        ;
+        // Macros have no standalone compiled method: they are emitted as closures
+        // stored in the macro registry by compileMacroFactory(), which ModuleNode
+        // calls from the generated loadDeclaredMacros() method.
```
stof: Why wouldn't the MacrosNode call `compile` instead of a custom `compileMacroFactory` ?
AFAICT, those MacroNode don't appear elsewhere in the AST (we have MacroDeclarationNode and MacroReferenceExpression in those places)

fabpot: Correct, changed.

src/ExpressionParser/Infix/DotExpressionParser.php
fabpot: Calling a macro without parentheses is existing behavior in Twig 3.x: both macros.foo and macros.foo() invoke the macro. The parentheses-free form is also required for macros.foo is defined. I've deprecated alling a macro without parentheses though.

src/Parser.php
```diff
@@ -128,7 +129,7 @@ public function parse(TokenStream $stream, $test = null, bool $dropNeedle = fals
             new BodyNode([$body]),
             $this->parent,
             $this->blocks ? new Nodes($this->blocks) : new EmptyNode(),
-            $this->macros ? new Nodes($this->macros) : new EmptyNode(),
+            new MacrosNode($this->macros),
```
stof, line 132: Shouldn't you also deprecate not passing a `Macros` instance to the `ModuleNode` ?

tests/Fixtures/macros/implicit_optional_arguments.legacy.test
```diff
@@ -0,0 +1,22 @@
+--TEST--
+macro arguments without a default value are implicitly optional (deprecated)
```
stof, line 2: shouldn't this test have a `--DEPRECATION--` section ?

tests/Fixtures/macros/varargs_implicit.legacy.test
```diff
@@ -0,0 +1,21 @@
+--TEST--
+macro with arbitrary arguments collected in the implicit "varargs" variable (deprecated)
```
stof, line 2: shouldn't this test have a `--DEPRECATION--` section ?

src/ExpressionParser/Infix/DotExpressionParser.php
fabpot: https://github.com/twigphp/Twig/pull/4877

src/Parser.php
fabpot: Good idea, done

tests/Fixtures/macros/implicit_optional_arguments.legacy.test
fabpot: indeed

tests/Fixtures/macros/varargs_implicit.legacy.test
fabpot: added

stof (approved):
Code looks good, but some tests are failing due to deprecations.

PR #4854, opened 2026-07-04
Make the sandbox a first-class citizen with a dedicated Sandbox class

fabpot:
I've been thinking about making the sabdbox feature as a first class citizen for years. With all the work that has been done recently on security issues, I spent some time on it again. Here is the result.

The main ideas:

* Currently, the sandbox is thigtly coupled to the "main" environment: `SandboxExtension` is registered on the environmen directly, so it instruments all compiled template, and adds runtime checks to all renders, trusted or not. As recommended in the docs, you should have a dedicated environment for sandboxes, different from the main one, but it's not really "enforced" nor natural to do.
* As a consequence, we store some state via `enableSandbox()`/`disableSandbox()` with try/finally patterns scattered across the codebase to support rendering sandboxed and non-sandboxed templates from a environment.
* When using one environment, a sandboxed template can `include` anything the loader can load, sees every application global, and inherits all extensions, this is a footgun (again, already not recommended in the docs).
* There are too maybe "knobs": global mode, `enableSandbox()`, `{% include(..., sandboxed: true) %}`, and `{% sandbox %}`.

The new `Twig\Sandbox\Sandbox` class renders untrusted templates through a dedicated, always-sandboxed environment crafted by the developer. Taht way, there is no state to toggle and nothing leaks between the main environment and the sandbox, in either direction.


src/Sandbox/Sandbox.php
```diff
@@ -0,0 +1,92 @@
...
+ * The sandbox takes ownership of an environment crafted specifically for it:
+ * its loader defines which templates are reachable, its extensions, filters,
+ * functions, tests, and globals define which capabilities exist, and the
+ * security policy defines what is allowed to execute. Never pass an
+ * application environment: the sandbox environment must be dedicated to
+ * rendering untrusted templates.
+ *
+ * @author Fabien Potencier <fabien@symfony.com>
+ */
+final class Sandbox
```
stof: As this is meant to be a public API that projects using Twig will likely inject through DI, it should probably have an interface for its public API (or it should not be final). Otherwise, the DI would force using the real sandbox all the time without any way to change things (no mocking, no inheritance, no decoration)

src/Sandbox/Sandbox.php
```diff
@@ -0,0 +1,92 @@
...
+        }
+
+        try {
+            $env->addExtension(new SandboxExtension($policy, true));
+        } catch (\LogicException $e) {
+            throw new \LogicException(\sprintf('The environment passed to "%s" must be dedicated to it: pass a freshly built environment that has no "%s" registered and has not been used yet.', self::class, SandboxExtension::class), 0, $e);
+        }
+    }
+
+    public function render(string $name, array $context = []): string
```
stof, line 47: I suggest adding the `array<string, mixed>` type for the context array to provide better type for static analysis (in all methods that have it)

src/Sandbox/SecurityChecker.php
```diff
@@ -0,0 +1,234 @@
...
+    }
+
+    public function checkSecurity(array $tags, array $filters, array $functions, array $tests, ?Source $source = null): void
+    {
+        if (!$this->isSandboxed($source)) {
+            return;
+        }
+
+        if ((new \ReflectionMethod($this->policy, 'checkSecurity'))->getNumberOfParameters() >= 4) {
+            $this->policy->checkSecurity($tags, $filters, $functions, $tests);
```
stof: shouldn't this catch the `SecurityNotAllowed*Error` exceptions (probably catching `SecurityError`) to set the source context in them ?

src/Sandbox/SecurityPolicy.php
```diff
@@ -87,6 +87,11 @@ public function setStrict(bool $strict): void
         $this->strict = $strict;
     }
 
+    public function isStrict(): bool
```
stof, line 93: should it be `@internal` ?

src/Sandbox/Sandbox.php
stof: btw, I think it might make sense to provide an official Symfony integration of the sandbox feature (either as an opt-in feature of `symfony/twig-bundle` or a new `twig/sandbox-bundle` in the twig extras). This would require thinking about how the registration of extensions should be managed when dealing with a separate environment in the container.

src/Sandbox/Sandbox.php
stof, line 47: The documented usage as `{{ sandbox.render() }}` in Twig by passing a `Sandbox` object in the context would break with auto-escaping. We should either return a Markup object from those methods or expose a native Twig API (from an opt-in extension taking the Sandbox as argument of its runtime) that could mark the output of functions as safe.

src/Sandbox/Sandbox.php
fabpot: Right, I've added an interface.

src/Sandbox/Sandbox.php
fabpot: Done in the interface

fabpot: I've added a proper function now.

src/Extension/SandboxBridgeExtension.php
```diff
@@ -0,0 +1,44 @@
...
+{
+    public function __construct(
+        private SandboxInterface $sandbox,
+    ) {
+    }
+
+    public function getFunctions(): array
+    {
+        return [
+            new TwigFunction('render_sandboxed', [$this, 'render'], ['is_safe_callback' => [RenderSandboxedFunction::class, 'getSafe'], 'node_class' => RenderSandboxedFunction::class]),
```
stof: This `render_sandboxed` function should have its own documentation page, like other functions.

src/Extension/CoreExtension.php
```diff
@@ -1499,16 +1500,20 @@ public static function testMapping($value): bool
      */
     public static function include(Environment $env, $context, $template, $variables = [], $withContext = true, $ignoreMissing = false, $sandboxed = false)
     {
+        if ($sandboxed) {
+            trigger_deprecation('twig/twig', '3.29', 'The "sandboxed" argument of the "include" function is deprecated, use "%s" to render untrusted templates instead.', Sandbox::class);
```
stof: this should rather mention the `render_sandboxed` Twig function as replacement for the Twig function.

src/Extension/SandboxBridgeExtension.php
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Exposes a sandbox to trusted templates.
+ *
+ * @author Fabien Potencier <fabien@symfony.com>
+ */
+final class SandboxBridgeExtension extends AbstractExtension
+{
+    public function __construct(
+        private SandboxInterface $sandbox,
```
stof: Should this use a runtime class instead, to allow lazy-loading the Sandbox service (and its associated Environment) ?

src/Extension/SandboxBridgeExtension.php
```diff
@@ -0,0 +1,44 @@
...
+    {
+        return [
+            new TwigFunction('render_sandboxed', [$this, 'render'], ['is_safe_callback' => [RenderSandboxedFunction::class, 'getSafe'], 'node_class' => RenderSandboxedFunction::class]),
+        ];
+    }
+
+    /**
+     * @param array<string, mixed> $context
+     */
+    public function render(string $name, array $context, string $outputStrategy): string
```
stof: what is this unused output strategy ?

src/Sandbox/SecurityChecker.php
stof: @fabpot what about this question ?

src/Extension/SandboxBridgeExtension.php
fabpot: Good idea

src/Extension/CoreExtension.php
fabpot: Right

src/Extension/SandboxBridgeExtension.php
fabpot: Done

src/Extension/SandboxBridgeExtension.php
fabpot: Removed

doc/functions/include.rst
```diff
@@ -72,12 +72,15 @@ inclusion. The first template that exists will be rendered:
...
+When including a template created by an end user, you should
+:doc:`sandbox<../sandbox>` it.
 
-.. code-block:: twig
+.. deprecated:: 3.29
 
-    {{ include('page.html.twig', sandboxed: true) }}
+    Sandboxing the included template via the ``sandboxed`` argument is
+    deprecated as of Twig 3.29. Render the untrusted template with the
+    ``Twig\Sandbox\Sandbox`` class from PHP or the ``render_sandboxed()``
```
stof: shouldn't `render_sandboxed` be a link to its doc ?

src/Runtime/SandboxRuntime.php
```diff
@@ -0,0 +1,31 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Twig\Runtime;
+
+use Twig\Extension\RuntimeExtensionInterface;
+use Twig\Sandbox\SandboxInterface;
+
+final class SandboxRuntime implements RuntimeExtensionInterface
```
stof: should it be named `SandboxBridgeRuntime` for consistency with the extension ?

doc/functions/include.rst
fabpot: Done

src/Runtime/SandboxRuntime.php
fabpot: Sure.

src/Sandbox/SecurityChecker.php
fabpot: Fixed.

src/Sandbox/SecurityChecker.php
```diff
@@ -0,0 +1,240 @@
...
+    {
+        $this->policy = $policy;
+    }
+
+    public function getSecurityPolicy(): SecurityPolicyInterface
+    {
+        return $this->policy;
+    }
+
+    public function checkSecurity(array $tags, array $filters, array $functions, array $tests, ?Source $source = null): void
```
stof: should the source be optional here ? Is there a case where we don't provide it ?

stof (approved):

tests/Extension/SandboxDeprecationsTest.php
```diff
@@ -69,4 +69,16 @@ public function testIncludeFunctionSandboxedArgumentIsDeprecated(): void
...
     }
+
+    public function testIncludeFunctionFalseSandboxedArgumentIsDeprecated(): void
+    {
+        $twig = new Environment(new ArrayLoader([
+            'index' => '{{ include("partial", sandboxed: false) }}',
+            'partial' => 'partial content',
+        ]), ['autoescape' => false]);
+
+        $this->expectDeprecation('Since twig/twig 3.29: The "sandboxed" argument of the "include" function is deprecated, use the "render_sandboxed" function to render untrusted templates instead.');
```
stof: When passing `false`, we should rather suggest removing the argument. Using `render_sandboxed` is not the right replacement in that case.

fabpot: I've change the message in case of false.

src/Sandbox/SecurityChecker.php
fabpot: SandboxExtension::checkSecurity() called by a third-party might send null here. Will change in 4.0 as this is deprecated. Removed `= null` for now.

PR #4857, opened 2026-07-08
Fix version in deprecation message for tag usage outside of root template

MrYamous:
#4292 has been introduced in 3.28 and related deprecations are shown for this version in changelog file

But in some part of code including deprecation message, it's about 3.27, this is confusing when upgrading Twig


fabpot (approved):

fabpot:
Thank you @MrYamous.

PR #4862, opened 2026-07-17
Update for.rst

marilenaRM:
Add context and explanation about loop.parent.

### DIFF
--- a/doc/tags/for.rst
+++ b/doc/tags/for.rst
@@ -80,6 +80,25 @@ Variable              Description
     ``loop.last`` variables are only available for PHP arrays, or objects that
     implement the ``Countable`` interface.
 
+When loops are nested, you can access the parent loop context via the
+``parent`` key of the ``loop`` variable:
+
+.. code-block:: twig
+
+    {% for topic, messages in topics %}
+       * {{ loop.index }} : {{ topic }}
+        {% for message in messages %}
+            - {{ loop.parent.loop.index }}.{{ loop.index }}: {{ message }}
+        {% endfor %}
+    {% endfor %}
+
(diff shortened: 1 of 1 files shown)


stof (approved):

doc/tags/for.rst
```diff
@@ -80,6 +80,25 @@ Variable              Description
     ``loop.last`` variables are only available for PHP arrays, or objects that
     implement the ``Countable`` interface.
 
+When loops are nested, you can access the parent loop context via the
+``parent`` key of the ``loop`` variable:
```
homersimpsons: For the record, this is not only on nested loop:
- (v3.x)
https://github.com/twigphp/Twig/blob/da8bd407000b6ef1adfce829ac9921f12b544617/src/Node/ForNode.php#L56
- (v4.x)
https://github.com/twigphp/Twig/blob/66cb19fa9850d9a147404fee4bfd7975ff4ba20e/src/Runtime/LoopContext.php#L34-L37

One use-case I found when documenting this in #4865 is about variable shadowing:
```twig
{% set users = [{"name": "Homer"}] %}
{% set user = {'name': 'Fabien'} %}
{% for user in users %}
    {{ loop.parent.user.name }} - {{ user.name }}
{% endfor %}
```

I do not have a proper wording to document this.

marilenaRM: Good point — loop.parent is indeed the whole enclosing context, not just nested loops. I've reworded the intro and added your variable-shadowing example. Thanks!

homersimpsons (approved):

fabpot (approved):

fabpot:
Thank you @marilenaRM.

PR #4864, opened 2026-07-18
:coffin: Drop php <8.0.11 gated code and tests

homersimpsons:
- Raised minimum version to PHP 8.0.11 as some code was gated behind this and I believe this is not an issue.
- Remove code gated behind version not allowed by composer.json
- Remove test and conditions gated behind version not allowed by composer.json
- Patch `src/Node/Expression/NameExpression.php` to remove useless `isset` call (was originally done for perf, it was improved in 5b5f5659a26412166d4f6ff785e6b2f586fa5bb8 but only for one branch). This was the original change I wanted to make.

Note that I noticed that there were a twig 4 branch with a bunch of this cleaned up, I'm not sure whether should I target 3.x or aim for 4.x directly?


fabpot:
This is the kind of cleanup we're not doing on 3.x anymore, but directly in 4.x. Thank you.

PR #4865, opened 2026-07-18
:memo: Add `loop.parent` example

homersimpsons:
The current wording does not give enough explanation to me. I had to find https://stackoverflow.com/questions/17534802/how-does-loop-parent-actually-work to understand its need.

I added an example with variable shadowing, should I add one with nested loops? such as:
```twig
{% for i in [1, 2, 3] %}
  {% for j in [1, 2, 3] %}
    {{ loop.parent.loop.index }} - {{ loop.index }}
  {% endfor %}
{% endfor %}
```

### DIFF
--- a/doc/tags/for.rst
+++ b/doc/tags/for.rst
@@ -74,6 +74,11 @@ Variable              Description
         {{ loop.index }} - {{ user.username }}
     {% endfor %}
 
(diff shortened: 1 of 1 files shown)


homersimpsons:
(this can be a duplicate of #4862)

fabpot:
Closing in favor of #4862

PR #4866, opened 2026-07-18
:coffin: Drop useless php comparisons

homersimpsons:


fabpot (approved):

fabpot:
Thank you @homersimpsons.

PR #4870, opened 2026-07-23
Add a docs option to the types and block tags

### LINKED ISSUE #4768: Standardize Twig comment annotations for documenting template variables and blocks

Amoifr:
Implements the design proposed by @fabpot in #4768:

- each `{% types %}` entry accepts an optional `docs="..."` option after the type string; it is stored in the `TypesNode` mapping (`docs` is `null` when not provided)
- the `{% block %}` tag accepts an optional `docs="..."` option after the block name; it is stored as a `docs` attribute on the `BlockNode` (new optional constructor argument, BC)

The tags themselves do not use this documentation: it is metadata for tools (IDEs, documentation generators, the Symfony UX Toolkit) that analyze the parsed nodes.

Notes:

- For `{% block %}`, `docs` is only treated as an option when followed by `=`, so the shortcut syntax `{% block title docs %}` (printing a variable named `docs`) keeps working; this is covered by a test.
- I kept `docs` as originally proposed; renaming it to `desc` is a one-line change if preferred.

Closes #4768

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
 # 3.29.0 (2026-XX-XX)
 
+ * Add a `docs` option to the `types` tag entries and to the `block` tag to document template variables and blocks
  * Fix `IntlExtension` ignoring explicit date/time formats and the `format_date`/`format_time` filters when a date formatter prototype is configured
  * Add a `format_list` filter to `IntlExtension` to format a list of strings using PHP 8.5's `IntlListFormatter`
  * Fix array access with a `Stringable` key coercing the key to string for `ArrayAccess` objects that use object keys (such as `SplObjectStorage`)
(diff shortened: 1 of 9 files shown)


willrowe:
I like `desc` better.

Kocal:
I forgot to close the issue, since I closed the initial PR  https://github.com/twigphp/Twig/pull/4770. We went for https://github.com/symfony/ux/pull/3694 in UX, and also in the PhpStorm plugin https://github.com/Haehnchen/idea-php-symfony2-plugin/pull/2826

fabpot:
See https://github.com/twigphp/Twig/pull/4871

fabpot:
Closing in favor of #4871
Thanks for the work you've done here, it helped me think about BC and I found another path without.

PR #4871, opened 2026-07-25
Attach documentation comments to nodes

### LINKED ISSUE #4768: Standardize Twig comment annotations for documenting template variables and blocks

fabpot:
Alternatives to #4870

To avoid BC breaks, I have another idea, using `##` as a new syntax, a bit like `/** */` in PHP vs `/* */`.

"Documentation" is attached as metadata to the next relevant node:

```twig
{## The main content displayed on the page #}
{% block content %}
    ...
{% endblock %}
```

Documentation comments can also describe variables declared with the `types` tag:

```twig
{% types {
    ## The unique identifier of the article
    id: 'string',

    ## Whether the article should be highlighted
    featured?: 'boolean',
} %}
```

Node visitors can access this metadata through `Node::getDocumentation()`, allowing IDEs, static analyzers, and documentation generators to consume it without affecting template rendering. Documentation is preserved when visitors or optimizations replace nodes.

Closes #4768
Closes #4870

### DIFF
--- a/src/Lexer.php
+++ b/src/Lexer.php
@@ -37,6 +37,8 @@ class Lexer
     private $position;
     private $positions;
     private $currentVarBlockLine;
+    /** @var list<string> */
+    private array $documentation = [];
     private array $openingBrackets = ['{', '(', '['];
     private array $closingBrackets = ['}', ')', ']'];
 
@@ -149,6 +151,21 @@ class Lexer
                 ')
             }sx',
 
+            // #} or ##}
+            'lex_documentation_comment' => '{
+                (?:'.
+                    preg_quote($this->options['whitespace_trim'].'#'.$this->options['tag_comment'][1], '#').'\s*\n?'. // -##}\s*\n?
+                    '|'.
+                    preg_quote($this->options['whitespace_line_trim'].'#'.$this->options['tag_comment'][1], '#').'['.$this->options['whitespace_line_chars'].']*'. // ~##}[ \t\0\x0B]*
+                    '|'.
+                    preg_quote($this->options['whitespace_trim'].$this->options['tag_comment'][1], '#').'\s*\n?'. // -#}\s*\n?
+                    '|'.
+                    preg_quote($this->options['whitespace_line_trim'].$this->options['tag_comment'][1], '#').'['.$this->options['whitespace_line_chars'].']*'. // ~#}[ \t\0\x0B]*
+                    '|'.
+                    preg_quote($this->options['tag_comment'][1], '#').'(?:\r\n?|\n)?'. // #}(?:\r\n?|\n)?
+                ')
+            }sx',
+
             // verbatim %}
             'lex_block_raw' => '{
                 \s*verbatim\s*
@@ -169,6 +186,8 @@ class Lexer
                     preg_quote($this->options['tag_variable'][0], '#'). // {{
                     '|'.
                     preg_quote($this->options['tag_block'][0], '#'). // {%
+                    '|'.
+                    preg_quote($this->options['tag_comment'][0].'#', '#'). // {##
                     '|'.
                     preg_quote($this->options['tag_comment'][0], '#'). // {#
                 ')('.
@@ -198,6 +217,7 @@ class Lexer
         $this->states = [];
         $this->brackets = [];
         $this->position = -1;
+        $this->documentation = [];
 
         // find all token starts in one go
         preg_match_all($this->regexes['lex_tokens_start'], $this->code, $matches, \PREG_OFFSET_CAPTURE);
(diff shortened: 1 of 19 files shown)


Kocal:
Will it works for `{% props %}` tag from UX TwigComponent?

Also, I forgot to close https://github.com/twigphp/Twig/issues/4768, but it was already resolved with our `@prop` and `@block` annotations (which are also properly highlighted in PHPStorm, thanks to the Symfony plugin). 

Since my initial issue is closed, do we really need this new feature here? 

It would requires UX to change again, and the Symfony PHPStorm plugin should be modified as well.

fabpot:
> Will it works for `{% props %}` tag from UX TwigComponent?
> 
> Also, I forgot to close #4768, but it was already resolved with our `@prop` and `@block` annotations (which are also properly highlighted in PHPStorm, thanks to the Symfony plugin).
> 
> Since my initial issue is closed, do we really need this new feature here?
> 
> It would requires UX to change again, and the Symfony PHPStorm plugin should be modified as well.

This is generic, so it should work just fine.

Kocal:
Cool, thanks!

When merged/released, I will work on UX & Symfony PHPStorm plugin to re-align things with this PR.

Kocal:
Opened https://github.com/Haehnchen/idea-php-symfony2-plugin/pull/2836, UX PR will follows when this one get merged/released.

stof:
@fabpot is it expected that this PR contains commits from https://github.com/twigphp/Twig/pull/4851 ?

src/Node/TypeNode.php
```diff
@@ -0,0 +1,28 @@
...
+
+use Twig\Attribute\YieldReady;
+
+/**
+ * Represents a type declaration node.
+ *
+ * @author Fabien Potencier <fabien@symfony.com>
+ */
+#[YieldReady]
+final class TypeNode extends Node
```
stof, line 22: should this class (or at least its constructor) be `@internal` if it is only meant to be instantiated by TypesNode ?

src/Node/NodeDocumentation.php
```diff
@@ -0,0 +1,54 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Twig\Node;
+
+use Twig\Token;
+
+/**
+ * @internal
```
stof: should this be `@internal` ? Looking at the implementation of token parsers, it looks like custom token parsers provided by extensions might need to use it as well if they want to support documentation on their nodes.

src/Lexer.php
```diff
@@ -478,7 +486,17 @@ private function lexComment(): void
             throw new SyntaxError('Unclosed comment.', $this->lineno, $this->source);
         }
 
-        $this->moveCursor(substr($this->code, $this->cursor, $match[0][1] - $this->cursor).$match[0][0]);
+        $comment = substr($this->code, $this->cursor, $match[0][1] - $this->cursor);
+        if (str_starts_with($comment, '#')) {
```
stof: how does the `##` marker combine with whitespace control markers ?

fabpot:
> @fabpot is it expected that this PR contains commits from #4851 ?

Yes, it was needed, but I've just rebased now that the PR has been merged, it's going to make reviewing easier.

src/Node/TypeNode.php
fabpot: Done

src/Node/NodeDocumentation.php
fabpot: Done

src/Lexer.php
fabpot: Fixed, it was `{#-#` which was ugly, now it's `{##-` which is what I would expect to work. Same for ~.

src/NodeVisitor/DocumentationNodeVisitor.php
```diff
@@ -0,0 +1,75 @@
...
+            $index = array_key_last($this->macroDeclarations);
+            if ($node->getTemplateLine() === $macro->getTemplateLine()) {
+                if (isset($this->macroDeclarations[$index][$node->getAttribute('name')])) {
+                    $this->macroDeclarations[$index][$node->getAttribute('name')]->setDocumentation(null);
+                }
+                $this->macroDeclarations[$index][$node->getAttribute('name')] = $node;
+            } else {
+                $node->setDocumentation(null);
+            }
+        }
```
Kocal: Can we put `$node->getAttribute('name')` in a variable? it's repeated 7 times here

stof: Not for all of them. We cannot read it before the instanceof checks, otherwise we would not be sure that the attribute exists.

fabpot: Done

Kocal:
Let's merge it? :D

fabpot:
@Kocal For your information, full support has been added (including types tags) in the Symfony LSP server in 0.12.0.

Kocal:
Awesome :D

PR #4873, opened 2026-07-27
Deprecate duplicate macro definitions

fabpot:


src/Parser.php
```diff
@@ -303,6 +303,10 @@ public function hasMacro(string $name): bool
 
     public function setMacro(string $name, MacroNode $node): void
     {
+        if (isset($this->macros[$name])) {
+            trigger_deprecation('twig/twig', '3.29', 'Defining the macro "%s" more than once in "%s" is deprecated and will throw a SyntaxError in Twig 4.0 (first definition at line %d, second at line %d).', $name, $this->stream->getSourceContext()->getName(), $this->macros[$name]->getTemplateLine(), $node->getTemplateLine());
```
stof, line 307: should this mention which one is actually used in Twig 3.x to make it easier to fix this ? I think the last one wins currently.

fabpot: Done in https://github.com/twigphp/Twig/pull/4883

PR #4877, opened 2026-07-28
Deprecate macro calls without parentheses

fabpot:


src/Node/Expression/MacroReferenceExpression.php
```diff
@@ -51,6 +53,11 @@ public function __construct(MacroVariable $template, string|AbstractExpression $
         parent::__construct($nodes, $attributes, $lineno);
     }
 
+    public function setHasParentheses(bool $hasParentheses): void
```
stof, line 59: should this be `@internal` ?

src/Node/Expression/MacroReferenceExpression.php
```diff
@@ -63,6 +70,10 @@ public function __clone()
 
     public function compile(Compiler $compiler): void
     {
+        if (!$this->hasParentheses && !$this->definedTest) {
+            trigger_deprecation('twig/twig', '3.29', 'Omitting parentheses when calling a macro is deprecated and will throw a SyntaxError in Twig 4.0; add parentheses after the macro name.');
```
stof: this should mention the location of the node to make it easier to fix the deprecation.

src/Node/Expression/MacroReferenceExpression.php
fabpot: Makes sense

src/Node/Expression/MacroReferenceExpression.php
fabpot: done

stof (approved):

PR #4892, opened 2026-08-21
Register the missing extra callables in MissingExtensionSuggestor

smnandre:
`Extensions.php` mirrors what each `twig/*-extra` package declares, so `MissingExtensionSuggestor` can answer an unknown name with `try running "composer require twig/string-extra"` rather than a bare `Unknown "slug" filter`.

This PR adds the twelve callables missing from it:
* `html-extra`: `html_attr_merge` and `html_attr_type` filters, `html_cva` and  `html_attr` functions
* `intl-extra`: `format_list` filter, `language_names`, `script_names`, `country_names`, `locale_names`, `currency_names` and `timezone_names` functions
* `string-extra`: `slug`, `plural` and `singular` filters

`ExtensionsTest` checks the catalog both ways: 
* every listed name resolves on an `Environment` holding the extension, 
* every declared filter, function and tag is listed


fabpot:
Thank you @smnandre.

PR #4895, opened 2026-08-26
Extract `htmlAttrValue()` from `html_attr` for standalone attribute rendering

Kocal:
The per-value resolution behind `html_attr` is extracted into a new public `HtmlExtension::htmlAttrValue()`, returning the unescaped value or `null` to omit the attribute; `html_attr()` now delegates to it, output unchanged, existing tests untouched. This lets third parties render a single attribute exactly like `html_attr` without a Twig `Environment`, since the resolution is escaper-free. symfony/ux#3820 and symfony/ux#3821 depend on this PR.

The `data-*` branch only tested `is_scalar()`, so a `\Stringable` was JSON-encoded instead of using its string representation; the same object already rendered its string form in `title` or `class`, and `AttributeValueInterface` was already excluded from that branch.

| Value in `data-value` | Before | After |
| --- | --- | --- |
| a `\Stringable` | `data-value="{}"` | `data-value="hello"` |
| a `\Stringable` that is also `JsonSerializable` | `data-value="&quot;01JABC&quot;"` | `data-value="01JABC"` |

### DIFF
--- a/doc/functions/html_attr.rst
+++ b/doc/functions/html_attr.rst
@@ -123,8 +123,11 @@ attributes are converted to strings ``"true"`` and ``"false"``.
 Data Attributes
 ---------------
 
-For ``data-*`` attributes, boolean ``true`` values will be converted to ``"true"``.
-Values that are not scalars are automatically JSON-encoded.
+For ``data-*`` attributes, a boolean ``true`` is converted to the string
+``"true"``, and any non-scalar value is JSON-encoded. Two exceptions behave as
+they do for any other attribute: an iterable is rendered as a token list, and
+a ``Stringable`` object is cast to its string representation. When an object
+is both, the iterable behavior wins.
 
 .. code-block:: html+twig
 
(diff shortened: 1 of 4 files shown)


stof (approved):

fabpot (commented):
A few comments:

* The documented `data-*` contract still says that non-scalar values are JSON encoded
* The tests do not cover the `Stringable` plus `JsonSerializable`. If the new precedence is intentional, could you update `doc/functions/html_attr.rst` and add a regression case?

Kocal:
@fabpot Oops you're right, I just pushed some modifications to cover what you reported.

doc/functions/html_attr.rst
```diff
@@ -124,7 +124,9 @@ Data Attributes
 ---------------
 
 For ``data-*`` attributes, boolean ``true`` values will be converted to ``"true"``.
-Values that are not scalars are automatically JSON-encoded.
+Values that are not scalars are automatically JSON-encoded, except objects
```
fabpot: Is it true for objects implementing both `Stringable` and `Traversable`? The iterable branch renders a token list instead of using `__toString()`.

Kocal: It should be more correct now, thx!

fabpot:
Thank you @Kocal.

PR #4910, opened 2026-08-28
Reject template wrappers from another environment

fabpot:
Twig now rejects `TemplateWrapper` instances created by another `Environment`.

This prevents templates from unexpectedly using another environment's loader, extensions, globals, or sandbox policy.

### DIFF
--- a/src/TemplateWrapper.php
+++ b/src/TemplateWrapper.php
@@ -11,6 +11,8 @@
 
 namespace Twig;
 
+use Twig\Error\RuntimeError;
+
 /**
  * Exposes a template to userland.
  *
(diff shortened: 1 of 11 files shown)


stof:
This will break the Symfony form renderer, which relies on the internal API: https://github.com/symfony/symfony/blob/f1072d8902e3d397ab5f9190877fb947d259b0c3/src/Symfony/Bridge/Twig/Form/TwigRendererEngine.php#L142

It might make sense to first refactor Symfony to stop using the internal API (which will be safer regarding semver guarantees)

src/TemplateWrapper.php
```diff
@@ -96,11 +98,21 @@ public function getTemplateName(): string
...
-    public function unwrap()
+    public function isOwnedBy(Environment $env): bool
+    {
+        return $this->env === $env && $this->template->isOwnedBy($env);
+    }
+
+    /**
+     * @internal
+     */
+    public function unwrap(Environment $env): Template
```
stof, line 110: This will still break existing Symfony releases that don't use the new BlockChain API yet but use the internal API instead.
Do you plan to migrate all Symfony branches to the BlockChain API ?

fabpot: See https://github.com/symfony/symfony/pull/66043

stof: Note that this means we should ensure that Symfony releases this update before the Twig release introducing this signature change.

PR #4915, opened 2026-09-01
Fix the default filter fallback reusing a null-safe temporary variable

### LINKED ISSUE #4914: Undefined variable $_vN when chaining the null-safe operator (?.) with the default() filter and the object is null
### Description

When a null-safe (`?.`) chain is used as both the primary expression and inside the fallback argument of the `default()` filter, and the underlying variable is actually `null` at runtime, the compiled template references a temporary variable (`$_v1`, `$_v2`, ...) that was only ever assigned in a branch of the ternary that never executes. This produces a PHP `Undefined variable $_vN` warning (and, depending on `error_reporting`/logging config, an application-level error).

This looks related to the null-safe short-circuiting work in #4748 / #4869 and to the `default()` filter's argument-evaluation strategy discussed in #4668 / #4885, but I could not find an existing issue for this specific combination.

### How to reproduce

```php
<?php
require 'vendor/autoload.php';

$loader = new \Twig\Loader\ArrayLoader([
    'bug.twig' => "{{ item?.label|default(item?.name) }}\n",
]);
$twig = new \Twig\Environment($loader, ['debug' => true, 'cache' => false]);

echo $twig->render('bug.twig', ['item' => null]);
```

Output (PHP 8.4.24, Twig 3.28.0):

```
PHP Warning:  Undefined variable $_v2 in .../Environment.php(420) : eval()'d code on line 43
```

Relevant excerpt of the generated code (`item?.label|default(item?.name)`):

```php
(((null === ($_v0 = ($context["item"] ?? null))) ? null : CoreExtension::getAttribute($this->env, $this->source, $_v0, "label", [], "any", true, true, false, 1))
    ? (Twig\Extension\CoreExtension::default(
        ((null === ($_v1 = ($context["item"] ?? null))) ? null : CoreExtension::getAttribute($this->env, $this->source, $_v1, "label", [], "any", false, false, false, 1)),
        ((null === ($_v2 = ($context["item"] ?? null))) ? null : CoreExtension::getAttribute($this->env, $this->source, $_v2, "name", [], "any", false, false, false, 1))
    ))
    : (CoreExtension::getAttribute($this->env, $this->source, $_v2, "name", [], "any", false, false, false, 1))
)
```

`$_v2` is only ever assigned inside the **true** branch of the outer ternary (as a side effect of evaluating the second argument to `default()`). The **false** branch — taken exactly when `item` is `null`, i.e. exactly when the null-safe chain is supposed to matter — reuses `$_v2` assuming it was already set. It wasn't, so PHP emits `Undefined variable $_v2`.

### Expected behavior

No PHP warning; the whole expression should evaluate to `null`/empty just like a plain `item?.name` would when `item` is null.

### Actual behavior

`Undefined variable $_vN` warning from the compiled template, surfacing as an application error depending on error/log handling (this is how we found it: Symfony logs warnings as errors in dev, and it breaks tests/CI in strict mode).

### Real-world case

This was found in `isotope/isotope-core`'s `content_element/iso_orderhistory.html.twig`, in the exact pattern:

```twig
{{ collection.deliveryStatus?.label|default(collection.deliveryStatus?.name)|trans([], 'isotope') }}
```

which crashes for every order collection where `deliveryStatus` (or `transactionStatus`) hasn't been set yet.

### Workaround

Resolve the chain into a local variable first so the null-safe access only appears once:

```twig
{% set deliveryStatus = collection.deliveryStatus %}
{{ (deliveryStatus ? (deliveryStatus.label|default(deliveryStatus.name)) : '')|trans([], 'isotope') }}
```

### Environment

- Twig version: 3.28.0
- PHP version: 8.4.24
- Twig `debug`: both true and false reproduce it (cache disabled in the repro above for clarity)

lazerg:
The `default` filter reuses the fallback argument node in both branches of the conditional it compiles to. Because `GetAttrExpression` records the temporary variable it allocated for a null-safe chain on the node itself, the second compilation skipped the assignment and emitted a bare `$_vN` reference, so `{{ item?.label|default(item?.name) }}` warned about an undefined variable whenever `item` was null.

The fallback node is now cloned, like the node used for the defined test already is, so each branch compiles its own temporary.

Fixes #4914

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
 # 3.29.0 (2026-XX-XX)
 
+ * Fix the `default` filter fallback emitting an undefined variable warning when it uses the null-safe operator
  * Fix the `matches` operator silently treating PCRE execution errors as non-matches
  * Add the `HtmlExtension::htmlAttrValue()` method to resolve a single HTML attribute value the way the `html_attr` function renders it
  * Fix `html_attr` JSON encoding a `Stringable` value in a `data-*` attribute instead of using its string representation
--- a/src/Node/Expression/Filter/DefaultFilter.php
+++ b/src/Node/Expression/Filter/DefaultFilter.php
@@ -55,7 +55,7 @@ class DefaultFilter extends FilterExpression
 
         if ('default' === $name && ($node instanceof ContextVariable || $node instanceof GetAttrExpression)) {
             $test = new DefinedTest(clone $node, new TwigTest('defined', null, ['always_allowed_in_sandbox' => true]), new EmptyNode(), $node->getTemplateLine());
-            $false = \count($arguments) ? $arguments->getNode('0') : new ConstantExpression('', $node->getTemplateLine());
+            $false = \count($arguments) ? clone $arguments->getNode('0') : new ConstantExpression('', $node->getTemplateLine());
 
             $node = new ConditionalTernary($test, $default, $false, $node->getTemplateLine());
         } else {
--- a/tests/ExpressionParserTest.php
+++ b/tests/ExpressionParserTest.php
@@ -445,6 +445,27 @@ class ExpressionParserTest extends TestCase
                 ['foo' => (object) ['bar' => (object) ['baz' => null]]],
                 '',
             ],
+            // default filter fallback
+            [
+                '{{ foo?.bar|default(foo?.baz) }}',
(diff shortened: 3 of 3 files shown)


fabpot (approved):

fabpot:
Thank you @lazerg.

Issue #3636, opened 2022-01-23
Cache invalidateTags issue with twig/extra-bundle and Redis


PanzerLlama:
I noticed a problem when using the {% cache "cache_name" tag('tag_name') %}.

This creates an entry in Redis in a form of:
`"...:tag_name\x00tags\x00"`

while "Adapter/AbstractTagAwareAdapter.php" `invalidateTags()` seems to look for`"...:\x00tags\x00tag_name"` so calling that method always fails to invalidate the tagged entries.

The "Adapter/TagAwareAdapter.php" uses the key name as `$tag.static::TAGS_PREFIX` in InvalidateTags while the other `self::TAGS_PREFIX.$tag.`

How to replicate:
1) Create a template with `{% cache "cache_name" tag('tag_name') %}`
2) Call `RedisTagAwareAdapter->invalidateTags(['tag_name']);`

No tagged entries get invalidated.

        "symfony/framework-bundle": "5.4.*",
        "twig/cache-extra": "^3.3",
        "twig/extra-bundle": "^3.3",
        "twig/twig": "^3.0"

edouardgat:
Have the same issue.
@PanzerLlama I solved it by using this config in `services.yaml`, most specifically `.twig.cache.inner` and `twig.cache`

```
  Symfony\Component\Cache\Adapter\RedisAdapter:
    arguments: ['@snc_redis.default']

  Symfony\Component\Cache\Adapter\RedisTagAwareAdapter:
    arguments: ['@snc_redis.default']

  Symfony\Component\HttpFoundation\Session\Storage\Handler\RedisSessionHandler:
    arguments: ['@snc_redis.session']

  .twig.cache.inner:
    class: Symfony\Component\Cache\Adapter\RedisTagAwareAdapter
    arguments: ['@snc_redis.default']

  twig.cache:
    class: Symfony\Component\Cache\Adapter\RedisTagAwareAdapter
    arguments: ['@snc_redis.default']
```

For the Arguments of RedisTagAwareAdapter, you choose the one you use. `snc_redis.default` is my Redis client :
```
snc_redis:
  clients:
    default:
      type: phpredis
      alias: default
      dsn: '%env(REDIS_HOST_URL)%/%env(int:REDIS_DB_DEFAULT)%'
      logging: '%kernel.debug%'
```

fabpot:
Looks like something related to Symfony, not Twig.

PR #4923, opened 2026-09-11
Fix wrapping the Twig cache pool in a second tag aware adapter

nicolas-grekas:
`twig/extra-bundle` wires the `{% cache %}` pool like this:

```php
->set('twig.cache', TagAwareAdapter::class)
    ->args([$service('.twig.cache.inner')])

->set('.twig.cache.inner')
    ->parent('cache.app')
    ->tag('cache.pool', ['name' => 'twig.cache'])
```

`.twig.cache.inner` is a child of `cache.app`, so when an application configures
`framework.cache.app` with a natively tag aware adapter (`cache.adapter.redis_tag_aware`,
or the valkey, pdo and mongodb variants), the child pool is *already* a
`TagAwareAdapterInterface` and `twig.cache` wraps it in a second `TagAwareAdapter`.

Two nested `TagAwareAdapter`s never read back what they wrote:

```php
$pool = new TagAwareAdapter(new TagAwareAdapter(new ArrayAdapter()));

$item = $pool->getItem('k');
$item->set('value')->tag('t1');
$pool->save($item);

var_dump($pool->getItem('k')->isHit()); // bool(false)
```

So every `{% cache %}` block silently misses, on every request. This was reported here as
twigphp/Twig#3636 (the workaround in that thread is to redefine `twig.cache` and
`.twig.cache.inner` by hand as a `RedisTagAwareAdapter`), and again on the Symfony side as
symfony/symfony#54339 by @rpkamp. Symfony first tried to fix it in symfony/symfony#57927 by
deprecating passing a tag aware pool to `TagAwareAdapter`, but @keulinho's analysis in
symfony/symfony#58830 showed that the deprecated thing was not the culprit, so that
deprecation was reverted in symfony/symfony#58950 and the defect was left where it is: in
this bundle. Symfony's own cache wiring already gets this right, it aliases
`cache.app.taggable` to `cache.app` when the configured adapter is natively tag aware
instead of decorating it.

## The fix

A small compiler pass resolves the parent chain of `.twig.cache.inner` and, when the
adapter it ends up on implements `TagAwareAdapterInterface`, drops the decorator and
aliases `twig.cache` to the pool. Nothing changes for the common case of a plain
`cache.app`, where the decorator is what makes the pool taggable and stays in place.

I picked this over declaring the pool through `framework.cache.pools` (with `tags: true`,
which would let Symfony make the same decision) because the pool is only registered when
the `cache` extension is enabled, and reproducing that condition inside a `prepend()` call
means processing the bundle's own configuration before `load()` runs. The compiler pass
stays entirely inside the bundle and only acts when `twig.cache` exists.

Detecting tag awareness from the resolved adapter class rather than from a list of adapter
ids keeps this working across `^5.4|^6.4|^7.0|^8.0` even though the set of natively tag
aware adapters grew over that range, and it also covers a `cache.app` overridden with a
custom tag aware adapter.

Two constraints are covered by the tests: `twig.cache` keeps its own namespace instead of
collapsing into the application pool, and the three argument aliases
(`TagAwareCacheInterface $twigCache`, `CacheInterface $twigCache`,
`CacheItemPoolInterface $twigCache`) keep resolving.

## Verification

Compiled containers with `framework.cache.app` left at its default and set to
`cache.adapter.redis_tag_aware`, before and after the patch:

| `framework.cache.app` | before | after |
| --- | --- | --- |
| default (filesystem) | `TagAwareAdapter(FilesystemAdapter)` | unchanged |
| `cache.adapter.redis_tag_aware` | `TagAwareAdapter(RedisTagAwareAdapter)` | `RedisTagAwareAdapter` |
| `cache.adapter.valkey_tag_aware` | `TagAwareAdapter(RedisTagAwareAdapter)` | `RedisTagAwareAdapter` |
| `cache.adapter.redis` | `TagAwareAdapter(RedisAdapter)` | unchanged |

In the fixed tag aware case the pool namespace stays distinct from `cache.app`'s.

`extra/twig-extra-bundle` test suite, on Symfony 8.2-dev and on framework-bundle 6.4:

```
PHPUnit 9.6.36 by Sebastian Bergmann and contributors.

Testing
...................                                               19 / 19 (100%)

Time: 00:00.534, Memory: 28.00 MB

OK (19 tests, 100 assertions)
```

The two new tests fail without the pass:

```
1) TwigCachePoolPassTest::testThePoolIsNotDecoratedWhenTheAppAdapterIsTagAware
-'@.twig.cache.inner'
+'Symfony\Component\Cache\Adapter\TagAwareAdapter'
```

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
 # 3.29.0 (2026-XX-XX)
 
+ * Fix `{% cache %}` always missing in the Symfony bundle when `framework.cache.app` uses a natively tag aware adapter
  * Fix the sandbox resolving `use` trait templates before checking that the `use` tag is allowed
  * Fix `html_attr` dropping `style` declarations whose value is `0`, `0.0` or `'0'`
  * Fix the `default` filter fallback emitting an undefined variable warning when it uses the null-safe operator
--- a/extra/twig-extra-bundle/DependencyInjection/Compiler/TwigCachePoolPass.php
+++ b/extra/twig-extra-bundle/DependencyInjection/Compiler/TwigCachePoolPass.php
@@ -1,0 +1,57 @@
+<?php
+
+/*
+ * This file is part of Twig.
+ *
+ * (c) Fabien Potencier
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Twig\Extra\TwigExtraBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\Cache\Adapter\TagAwareAdapterInterface;
+use Symfony\Component\DependencyInjection\ChildDefinition;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+/**
+ * Drops the "twig.cache" TagAwareAdapter when the pool it decorates is already tag aware.
+ *
+ * Wrapping a tag aware adapter in a TagAwareAdapter makes every read a miss, so the
+ * decorator must go when "cache.app" uses a natively tag aware adapter, the same way
+ * Symfony aliases "cache.app.taggable" to "cache.app" instead of decorating it.
+ *
+ * @internal
+ */
+final class TwigCachePoolPass implements CompilerPassInterface
+{
+    public function process(ContainerBuilder $container): void
+    {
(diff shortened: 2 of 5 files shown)


stof (approved):

fabpot:
Thank you @nicolas-grekas.

rpkamp:
Nice solution @nicolas-grekas, thank you :slightly_smiling_face:

PR #4924, opened 2026-09-11
Resolve constant parent templates once instead of on every lookup

fabpot:
Currently, `Template::$parent` is only memoized by the compiled `doDisplay()`. Anything that reaches a template through `getParent()` without rendering it (`TemplateWrapper::hasBlock()`, `getBlockNames()`, `renderBlock()`, `yieldParentBlock()`, `MacroNamespace::getParent()`) re-evaluates `doGetParent()` and re-runs the sandbox check on every call, even when the parent is a string literal.

`doGetParent()` now does `$this->parent ??= $this->load("name", $line)` for a constant parent, which is what `doDisplay()` already does one line later. Dynamic parents are untouched and still resolve per context.

This also fixes a bug: `getParent()` loaded a constant parent with line `-1`, so a missing parent reported through `hasBlock()`/`getBlockNames()` lost its line number, while `render()` reported the `{% extends %}` line. All three now agree.

### DIFF
--- a/src/Node/ModuleNode.php
+++ b/src/Node/ModuleNode.php
@@ -143,16 +143,17 @@ final class ModuleNode extends Node implements CoercesChildrenToStringInterface
         ;
 
         if ($parent instanceof ConstantExpression) {
-            $compiler->subcompile($parent);
-        } else {
-            $compiler
-                ->raw('$this->load(')
-                ->subcompile($parent)
-                ->raw(', ')
-                ->repr($parent->getTemplateLine())
-                ->raw(')')
-            ;
+            // a constant parent never depends on the context, so resolve it once
+            $compiler->raw('$this->parent ??= ');
         }
+
+        $compiler
+            ->raw('$this->load(')
+            ->subcompile($parent)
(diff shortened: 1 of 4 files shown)


stof (approved):

src/Node/ModuleNode.php
```diff
@@ -143,17 +143,18 @@ protected function compileGetParent(Compiler $compiler): void
...
-        } else {
-            $compiler
-                ->raw('$this->load(')
-                ->subcompile($parent)
-                ->raw(', ')
-                ->repr($parent->getTemplateLine())
-                ->raw(')')
-            ;
+            // a constant parent never depends on the context, so resolve it once
+            $compiler->raw('$this->parent ??= ');
```
GromNaN, line 147: This property should be defined in the template class?

GromNaN: It is: https://github.com/twigphp/Twig/blob/4de6bc3b06e20b7175cda2a6c6046578d434e981/src/Template.php#L35

PR #4917, opened 2026-09-05
Template runtime and block composition

fabpot:
This PR addresses the Symfony compatibility break from #4910

It introduces runtime composition of templates used as collections of named block renderers: The renderer provides an ordered set of unrelated templates. The first matching block wins, nested `block()` calls see the complete composed set, and `parent()` remains within the block’s own inheritance or `use` hierarchy.

This feature is going to be useful for more than just Symfony.

### Ibexa Core

**Project:** `ibexa/core`  
**Feature:** CMS field rendering through `FieldBlockRenderer`

Ibexa maintains prioritized field templates, selects blocks such as `ibexa_string_field`, walks parent templates, constructs a block map and passes it to `renderBlock()`.

This is the strongest independent fit for `BlockChain`:

```php
$blocks = new BlockChain($twig, [
    $localTemplate,
    ...$projectFieldThemes,
    ...$vendorFieldThemes,
]);

return $blocks->renderBlock($fieldType.'_field', $context);
```

### Data-grid and listing renderers

The audit found the same broad mechanism in:

- `Prezent/prezent-grid`, `src/Twig/GridRenderer.php`
- `pawellen/listing`, `Renderer/ListingRenderer.php`
- `Braunstetter/data-grid-bundle`, `src/GridRendererEngine.php`
- `AnoDataGrid`, `DataGridExtension.php`

Their common feature is **layered grid themes**:

1. Configure default grid templates.
2. Add per-grid or per-view overrides.
3. Map a column type to a block name.
4. Walk template inheritance.
5. Merge or cache available blocks.
6. Render the selected cell, header or filter block.

Several accessed `unwrap()`, `getBlocks()` or `getParent()` directly; others passed manually assembled block maps into `renderBlock()` or `displayBlock()`.

## Adjacent use cases

The audit also found block-library patterns that could benefit if they grow into multi-template composition:

- **iTop:** plugin-contributed login blocks such as `login_input`, `login_submit`, `login_form_footer` and `login_links`; independently renders `body`, `script`, `ready_script` and `css`.
- **Email renderers:** independently render `subject`, `body_text` and `body_html` blocks.
- **Runtime theme overlays:** tenant branding, application skins, email themes, reports and configurable admin interfaces.
- **Extension-provided block libraries:** enabled modules contribute blocks such as `toolbar`, `field_text`, `dashboard_metric` or `login_footer`.
- **Testing and preview tooling:** render a block against an exact theme stack without generating a synthetic host template.

## Important negative finding

[...]

PR #4925, opened 2026-09-11
Resolve block chains against the render context

fabpot:
While working on the Symfony PR for #4917, I realize that the performance was worse with the Twig's way. #4924 fixes part of the performance "regression". This one closes the gap.

`BlockChain` currently freezes each template's lineage by cloning it, so the block map and `parent()` both resolve `{% extends %}` against the constructor context.

This PR changes that to resolve the chain against the render context instead. `hasBlock()` and `getBlockNames()` take a context, like `TemplateWrapper` already does, and the third constructor argument becomes a default the render context can override. A lineage whose templates all have a fixed parent (none, or constant) is resolved once and cached, so the common case costs nothing; a dynamic `{% extends %}` re-resolves per call (not use by Symfony anyway).

This fixes also an inconsistency: a chained template with a dynamic parent now behaves exactly as it does when rendered directly. It also removes all changes in the "core" logic of Template.

```
                       construct    render   1 chain + 40 renders
Symfony's engine today   13.44us    0.78us         50.3us
#4917                    ~70us      0.96us        ~117us
this PR                   0.61us    0.84us         53.0us
```


src/BlockChain.php
```diff
@@ -18,21 +18,24 @@
...
+
+    /** @var array<string, array<string, array{Template, string}>> */
+    private array $blocks = [];
+
+    /** @var array<string, array{Template, string}>|null */
+    private ?array $fixedBlocks = null;
 
     /**
      * @param iterable<string|TemplateWrapper> $templates Templates ordered from highest to lowest precedence
+     * @param array                            $context   Default variables used to resolve dynamic parent expressions
```
stof, line 32:
```suggestion
     * @param array<string, mixed>             $context   Default variables used to resolve dynamic parent expressions
```

src/BlockChain.php
```diff
@@ -41,67 +44,127 @@ public function __construct(
...
-        if (!isset($this->template)) {
+        if (!$this->templates) {
             throw new \InvalidArgumentException('A block chain requires at least one template.');
         }
-
-        $this->blocks = $blocks;
     }
 
-    public function hasBlock(string $name): bool
+    public function hasBlock(string $name, array $context = []): bool
```
stof, line 60: I suggest adding `@param array<string, mixed> $context` on this method to give more precise type (phpstan level 6+ does not like arrays that don't specify their types)

src/BlockChain.php
```diff
@@ -41,67 +44,127 @@ public function __construct(
...
+        $context += $this->context + $this->env->getGlobals();
+
+        return isset(($this->fixedBlocks ?? $this->resolveBlocks($context))[$name]);
     }
 
     /**
      * @return string[]
      */
-    public function getBlockNames(): array
+    public function getBlockNames(array $context = []): array
```
stof, line 70: I suggest adding `@param array<string, mixed> $context` on this method to give more precise type (phpstan level 6+ does not like arrays that don't specify their types)

src/BlockChain.php
```diff
@@ -41,67 +44,127 @@ public function __construct(
...
         }
-
-        $this->blocks = $blocks;
     }
 
-    public function hasBlock(string $name): bool
+    public function hasBlock(string $name, array $context = []): bool
     {
-        return isset($this->blocks[$name]);
+        $context += $this->context + $this->env->getGlobals();
```
stof, line 62: should this be done only when `fixedBlocks` is not set, as this would not use the context otherwise ?

src/BlockChain.php
```diff
@@ -41,67 +44,127 @@ public function __construct(
...
-        if (isset($this->blocks[$name])) {
-            return $this->blocks[$name][0];
+        $key = '';
+        $lineage = [];
+        $fixed = true;
+
+        foreach ($this->templates as $template) {
+            $seen = [];
+            do {
+                if (!$template->isOwnedBy($this->env)) {
```
stof, line 146: Do we need to do this for each resolution of the lineage ? The constructor already enforces it for chain templates. And `$template->getParent()` returns a template owned by the same Environment

src/Template.php
```diff
@@ -379,42 +373,16 @@ public function isOwnedBy(Environment $env): bool
     }
 
     /**
+     * Returns whether getParent() has stopped depending on the context.
+     *
+     * Only a template with no parent or with a constant one ever resolves its
+     * parent for good, so this also answers whether its lineage can still move.
+     *
      * @internal
```
stof, line 381: This `@internal` tag is useless. The whole class is internal.

PR #4926, opened 2026-09-14
Allow block chains to be composed of other block chains

fabpot:
This is something that is needed by Symfony (and other OSS projects), so better to have it directly in Twig.


src/BlockChain.php
```diff
@@ -33,20 +33,29 @@ final class BlockChain
...
      */
     public function __construct(
         private Environment $env,
         iterable $templates,
         private array $context = [],
     ) {
         foreach ($templates as $template) {
+            if ($template instanceof self) {
+                if ($env !== $template->env) {
+                    throw new \LogicException('A block chain cannot contain templates from different Twig environments.');
```
stof, line 47: should this talk about a block chain rather than templates ?

fabpot: Good catch, fixed in 56e5c0794b9ace8414c4e525221b5b203d65cf10

PR #4933, opened 2026-09-18
Improve macro call performance

fabpot:
This is mostly about getting back to the same performance we had in 3.28.


fabpot:
@upsun-dispatch review

PR #4934, opened 2026-09-19
Expose the escaping strategy a template was compiled with

fabpot:
Compiled templates now record the escaping strategy they were compiled with and expose it via `TemplateWrapper::getDefaultEscapeStrategy()`, so the strategy of a template can be known without deriving it again from its name, which is both slower and can disagree with how its content was actually escaped.


fabpot:
@upsun-dispatch review

stof (approved):

